PR #42, opened 2014-12-09
Make MaxChanLen a per Conn option

chendo:
Our particular use case uses many low throughput subscriptions. The default channel length of 64k uses quite a lot of memory and isn't necessary for us.

This PR adds the ability to override `MaxChanLen`


derekcollison:
I fixed this separately..

chendo:
Whoops, I forgot about this PR. Works for me, thanks!

PR #43, opened 2014-12-30
Set Request() call's inbox subscription chan length to 1.

lming:
The default inbox chan buffer size is 64K while Request() only need 1. CPU wastes may cycles on mallocing and gc if Request() is called frequently.


derekcollison:
I fixed this separately under #46..

PR #45, opened 2015-01-15
Add instruction to install nats sub/pub cli

longnguyen11288:


derekcollison:
I think these should go under gnatsd as util or something like that..

PR #49, opened 2015-02-10
Test on travis with go 1.4

klobucar:


.travis.yml
```diff
@@ -8,8 +8,7 @@ before_script:
 - which nats-server
 - nats-server -v
 install:
-- go get code.google.com/p/go.tools/cmd/vet
-- go get code.google.com/p/go.tools/cmd/cover
+- go get golang.org/x/tools/cmd/...
```
lloydde: This would seem to download and install quite a few other packages, https://godoc.org/golang.org/x/tools/cmd .
For now does it make sense just to continue just to get the two packages used?

PR #55, opened 2015-03-16
parser: rm unused OP_CONNECT & CONNECT_ARG.

lloydde:
I was looking at the test coverage and this caught my eye.

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -37,14 +37,6 @@ const (
 	OP_MINUS_ERR
 	OP_MINUS_ERR_SPC
 	MINUS_ERR_ARG
-	OP_C
-	OP_CO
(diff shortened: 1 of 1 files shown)


krobertson:
Unsure about removing this. It is still on the server side, and while it is only referenced in the parser on the server side, it is essentially part of the protocol.

lloydde:
@krobertson ok, let me know as we can add the rest of the states as there is no where to go from nc.ps.state = OP_C . And add a test case.

lloydde:
I did notice that the node client using a different model does use CONNECT, https://github.com/derekcollison/node-nats/blob/master/lib/nats.js#L338 .

Issue #53, opened 2015-03-14
.travis.yml should not use repotoken


lloydde:
.travis.yml should not use repotoken

The current behavior is travic-ci pull requests from ppl not in the aperca org fail.

Example build failing of change to README.md:
https://travis-ci.org/apcera/nats/builds/54365695

> Setting service_name to "travis-ci" and service_job_id to the Travis Job ID will automatically look up the appropriate build and repository, and assign the job correctly on Coveralls.
-- <cite>[THE COVERALLS API][1]</cite>

[1]:https://coveralls.zendesk.com/hc/en-us/articles/201774865-API-Introduction

PR #56, opened 2015-03-16
Golint a little happier

lloydde:


lloydde:
The tests still all pass. The travis build fails on goveralls, because of #53 .

krobertson:
Will discuss with @derekcollison and internally. A number of the renames may easily break compatibility for consumers of the library, and Go currently lacks a good way to do versioning around these things.

lloydde:
Thanks @krobertson. I'm not passionate about any of these changes. If you do find an example of a breakage, I'd like to add a test case.

krobertson:
I think its more because nats is a library rather than a server... so changing public naming can break consumers. I know merging this right now would break some of our internal repos which are using NATS.

lloydde:
Thanks for the further explaination @krobertson . I'm new to golang. I just researched the current state of  dependency versioning in go, and so will close this pull request. The ship here has sailed.

PR #58, opened 2015-06-29
Add protobuf encoder in nats client

aantelov87:


pkieltyka:
I agree, I don't think its worth introducing the external dependency. Perhaps put this in github.com/nats-io/nats/encoders/pbe or anywhere, and use `nats.RegisterEncoder()` to register the encoder outside of the core package.

derekcollison:
We could just do a simple example of how to use protobufs in the README?

pkieltyka:
IMHO, I'd like to see the protobuf encoder offered somewhere as a package. It also will create awareness for people that protobufs can be used with NATS. Then say, each of my services that wants to use protobufs, I'll have to make a package for the encoder somewhere anyways.

derekcollison:
Should we abandon this PR and create a new repo for encoders?

pkieltyka:
hmm.. I was thinking about that, but the github.com/nats-io org is for all nats clients, and if you have github.com/nats-io/nats-encoders for just Go encoders, then that would work, but then each client ie node-nats would have the repo nats-io/node-nats-encoders/ ...? or what're you thinking?

btw, (just checking in case it's not known), but if you did put encoders in `github.com/nats-io/nats/encoders/` like a `nats/encoders/pbe` package, it wouldn't add a dependency to the nats go client. Even then, doing a `go get github.com/nats-io/nats` wouldn't install the pbe package. It would only install all packages if you did `go get github.com/nats-io/nats/...` which people shouldnt be doing.

derekcollison:
Only the go client has encoders, so could do a sub-package.

pires:
I vote for sub-package. I actually implemented such an encoder as well and am very eager to see this merged.

pkieltyka:
:+1: on sub-package :)

derekcollison:
Any thoughts on the sub-package name? Since we include some encoders by default, should it be
x-encoders?

pires:
I have nothing against so :+1:

pkieltyka:
@derekcollison how about `github.com/nats-io/nats/encoders/npb` ?

the encoders/ is just a directory for holding other encoder sub-packages

pkieltyka:
and npb stands for "nats protobuff"

derekcollison:
We have built-in ones, and people may confuse a more generic encoders with
where they are all located. But to move the ones now to there would break
backward compatibility, which is a no-no in my book.. ;) Hence the
x-encoders for extra.

derekcollison:
Resolved as per PR and merge on 6d6f2c4f31f6bba97cefafed57929153eb475196

PR #64, opened 2015-09-01
Enforce MaxPayload on the client

wallyqs:
Makes the client proactively follow the `MaxPayload` size which was announced by the server upon connect. After this change, the client would get an `ErrMaxPayload` error when trying to publish a message which is larger than this limit.


derekcollison:
Should it be settable by the client at all? Should simply be something that
the client can query but its set by the server and enforced by the client
upon send.

wallyqs:
Thanks for the feedback, I modified it so that the client follows the max_payload size that is sent from server in the INFO message and removed the `EnforceMaxPayload`, reworded the PR to reflect this.

derekcollison:
We should have a MaxPayload() method on the connection, that way its read
only.

wallyqs:
Thanks for the comments, I made some updates to the PR addressing the issues you mentioned in case you can take a look.

PR #82, opened 2015-10-07
Don't fail to publish subsequent messages after sending one that's too big

reillywatson:
We have some messages that might be quite large.  Publishing any of these causes the connection to never send a message again, because publish() checks nc.err != nil and then bails.  ErrMaxPayload is only relevant to the individual message, it shouldn't make the whole connection bad.

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -1202,10 +1202,8 @@ func (nc *Conn) publish(subj, reply string, data []byte) error {
 	var msgSize int64
 	msgSize = int64(len(data))
 	if msgSize > nc.info.MaxPayload {
-		nc.err = ErrMaxPayload
-		err := nc.err
 		nc.mu.Unlock()
-		return err
+		return ErrMaxPayload
 	}
 
 	if nc.isClosed() {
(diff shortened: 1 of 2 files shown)


derekcollison:
This can be applicable if the client handles the error, but if the server handles it the connection will be closed and not useable. I think your suggestion is a good one, but need to understand how to signal the error and regain useable state.

reillywatson:
I'm not totally sure I follow.  This code is entirely in the client.  Is there a situation where a message larger than MaxPayload would even make it to the server?  For instance, is there a means of changing the server configuration while it's running in some way that wouldn't notify the client?

derekcollison:
Yes the code to catch it in the client is fairly recent. But I agree with you, just need to think through how to properly signal the error and allow the connect to continue.

reillywatson:
@derekcollison that seems unrelated, Travis CI is failing on master (see #83).

reillywatson:
Err, or rather see https://travis-ci.org/nats-io/nats/jobs/84114034, Github hid the Travis CI info when I closed that pull.

kozlovic:
LGTM

PR #89, opened 2015-10-16
"Fix" some tests.

kozlovic:
It looks like the default connect timeout of 2 seconds sometimes cause Travis build to fail. Not changing it at this time, but may have to do something special for Travis.


kozlovic:
Abandoning since the tests seem to fail randomly. Passed during the push, failed at the pr...

PR #92, opened 2015-10-19
Change default wait for tests

kozlovic:
I think that those changes may help prevent random build failures on Travis. Note that the change to the test's Wait() function is something to keep in mind when writing new tests, if those tests rely on Wait() to timeout quickly. You should then use the WaitTime() version that allows you to set a small timeout.

cc @derekcollison

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -678,11 +678,13 @@ func (nc *Conn) sendConnect() error {
 		nc.mu.Unlock()
 		return err
 	}
+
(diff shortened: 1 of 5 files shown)


kozlovic:
Yes, but that seems to help Travis... and it is a long time only for cases that such a wait is needed. If the test behaves ok and the notification is received in less than a mill, then that does not hurt.

PR #101, opened 2015-10-29
Added a queueing subscriber to the example programs

jduhamel:
Created a nats-qsub.go example. Main difference from nats-sub.go is that I called QueueSubscribe and printed out the queue name as well as the pid of the executable in the printMsg function for demonstration purposes otherwise it's identical to nats-sub.go

Changes:
*Second argument is queue name. 
*Subscribe call is to QueueSubscribe
*printMsg prints 2 additional pieces of information. queue name and process id.

### DIFF
--- a/examples/nats-qsub.go
+++ b/examples/nats-qsub.go
@@ -1,0 +1,63 @@
+// Copyright 2012-2015 Apcera Inc. All rights reserved.
+// +build ignore
+
+package main
+
+import (
+	"flag"
+	"log"
+	"os"
+	"runtime"
+	"strings"
+
+	"github.com/nats-io/nats"
+)
+
+func usage() {
+	log.Fatalf("Usage: nats-sub [-s server] [--ssl] [-t] <subject> <queue-group>\n")
+}
+
+func printMsg(m *nats.Msg, i int) {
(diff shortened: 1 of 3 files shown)


derekcollison:
Why not just add a queue group flag to nats-sub?

jduhamel:
Can do very easily. I was trying to keep the examples as clean as possible. Obviously there are only 3 changes in the code so I can wrap them with a if queueExample construct. 

Question is more to you about do you want to provide a minimal number of example or provide a set of examples that cover the major modalities. pub/sub queuing and request reply. 

-Joe

derekcollison:
You might have a point.

jduhamel:
Following up on this you would only need (or I can send the ones I have) a simple nats-req.go and a nats-rply.go to complete the set. My major driver for doing the nats-qsub was the documentation on the doc site was wrong and just wanted to put the raw materials out there.

derekcollison:
sure send them along too, or add them to this PR.

jduhamel:
Let me know what you want to do. -Joe

mcqueary:
LGTM. We'll want to include equivalent examples in the other clients.

/cc @nats-io/dev

PR #102, opened 2015-11-04
remove publish func defter

yireyun:


derekcollison:
What's the reasoning behind the change? Performance?

yireyun:
Avoid using defter, improve performance

PR #107, opened 2015-11-09
Performance improvement for async subscribers

kozlovic:
* Introduced a 'closed' flag in the subscription object, eliminating the need to grab the connection lock in the deliveryMsgs routine.
* Moved the creation of the message outside of the subscriber lock in processMsg


kozlovic:
@derekcollison I have revert to use s.conn for most of the tests. So that reduces the changes to use closed flag to break out of the delivery msg routine. A test is failing, but again that's probably timing, not related to the changes.

derekcollison:
Where were most of the performance gains coming from in your opinion?

kozlovic:
Like I described in the very first commit:

* The 'closed' flag eliminates the need to grab the connection lock in the deliveryMsgs routine, so less contention with the readLoop invoking processMsg
* Moving the creation of the message outside of the subscriber lock in processMsg, again, reduce lock contention with the sub's lock between the readLoop and deliveryMsgs routine.

That's what I discovered writing the C client (these changes come from the C client ;-)).

kozlovic:
.. and now the test fails ;-). Maybe we should wait on PR #110 to be merged ;-)

derekcollison:
I merged..

kozlovic:
@derekcollison There would be no need to pass sub.mch now, since we pass the sub itself. I could change that if you want.

By the way, I thought I broke something because subscribe() is done under connection's lock and deliveryMsgs() was grabbing that lock, which would give us a synchronization point (happens-before order).
I don't grab the connection lock anymore in that routing, but that's ok, since according to the spec (https://golang.org/ref/mem#tmp_5): "The go statement that starts a new goroutine *happens before* the goroutine's execution begins.", and: "Within a single goroutine, the *happens-before* order is the order expressed by the program". So we are sure that sub's max (and mch if I modify the routine to not pass it as a parameter) are properly updated.

derekcollison:
Want me to merge this one?

kozlovic:
I did not know you were waiting on me to do it ;-)
Sure, unless you want me to make the change to not pass the sub.mch as a parameter.

derekcollison:
Make that change to on this branch then I will merge when the tests clear.

kozlovic:
Tests are green, but let me do the change...

kozlovic:
Hold on: I am getting failures (even without the change) on TestRecvChanAsyncLeakGoRoutines() with -cpu=1,2,3,4. Need to take a look.

kozlovic:
@derekcollison The failures I got sometimes are due to runtime.NumGoroutine() when running with -cpu > 1. NumGoroutine() is not accurate (since it checks on things that are changed concurrently). Getting the stacks before and after (runtime.Stack([]byte, bool all) - which stops the world) was always consistent and not showing any leakage of goroutine. Plus, a good old "printf" at the end of the deliveryMsgs routine was always showing that it was properly exiting.

derekcollison:
ok keep me posted..

kozlovic:
@derekcollison Sorry for not being more clear. We are good to go. The failures (which happen only with -cpu) are false positive. You can check runtime.NumGoroutine() and will see that this is not reliable. I am confident that we don't have a problem in shutting down the deliveryMsgs routine.

Issue #74, opened 2015-09-17
AutoUnsubscribe() leaves deliverMsgs() go routine running


kozlovic:
Request() uses AutoUnsubscribe(1) and after the reply is received (or has timed-out), the subscription is closed: s.Unsubscribe(). This will close the channel and the go routine will fall out.

However, if an user uses AutoUnsubscribe() because he wants to receive at most 10 messages, and does not worry about how many messages have been received, then the go routine will stay even though it is not doing anything.

I wonder if deliverMsgs() should not exit when (delivered > max)

kozlovic:
Note: Regardless if the user should still call s.Unsubscribe() when done with the subscription, there is no need to keep the routine alive when no more message will ever be delivered.

kozlovic:
Furthermore, the subscription is not removed. So on a reconnect, the subscription is sent back with the auto unsubscribe count, but the server has received == 0. So it will send messages to the client that will discard them, since locally, the client knows that its number of delivered messages is greater than the max.
So not only the deliverMsgs() should exit when (delivered > max) but also remove itself from the connection's subscription list to avoid the above?

PR #114, opened 2015-11-27
Clean up logic around max handling to regress

camros:
I'm trying to test changes against my new Java client and believe that this is a little cleaner and fixes some edge race conditions. There are several more and not sure that these changes should be applied until a wider look at delivered/max logic is looked at here...

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -1031,8 +1031,8 @@ func (nc *Conn) deliverMsgs(s *Subscription) {
 		s.mu.Unlock()
 
 		// Use the same semantics everywhere for delivered
-		// Increments... I know this is in a lock vs. elsewhere
-		// Is this less racing???
+		// Increments... This was in the lock above but race
+		// checks were failing.
 		delivered = atomic.AddUint64(&s.delivered, 1)
 
 		if closed {
--- a/nats.go
+++ b/nats.go
@@ -1028,10 +1028,12 @@ func (nc *Conn) deliverMsgs(s *Subscription) {
 		s.mu.Lock()
 		max = s.max
 		closed = s.closed
+		s.mu.Unlock()
+
 		// Use the same semantics everywhere for delivered
 		// Increments... I know this is in a lock vs. elsewhere
+		// Is this less racing???
 		delivered = atomic.AddUint64(&s.delivered, 1)
-		s.mu.Unlock()
 
 		if closed {
 			break
--- a/nats.go
+++ b/nats.go
@@ -1028,8 +1028,9 @@ func (nc *Conn) deliverMsgs(s *Subscription) {
 		s.mu.Lock()
 		max = s.max
 		closed = s.closed
-		s.delivered++
-		delivered = s.delivered
+		// Use the same semantics everywhere for delivered
+		// Increments... I know this is in a lock vs. elsewhere
+		delivered = atomic.AddUint64(&s.delivered, 1)
 		s.mu.Unlock()
 
 		if closed {
@@ -1083,14 +1084,11 @@ func (nc *Conn) processMsg(data []byte) {
 
 	sub.mu.Lock()
 
-	// This is a catch all for more than max messages delivered.
-	if sub.max > 0 && sub.msgs > sub.max {
-		sub.mu.Unlock()
-		nc.removeSub(sub)
-		nc.mu.Unlock()
-		return
-	}
-
+	// Let the deliver loop and logic handle unsubscribing
+	// when we deliver / dequeue messages vs. producing them 
+	// to the consumer. Open new issue here around max handling semantics 
+	// overall?
+	
 	// Sub internal stats
 	sub.msgs += 1
 	sub.bytes += uint64(len(data))
@@ -1542,6 +1540,8 @@ func (s *Subscription) NextMsg(timeout time.Duration) (msg *Msg, err error) {
 			return nil, ErrConnectionClosed
 		}
 		delivered := atomic.AddUint64(&s.delivered, 1)
+		// FIXME: jam: off to the races on max vs. delivered
+		// FIXME: jam: cached above but possibly updated...
 		if max > 0 {
 			if delivered > max {
 				return nil, errors.New("nats: Max messages delivered")


camros:
Race check passes.

nats.go
```diff
@@ -1028,10 +1028,13 @@ func (nc *Conn) deliverMsgs(s *Subscription) {
 		s.mu.Lock()
 		max = s.max
 		closed = s.closed
-		s.delivered++
-		delivered = s.delivered
 		s.mu.Unlock()
 
+		// Use the same semantics everywhere for delivered
```
kozlovic: I think it was fine leaving delivered above. In case on async subscription, this variable is updated only here. It does not even need any locking, but clearly no need for atomic.

nats.go
```diff
@@ -1083,14 +1086,11 @@ func (nc *Conn) processMsg(data []byte) {
 
 	sub.mu.Lock()
 
-	// This is a catch all for more than max messages delivered.
```
kozlovic: You can't do that otherwise you would let messages accumulate to the channel, to a point probably of blocking down there, when trying to put it in the channel (since it is buffered).

kozlovic:
Could you print the race conditions that you got? I am not sure what race there was there. Thanks!

camros:
The code in various places around delivered/msgs/max does not work when pushing the client hard and is not correct. This fix helps. I understand a async and sync subscribers are different and not valid for the same subscription. 

All logic around auto-unsubscribe handling needs to be consistent in its results. It's not now. Will check in a larger test suite later that demos this with the my java client code m

kozlovic:
I retract my comment about the channel. Your changes would not alter the behavior of the channel, and so there would still be code to check that the SubChanLen limit is reached and process a slow consumer there. 
However, in the case of a sync subscription that would stop calling NextMsg(), removing the "removeSub" from processMsg() would lead to issue #74, whereby the subscription would not be removed, which would be a problem in case of reconnect.

camros:
My proposed change does not affect NextMesg from calling nc.removeSub. If you look at all this code its flawed, sometimes checking sub.delivered, sometimes sub.msgs and comparing to sub.max.

We have two overall issues in this code IMO and I am trying to vent out whether my view is right or wrong:
1. We should only compare delivered to max. Period no references to msgs, etc. that is to track deliveries to the subscription overall, NOT delivered messages
2. We should only decrement delivered messages/do this comparison logically ONCE, the previous code does it in two different ways with different concurrency profiles. This check should be done when WE DELIVER a message. That is the only time its valid to look at delivered.

I suspect the previous code was to optimize putting messages on the pending channel that may not be processed for auto unsubscribed subscriptions. That code is not correct from my understanding of how this is supposed to work. 

Can we put together a doc/issue/clarity on this? Am happy to do so but want to make sure I understand the semantics or my description above is correct -- or close ;-).

nats.go
camros: Above code not changed.

nats.go
```diff
@@ -1083,14 +1086,11 @@ func (nc *Conn) processMsg(data []byte) {
...
-	if sub.max > 0 && sub.msgs > sub.max {
-		sub.mu.Unlock()
-		nc.removeSub(sub)
-		nc.mu.Unlock()
-		return
-	}
-
```
camros: Deleted block here shows race condition. Why are we comparing sub.msgs here vs. sub.delivered? Why do this here, we DO NOT KNOW the message is going to be delivered. What if unsubscribe/etc... is called again, what if max is changed, etc...

nats.go
```diff
@@ -1542,6 +1542,8 @@ func (s *Subscription) NextMsg(timeout time.Duration) (msg *Msg, err error) {
 			return nil, ErrConnectionClosed
 		}
 		delivered := atomic.AddUint64(&s.delivered, 1)
+		// FIXME: jam: off to the races on max vs. delivered
+		// FIXME: jam: cached above but possibly updated...
 		if max > 0 {
 			if delivered > max {
 				return nil, errors.New("nats: Max messages delivered")
```
camros: Just commenting here on other race conditions. We do not want to check delivered/etc. in all these different ways. Longer term fix to talk through/discuss with test cases.

nats.go
kozlovic: Are you saying that there is a race and this is why you remove or you get the race after removing? Could you please include the race report?

The point of this check was to make sure that the connection does not "deliver" (that is pass it to the deliverMsgs routing or NextMsg) more messages than the set max number. 
Max can be changed, but that would be done under sub.mu.lock(), so we are fine.

nats.go
kozlovic: If you don't check delivered here, and you propose to remove from processMsg() then where is it going to happen? 
In case of sync subscriptions (use of NextMsg), the user can call it from different places at the same time. The use of atomic guarantees that only the first caller to go over the max will remove the subscription. The other calls to NextMsg() would have delivered strictly greater than max and will simply return the exception.

kozlovic:
You wrote:  "My proposed change does not affect NextMesg from calling nc.removeSub."

But your proposed changes would cause an issue when the user is not calling NextMsg(). If you remove the block where we call nc.removeSub() when sub.msgs > sub.max in processMsg(), then when messages arrive from the connection and the max is reached, without the code you are proposing to remove, the subscription will stay (again, if the user does not call NextMsg()). If a reconnect occurs, the library will resend a SUB for that subscription, because it still has it. So the code there ensures that the connection clean up the subscription that has received the given max amount. At this level, the connection doesn't care if the messages were actually delivered to the application. The connection simply knows that it has seen sub.msgs messages and that it is more than the sub.max allowed.

derekcollison:
I think the best way to proceed on these types of issues is to present test cases first that show the bad behavior, along with the fix. More tests are always good IMO.

camros:
Agreed. Just want to confirm expected logical behavior of max with regard to a subscribers view.  Ignore client libraries and think of the logical protocol spec. Is the below a reasonable draft of expected behavior of unsub and sub?

If so will clarify and get a test in. 

1. We guarantee at most _max messages_ will be delivered to a given subscriber
2. By default _max messages_ is infinite
3. Request response subscribers will automatically set _max messages_ to 1
4. Clients may override the default using language specific methods (typically passing _max_ to the unsubscribe method on connection). 
5. Nats guarantees that if _max messages_, at the exact time of a message is to be consumed by the subscriber, is less than or equal to zero the message will not be delivered (sic., discarded) to the subscriber. (Need to mention being removed as a subscription/logically closed?)
6. Nats guarantees that the value of _max messages_ for the subscriber is decremented right before it is consumed and in an atomic fashion with regard to step 5
7. If messages are to be delivered again to the same subject, the subscriber must use one of the mechanisms in steps 2-4 (necessary language?)

derekcollison:
Issue 3 above is cliebt specific, for example Node.js does not set this. Also, if set, performs an UNSUB.

Issue 5 I think you mean delivered. And yes if we receive it on the client side, UNSUB again as appropriate if we believe we did the auto UNSUB before.

Issue 6 is for delivered, but yes.

Not following Issue 7.

camros:
Thanks that helps and confirms my understanding. Again ignoring specific client framework specifics and their behavior -- the primary clarity I'm looking for is the semantics of unsub with respect to max != 0 

For point 7 (when max can change, delivery starts if ever again for a subscriber) was just trying to clarify delivery will not happen until a new subscription is created or the client changes max (if the client lib allows that, none do that I've reviewed)

The condition I found and will demo is that if the subscriber calls unsub themself (and other cases), the go and other client clients drop messages they should not because that check of max/delivered is not in the same unit of work as the as the actual callback or sync call to next message. There could be an undetermined amounted time when the eval off all that happens from the parser call back to the other routines to deliver/get a message to/from the calling client.

derekcollison:
UNSUB with max != 0 and present for the protocol signals the server to auto-unsubscribe when max has been delivered to the client's connection.

Currently I do not think any client supports resetting max after it is initially set. The server should be fine with it.

For the last condition, test will be good to show the bad behavior.

mcqueary:
Agreed, I don't think any client allows modifying max. Since Java, C#, and
C are modeled directly on the Go client, I'm pretty sure of it.

On Sunday, November 29, 2015, Derek Collison <notifications@github.com>
wrote:

> UNSUB with max != 0 and present for the protocol signals the server to
> auto-unsubscribe when max has been delivered to the client's connection.
>
> Currently I do not think any client supports resetting max after it is
> initially set. The server should be fine with it.
>
> For the last condition, test will be good to show the bad behavior.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/nats-io/nats/pull/114#issuecomment-160505086>.
>


-- 
*Larry McQueary* | Director, Messaging Technology
larry@apcera.com | @mcqueary | github.com/mcqueary

camros:
Umm, what if someone calls unsub in their message callback?

derekcollison:
I believe I test for that.

mcqueary:
There is a unit test for that in Go, Java, C and C#.

On Mon, Nov 30, 2015 at 9:34 AM, Derek Collison <notifications@github.com>
wrote:

> I believe I test for that.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/nats-io/nats/pull/114#issuecomment-160681244>.
>



-- 
*Larry McQueary* | Director, Messaging Technology
larry@apcera.com | @mcqueary | github.com/mcqueary

mcqueary:
Actually, re-reading Jeff's question, I'm not sure I'm interpreting
correctly:

 "Umm, what if someone calls unsub in their message callback?",

I am not sure I understand the question. Are you asking whether max gets
reset when unsubscribing from within a callback?


On Mon, Nov 30, 2015 at 9:37 AM, Larry McQueary <larry@apcera.com> wrote:

> There is a unit test for that in Go, Java, C and C#.
>
> On Mon, Nov 30, 2015 at 9:34 AM, Derek Collison <notifications@github.com>
> wrote:
>
>> I believe I test for that.
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/nats-io/nats/pull/114#issuecomment-160681244>.
>>
>
>
>
> --
> *Larry McQueary* | Director, Messaging Technology
> larry@apcera.com | @mcqueary | github.com/mcqueary
>



-- 
*Larry McQueary* | Director, Messaging Technology
larry@apcera.com | @mcqueary | github.com/mcqueary

camros:
I'm wasting time without showing code failing in  patch-1. here's one simple scenario, this stuff matters if we are pushing on more complex protocols like state management...

expected = 10000
received = AtomicInt (0);
// get the radar loop to drop messages that shouldn't be
handleMessage (Sub s, ...)
{
Int count = AtomicInc (&received);
  if count == 0: {
    Sleepseconds (5); // or long DB call, etc
    myconn.Unsub (s, 0); // or s.AutoUnsubscribe (0)
  }
}

assert (expected / 10 > 0);
sub = myconn.subscribe ("subj"' handleMessage, expected / 10);
expected.times.do { myconn.Publish ( "topple"); }
wait_seconds (10);
assert (expected == received.get ());

derekcollison:
This still relevant?

camros:
No longer relevant imo for the go client.

PR #117, opened 2015-12-22
Removed ineffective nc.err settings

mcqueary:
The affected lines set nc.err, but the calling function immediately reassigns nc.err, rendering these statements useless.

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -778,11 +778,6 @@ func (nc *Conn) processDisconnect() {
 	if nc.err != nil {
 		return
 	}
-	if nc.info.TLSRequired {
-		nc.err = ErrSecureConnRequired
-	} else {
-		nc.err = ErrConnectionClosed
-	}
 }
 
 // flushReconnectPending will push the pending items that were


mcqueary:
@derekcollison If you mean in between `nc.err` assignments? No, at least not in a way that I wouldn't consider a programming error.

- The user hasn't been notified of an error yet, therefore they'd have no reason to call `LastError()`
- The connection thread is locked throughout `processDisconnect()` and the subsequent `nc.err=err`. Any thread observing the lock wouldn't be able to access `nc.err` in between these two assignments.
- We're talking about the nanoseconds(?) between a variable assignment, stack frame pop, and variable reassignment. Even a non-threadsafe call to `LastError()` would have to happen in the nanosecond(s)between assignments.

derekcollison:
Not in between, but if the client get's a callback that they have registered for, say disconnect or closed, they could use LastError() to understand the underlying issue. The Go client's api is decent, but needs a redo on async callbacks around errors and encodings, etc.

mcqueary:
OK, I may have misunderstood the context of your question -- I was answering in the context of this specific Issue. Yes, `LastError()` returns `nc.err`, and the Java, C# and C clients do the equivalent.

derekcollison:
So we should leave it being set then, correct?

mcqueary:
No. `processOpErr()` and `processDisconnect()` are internal functions, not public API. `nc.err` is assigned and immediately reassigned in (basically) adjacent instructions. Then the mutex lock is released, and then the connection is closed. In the instant all this is happening, the user hasn't even been notified that anything is wrong yet, so they'd have no reason to call `LastError()`, nor any time to do it.

derekcollison:
ok gotcha..

Issue #118, opened 2015-12-26
sendConnect() doesn't handle +OK when verbose is set


mcqueary:
If the client's initial `CONNECT` message has `"verbose":true`, the server will reply with `+OK`. However, `sendConnect()` doesn't expect this, and will return an error.

kozlovic:
Sounds like I introduced this when making the (re)connect process synchronous, right?

-- 
Ivan Kozlovic | Principal Engineer
http://github.com/kozlovic
Apcera Inc.

derekcollison:
@kozlovic I think so..

kozlovic:
Will send something soon...

kozlovic:
Fixed by #120

PR #120, opened 2015-12-26
Fix handling of Verbose option during connect/reconnect

kozlovic:
* Since the (re)connect process was made synchronous, we now need to handle the possible return of +OK by the server.

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -725,6 +725,15 @@ func (nc *Conn) sendConnect() error {
 		return err
 	}
 
+	// If opts.Verbose is set, handle +OK
+	if nc.Opts.Verbose && strings.HasPrefix(line, _OK_OP_) {
+		// Read the rest now...
+		line, err = br.ReadString('\n')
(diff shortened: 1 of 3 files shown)


kozlovic:
This is for issue #118

mcqueary:
LGTM

mcqueary:
Actually, I wonder if that shouldn't be `line == _OP_OK_` instead of just checking for a prefix match? :)

kozlovic:
I thought about it, but that would not work as-is: _OK_OP_ is "+OK", but line is "+OK\r\n". I could add a new const okProto which is "+OK\r\n". Let me think about it.

mcqueary:
I'm sorry, of course you're correct. In java the CRLF is already stripped by this point. I forgot that's not true in the go client.

kozlovic:
Still, I had the same thought about checking exact value vs check prefix. So I modified the fix.

PR #124, opened 2016-01-04
Fix incorrect ErrConnectionClosed errors

### LINKED ISSUE #123: Incorrect error from NextMsg(timeout) when max reached
The `NextMsg(timeout)` function interprets `s.mch == nil` as indicating that the connection is closed, and returns `ErrConnectionClosed`. This is an invalid assumption. `s.mch` having the value `nil` only means that the subscription has become invalid (removed via `removeSub(s)`), which will be true if the connection `close(...)` has been called, but it will also be true if the subscription's auto unsubscribe `max` has been reached. In this latter case, erroneously reporting that the connection is closed results in confusion.

Fortunately we have a reliable indicator of whether the subscription is invalid: the `closed` field. I propose that we remove the Conn checks from this method and simply check `closed` to see if the sub is still valid.

Therefore I propose replacing this:
```go
	if s.mch == nil {
		s.mu.Unlock()
	}
	if s.mcb != nil {
		s.mu.Unlock()
		return nil, errors.New("nats: Illegal call on an async Subscription")
	}
	if s.conn == nil {
		s.mu.Unlock()
		return nil, ErrBadSubscription
	}
	if s.sc {
		s.sc = false
		s.mu.Unlock()
		return nil, ErrSlowConsumer
	}
```
with this:
```go
	if s.closed {
		s.mu.Unlock()
		return nil, ErrBadSubscription
	}
	if s.mcb != nil {
		s.mu.Unlock()
		return nil, errors.New("nats: Illegal call on an async Subscription")
	}
	if s.sc {
		s.sc = false
		s.mu.Unlock()
		return nil, ErrSlowConsumer
	}
```

This results in two test changes. `TestClosedConnections` needs to test for `sub.NextMsg(10)` to return `nats.ErrBadSubscription` rather than `nats.ErrConnectionClosed`, and `TestClientSyncAutoUnsub` should test that the error returned from `sub.NextMsg(10 * time.Millisecond)` is `nats.ErrBadSubscription` (right now it only tests `err != nil`).

mcqueary:
Fixes #123 

/cc @nats-io/dev

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -1509,17 +1509,13 @@ func (s *Subscription) AutoUnsubscribe(max int) error {
 // message has been delivered.
 func (s *Subscription) NextMsg(timeout time.Duration) (msg *Msg, err error) {
 	s.mu.Lock()
-	if s.mch == nil {
+	if s.closed {
 		s.mu.Unlock()
-		return nil, ErrConnectionClosed
+		return nil, ErrBadSubscription
 	}
 	if s.mcb != nil {
 		s.mu.Unlock()
 		return nil, errors.New("nats: Illegal call on an async Subscription")
-	}
-	if s.conn == nil {
-		s.mu.Unlock()
-		return nil, ErrBadSubscription
 	}
 	if s.sc {
 		s.sc = false
--- a/nats.go
+++ b/nats.go
@@ -1509,19 +1509,30 @@ func (s *Subscription) AutoUnsubscribe(max int) error {
 // message has been delivered.
 func (s *Subscription) NextMsg(timeout time.Duration) (msg *Msg, err error) {
 	s.mu.Lock()
-	if s.closed {
-		s.mu.Unlock()
-		return nil, ErrBadSubscription
-	}
-	if s.mcb != nil {
-		s.mu.Unlock()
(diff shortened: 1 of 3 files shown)


kozlovic:
Commented on the PR: I have been there, done that, but we decided to stick with s.mch test instead.

mcqueary:
I have changed this to reflect what I most recently suggested in the comments of #123. Preserves existing logic, but tests for max messages reached when mch is nil. The question is whether the error should be more specific? Maybe turn "nats: Max messages delivered" into a full-fledged error?

mcqueary:
Summary of accumulated changes in this PR:
- add `s.connClosed`, which is set during `Conn.close()` to indicate that the subscription was closed due to the Conn being closed
- removed check for `s.conn == nil` in `NextMsg()`
- add `ErrMaxMessages` to indicate autoUnsubscribe caused a subscription to be closed
- add `ErrSyncSubRequired` for when `NextMsg()` is called on an async subscription.
- add a test for calling NextMsg on a closed subscription

Each condition is now discretely identified and tested for.

derekcollison:
Run gofmt on the source files?

mcqueary:
@derekcollison gofmt done.

PR #126, opened 2016-01-08
Fixes issue where nats connection cannot be made

aemengo:
Running into issues connecting to a locally hosted nats server, which is run with nats gem:
`nats (0.5.1)` on OS X: `10.9.5`. This is due to a `+OK` line being returned as opposed to the other pre-selected
errors, which should be permissible.

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -737,13 +737,13 @@ func (nc *Conn) sendConnect() error {
 		}
 	}
 
-	// We expect a PONG
-	if line != pongProto {
-		// But it could be something else, like -ERR
-		if strings.HasPrefix(line, _ERR_OP_) {
-			return errors.New("nats: " + strings.TrimPrefix(line, _ERR_OP_))
-		}
-
+	// We expect a PONG, the +OK line is also fine
+	switch {
+	case line == pongProto, strings.HasPrefix(line, _OK_OP_):
+		break
+	case strings.HasPrefix(line, _ERR_OP_):
+		return errors.New("nats: " + strings.TrimPrefix(line, _ERR_OP_))
+	default:
 		return errors.New("nats: " + line)
 	}
 


kozlovic:
Please look at PR #120. You may want to update your local repo.

aemengo:
@kozlovic Thanks for the response. I am definitely getting the `+OK` at that line of the code. And I am following the [basic usage](https://github.com/nats-io/nats#basic-usage), so I don't see how the client is asking for it. I am using server provided by the [gem](https://rubygems.org/gems/nats/versions/0.5.1) `0.5.1` which I installed via the `gem install nats` command.

I have inspected further and see that it's not the `okProto` condition that it failing but the `nc.Opts.Verbose` command. It is returning `false` for me.

wallyqs:
Hi @aemengo , the `nats-server` included in the ruby nats gem is deprecated, I recommend you to try with a release of the latest version of gnatsd instead: https://github.com/nats-io/gnatsd/releases

kozlovic:
That means that the server is sending +OK although the client `Verbose` is not set. I would consider this a protocol error, so the API as it is (without your fix) is working correctly.

kozlovic:
Also, even if we wanted to "support" the fact that the server is sending `+OK` even when not asked, the fix would not be appropriate. We want in `sendConnect()` to make sure that we receive a `PONG` to our initial `PING` - this is part of the connect protocol if you will.

Your change will return after getting `+OK`, but there is no guarantee that `PONG` will arrive after that. So we would have to say if `+OK`, then read the next line and make sure we get `PONG` before returning ok. Anything else will be considered an error.

aemengo:
Understood. Perhaps a pull request wasn't the most appropriate course of action. Ultimately, I was just puzzled as to why the library couldn't connect to my nats server, since I was completely using defaults for everything and basic usage. When I inspected the golang err returned. The only thing that was shown was: 'nats: +OK'

kozlovic:
Don't worry. We encourage people to submit PRs! I am sure that if you try with latest release of the server, it will work fine.

aemengo:
Confirmed. The library works swimmingly with the gnatsd release.

PR #133, opened 2016-01-21
Schedule callbacks on a connection in a go routine to preserve order

kozlovic:
@derekcollison This seems overly complex to me, and not really go-like. Would appreciate pointers on how to do this differently.

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -135,23 +135,10 @@ type Options struct {
 	SubChanLen int
 }
 
-type asyncCBType int
-
-const (
-	CONN_CB = asyncCBType(iota)
-	ERR_CB
-)
-
-type asyncCB struct {
-	cbType asyncCBType
-
-	connCb ConnHandler
-
-	errCb ErrHandler
-	sub   *Subscription
-	err   error
-
-	next *asyncCB
+type errHandlerInfo struct {
+	cb  ErrHandler
+	sub *Subscription
+	err error
 }
 
 const (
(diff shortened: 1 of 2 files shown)


derekcollison:
I agree, this is pretty complex. I think we can come up with something simpler and have one Go routine per Connection. It selects over ConnHandler chan (connect, reconnect, disconnect) and an ErrHandler chan and a quit condition. Let me give this a pass and see what I can come up with.

kozlovic:
Appreciate it. I did not want to use a channel (other than to notify that the list should be processed) because of the size of the channel. If it is unbuffered, we would potentially block. If buffered, what size to use (reference to removing channel for async msgs come to mind)?

derekcollison:
The connHandler is not very many events, so 4 or 8 buffered size. The
ErrHandler may be more, but if it comes from the server we will be
disconnected, and if it is from client side slowConsumer we gate it.

kozlovic:
I thought about it, but technically, the same connection could go through a list of disconnect - reconnect -disconnect - reconnect - disconnect - etc... possibly rapidly.

derekcollison:
Not sure how rapid or how common, but I understand what you are saying.

kozlovic:
If we have a select for conn handler and one for async err handler, then the total order would not be respected. If we are ok with that, I can have another go at trying to implement this.
That being said, for async, we still need to not only push a function, but there are the sub and err to pass around that may be different each time.

derekcollison:
You can close over the arguments.

kozlovic:
@derekcollison Not sure if this is closer to what you had in mind.

The issue I have with that approach is that async cbs may be out-of-order compared to connection callbacks (due to the use of different channels). Also, the use of a buffered channel scares me, especially that for async err, we send to the channel from readLoop thread, and if we block there (channel full) while the user tries to close the connection from async err callback, we'll deadlock.

The initial approach was ugly, but it did not have any of the above issues.

kozlovic:
Closing this PR since covered in other one.

PR #135, opened 2016-01-23
Preserve ordering of callbacks, fixes #30

### LINKED ISSUE #30: Order or lack thereof of callbacks
It seems a bit weird that in most cases I will see a ClosedCB and maybe a ReconnectedCB before the DisconnectedCB.  Because DCB is async, there is no order guarantee and it makes it difficult to actually do anything when you do see it since you cannot easily determine a reliable sequencing between the various callbacks.

derekcollison:
/cc @kozlovic 

Can you create an ordered test, possibly pulling from your other PR?

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -37,7 +37,7 @@ const (
 	DefaultTimeout          = 2 * time.Second
 	DefaultPingInterval     = 2 * time.Minute
 	DefaultMaxPingOut       = 2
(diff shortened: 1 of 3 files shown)


kozlovic:
Will do. I found some problems with the code (forgot to do the Disconnected on doReconnect, issue with closing with delayed nillify of ach, etc..) that I need to fix and test for it. Will submit something later this week-end.

Issue #136, opened 2016-01-24
Possible nil pointer dereference during reconnect


kozlovic:
Publish() checks if we have reached the max size of the pending buffer by checking `isReconnecting()`. The status is changed to `RECONNECTING` before starting the go routine doReconnect() where the pending buffer is actually created. This can result in `nc.pending` being nil when checking `nc.pending.Len()`, in `publish`.
Possible fix is to check for `nc.pending != nil` instead of `isReconnecting()`, or move creation of the pending buffer earlier.

kozlovic:
Fixed in #137

PR #137, opened 2016-01-24
Fix possible nil dereference when publish during reconnect.

kozlovic:
Addresses issue #136.

This also could be addressed by moving creation of pending buffer in `processOpErr`?

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -1525,7 +1525,9 @@ func (nc *Conn) publish(subj, reply string, data []byte) error {
 
 	// Check if we are reconnecting, and if so check if
 	// we have exceeded our reconnect outbound buffer limits.
-	if nc.isReconnecting() {
+	// (note that isReconnecting() could return true but nc.pending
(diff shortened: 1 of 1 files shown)


kozlovic:
@derekcollison Not true. `RECONNECTING` is set in `processOpErr` (https://github.com/nats-io/nats/blob/master/nats.go#L1157), then the go routine `doReconnect` is started, and the lock is released. By the time the lock is acquired in `doReconnect`, and the buffer created, there is a window where one can publish and hit the condition where `RECONNECTING` is true, and `nc.pending` is nil.

derekcollison:
Yes you are correct, I think we should flush regardless. And you might be right to move where we create pending.

kozlovic:
Moved the creation of the pending buffer and restored the `isReconnecting` test. Running tests to confirm this is ok and will push changes in a bit.

derekcollison:
Thanks @kozlovic!

PR #139, opened 2016-01-25
Optimize for larger messages.

kozlovic:
This was already present in the server.

### DIFF
--- a/enc_test.go
+++ b/enc_test.go
@@ -1,6 +1,7 @@
 package nats_test
 
 import (
+	"fmt"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


kozlovic:
Ok, I guess you mean writing a new test that can access internals, right?

derekcollison:
If we need to. I would check current coverage, just want to make sure all new code is covered with tests.

kozlovic:
It's not. I will work on a test that check those.

derekcollison:
Thanks.

derekcollison:
Can you rebase and resolve conflicts?

kozlovic:
Will do. I knew this was going to happen with the other PR ;-)

PR #140, opened 2016-01-27
Fix parser to check for missing sid.

kozlovic:
Adding tests for code coverage, I discovered that we would not check for missing sid (sid would be -1 due to parseInt64). I believe sid is mandatory.


kozlovic:
Note that I did this outside of #139, since there is a fix for the parser. That being said, these tests combined to the one in the other PR will increase code coverage of the parser to above 90%.

derekcollison:
This is great! Thanks.

PR #141, opened 2016-01-28
Fix some encoders issues.

kozlovic:
- Return error when Handler has no argument (would cause nil pointer panic)
- Schedule async error callbacks when possible
- Add more tests and increase test coverage


derekcollison:
Your on fire! ;)

PR #142, opened 2016-01-28
Fix netchan encoder's call to async err callback.

kozlovic:
Check closed state and send into async cb channel under conn's lock.


derekcollison:
Should we have a test for this fix?

PR #145, opened 2016-01-29
Fix "Stale Connection" error check.

kozlovic:
The case of the error message was changed to all lower case in the client. The server is using mixed case. This would cause the check to fail, resulting in connection close, instead of possibly trying to reconnect.

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -44,7 +44,7 @@ const (
 )
 
 // STALE_CONNECTION is for detection and proper handling of stale connections.
-const STALE_CONNECTION = "Stale Connection"
+const STALE_CONNECTION = "stale connection"
 
 // Errors
 var (
@@ -1489,7 +1489,7 @@ func (nc *Conn) LastError() error {
 // sets the connection's lastError.
 func (nc *Conn) processErr(e string) {
 	// FIXME(dlc) - process Slow Consumer signals special.
-	if e == STALE_CONNECTION {
+	if strings.ToLower(e) == STALE_CONNECTION {
 		nc.processOpErr(ErrStaleConnection)
 	} else {
 		nc.mu.Lock()
--- a/nats.go
+++ b/nats.go
@@ -44,7 +44,7 @@ const (
 )
 
 // STALE_CONNECTION is for detection and proper handling of stale connections.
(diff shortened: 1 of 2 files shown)


kozlovic:
They are all lower cases because you change them in 90f704d8165af6cb368abcec3522b949c5df67a8. Are you saying that I should revert the change to the errors from this commit?

kozlovic:
If you ask to leave "Stale Connection" as "stale connection" in the client to match the case of other errors in master, and ask to change the server error "Stale Connection" to "stale connection", you are going to break older (that is, anything outside of master) clients. Is that what you want?

derekcollison:
I thought you did revert it back to CamelCase.. no?

kozlovic:
In this commit, I did not. Just fixed the "stale connection" one. Do you want me to revert 90f704d8165af6cb368abcec3522b949c5df67a8 (the errors part) in this PR?

kozlovic:
@derekcollison To re-state the issue: the server is sending "Stale Connection", so the check for "Stale Connection" == "stale connection" will fail. I can convert to lower case both before test to fix this issue.
The question is, regarding the other errors, do you think it may be a problem for customers with existing applications that may compare error's string, or do you believe that they would simply check against the error (say `nc.LastError() == ErrConnectionClosed` instead of checking the string value)? Depending on the answer, there is no point of rolling back your change and I will modify the fix to do lower case before check.

derekcollison:
Let's make the const, and hence the client error all lowercase. Let's convert server to lowercase in test suite to match. Thanks.

kozlovic:
How is that helpful? If the parse error test uses lower case, then the test would pass, but that's not what a real server will send. If we do change gnatsd error to lower case, then it will break older nats clients.

kozlovic:
I misread. So you say we convert to lower case in `processErr` to check and leave the errs in client in lower case.

derekcollison:
yes.

kozlovic:
Got it. Thanks!

kozlovic:
Disregard the coverage "failure". The decrease is in master (due to the fix for the parser). The coverage for the actual branch is unchanged: https://coveralls.io/builds/4904479

PR #150, opened 2016-01-29
Add nats.ClientCert helper function.

kozlovic:
Useful when server requires client certificate.


kozlovic:
Not in the client repo (copied over from server). Good point about README.md, will do that.

derekcollison:
Let me know when README.md changes make it in and will merge. Thanks.

Issue #152, opened 2016-01-30
On reconnect, client may send incorrect UNSUB value to server


kozlovic:
Setup
- Create Subscription
- Call AutoUnsubscribe(N)
- Send/Consume "N - m" messages
- Restart server and wait for client to reconnect
- Send more messages.

The client auto-unsubscribes after the correct number of messages, however, we can see in the server log that the original (N) number is sent as part of the UNSUB protocol. The server will therefore send more messages to the library, which will drop them.

(filed for tracking purposes - fix coming up)

kozlovic:
Closed by #153

Issue #151, opened 2016-01-30
NextMsg() on SyncSubscription may not return until Timeout when max is reached


kozlovic:
Setup:
- Create sync subscription
- call AutoUnsubscribe
- Have multiple go routines sitting on NextMsg(Timeout)
- Send enough messages to reach the max
- All but one NextMsg() will wait until Timeout before exiting, although the subscription is being removed.

(created for tracking purposes - fix coming up)

kozlovic:
Closed by #153

PR #153, opened 2016-01-30
Fix for issues #151 & #152.

kozlovic:
- Calls to NextMsg() for SyncSubscriptions may block until Timeout even when max messages have been received and subscription is removed.
- During reconnect, when resending the subscriptions, update the 'max' value with 'max - delivered' to prevent server for sending messages for nothing.


derekcollison:
We ready to merge from your standpoint?

kozlovic:
Yes.

PR #159, opened 2016-02-04
Change directory layout to support 'go get' for the nats client commands

tw4dl:
With this, we can do: go get github.com/nats-io/nats/… 
and these go binaries will be built in $GOPATH/bin

### DIFF
--- a/cmd/nats-bench/nats-bench.go
+++ b/cmd/nats-bench/nats-bench.go
@@ -1,5 +1,4 @@
 // Copyright 2015 Apcera Inc. All rights reserved.
-// +build ignore
 
 package main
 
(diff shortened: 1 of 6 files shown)


tw4dl:
No "go install" is needed. Currently these are treated as examples, and are not built. Some or all could be moved to be client commands which user can easily get to help testing and validating gnatsd with the suggested change. It is more of a WIP PR, need related doc changes.

derekcollison:
I think we can add some of these to server (gnatsd) codebase to install and have some utilities. Don't think we do that at the Go client layer.

tw4dl:
SGTM

PR #160, opened 2016-02-04
Protobuf encoder improvement

AnatoliiPrylutskyi:
Hi

I found it convenient and intuitive to subscribe with empty interface{} in case when you don't need arguments to reply with encoded response. Also it's convenient to request with nil argument.

I used this modified proto encoder for our application.
I create failed test cases in commit: https://github.com/AnatoliiPrylutskyi/nats/commit/a3c60694003763933585705ca6f89b18599a8367

Thanks

### DIFF
--- a/encoders/protobuf/protobuf_enc.go
+++ b/encoders/protobuf/protobuf_enc.go
@@ -33,6 +33,9 @@ var (
 
 // Encode
 func (pb *ProtobufEncoder) Encode(subject string, v interface{}) ([]byte, error) {
+	if v == nil {
+		return nil, nil
+	}
 	i, found := v.(proto.Message)
 	if !found {
 		return nil, ErrInvalidProtoMsgEncode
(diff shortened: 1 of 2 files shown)


AnatoliiPrylutskyi:
Strange error message on build, please inform me if I can fix this error.

AnatoliiPrylutskyi:
For example:
Service "A" hold currency rates info and do not need arguments to return rates, it return currency rates on request in protobuf on topic "currencyRates".
Service "B" need to load all currency rate, so Service "B" need to request Service "A" on "currencyRates" topic with no arguments.
In current implementation I can't pass nil to Request function with protobuf encoder, cause I will receive error, but it's convenient to use proto-encoded connection to auto-unmarshall response from protobuf. Also I can't subscribe with empty interface{} with proto-encoded connection.

AnatoliiPrylutskyi:
> natsResp := new(f9_proto.CurrencyResponse)
> Nats.ProtoConn().Request(currencyRatesTopic, nil, natsResp, timeout)

PR #170, opened 2016-03-07
Fixed reconnect loop possibly not yielding to Close()

kozlovic:
The issue manifests on Windows with a ReconnectWait lower than a second, because it takes more than a second for createConn() to return when no server is running. Therefore, the loop would never sleep (and never release the lock), which would prevent a Close() from breaking the reconnect loop.

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -16,6 +16,7 @@ import (
 	"net"
 	"net/url"
 	"regexp"
+	"runtime"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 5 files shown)


kozlovic:
@derekcollison Thanks for feedback. Updated tests. Did you see nats.go, though, and use of runtime.Gosched(). Is that ok? The idea is that we want to release lock, even if we don't sleep, to give a chance to Close() to break out the loop.

derekcollison:
I did not see any changes to nats.go..

kozlovic:
That's what I figured. I had 2 commits originally on the PR (wanted to separate the actual fix and changes to the tests for Windows tests to pass).

kozlovic:
Should have done 2 PRs, sorry about that.

kozlovic:
@derekcollison With Gosched(), it seems to be working, although Gosched() says:

PR #173, opened 2016-03-09
Fixed panic on channel receive after connection closed

kozlovic:
-When creating a Chan subscriber, the user provides the channel.
 We should not close that channel on connection close.
-Add test to check that connection close does not close the user
 channel.


kozlovic:
@derekcollison Not sure if it was an oversight or not. We do ensure that we don't close the channel on `Unsubscribe`, so I figure we should have done the same on `Close`?

PR #176, opened 2016-04-07
Update .gitignore to ignore vim swp files and eclipse project artifacts

mcqueary:


kozlovic:
Tests fail because of `vet` that is being addressed in other PR. Merging this one.

PR #188, opened 2016-05-25
Reconnect upon receiving an authorization timeout error

ColinSullivan1:
Begin the reconnect process upon receiving an authorization timeout error from the server.

In situations where a server is under high load, there can be a delay between the authorization timeout message being processed in the client and the connection being closed by the server.  Normally, the connection was being closed quickly, triggering the reconnect process.  However, if the authorization timeout was processed first, the client closed it's connection and did not reconnect when configured to.

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -42,8 +42,11 @@ const (
 	LangString              = "go"
 )
 
-// STALE_CONNECTION is for detection and proper handling of stale connections.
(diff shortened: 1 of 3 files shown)


ColinSullivan1:
Need to rewrite the test to account for locking...

ColinSullivan1:
Closing pending further review...

PR #199, opened 2016-07-05
[BUG] Modified examples/nats-bench so that specified message count is…

aricart:
… distributed equally between the number of publisher and subscriber clients.

[NEW] Added the flag -csv <filepath> to save a per-client report of the benchmark. This allows for the data to be saved and compared to with a different run.

[NEW] Added a the flag -o <filename> to save a JSON version of the report. Renamed some of the internal fields of the structs to enable the JSON serialization to work.

[CHANGED] The throughtput values are calculated based on the message payload. If additional messages were received by the connection these are categorized as "protocol messages".

@kozlovic @ColinSullivan1

### DIFF
--- a/examples/nats-bench.go
+++ b/examples/nats-bench.go
@@ -1,17 +1,21 @@
 // Copyright 2015 Apcera Inc. All rights reserved.
-// +build ignore
 
 package main
 
 import (
+	"bytes"
 	"flag"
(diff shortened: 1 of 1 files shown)


aricart:
I am going to refactor some of the benchmark utils since streaming uses it. Let's not duplicate code.

PR #201, opened 2016-07-07
Update nats bench

aricart:
- Refactored `nats-bench` 'bench' functionality into a small library. Changed nats-bench to use bench package
- Added 'bench' package to github.com/nats-io/nats/bench 
- Added support for saving samples in CSV format with the intention of incorporating running of bench as part of the build and save historic performance information
- Added tests to bench package

@kozlovic @ColinSullivan1

### DIFF
--- a/bench/bench.go
+++ b/bench/bench.go
@@ -1,0 +1,335 @@
+// Copyright 2012-2016 Apcera Inc. All rights reserved.
+
+package bench
+
+import (
+	"bytes"
+	"encoding/csv"
+	"fmt"
+	"log"
+	"math"
+	"strconv"
+	"time"
+
+	"github.com/nats-io/nats"
+	"github.com/nats-io/nuid"
+)
+
+// A Sample for a particular client
+type Sample struct {
+	JobMsgCnt int
+	MsgCnt    uint64
+	MsgBytes  uint64
+	IOBytes   uint64
+	Start     time.Time
+	End       time.Time
+}
+
+// SampleGroup for a number of samples, the group is a Sample itself agregating the values the Samples
+type SampleGroup struct {
+	Sample
+	Samples []*Sample
+}
+
+// Benchmark to hold the various Samples organized by publishers and subscribers
+type Benchmark struct {
+	Sample
+	RunID      string
+	Pubs       *SampleGroup
+	Subs       *SampleGroup
+	subChannel chan *Sample
+	pubChannel chan *Sample
+}
+
+// NewBenchmark initializes a Benchmark. After creating a bench call AddSubSample/AddPubSample.
+// When done collecting samples, call EndBenchmark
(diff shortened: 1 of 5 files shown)


bench/bench.go
```diff
@@ -0,0 +1,335 @@
+// Copyright 2012-2016 Apcera Inc. All rights reserved.
```
kozlovic: Copyright 2016 (not 2012-2016) since this is a new file.

bench/bench.go
```diff
@@ -0,0 +1,335 @@
...
+	writer.Write([]string{"#RunID", "ClientID", "MsgCount", "MsgBytes", "MsgsPerSec", "BytesPerSec", "DurationSecs"})
+	groups := []*SampleGroup{bm.Subs, bm.Pubs}
+	pre := "S"
+	for i, g := range groups {
+		if i == 1 {
+			pre = "P"
+		}
+		for j, c := range g.Samples {
+			r := []string{bm.RunID, fmt.Sprintf("%s%d", pre, j), fmt.Sprintf("%d", c.MsgCnt), fmt.Sprintf("%d", c.MsgBytes), fmt.Sprintf("%d", c.Rate()), fmt.Sprintf("%f", c.Throughput()), fmt.Sprintf("%f", c.Duration().Seconds())}
+			err := writer.Write(r)
```
kozlovic: Do it in one line: `if err := writer.Write(r); err != nil {`..

bench/bench.go
```diff
@@ -0,0 +1,335 @@
...
+			}
+		}
+	}
+
+	writer.Flush()
+	return buffer.String()
+}
+
+// NewSample creates a new Sample initialized to the provided values. The nats.Conn information captured
+func NewSample(jobCount int, msgSize int, start, end time.Time, nc *nats.Conn) *Sample {
```
kozlovic: Although I don't see it as a problem, if bringing this dependency proves to be problematic, you would have to pass the In/Out stats values as parameters here...

bench/benchlib_test.go
```diff
@@ -0,0 +1,221 @@
...
+	s.MsgBytes = uint64(messages * Byte)
+	s.IOBytes = s.MsgBytes
+	return s
+}
+
+func TestDuration(t *testing.T) {
+	s := millionMessagesSecondSample(1)
+	duration := s.End.Sub(s.Start)
+	if duration != s.Duration() || duration != time.Second {
+		t.Fail()
```
kozlovic: Would be nice to put some text and explain what failed, following a model such as `t.Fatalf("Expected %v, got %v", expectedValue, actualValue)`, when applicable of course.

kozlovic: True for all other `t.Fail()` that follow...

kozlovic: Also, note that `t.Fail()` marks the function as failed, but continue execution, which may not be desired.

PR #202, opened 2016-07-07
Refactor nats-bench functionality into a small library.

aricart:
[NEW] Added 'bench' package to nats
[NEW] Added support for saving samples in CSV format
[NEW] Added tests to bench package
[CHANGED] nats-bench to use bench package

### DIFF
--- a/examples/nats-bench.go
+++ b/examples/nats-bench.go
@@ -55,7 +55,7 @@ func main() {
 	}
 	opts.Secure = *tls
 
-	benchmark = bench.NewBenchmark(*numSubs, *numPubs)
+	benchmark = bench.NewBenchmark("NATS", *numSubs, *numPubs)
 
 	var startwg sync.WaitGroup
 	var donewg sync.WaitGroup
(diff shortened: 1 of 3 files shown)


examples/nats-bench.go
```diff
@@ -134,22 +104,22 @@ func runPublisher(startwg, donewg *sync.WaitGroup, opts nats.Options, numMsgs in
...
+		msg = make([]byte, msgSize)
 	}
 
 	start := time.Now()
 
 	for i := 0; i < numMsgs; i++ {
 		nc.Publish(subj, msg)
 	}
-	pubStatChan <- newStats(start, time.Now(), nc, true)
+	benchmark.AddPubSample(bench.NewSample(numMsgs, msgSize, start, time.Now(), nc))
```
kozlovic: For the publisher, you should move the `nc.Flush()` ahead of the capture of the time here. That will be more representative of all data flushed to the server.

kozlovic:
Other than the note, LGTM.

aricart:
removing to collapse, and to make a fix that Ivan caught.

PR #206, opened 2016-07-14
Changes to nats-bench and bench lib

aricart:
Modified nats-bench so that subscribers wait to receive the total message count.

Modified the bench library so that it adds the number of messages that the client received or published.


kozlovic:
LGTM

ColinSullivan1:
LGTM

PR #209, opened 2016-08-01
[ADDED] Support for processing of asynchronous INFO protocols

kozlovic:
This would be used in conjunction with server's PR #314.
The client may receive INFO protocols with a field containing
a possible array of URLs that correspond to server addresses in
the full mesh cluster that clients can connect to.
The server pool is updated with these URLs.
3 new Options and 2 new Options setters have been introduced to
manage username/password and/or tokens when dealing with these
bare URLs.

/cc @derekcollison

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -141,6 +141,10 @@ type Options struct {
 	// NOTE: This does not affect AsyncSubscriptions which are
 	// dictated by PendingLimits()
 	SubChanLen int
+
+	User     string
+	Password string
+	Token    string
 }
 
 const (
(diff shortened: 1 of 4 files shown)


kozlovic:
@aricart @ColinSullivan1 @mcqueary @wallyqs Please follow this PR since if this is merged, you will have to update your respective clients that want to offer support for async INFOs.

kozlovic:
In the process of updating tests for coveralls, noticed a bug in that I don't create the argBuf in split scenarios. Fixing that. Will update later.

PR #212, opened 2016-08-17
Add option for allowing custom Dialers

wallyqs:
The Golang [Dialer](https://golang.org/pkg/net/#Dialer) has a number of options which would be convenient to allow users customizing them.
This changeset adds a `nats.Dialer` func for setting custom dialers (e.g. to activate opt-in `DualStack` support) which would be used when attempting to establish a connection to the server, still using by default the [DialTimeout](https://golang.org/pkg/net/#DialTimeout) behavior as it is currently.

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -84,6 +84,9 @@ var DefaultOptions = Options{
 	MaxPingsOut:      DefaultMaxPingOut,
 	SubChanLen:       DefaultMaxChanLen,
 	ReconnectBufSize: DefaultReconnectBufSize,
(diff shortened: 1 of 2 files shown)


derekcollison:
Check why travis tests failed?

kozlovic:
@wallyqs Could you rebase?

wallyqs:
@kozlovic thanks for following up, rebased and tests are passing now :)

kozlovic:
LGTM

PR #214, opened 2016-08-22
[ADDED] Add Conn.Servers API

ColinSullivan1:
Returns a string array of all known servers including those updated by the async INFO protocol.

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -2534,10 +2534,10 @@ func (nc *Conn) Servers() []string {
 	return nc.getServers(false)
 }
 
-// ImplicitServers returns only the server urls that have been discovered
+// DiscoveredServers returns only the server urls that have been discovered
 // after a connection has been established. If authentication is enabled,
 // use UserInfo or Token when connecting with these urls.
-func (nc *Conn) ImplicitServers() []string {
+func (nc *Conn) DiscoveredServers() []string {
 	nc.mu.Lock()
 	defer nc.mu.Unlock()
 	return nc.getServers(true)
--- a/nats.go
+++ b/nats.go
@@ -2507,8 +2507,10 @@ func (nc *Conn) IsConnected() bool {
 	return nc.isConnected()
 }
 
-// Servers returns the list of known servers, including additional
-// servers discovered after a connection has been established.
+// Servers returns the list of known server urls, including additional
+// servers discovered after a connection has been established.  URLs passed
+// into NATS through the API be preserved in their original form, but any
+// credentials in a URL will never be visible outside of the application.
 func (nc *Conn) Servers() []string {
 	nc.mu.Lock()
 	defer nc.mu.Unlock()
--- a/nats.go
+++ b/nats.go
@@ -2511,7 +2511,7 @@ func (nc *Conn) IsConnected() bool {
 // caller must lock
 func (nc *Conn) getServers(implicitOnly bool) []string {
 	poolSize := len(nc.srvPool)
-	servers := make([]string, 0)
+	var servers = make([]string, 0)
 	for i := 0; i < poolSize; i++ {
 		if implicitOnly && !nc.srvPool[i].isImplicit {
 			continue
--- a/nats.go
+++ b/nats.go
@@ -2507,6 +2507,19 @@ func (nc *Conn) IsConnected() bool {
 	return nc.isConnected()
 }
 
+// Servers returns the list of known servers, including additional
+// servers discovered after a connection has been established.
+func (nc *Conn) Servers() []string {
+	nc.mu.Lock()
+	defer nc.mu.Unlock()
+	poolSize := len(nc.srvPool)
+	servers := make([]string, poolSize)
+	for i := 0; i < poolSize; i++ {
+		servers[i] = nc.srvPool[i].url.String()
+	}
+	return servers
+}
+
 // Status returns the current state of the connection.
 func (nc *Conn) Status() Status {
 	nc.mu.Lock()
--- a/nats.go
+++ b/nats.go
@@ -2508,9 +2508,7 @@ func (nc *Conn) IsConnected() bool {
 }
 
 // Servers returns the list of known server urls, including additional
-// servers discovered after a connection has been established.  URLs passed
-// into NATS through the API be preserved in their original form, but any
-// credentials in a URL will never be visible outside of the application.
+// servers discovered after a connection has been established.
 func (nc *Conn) Servers() []string {
 	nc.mu.Lock()
 	defer nc.mu.Unlock()
--- a/nats.go
+++ b/nats.go
@@ -263,6 +263,7 @@ type srv struct {
 	didConnect  bool
 	reconnects  int
 	lastAttempt time.Time
+	isImplicit  bool
 }
 
 type serverInfo struct {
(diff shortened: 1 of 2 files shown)


nats.go
```diff
@@ -2507,6 +2507,19 @@ func (nc *Conn) IsConnected() bool {
 	return nc.isConnected()
 }
 
+// Servers returns the list of known servers, including additional
```
derekcollison: Do we include credentials?

ColinSullivan1: No, just what the server returns - hostname and port.

ColinSullivan1: Hold on - let me test to see if the local user provided credentials are returned.

kozlovic: Well, not really, we return all in srvPool, right? So there can be the one that the user added through Options.Servers

kozlovic: Question to @derekcollison what difference it makes? The user adds server urls (with credentials) in Options.Servers. The user can call Options.Servers and will see those. The Conn.Servers returns Options.Servers + the one that it got from INFO. *Those* don't contain credentials.

ColinSullivan1: When the user/pass is passed locally in the options, then the credentials are included.  Updates from an async INFO will not have any credentials.  Shall I strip them out?  (The application can read the credentials elsewhere)

derekcollison: Maybe we should tag servers in a structure that denotes whether they were explicit or implicit? We should be sensitive to credentials IMO. 

How do we handle applying credentials to the implicit, sever delivered, information?

derekcollison: @kozlovic I hear what you are saying. Maybe if we mark type, and document that only explicit servers (pre-configured) can contain credentials that might suffice. We would want to be consistent across all clients we support.

kozlovic: For the implicit servers, we have introduced `nats.UserInfo(string, string)` and `nas.Token(string)` to set credentials exactly for this purpose, that is, servers for which we don't have credentials.
From the README.md, Clustered Usage:
```
// When connecting to a mesh of servers with auto-discovery capabilities,
// you may need to provide a username/password or token in order to connect
// to any server in that mesh when authentication is required.
// Instead of providing the credentials in the initial URL, you will use
// new option setters:
nc, err = nats.Connect("nats://localhost:4222", nats.UserInfo("foo", "bar"))

// For token based authentication:
nc, err = nats.Connect("nats://localhost:4222", nats.Token("S3cretT0ken"))

// You can even pass the two at the same time in case one of the server
// in the mesh requires token instead of user name and password.
nc, err = nats.Connect("nats://localhost:4222",
    nats.UserInfo("foo", "bar"),
    nats.Token("S3cretT0ken"))

// Note that if credentials are specified in the initial URLs, they take
// precedence on the credentials specfied through the options.
// For instance, in the connect call below, the client library will use
// the user "my" and password "pwd" to connect to locahost:4222, however,
// it will use username "foo" and password "bar" when (re)connecting to
// a different server URL that it got as part of the auto-discovery.
nc, err = nats.Connect("nats://my:pwd@localhost:4222", nats.UserInfo("foo", "bar"))
```

ColinSullivan1: To handle the implicit case, there are new options to set credentials to be used when none are specified in the url.  Would we have to mark?  Could this just be documented, where we say something like _"Local credentials will be included if specified by the application in a connection url, but credentials will never be included in server updated urls."_

ColinSullivan1: @derekcollison Is it OK to return urls as they are (I can update the documentation to say user passed credentials will be present), or should we tag/filter?

derekcollison: If they pass it in to us, yes I think its ok. We hold onto it for auto-reconnect.

nats.go
```diff
@@ -2507,8 +2507,10 @@ func (nc *Conn) IsConnected() bool {
 	return nc.isConnected()
 }
 
-// Servers returns the list of known servers, including additional
-// servers discovered after a connection has been established.
+// Servers returns the list of known server urls, including additional
+// servers discovered after a connection has been established.  URLs passed
+// into NATS through the API be preserved in their original form, but any
+// credentials in a URL will never be visible outside of the application.
```
derekcollison: Why do we say this? If we return it as part of URL?

ColinSullivan1: Yes.  Or should we leave that out?

derekcollison: I would leave that comment out. Did we decide if we wanted to mark which ones were explicit vs implicit via wrapping in a struct? @kozlovic what do you think?

kozlovic: I think that if the user specifies the URLs with credentials through Options.Servers (or the connect URL), I don't see a problem returning those URLs with credentials (as they were entered). The one discovered will not have credentials. So I don't see the need to distinguish between explicit vs implicit.

kozlovic: That being said, it would be very easy to return: "nats://host:port" for all urls in the server pool. The thing I don't know is what is the use case of this new API. If the user is going to use those to issue a connect, then getting the full URL (possibly with credentials) will help. The user can use UserInfo() and/or Token() to help with the URLs that would not have credentials.

ColinSullivan1: The use case is to provide information, and one could use the results of this when creating a new connection for connection balancing.

derekcollison: If there is a way to get userInfo, then servers should IMO return plain urls.. nats://host:port.

kozlovic: Indeed, you could have something like:
```
var nc nats.Conn
urls := nc.Servers()
// Try to connect to any server
for _, url := range urls {
  if c, err := nats.Connect(url, nats.UserInfo("ivan", "pwd")); err != nil {
     continue
   } else {
     nc = c
    break
   }
}
```
But the issue is if the credentials are not the same for some of the servers entered explicitly, than the one provided through `UserInfo`, then you would lose the benefit of explicit credentials for explicit servers. Again, this is a bit convoluted, since I don't know how you would manage to have different credentials for a given use in a cluster, but...

nats.go
```diff
@@ -678,7 +679,7 @@ func (nc *Conn) setupServerPool() error {
 
 	// Place default URL if pool is empty.
 	if len(nc.srvPool) <= 0 {
-		if err := nc.addURLToPool(DefaultURL); err != nil {
+		if err := nc.addURLToPool(DefaultURL, false); err != nil {
```
kozlovic: Technically, this one is implicit since we add to the pool if the pool is empty at this point (based on Options.Servers). But agree that in this context we consider "implicit" only the ones returned by the server's INFO protocol messages.

kozlovic:
LGTM

kozlovic:
@ColinSullivan1 Unfortunately, the merge of #219 caused conflicts with your PR. You will need to resolve.

kozlovic:
@wallyqs Could you have a look to this one too? Thanks!

kozlovic:
@wallyqs Not sure if the thumbs up was that you will review, or for a LGTM?

wallyqs:
LGTM

nats.go
```diff
@@ -2509,6 +2510,39 @@ func (nc *Conn) IsConnected() bool {
...
+// servers discovered after a connection has been established.  If
+// authentication is enabled, use UserInfo or Token when connecting with
+// these urls.
+func (nc *Conn) Servers() []string {
+	nc.mu.Lock()
+	defer nc.mu.Unlock()
+	return nc.getServers(false)
+}
+
+// ImplicitServers returns only the server urls that have been discovered
```
ColinSullivan1: That sounds good to me, and might make more sense to the user.  Any objections to changing it to `DiscoveredServers`?

kozlovic: 👍

wallyqs: here is a patch passing tests in case we want to do the rename :) https://github.com/wallyqs/nats/commit/769ac302fd292fb7b47b6e4dc762d5f11e56bb9b

ColinSullivan1:
@derekcollison Based on recent comments, I renamed the API from ImplicitServers to DiscoveredServers.  Does that work?

PR #216, opened 2016-08-31
Add `ConnectedHost` to check current connected server

wallyqs:
This can be helpful when only wanting to check against which server endpoint we are connected/reconnected as `ConnectedUrl` could include the credentials in the url string.

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -837,7 +837,7 @@ func (nc *Conn) ConnectedHost() string {
 	if nc.status != CONNECTED {
 		return _EMPTY_
 	}
-	return fmt.Sprintf("nats://%s", nc.url.Host)
+	return fmt.Sprintf("%s://%s", nc.url.Scheme, nc.url.Host)
 }
 
 // ConnectedServerId reports the connected server's Id
--- a/nats.go
+++ b/nats.go
@@ -814,7 +814,7 @@ func (nc *Conn) spinUpGoRoutines() {
 	nc.mu.Unlock()
 }
 
-// Report the connected server's Url
+// ConnectedUrl reports the connected server's Url
 func (nc *Conn) ConnectedUrl() string {
 	if nc == nil {
 		return _EMPTY_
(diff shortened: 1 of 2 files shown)


nats.go
```diff
@@ -827,7 +827,20 @@ func (nc *Conn) ConnectedUrl() string {
...
+func (nc *Conn) ConnectedHost() string {
+	if nc == nil {
+		return _EMPTY_
+	}
+	nc.mu.Lock()
+	defer nc.mu.Unlock()
+	if nc.status != CONNECTED {
+		return _EMPTY_
+	}
+	return nc.url.Host
```
kozlovic: To be consistent with the other APIs where we return an URL including scheme, you should maybe include it here.

test/auth_test.go
```diff
@@ -45,11 +45,27 @@ func TestAuth(t *testing.T) {
...
+	// Confirm credentials are part of the url
+	got := nc.ConnectedUrl()
+	expected := "nats://derek:foo@localhost:8232"
+	if got != expected {
+		t.Errorf("Expected url as %s got: %s", expected, got)
+	}
+
+	// Confirm we can get the connected host without credentials as well
+	got = nc.ConnectedHost()
+	expected = "localhost:8232"
```
kozlovic: Should we include the scheme? (`nats://`)

kozlovic:
@ColinSullivan1 is adding `Servers()` and `ImplicitServers()` and in those we are returning URLs - some without credentials - but in the form: `nats://host:port`, should you do the same here? Otherwise, LGTM.

nats.go
wallyqs: sounds good to me, just modified 👍

wallyqs:
Just updated, thanks!

kozlovic:
LGTM

nats.go
```diff
@@ -827,7 +827,20 @@ func (nc *Conn) ConnectedUrl() string {
...
+func (nc *Conn) ConnectedHost() string {
+	if nc == nil {
+		return _EMPTY_
+	}
+	nc.mu.Lock()
+	defer nc.mu.Unlock()
+	if nc.status != CONNECTED {
+		return _EMPTY_
+	}
+	return fmt.Sprintf("%s://%s", nc.url.Scheme, nc.url.Host)
```
wallyqs: Updated here to use the same scheme as the url here as in other apis (should be `nats` most likely)

kozlovic: Or "tls://" ;-). Colin did that on the other PR too.

ColinSullivan1:
LGTM

derekcollison:
Already have this..

```
func (nc *Conn) ConnectedUrl() string
```

wallyqs:
yes, only small change is that `ConnectedUrl` includes the credentials in the string if present and the `ConnectedHost` method here does not, thought it would be helpful to have an option which shows only the host in case of logging to which server it was connected or client reconnected for example.

derekcollison:
I think there is too much overlap IMO. Not needed.

PR #220, opened 2016-09-08
Removed dead code from previous iteration.

aricart:
Previous version of the code formatted differently. Some dead logic remained behind.
@mcqueary @kozlovic

### DIFF
--- a/bench/bench.go
+++ b/bench/bench.go
@@ -254,16 +254,16 @@ func (bm *Benchmark) Report() string {
 	var buffer bytes.Buffer
 
 	indent := ""
+	if !bm.Pubs.HasSamples() && !bm.Subs.HasSamples() {
+		return "No publisher or subscribers. Nothing to report."
+	}
+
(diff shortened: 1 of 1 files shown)


bench/bench.go
```diff
@@ -254,16 +254,16 @@ func (bm *Benchmark) Report() string {
 	var buffer bytes.Buffer
 
 	indent := ""
+	if !bm.Pubs.HasSamples() && !bm.Subs.HasSamples() {
+		return "No publisher or subscribers. Nothing to report."
```
ColinSullivan1: When could this happen?  If it'd be based on a configuration issue, that should be detected up front?  Perhaps "Error:  No publisher or subscriber samples obtained."?

aricart: yes if the tool didn't check the args, and simply passes them. This will catch it. In the case the client code didn't perform the check, the op still runs, and the message at least will let them know.

kozlovic:
LGTM

ColinSullivan1:
LGTM

PR #228, opened 2016-10-07
examples: go-bench should pass random data

arekkas:
Benchmarks are more impressive with random data!

### DIFF
--- a/examples/nats-bench.go
+++ b/examples/nats-bench.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/nats-io/nats"
 	"github.com/nats-io/nats/bench"
+	"crypto/rand"
 )
 
 // Some sane defaults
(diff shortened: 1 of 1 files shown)


examples/nats-bench.go
```diff
@@ -102,14 +103,11 @@ func runPublisher(startwg, donewg *sync.WaitGroup, opts nats.Options, numMsgs in
 
 	args := flag.Args()
 	subj := args[0]
-	var msg []byte
-	if msgSize > 0 {
```
arekkas: this doesn't make sense as it's the long version of: `var msg = make([]byte, 0)`

arekkas:
Not sure why tests fail on Go 1.6 - works locally.

aricart:
One thing with the random bytes is that while that sounds useful to simulate performance, effectively we are inserting an 'application' delay to generate the bytes, as every message will now have to generate some random number of bytes.

In the bench we are trying to measure the number of bytes/messages that the messaging system can pass through. Additional delays are understood, but not the goal of the bench.

arekkas:
That is very much possible, unfortunately I don't know the internals of the benchmarking tool used.

kozlovic:
I second @aricart comments. Creating a slice in that loop with random data will be really costly and skew the results of the messaging layer that we are benchmarking here. Those slices will put pressure on GC and aren't you going to deplete the entropy pool doing that in a tight loop?

arekkas:
If benchmark includes this loop it doesn't make sense, I thought benchmarking would happen on flush only.

kozlovic:
@arekkas I see. No, the benchmark if measuring all line of codes before the `for` loop (notice the capture of start time). We *could* make ONE message before the capture of the start with a slice of random data, but again, not sure in what this is useful for the purpose of the benchmark.

aricart:
Effectively the publisher is making a byte array of the specified size. So whether random or not, that number of bytes are sent...

PR #229, opened 2016-10-07
[FIX #227] Added check to insure message count is greater than zero

### LINKED ISSUE #227: nats-bench panics when message amount is 0
```
go run nats-bench.go -n 0 msg.test
panic: runtime error: index out of range

goroutine 1 [running]:
panic(0x5f9940, 0xc042008070)
        C:/Go/src/runtime/panic.go:500 +0x1af
main.main()
        D:/workspace/go/src/github.com/nats-io/nats/examples/nats-bench.go:76 +0x68e
exit status 2
```

aricart:


kozlovic:
LGTM

examples/nats-bench.go
```diff
@@ -47,6 +47,10 @@ func main() {
 		usage()
 	}
 
+	if *numMsgs < 1 {
```
derekcollison: <=0?

derekcollison: Does it have to be >0?

aricart: Yes - test is to abort run if not 1 or more.

examples/nats-bench.go
```diff
@@ -47,6 +47,10 @@ func main() {
 		usage()
 	}
 
+	if *numMsgs <= 0 {
```
zquestz: NIT: For what it is worth. I find *numMsgs < 1 much easier to read.

PR #231, opened 2016-10-20
[FIXED] assignment copies lock value for crypto/tls.Config

kozlovic:
Running `go vet ./...` with `go 1.7.3` would report the following:

```
nats.go:769: assignment copies lock value to tlsCopy: crypto/tls.Config contains sync.Once contains sync.Mutex
```

Add a “clone” function while waiting for this to be addressed
by the language itself (https://go-review.googlesource.com/#/c/28075/)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: go
 sudo: false
 go:
-- 1.5
-- 1.6
+- 1.6.3
+- 1.7.3
 install:
 - go get -t ./...
 - go get github.com/nats-io/gnatsd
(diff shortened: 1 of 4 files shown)


kozlovic:
@derekcollison Alternatively, I could have created 2 files with `+build go1.5,!go1.7` in one and `+build go1.7` on the other and so not using reflection. Would that be better? (If so, I could update the server with same change).

derekcollison:
Might be good in case others build on 1.5.x.

kozlovic:
@derekcollison Ok, used 2 files instead of reflection. Keep in mind that we will remove those when `go 1.8` comes out (or sooner).

PR #232, opened 2016-10-26
Add staticcheck static analyzer to Travis

y0ssar1an:
[`staticcheck`](https://github.com/dominikh/go-staticcheck) is a popular static analyzer. It's a free bug finder that catches some things that pass `go vet`.

PR #233, opened 2016-10-26
Add staticcheck static analyzer to travis

y0ssar1an:
[`staticcheck`](https://github.com/dominikh/go-staticcheck) is a popular static analyzer. It's a free bug finder that catches some things that pass `go vet`.

This is a resubmit of #232. I force-pushed to that fork, which apparently makes it impossible to reopen the PR.


kozlovic:
@y0ssar1an It is working, it's just that staticcheck founds an error that you did not address: 
```
/home/travis/gopath/src/github.com/nats-io/nats/test/netchan_test.go:335:2: should check returned error before deferring ec.Close()
```
That being said, if we set `staticcheck` in Travis, we would do that for all nats-io Go repos. The problem is that we have some that are now vendoring third party repos. If it finds error in those, this would make the Travis build to fail, which would block us. I suggest that we do not add this to Travis until we can make a decision on how to handle that. As a side note, I am now using this locally and thanks for the tip.

So instead of adding it to Travis, do you want to post a PR with the resolution of the two issues found? I can do it if you don't want or don't have time.
Thanks again!

kozlovic:
@y0ssar1an Or let me ask you the question: what do you do if you use this tool and have vendor directory and the tool finds possible issues (sometimes benign like unnecessary break in a case statement)? You may not be able to have the vendor update/fix their code quickly. Yet, this is making your Travis build fail. What do you do then?

kozlovic:
@y0ssar1an After internal discussions, we have decided to include this into the nats-io go repos. Would you mind fixing the remaining issue with that PR or I can make a PR out of the branch that I have created, leaving up to you.

y0ssar1an:
Thank you for the catch! Strangely, when I ran `staticcheck ./...` on my local machine, the repo was 100% clean.

The way to avoid scanning the `vendor/` directory is with this one-liner.
```sh
staticcheck $(go list ./... | grep -v /vendor/)
```
I got that tip from [Andrew Gerrand](https://github.com/nf). This also works with `go vet`, `golint`, etc.

kozlovic:
LGTM, and thanks for the vendor exclusion tip!

PR #235, opened 2016-11-01
Add SetSlowConsumerTimeout for channel based subscriptions

wallyqs:
Currently `ChanSubscribe` and `ChanQueueSubscribe` APIs are prone to slow consumer errors on burst of messages when a channel is either unbuffered or has a small buffer, leading to received messages become dropped and not processing them.

This change adds a new `SetSlowConsumerTimeout` API that allows the client to set an opt-in _slow consumer timeout_ on its channel to customize how long to wait for a message to be consumed via the channel before dropping them and triggering the async error callback.

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -243,6 +243,9 @@ type Subscription struct {
 	pMsgsLimit  int
 	pBytesLimit int
 	dropped     int
+
+	// SlowConsumer timeout for channel based subscriptions.
+	sct time.Duration
 }
 
 // Msg is a structure used by Subscribers and PublishMsg().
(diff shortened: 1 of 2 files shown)


nats.go
```diff
@@ -2266,6 +2283,14 @@ func (s *Subscription) SetPendingLimits(msgLimit, bytesLimit int) error {
 	return nil
 }
 
+// SetSlowConsumerTimeout sets the deadline to wait before dropping mesages
+// when mode of delivery for subscription is a channel.
+func (s *Subscription) SetSlowConsumerTimeout(timeout time.Duration) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.sct = timeout
```
ColinSullivan1: Should there be checks for validity and to see if this is a channel subscription, as in `SetPendingLimits`?

```
 	if s.conn == nil {
 		return ErrBadSubscription
 	}
 	if s.typ != ChanSubscription {
 		return ErrTypeSubscription
 	}
```

kozlovic:
@wallyqs The danger with this approach is that it is going to increase the risk of the server closing the connection as a slow consumer since you are blocking the routine dequeuing from the socket.

If the user specifies a big enough timeout and does not have its channel ready, and if messages are sent to the server, the `readLoop` won't dequeue data since it will be waiting in that `select`, causing the server to drop the connection. 

Overall, I think that the risk out-weight the benefit.

wallyqs:
@ColinSullivan1  @kozlovic thanks for checking.  I agree with the feedback and think it needs some rethought as this blocks the reading loop so affects the ping interval even... I will close for now.

PR #237, opened 2016-11-15
Static check build fixes

wallyqs:
Addresses the nits from staticcheck which show up in the build

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -1445,7 +1445,7 @@ func (nc *Conn) waitForMsgs(s *Subscription) {
 		}
 
 		// Deliver the message.
-		if m != nil && (max <= 0 || delivered <= max) {
+		if m != nil && (max == 0 || delivered <= max) {
 			mcb(m)
 		}
 		// If we have hit the max for delivered msgs, remove sub.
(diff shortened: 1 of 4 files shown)


test/conn_test.go
```diff
@@ -375,7 +375,7 @@ func TestErrOnConnectAndDeadlock(t *testing.T) {
 		nc, err := nats.Connect(natsURL)
 		if err == nil {
 			nc.Close()
-			t.Fatal("Expected bad INFO err, got none")
```
derekcollison: Again I think we want fatal here.

nats.go
```diff
@@ -1445,7 +1445,7 @@ func (nc *Conn) waitForMsgs(s *Subscription) {
 		}
 
 		// Deliver the message.
-		if m != nil && (max <= 0 || delivered <= max) {
```
derekcollison: What if max < 0 ?

wallyqs: This addresses following staticcheck report `github.com/nats-io/nats/nats.go:1448:19: 'x <= 0' for unsigned values of x is the same as 'x == 0'`. Adding a `max < 0` reports that `unsigned values are never < 0` too...

wallyqs:
Closing as needs to address first with proper fix for usage of T.Fatalf in goroutine error (`the goroutine calls T.Fatalf, which must be called in the same goroutine as the test` issue)

nats.go
derekcollison: Gotcha..

PR #252, opened 2017-01-14
Add FlusherTimeout to prevent blocking for long on unhealthy connections

wallyqs:
This adds a customizable `FlusherTimeout` option for the flusher loop to be able to yield after a certain time in case a write to the underlying writer takes too long due to an unhealthy connection.

This prevents the client blocking for an indefinite amount of time until reaching an OS level `i/o timeout` since the `bufio.Flush` is happening while the lock is being hold, thus giving room for the ping interval to be able to `processOpErr` and disconnect once it reaches the max amount of missing ping replies from the server (with an unhealthy connection the reading loop would be stuck on `conn.Read` too because there is no deadline either, so it would take as a long as the ping interval check closes the connection or `i/o timeout`).

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -60,6 +60,7 @@ var (
 	ErrSlowConsumer         = errors.New("nats: slow consumer, messages dropped")
 	ErrTimeout              = errors.New("nats: timeout")
 	ErrBadTimeout           = errors.New("nats: timeout invalid")
+	ErrFlusherTimeout       = errors.New("nats: flusher timeout")
 	ErrAuthorization        = errors.New("nats: authorization violation")
 	ErrNoServers            = errors.New("nats: no servers available for connection")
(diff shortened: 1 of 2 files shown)


nats.go
```diff
@@ -1598,6 +1605,12 @@ func (nc *Conn) flusher() {
 			return
 		}
 		if bw.Buffered() > 0 {
+			// Allow customizing how long we should wait for a flush to be done
+			// to prevent unhealthy connections blocking the client for too long.
+			if flusherTimeout > 0 {
+				conn.SetWriteDeadline(time.Now().Add(flusherTimeout))
```
derekcollison: Where do we clear it?

wallyqs: Thanks for checking, I added resetting the WriteDeadline on successful flush now

PR #258, opened 2017-01-26
[FIXED] Do not shuffle entire pool when adding URL from INFO

kozlovic:
When connecting to a server, the received INFO may contain an array
of URLs. The current code was adding those and then shuffeling the
entire pool (if NoRandomize is false). This is wrong since the
connect (and reconnect) loop would then potentially skip never
tried servers and try again servers that had just failed.


kozlovic:
Closing this PR for now due to constant random failures of various tests (related mostly to TLS), which I believe are not related to this. Will resubmit later when/if I find what's the issue.

PR #274, opened 2017-04-06
wip: Adding Context aware APIs to the client

wallyqs:
This follows from the idea by @colinsullivan1 at https://github.com/nats-io/go-nats/issues/221#issuecomment-288129861 and needs another pass but opening wip PR already for early comments.

Currently it supports the following 2 APIs (plus an extra one):

```
func (nc *Conn) RequestWithContext(ctx context.Context, subj string, data []byte) (*Msg, error)
func (s *Subscription) NextMsgWithContext(ctx context.Context) (*Msg, error)
+ func (s *Subscription) SetContext(ctx context.Context)
```

Since `"context"` package is supported only after Go1.7, I added a `+build go1.7` tag just in case so that it does not break if using older versions of the runtime plus a small `cntxt` interface as part of the `Subscription` state which also implements `Context` interface, this way in those clients it just does not get used unless using more recent versions of Go.

### DIFF
--- a/context.go
+++ b/context.go
@@ -1,0 +1,60 @@
+// Copyright 2012-2017 Apcera Inc. All rights reserved.
+
+// +build go1.7
+
+// A Go client for the NATS messaging system (https://nats.io).
+package nats
+
+import "context"
+
+// RequestWithContext takes a context, a subject and payload in bytes
+// and request expecting a single response.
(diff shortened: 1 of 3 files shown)


wallyqs:
Closing in favor of https://github.com/nats-io/go-nats/pull/275

PR #275, opened 2017-04-10
[ADDED] context.Context support

### LINKED ISSUE #221: context.Context integration
Hi. 

In go 1.7 it was introduced context.Context interface. Are you going to support it in client?

wallyqs:
Adds `context.Context` support to the +Go1.7 clients implementing the APIs suggested at https://github.com/nats-io/go-nats/issues/221#issuecomment-288193653

```go
func (c *EncodedConn) RequestWithContext(ctx context.Context, subject string, v interface{}, vPtr interface{}) error
func (nc *Conn) RequestWithContext(ctx context.Context, subj string, data []byte) (*Msg, error)
func (s *Subscription) NextMsgWithContext(ctx context.Context) (*Msg, error)
```

Fixes #221

### DIFF
--- a/context.go
+++ b/context.go
@@ -1,0 +1,175 @@
+// Copyright 2012-2017 Apcera Inc. All rights reserved.
+
+// +build go1.7
+
+// A Go client for the NATS messaging system (https://nats.io).
+package nats
+
+import (
+	"context"
+	"reflect"
+)
+
+// RequestWithContext takes a context, a subject and payload
+// in bytes and request expecting a single response.
+func (nc *Conn) RequestWithContext(ctx context.Context, subj string, data []byte) (*Msg, error) {
+	if ctx == nil {
+		panic("nil context")
+	}
+	inbox := NewInbox()
+	ch := make(chan *Msg, RequestChanLen)
+
+	s, err := nc.subscribe(inbox, _EMPTY_, nil, ch)
+	if err != nil {
+		// If we errored here but context has been canceled already
+		// then return the error from the context instead.
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		default:
+		}
+
+		return nil, err
+	}
+	s.AutoUnsubscribe(1)
+	defer s.Unsubscribe()
+
+	err = nc.PublishRequest(subj, inbox, data)
+	if err != nil {
+		// Use error from context instead if already canceled.
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		default:
+		}
+
+		return nil, err
+	}
+
+	return s.NextMsgWithContext(ctx)
+}
+
+// NextMsgWithContext takes a context and returns the next message available
+// to a synchronous subscriber, blocking until either one is available or
+// context gets canceled.
+func (s *Subscription) NextMsgWithContext(ctx context.Context) (*Msg, error) {
+	if ctx == nil {
+		panic("nil context")
+	}
+	if s == nil {
+		return nil, ErrBadSubscription
+	}
+
+	// Add validateNextMsgState func and reuse
+	s.mu.Lock()
+	if s.connClosed {
+		s.mu.Unlock()
+		return nil, ErrConnectionClosed
+	}
+	if s.mch == nil {
+		if s.max > 0 && s.delivered >= s.max {
+			s.mu.Unlock()
+			return nil, ErrMaxMessages
+		} else if s.closed {
+			s.mu.Unlock()
+			return nil, ErrBadSubscription
+		}
+	}
+	if s.mcb != nil {
+		s.mu.Unlock()
+		return nil, ErrSyncSubRequired
+	}
+	if s.sc {
+		s.sc = false
+		s.mu.Unlock()
+		return nil, ErrSlowConsumer
+	}
+
+	// snapshot
+	nc := s.conn
+	mch := s.mch
+	max := s.max
+	s.mu.Unlock()
+
+	var ok bool
+	var msg *Msg
+
+	select {
+	case msg, ok = <-mch:
+		if !ok {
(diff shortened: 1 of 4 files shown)


context.go
```diff
@@ -0,0 +1,175 @@
...
+import (
+	"context"
+	"reflect"
+)
+
+// RequestWithContext takes a context, a subject and payload
+// in bytes and request expecting a single response.
+func (nc *Conn) RequestWithContext(ctx context.Context, subj string, data []byte) (*Msg, error) {
+	if ctx == nil {
+		panic("nil context")
```
derekcollison: In general I don't think we have panics in the client code. We should either error or ignore a nil context.

context.go
```diff
@@ -0,0 +1,175 @@
...
+	ch := make(chan *Msg, RequestChanLen)
+
+	s, err := nc.subscribe(inbox, _EMPTY_, nil, ch)
+	if err != nil {
+		// If we errored here but context has been canceled already
+		// then return the error from the context instead.
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		default:
```
derekcollison: should return statement be under default?

context.go
```diff
@@ -0,0 +1,175 @@
...
+	s.AutoUnsubscribe(1)
+	defer s.Unsubscribe()
+
+	err = nc.PublishRequest(subj, inbox, data)
+	if err != nil {
+		// Use error from context instead if already canceled.
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		default:
```
derekcollison: Same here.

context.go
```diff
@@ -0,0 +1,175 @@
...
+
+	return s.NextMsgWithContext(ctx)
+}
+
+// NextMsgWithContext takes a context and returns the next message available
+// to a synchronous subscriber, blocking until either one is available or
+// context gets canceled.
+func (s *Subscription) NextMsgWithContext(ctx context.Context) (*Msg, error) {
+	if ctx == nil {
+		panic("nil context")
```
derekcollison: Same here for panic in client code.

context.go
```diff
@@ -0,0 +1,175 @@
...
+
+// RequestWithContext will create an Inbox and perform a Request
+// using the provided cancellation context with the Inbox reply
+// for the data v. A response will be decoded into the vPtrResponse.
+func (c *EncodedConn) RequestWithContext(
+	ctx context.Context,
+	subject string,
+	v interface{},
+	vPtr interface{},
+) error {
```
derekcollison: This proper formatting?

context.go
```diff
@@ -0,0 +1,175 @@
...
+	if ctx == nil {
+		panic("nil context")
```
derekcollison: Again on panics.

context.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+	b, err := c.Enc.Encode(subject, v)
+	if err != nil {
+		// Use error from context instead if already canceled.
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
```
derekcollison: Return err as part of default.

test/context_test.go
```diff
@@ -0,0 +1,380 @@
...
+
+	// Cancel the context already so that rest of requests fail.
+	cancelCB()
+
+	_, err = nc.RequestWithContext(ctx, "fast", []byte("world"))
+	if err == nil {
+		t.Fatalf("Expected request with timeout context to fail: %s", err)
+	}
+
+	// Reported error is "context canceled" from Context package,
```
derekcollison: Cancelled vs canceled, I think both are valid, but we should be consistent in spelling.

test/context_test.go
```diff
@@ -0,0 +1,380 @@
...
+	// which is not a timeout error.
+	type timeoutError interface {
+		Timeout() bool
+	}
+	if _, ok := err.(timeoutError); ok {
+		t.Errorf("Expected to not have a timeout error")
+	}
+	expected = `context canceled`
```
derekcollison: Same here.

test/context_test.go
wallyqs: Thanks Derek, I changed throughout to use `canceled` which is the one used in the go repo:

https://github.com/golang/go/blob/dc6af19ff8b44e56abc1217af27fe098c78c932b/src/context/context.go#L155

context.go
wallyqs: Thanks, fixed formatting here...

context.go
wallyqs: I agree on that, so for now I have removed the checks to not introduce calling `panic` in the nats client either.

Here I was borrowing from the `http`, `net` and other libraries where `panic` is being called...

- https://github.com/golang/go/blob/3792db518327c685da17ca6c6faa4e1d2da4c33c/src/net/dial.go#L316-L319
- https://github.com/golang/go/blob/094498c9a13cd711ed45a65b153393eb8ae1566b/src/net/http/request.go#L323-L334
- https://github.com/go-redis/redis/blob/f33571c93c72d3c4717a7b3c0bce95136496f772/redis_context.go#L28-L32

...though in other packages like  `sql` package those checks aren't introduced either so switched to follow that style instead:

- https://github.com/golang/go/blob/master/src/database/sql/sql.go#L1097-L1118

derekcollison:
LGTM

context.go
kozlovic: @wallyqs You have removed the panic on nil context, but is it not going to panic any way when dereferencing it in the code? (unless context's method themselves do nothing if context is nil).

wallyqs: yes if `nil` it would panic when calling `ctx.Done()` anyway as you mention

kozlovic:
LGTM

PR #276, opened 2017-04-13
Add output to example_test.go

wallyqs:
Noticed that we are missing the `Output` comments section in the examples from `example_test`, so this PR includes them in order to run these examples as part of the build.

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -28,6 +28,9 @@ func ExampleConnect() {
 
 	nc, _ = opts.Connect()
 	nc.Close()
+
+	// Output:
+	//
 }
 
 // This Example shows an asynchronous subscriber.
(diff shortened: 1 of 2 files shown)


kozlovic:
No big deal, but there are 2 examples that use 1 second for the NextMsg() timeout knowing that the call is going to timeout. It's 2 seconds more for the test suite to pass, again no big deal, but do you think we could reduce this to a low value? We could put a comment saying that we use a small timeout on purpose since we know we are not going to get any message anyway.

kozlovic:
LGTM

wallyqs:
thanks @kozlovic ok updated to shorten the timeouts in those.

kozlovic:
@wallyqs Saw that, thanks, and LGTM.

wallyqs:
Closing this one for now, might revisit after fixing conflicts.

PR #279, opened 2017-04-21
Update tests following auth changes in gnatsd

kozlovic:
Also added misspell check and updated Go versions in .travis.yml

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,20 +1,22 @@
 language: go
 sudo: false
 go:
-- 1.8
+- 1.8.1
 - 1.6.4
-- 1.7.4
+- 1.7.5
 install:
 - go get -t ./...
 - go get github.com/nats-io/gnatsd
 - go get github.com/mattn/goveralls
 - go get github.com/wadey/gocovmerge
-- go get honnef.co/go/tools/cmd/staticcheck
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -1,20 +1,22 @@
...
 install:
 - go get -t ./...
 - go get github.com/nats-io/gnatsd
 - go get github.com/mattn/goveralls
 - go get github.com/wadey/gocovmerge
-- go get honnef.co/go/tools/cmd/staticcheck
-- go get honnef.co/go/tools/cmd/gosimple
+- go get -u honnef.co/go/tools/cmd/staticcheck
+- go get -u honnef.co/go/tools/cmd/gosimple
+- go get -u github.com/client9/misspell/cmd/misspell
```
derekcollison: Should add this to gnatsd too! ;)

kozlovic: Will do

derekcollison: Already on it, and removed need for the ignore clause for staticcheck. PR forthcoming.

kozlovic: Ok, just notice that misspell needs to be invoked with `.` not `./...` otherwise it is not going recursively into directories. It may then find issues in vendor, in which case we will have to explicitly specify all our directories, etc..

PR #285, opened 2017-04-24
Reduce risk of server closing conn (slow consumer) if sending to itself

kozlovic:
When a client sends data and has at least a matching subscriber,
there may be a case where the flusher is holding the connection's
lock while trying to flush the write buffer, which cannot complete
because the server is not draining data at the moment since it is
itself trying to send data to the client, which processMsg() function
is waiting on the connection's lock. After 2 seconds, the server
will close the client connection.
Using a dedicated lock for the subscriptions map that processMsg
needs to access is reducing the risk of lock contention between
publish/flusher and the readloop's processMsg function.
There would still be a case where if the subscription in the client
is considered a slow consumer and messages are dropped, we need to
acquire the connection lock there. But in cases where subscribers
keep up, we would reduce the risk of server's slow consumer error.

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -241,8 +241,6 @@ type Conn struct {
 	// struct and make sure they are all 64bits (or use padding if necessary).
 	// atomic.* functions crash on 32bit machines if operand is not aligned
 	// at 64bit. See https://github.com/golang/go/issues/599
-	ssid int64
-
 	Statistics
 	mu      sync.Mutex
 	Opts    Options
@@ -255,6 +253,8 @@ type Conn struct {
 	pending *bytes.Buffer
 	fch     chan bool
 	info    serverInfo
+	ssid    int64
+	subsMu  sync.RWMutex
 	subs    map[int64]*Subscription
 	mch     chan *Msg
 	ach     chan asyncCB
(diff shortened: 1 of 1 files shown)


kozlovic:
@derekcollison The main goal is to avoid the server cutting off the connection (slow consumer), but the side effect is an improvement in performance for the PubSub using same connection:
```
Ivans-MacBook-Pro:go-nats ivan$ benchcmp old.txt new.txt 
benchmark                                     old ns/op     new ns/op     delta
BenchmarkPublishSpeed-8                       125           129           +3.20%
BenchmarkPubSubSpeed-8                        1495          579           -61.27%
BenchmarkAsyncSubscriptionCreationSpeed-8     28716         29348         +2.20%
BenchmarkSyncSubscriptionCreationSpeed-8      43834         33926         -22.60%
BenchmarkInboxCreation-8                      122           123           +0.82%
BenchmarkRequest-8                            67531         67824         +0.43%
BenchmarkPublishSpeedViaChan-8                1667          1697          +1.80%

benchmark                                     old allocs     new allocs     delta
BenchmarkAsyncSubscriptionCreationSpeed-8     14             14             +0.00%
BenchmarkSyncSubscriptionCreationSpeed-8      13             13             +0.00%
BenchmarkRequest-8                            58             58             +0.00%

benchmark                                     old bytes     new bytes     delta
BenchmarkAsyncSubscriptionCreationSpeed-8     1326          1325          -0.08%
BenchmarkSyncSubscriptionCreationSpeed-8      66301         66299         -0.00%
BenchmarkRequest-8                            2666          2666          +0.00%
```

PR #286, opened 2017-04-28
Remove atomic usage for connection statistics

kozlovic:
Statistics are updated either under Conn.mu or Conn.subsMu. Instead
of using atomics when updating the stats and when getting them,
simply lock both Conn.mu and Conn.subsMu in Conn.Stats().
This saves 2 atomic operations on publish and 2 on processMsg.

Related to #285

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -20,7 +20,6 @@ import (
 	"strconv"
 	"strings"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/nats-io/go-nats/util"
@@ -1333,7 +1332,7 @@ func (nc *Conn) doReconnect() {
 		}
 
 		// We are reconnected
-		atomic.AddUint64(&nc.Reconnects, 1)
+		nc.Reconnects++
 
 		// Process connect logic
 		if nc.err = nc.processConnectInit(); nc.err != nil {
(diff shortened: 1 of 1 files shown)


kozlovic:
@derekcollison Sorry I missed that in #285. There is no need for atomic operations as long as we lock the 2 mutexes when getting the stats.

nats.go
```diff
@@ -2728,13 +2727,20 @@ func (nc *Conn) isConnected() bool {
...
-	return Statistics{
-		InMsgs:     atomic.LoadUint64(&nc.InMsgs),
-		InBytes:    atomic.LoadUint64(&nc.InBytes),
-		OutMsgs:    atomic.LoadUint64(&nc.OutMsgs),
-		OutBytes:   atomic.LoadUint64(&nc.OutBytes),
-		Reconnects: atomic.LoadUint64(&nc.Reconnects),
+	// Stats are updated either under connection's mu or subsMu mutexes.
+	// Lock both to safely get them.
+	nc.mu.Lock()
+	nc.subsMu.RLock()
```
derekcollison: Why read lock here?

kozlovic: I created `subsMu` as a `sync.RWMutex` (I could change it to regular `sync.Mutex`). In this case, we are "reading" the stats, not making any modification to either stats or subs map, so read-lock is enough.

PR #289, opened 2017-05-11
[FIXED] Locking in processPermissionsViolation() and added test

kozlovic:
Connection locking was missing in processPermissionsViolation().
Added a test that verifies that async cb error is invoked on
invalid publish or subscribe, and that the connection is not
closed neither by the server nor client.

@derekcollison Opening this for discussion: was it the proper intent for the server to keep the client connection (not closing it at the moment)? If so, then it makes sense for the client also to not close it and treat this as a non fatal error. This PR then simply fixes the locking and add a test (coverage was non existent for this function).

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -1655,10 +1655,12 @@ slowConsumer:
 // processPermissionsViolation is called when the server signals a subject
 // permissions violation on either publish or subscribe.
 func (nc *Conn) processPermissionsViolation(err string) {
+	nc.mu.Lock()
 	nc.err = errors.New("nats: " + err)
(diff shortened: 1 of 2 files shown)


derekcollison:
I think the client can not assume an error is fatal and should wait for the socket to be closed to clean up.

kozlovic:
@derekcollison Ok, so for this error, the client is not closing the connection. For stale connection, it reconnects, for all others it closes. If this needs to be changed, it should be in separate PR. Right now, this one seems to be doing the right thing then.

Issue #294, opened 2017-05-23
Requests are too protocol chatty, should be optimized.


derekcollison:
Every request forms a new INBOX subscription for each request and then tears them down. Since NATS has to flood the interest graph to clusters and possible future bridged clusters, this should be redesigned to reuse a single subscription but still be semantically be the same from a client perspective. One word of caution is that we use the circuit breaker pattern such that the server auto-unsubscribes after N responses (usually 1), so if you are requesting from a large set of normal subscribers, the client may be getting a higher number of responses that will be ignored. However I feel the normal case should be a request is always sent to a queue group, which does not need the circuit breaker assist. 

Example trace from a single request / response today.

[TRC] ::1:50431 - cid:2 - ->> [SUB _INBOX.Gmw3KasVkAqYFpskoKCV79  1]
[TRC] ::1:50431 - cid:2 - ->> [UNSUB 1 1]
[TRC] ::1:50431 - cid:2 - ->> [PUB foo _INBOX.Gmw3KasVkAqYFpskoKCV79 4]
[TRC] ::1:50431 - cid:2 - ->> MSG_PAYLOAD: [help]
[TRC] ::1:50426 - cid:1 - <<- [MSG foo 1 _INBOX.Gmw3KasVkAqYFpskoKCV79 4]
[TRC] ::1:50426 - cid:1 - ->> [PUB _INBOX.Gmw3KasVkAqYFpskoKCV79 5]
[TRC] ::1:50426 - cid:1 - ->> MSG_PAYLOAD: [hello]
[TRC] ::1:50431 - cid:2 - <<- [MSG _INBOX.Gmw3KasVkAqYFpskoKCV79 1 5]
[TRC] ::1:50431 - cid:2 - <-> [DELSUB 1]

kozlovic:
Resolved with #295

PR #295, opened 2017-05-25
More protocol friendly request pattern. Fix for #294.

derekcollison:
This request pattern is sematically the same but utilizes a single
wildcard subscription and the last token context for the response
subject binding to each request. Since NATS floods interest (subscriptions),
the old requestor pattern would create a new inbox subscription for each request.
It would then auto-unsubscribe and delete the subscription after the
response was received, causing quite a bit of protocol traffic.

So although this does have a performance gain as can be seen in the benchmarks,
it's implementation is specifically designed to be more friendly for NATS
clusters.

### DIFF
--- a/context.go
+++ b/context.go
@@ -16,7 +16,58 @@ func (nc *Conn) RequestWithContext(ctx context.Context, subj string, data []byte
 	if ctx == nil {
 		return nil, ErrInvalidContext
 	}
+	if nc == nil {
+		return nil, ErrInvalidConnection
+	}
 
+	// snapshot
+	var doSetup, useOldRequestStyle bool
+	nc.mu.Lock()
+	useOldRequestStyle = nc.Opts.UseOldRequestStyle
+	doSetup = (nc.respMux == nil)
+	nc.mu.Unlock()
+
+	// If user wants the old style.
+	if useOldRequestStyle {
+		return nc.oldRequestWithContext(ctx, subj, data)
+	}
+
+	// Make sure scoped subscription is setup at least once on first
+	// call to Request(). Will handle duplicates in createRespMux.
+	if doSetup {
+		if err := nc.createRespMux(); err != nil {
+			return nil, err
+		}
+	}
+	// Create literal Inbox and map to a chan msg.
+	mch := make(chan *Msg, RequestChanLen)
+	nc.mu.Lock()
+	respInbox := nc.newRespInbox()
+	nc.respMap[respToken(respInbox)] = mch
+	nc.mu.Unlock()
+
+	err := nc.PublishRequest(subj, respInbox, data)
+	if err != nil {
+		return nil, err
+	}
+
+	var ok bool
+	var msg *Msg
+
+	select {
+	case msg, ok = <-mch:
+		if !ok {
+			return nil, ErrConnectionClosed
+		}
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	}
+
+	return msg, nil
+}
+
+// oldRequestWithContext utilizes inbox and subscription per request.
+func (nc *Conn) oldRequestWithContext(ctx context.Context, subj string, data []byte) (*Msg, error) {
 	inbox := NewInbox()
 	ch := make(chan *Msg, RequestChanLen)
 
--- a/nats.go
+++ b/nats.go
@@ -2751,9 +2751,10 @@ func (nc *Conn) clearPendingRequestCalls() {
 	if nc.respMap == nil {
 		return
 	}
-	for _, ch := range nc.respMap {
+	for key, ch := range nc.respMap {
 		if ch != nil {
 			close(ch)
+			delete(nc.respMap, key)
 		}
 	}
 }
--- a/nats.go
+++ b/nats.go
@@ -28,7 +28,7 @@ import (
 
 // Default Constants
 const (
-	Version                 = "1.2.2"
+	Version                 = "1.3.0"
 	DefaultURL              = "nats://localhost:4222"
 	DefaultPort             = 4222
 	DefaultMaxReconnect     = 60
@@ -119,72 +119,72 @@ type Option func(*Options) error
 type Options struct {
 
 	// Url represents a single NATS server url to which the client
-	// will be connecting. If Servers is also set, it then becomes
-	// the first server in the array.
+	// will be connecting. If the Servers option is also set, it
+	// then becomes the first server in the Servers array.
 	Url string
 
-	// Servers is the configured set of servers which are
-	// available when attempting to connect.
+	// Servers is a configured set of servers which this client
+	// will use when attempting to connect.
 	Servers []string
 
-	// NoRandomize configures whether we will be randomizing
-	// the server pool of servers.
+	// NoRandomize configures whether we will randomize the
+	// server pool.
 	NoRandomize bool
 
-	// Name is the optional name label which will be sent to the server
+	// Name is an optional name label which will be sent to the server
 	// on CONNECT to identify the client.
 	Name string
 
-	// Verbose enables the server whether it should reply back
-	// OK on commands successfully being processed.
+	// Verbose signals the server to send an OK ack for commands
+	// successfully processed by the server.
 	Verbose bool
 
-	// Pedantic sets pedantic flag option sent on connect to signal server
-	// whether it should be doing further validation of subjects.
+	// Pedantic signals the server whether it should be doing further
+	// validation of subjects.
 	Pedantic bool
 
-	// Secure enable TLS secure connections that skip server
-	// verification by default.
+	// Secure enables TLS secure connections that skip server
+	// verification by default. NOT RECOMMENDED.
 	Secure bool
 
-	// TLSConfig is the custom TLS configuration to use for
-	// the secure transport.
+	// TLSConfig is a custom TLS configuration to use for secure
+	// transports.
 	TLSConfig *tls.Config
 
-	// AllowReconnect enables reconnection logic for when server we were
-	// connected to fails.
+	// AllowReconnect enables reconnection logic to be used when we
+	// encounter a disconnect from the current server.
 	AllowReconnect bool
 
-	// MaxReconnect sets the number of connect attempts that will be
-	// tried before giving up connecting further to a server in the pool.
-	// If negative, then it will never give up trying to connect.
+	// MaxReconnect sets the number of reconnect attempts that will be
+	// tried before giving up. If negative, then it will never give up
+	// trying to reconnect.
 	MaxReconnect int
 
-	// ReconnectWait sets the time to backoff after attempting to reconnect
+	// ReconnectWait sets the time to backoff after attempting a reconnect
 	// to a server that we were already connected to previously.
 	ReconnectWait time.Duration
 
-	// Timeout sets the timeout for Dial on a connection.
+	// Timeout sets the timeout for a Dial operation on a connection.
 	Timeout time.Duration
 
 	// FlusherTimeout is the maximum time to wait for the flusher loop
-	// to be able to finish writing to the underlying socket.
+	// to be able to finish writing to the underlying connection.
 	FlusherTimeout time.Duration
 
-	// PingInterval is the period at which the server will be sending ping
+	// PingInterval is the period at which the client will be sending ping
 	// commands to the server, disabled if 0 or negative.
 	PingInterval time.Duration
 
-	// MaxPingsOut is the maximum number of pending ping commands waiting
-	// for a response back before raising a ErrStaleConnection error.
+	// MaxPingsOut is the maximum number of pending ping commands that can
+	// be awaiting a response before raising an ErrStaleConnection error.
 	MaxPingsOut int
 
-	// ClosedCB sets the closed handler called when client will
+	// ClosedCB sets the closed handler that is called when a client will
 	// no longer be connected.
 	ClosedCB ConnHandler
 
-	// DisconnectedCB sets the disconnected handler called whenever we
-	// are disconnected.
+	// DisconnectedCB sets the disconnected handler that is called
+	// whenever we are disconnected.
 	DisconnectedCB ConnHandler
 
 	// ReconnectedCB sets the reconnected handler called whenever
@@ -198,8 +198,8 @@ type Options struct {
 	// AsyncErrorCB sets the async error handler (e.g. slow consumer errors)
 	AsyncErrorCB ErrHandler
 
-	// ReconnectBufSize of the backing bufio buffer during reconnect. Once this
-	// has been exhausted publish operations will error.
+	// ReconnectBufSize is the size of the backing bufio during reconnect.
+	// Once this has been exhausted publish operations will return an error.
 	ReconnectBufSize int
 
 	// SubChanLen is the size of the buffered channel used between the socket
@@ -208,7 +208,7 @@ type Options struct {
 	// dictated by PendingLimits()
 	SubChanLen int
 
-	// User sets the user to be used when connecting to the server.
+	// User sets the username to be used when connecting to the server.
 	User string
 
 	// Password sets the password to be used when connecting to a server.
@@ -217,8 +217,12 @@ type Options struct {
 	// Token sets the token to be used when connecting to a server.
 	Token string
 
-	// Dialer allows users setting a custom Dialer
+	// Dialer allows a custom Dialer when forming connections.
 	Dialer *net.Dialer
+
+	// UseOldRequestStyle force older method of Requests that utilize a new Inbox
+	// and Subscription for each request.
+	UseOldRequestStyle bool
 }
 
 const (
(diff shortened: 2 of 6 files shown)


nats.go
```diff
@@ -119,72 +119,72 @@ type Option func(*Options) error
...
 
-	// ClosedCB sets the closed handler called when client will
+	// ClosedCB sets the closed handler that is called when a client will
 	// no longer be connected.
 	ClosedCB ConnHandler
 
-	// DisconnectedCB sets the disconnected handler called whenever we
-	// are disconnected.
+	// DisconnectedCB sets the disconnected handler that is called
+	// whenever we are disconnected.
```
kozlovic: Should we replaced "we are" by "the client is"?

nats.go
```diff
@@ -217,8 +217,12 @@ type Options struct {
 	// Token sets the token to be used when connecting to a server.
 	Token string
 
-	// Dialer allows users setting a custom Dialer
+	// Dialer allows a custom Dialer when forming connections.
 	Dialer *net.Dialer
+
+	// UseOldRequestStyle force older method of Requests that utilize a new Inbox
```
kozlovic: I see that you have changed some of the verbs for other comments, so should it be "forces" here?

nats.go
```diff
@@ -1941,10 +1961,122 @@ func (nc *Conn) publish(subj, reply string, data []byte) error {
 	return nil
 }
 
-// Request will create an Inbox and perform a Request() call
+// respHandler is the global respnse handler. It will look up
+// the apprioriate channel based on the last token and place
```
kozlovic: typos: "response handler" and "appropriate"

nats.go
```diff
@@ -1941,10 +1961,122 @@ func (nc *Conn) publish(subj, reply string, data []byte) error {
...
+
+	nc.mu.Lock()
+	// Just return if closed, let Request timeout.
+	if nc.isClosed() {
+		nc.mu.Unlock()
+		return
+	}
+
+	// Grab mch
+	mch := nc.respMap[rt]
```
kozlovic: Should we check for `nil` here? We remove the channel from the map next line, so if the responder was to send 2 responses, the second response would get a nil channel.

nats.go
```diff
@@ -1941,10 +1961,122 @@ func (nc *Conn) publish(subj, reply string, data []byte) error {
...
+	// Delete the key regardless, one response only.
+	// FIXME(dlc) - should we track responses past 1
+	// just statistics wise?
+	delete(nc.respMap, rt)
+	nc.mu.Unlock()
+
+	// Don't block, let Request timeout instead, mch is
+	// buffered and we should delete the key before a
+	// second response is processed.
+	select {
```
kozlovic: If we allow only 1 response per request, then why make the backing channel more than size 1, and again, since there is only 1 response (we remove the channel from the map), we should be sure to be able to put the message in the channel and not need the select, right?

nats.go
```diff
@@ -1941,10 +1961,122 @@ func (nc *Conn) publish(subj, reply string, data []byte) error {
...
+	// _INBOX wildcard
+	ginbox := fmt.Sprintf("%s.*", NewInbox())
+	s, err := nc.Subscribe(ginbox, nc.respHandler)
+	if err != nil {
+		return err
+	}
+	// We could be racing here. So will we double check
+	// respMux here and discard the new one if set.
+	nc.mu.Lock()
+	defer nc.mu.Unlock()
```
kozlovic: Don't use `defer` here because otherwise you will call `s.Unsubscribe()` under `nc.mu.Lock()` which will deadlock.

nats.go
```diff
@@ -1941,10 +1961,122 @@ func (nc *Conn) publish(subj, reply string, data []byte) error {
...
+	if nc.respMux == nil {
+		nc.respSub = ginbox
+		nc.respMux = s
+		nc.respMap = make(map[string]chan *Msg)
+	} else {
+		// Discard duplicate, don't set others.
+		defer s.Unsubscribe()
```
kozlovic: Need to explicitly release lock before calling `s.Unsubscribe()`

nats.go
```diff
@@ -1941,10 +1961,122 @@ func (nc *Conn) publish(subj, reply string, data []byte) error {
...
+
+	// Make sure scoped subscription is setup at least once on first
+	// call to Request(). Will handle duplicates in createRespMux.
+	if doSetup {
+		if err := nc.createRespMux(); err != nil {
+			return nil, err
+		}
+	}
+	// Create literal Inbox and map to a chan msg.
+	mch := make(chan *Msg, RequestChanLen)
```
kozlovic: Why not just size 1 here since we handle only 1 response in response handler?

derekcollison:
I will take a look. Thought we had spell checks, etc. Good call on context, will take a look to make sure we have the appropriate tests in place, etc.

nats.go
kozlovic: By the way, your test of simultaneous requests did expose that (I was lucky that I got it the first time, but sometimes it would pass, but most times it would not on my machine).

derekcollison: Yes my slow laptop on the plane always passed but my office machine shows it. Good catch. Will fix later tonight when the schedule frees up..

derekcollison: And travis is even slower so they all passed too, but once on a fast machine you see it right away.

derekcollison: At least I did write the test.. ;)

nats.go
derekcollison: paranoia around blocking channels, prefer buffered ones, even if small for safety. I also believe it changes behavior of the Go scheduler to some degree when unbuffered.

kozlovic: I did not say unbuffered, since we would lose the message or block. I am saying why not size 1 (RequestChanLen is 8 I believe).

derekcollison: I believe 1 or 8 is same underneath the covers memory wise.. But I could be wrong, I noticed I used 8 elsewhere and figured I picked it for a reason.

kozlovic: It's fine, I just wanted to make sure that I understood the code correctly and that only one message should ever be put into the channel.

nats.go
```diff
@@ -1941,10 +1961,122 @@ func (nc *Conn) publish(subj, reply string, data []byte) error {
...
+		nc.respMux = s
+		nc.respMap = make(map[string]chan *Msg)
+	} else {
+		// Discard duplicate, don't set others.
+		defer s.Unsubscribe()
+	}
+	return nil
+}
+
+// New style request that will mux on a single Subscription.
```
tylertreat: Not true if `UseOldRequestStyle` is set.

Since this is public, seems like the docstring should describe what this does/is for rather than impl details.

tylertreat: Also, is there a reason we need to expose the old style? There shouldn't be any compatibility issues with the switch, right? New style vs. old style is confusing from an API "UX" perspective.

derekcollison: Good point on comment for public doc, will update.

derekcollison: Since this is such a big change, I want to make sure clients who need to go revert back to the old way. Adding option was simple and will not affect any code unless the client is stuck for some unknown reason and wants the old way. Only situation I can think of is use of circuit breaker with 1:N where all N respond and N is large.

nats.go
derekcollison: ok

nats.go
derekcollison: yes

nats.go
derekcollison: Fixed, wonder why misspell did not catch?

nats.go
derekcollison: Sending to a nil channel will block forever, so select below will do the right thing.

nats.go
derekcollison: See above.

nats.go
derekcollison: Fixed.

nats.go
derekcollison: Fixed.

nats.go
derekcollison: Update doc, keeping ability to set option to force old style semantics.

context.go
```diff
@@ -16,7 +16,58 @@ func (nc *Conn) RequestWithContext(ctx context.Context, subj string, data []byte
 	if ctx == nil {
 		return nil, ErrInvalidContext
 	}
+	if nc == nil {
```
kozlovic: Code coverage wise, this is not tested.

context.go
```diff
@@ -16,7 +16,58 @@ func (nc *Conn) RequestWithContext(ctx context.Context, subj string, data []byte
...
+
+	// snapshot
+	var doSetup, useOldRequestStyle bool
+	nc.mu.Lock()
+	useOldRequestStyle = nc.Opts.UseOldRequestStyle
+	doSetup = (nc.respMux == nil)
+	nc.mu.Unlock()
+
+	// If user wants the old style.
+	if useOldRequestStyle {
```
kozlovic: Code coverage wise, this is not tested.

context.go
```diff
@@ -16,7 +16,58 @@ func (nc *Conn) RequestWithContext(ctx context.Context, subj string, data []byte
...
+			return nil, ErrConnectionClosed
+		}
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	}
+
+	return msg, nil
+}
 
+// oldRequestWithContext utilizes inbox and subscription per request.
```
kozlovic: This whole function shows up as not covered, which brings the code coverage of this file down to 71%.

nats.go
```diff
@@ -1941,10 +1961,123 @@ func (nc *Conn) publish(subj, reply string, data []byte) error {
...
+		defer s.Unsubscribe()
+	}
+	nc.mu.Unlock()
+	return nil
+}
+
+// Request will send a request payload and deliver the response message,
+// or an error, including a timeout if no message was received properly.
+func (nc *Conn) Request(subj string, data []byte, timeout time.Duration) (*Msg, error) {
+	if nc == nil {
```
kozlovic: Test for this?

nats.go
```diff
@@ -1941,10 +1961,123 @@ func (nc *Conn) publish(subj, reply string, data []byte) error {
...
+	t := time.NewTimer(timeout)
+	defer t.Stop()
+
+	var ok bool
+	var msg *Msg
+
+	select {
+	case msg, ok = <-mch:
+		if !ok {
+			return nil, ErrConnectionClosed
```
kozlovic: Would like to see a test where we spawn a go routine that will close the connection after a delay, and issue a request in while there is no responder running, and make sure that Request() returns ErrConnectionClosed.

derekcollison:
@kozlovic @tylertreat @wallyqs Take a look now, should be good.

derekcollison:
I will add some tests for RequestContext and switch to old style, and check for nil.

nats.go
derekcollison: Agree, but hard to make timing work correctly I think. I did consider it.

context.go
derekcollison: Will fix..

nats.go
kozlovic: What about something like this:
```
...
wg := &sync.WaitGroup{}
wg.Add(1)
go func() {
    defer wg.Done()
    time.Sleep(250 * time.Millisecond)
    nc.Close()
}()
if _, err := nc.Request("foo", []byte("help"), 3*time.Second); err != nats.ErrInvalidConnection || err != nats.ErrConnectionClosed {
}
wg.Wait()
```
The test for `nats.ErrInvalidConnection` would be to avoid a flapping test where the connection would be closed in the go routine before the `Request()` call has a chance to execute.

derekcollison:
Am I good for a merge? /cc @kozlovic @tylertreat @wallyqs

nats.go
derekcollison: That may work..

kozlovic: I quickly tried that and looks like the Request is not kicked out...

kozlovic: I don't think we do anything with respMux and all channels in map during close, so that makes sense...

derekcollison: Currently this always gives us a timeout error, by design.

kozlovic: So we would never get this ErrConnectionClosed for a closed channel, since this channel is never closed?

tylertreat:
lgtm

nats.go
derekcollison: Fixed.

kozlovic: Actually, if you look at Close(), we do go through the list of subscriptions and close the mch or signal in case of async. So NextMsg() (in old Request style) would get kicked out and will get ErrConnectionClosed. You are changing the behavior in that the Request will now return ErrConnectionClosed but only after waiting for the whole timeout.

nats.go
```diff
@@ -1941,10 +1961,126 @@ func (nc *Conn) publish(subj, reply string, data []byte) error {
...
+		}
+	case <-t.C:
+		if nc.isClosed() {
```
kozlovic: I don't think this is right (see previous comment). Also, this version of is closed is not locked.

kozlovic: And for the test I described above (doing the close in go routine), I would also take the time before the request and after it returned and make sure that it is less than the timeout given to the request call, to make sure that the request is really kicked out ;-)

nats.go
derekcollison: Agree, fixed.

nats.go
derekcollison: Agree, fixed.

nats.go
```diff
@@ -2746,6 +2748,20 @@ func (nc *Conn) clearPendingFlushCalls() {
...
+func (nc *Conn) clearPendingRequestCalls() {
+	if nc.respMap == nil {
+		return
+	}
+	for _, ch := range nc.respMap {
+		if ch != nil {
+			close(ch)
+		}
+	}
+	nc.pongs = nil
```
kozlovic: did you meant nc.respMap here?

nats.go
```diff
@@ -2068,7 +2070,7 @@ func (nc *Conn) Request(subj string, data []byte, timeout time.Duration) (*Msg,
 			return nil, ErrConnectionClosed
 		}
 	case <-t.C:
-		if nc.isClosed() {
+		if nc.IsClosed() {
```
kozlovic: I would not put that here. It is trued that there is possibly a race that the connection is closed while the request just timed-out, but we don't do that in NextMsg, etc..

nats.go
```diff
@@ -1968,18 +1968,20 @@ func (nc *Conn) respHandler(m *Msg) {
...
 	// Delete the key regardless, one response only.
 	// FIXME(dlc) - should we track responses past 1
 	// just statistics wise?
 	delete(nc.respMap, rt)
+
+	// Just return if closed, kick out Request by
+	// closing channel.
+	if nc.isClosed() {
+		nc.mu.Unlock()
+		close(mch)
```
kozlovic: Is close of a nil channel valid? (remember, mch could be nil here...)

nats.go
derekcollison: ok fixed.

nats.go
derekcollison: Believe so..

derekcollison: Actually no, but removed it anyway.

kozlovic: No, it will panic, so this needs to be fixed.

nats.go
```diff
@@ -2592,6 +2745,19 @@ func (nc *Conn) clearPendingFlushCalls() {
...
+// This will clear any pending Request calls.
+// Lock is assumed to be held by the caller.
+func (nc *Conn) clearPendingRequestCalls() {
+	if nc.respMap == nil {
+		return
+	}
+	for _, ch := range nc.respMap {
+		if ch != nil {
+			close(ch)
+		}
```
kozlovic: You should empty the map. In a test, you got this:
```
github.com/nats-io/go-nats.(*Conn).respHandler(0xc420406a00, 0xc420ac21e0)
	/home/travis/gopath/src/github.com/nats-io/go-nats/nats.go:1982 +0x23a
github.com/nats-io/go-nats.(*Conn).(github.com/nats-io/go-nats.respHandler)-fm(0xc420ac21e0)
	/home/travis/gopath/src/github.com/nats-io/go-nats/nats.go:2004 +0x4c
github.com/nats-io/go-nats.(*Conn).waitForMsgs(0xc420406a00, 0xc4200f6f00)
	/home/travis/gopath/src/github.com/nats-io/go-nats/nats.go:1559 +0x1f9
created by github.com/nats-io/go-nats.(*Conn).subscribe
	/home/travis/gopath/src/github.com/nats-io/go-nats/nats.go:2223 +0x3d9
FAIL	github.com/nats-io/go-nats/test	6.891s
?   	github.com/nats-io/go-nats/util	[no test files]

kozlovic: That's because the channel was closed, but left in the map, then respHandler tried to close again.

derekcollison: Fixed.

nats.go
kozlovic: @derekcollison Seems that you missed above comment: closing a nil channel will panic

derekcollison: Reverted back to just returning.

kozlovic:
LGTM!

derekcollison:
Sorry that was so much a PITA.. old and rusty, uug.

PR #297, opened 2017-05-31
Pool timers to avoid allocations

nussjustin:
Requests are somewhat allocation heavy. This should help a bit. Also has a small positive effect on requests performance (more for the old request style).

Benchstat (base 9147cfac69833916f9c73353ba43f939815c330a):

```
name                              old time/op    new time/op    delta
Request-8                           43.9µs ± 5%    43.4µs ± 1%     ~     (p=0.690 n=5+5)
OldRequest-8                        60.0µs ± 3%    57.5µs ± 2%   -4.30%  (p=0.008 n=5+5)

name                              old alloc/op   new alloc/op   delta
Request-8                             885B ± 0%      693B ± 0%  -21.71%  (p=0.008 n=5+5)
OldRequest-8                        2.63kB ± 0%    2.44kB ± 0%   -7.27%  (p=0.000 n=5+4)

name                              old allocs/op  new allocs/op  delta
Request-8                             17.0 ± 0%      14.0 ± 0%  -17.65%  (p=0.008 n=5+5)
OldRequest-8                          55.0 ± 0%      52.0 ± 0%   -5.45%  (p=0.008 n=5+5)
```

Other benchmarks show no changes.

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -2056,8 +2056,8 @@ func (nc *Conn) Request(subj string, data []byte, timeout time.Duration) (*Msg, 
 		return nil, err
 	}
 
-	t := time.NewTimer(timeout)
-	defer t.Stop()
+	t := globalTimerPool.Get(timeout)
+	defer globalTimerPool.Put(t)
 
 	var ok bool
 	var msg *Msg
@@ -2380,8 +2380,8 @@ func (s *Subscription) NextMsg(timeout time.Duration) (*Msg, error) {
 	var ok bool
 	var msg *Msg
 
-	t := time.NewTimer(timeout)
-	defer t.Stop()
+	t := globalTimerPool.Get(timeout)
+	defer globalTimerPool.Put(t)
 
 	select {
 	case msg, ok = <-mch:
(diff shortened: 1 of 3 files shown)


nussjustin:
Test failure is unrelated and also happens on master (though only 2 times in 20 local runs, 10 with and 10 without -race).

nussjustin:
Since this works with sync.Pool testing is somewhat tricky. The test would need to disable GC (and reenable afterwards) and set GOMAXPROCS to 1 (and reset it afterwards) to prevent the pool from being emptied and to avoid rescheduling onto another P.

If you want I can push a test.

tylertreat:
lgtm. Would be nice to have some basic tests around `timerPool`, though I agree it's a bit tricky to test with the GC behavior.

nussjustin:
Pushed a new commit with a test that should work.

nussjustin:
Ah, okay. sync.Pool will randomly drop values when compiling with -race to prevent assuming it's behaviour. I changed the test to just check the behaviour of the timer instead.

tylertreat:
+1

PR #299, opened 2017-06-02
Fix new-style request leak

tylertreat:
The new request style was leaking channels in the respMap when requests
timed out. Now we remove them from the map when the request times out.

This also simplifies the respMux setup and prevents multiple resp
subscriptions from being created on the server.

@nats-io/core

### DIFF
--- a/context.go
+++ b/context.go
@@ -40,12 +40,13 @@ func (nc *Conn) RequestWithContext(ctx context.Context, subj string, data []byte
 	token := respToken(respInbox)
 	nc.respMap[token] = mch
 	createSub := nc.respMux == nil
+	ginbox := nc.respSub
 	nc.mu.Unlock()
 
 	if createSub {
 		// Make sure scoped subscription is setup only once.
 		var err error
-		nc.respSetup.Do(func() { err = nc.createRespMux() })
+		nc.respSetup.Do(func() { err = nc.createRespMux(ginbox) })
 		if err != nil {
 			return nil, err
 		}
--- a/context.go
+++ b/context.go
@@ -65,6 +65,9 @@ func (nc *Conn) RequestWithContext(ctx context.Context, subj string, data []byte
 			return nil, ErrConnectionClosed
 		}
 	case <-ctx.Done():
+		nc.mu.Lock()
+		delete(nc.respMap, token)
+		nc.mu.Unlock()
 		return nil, ctx.Err()
 	}
 
--- a/context.go
+++ b/context.go
@@ -7,6 +7,7 @@ package nats
 
 import (
 	"context"
+	"fmt"
 	"reflect"
 )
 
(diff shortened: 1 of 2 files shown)


derekcollison:
So of course I looked at that as well, and coded it up that way first but realized I was using the Once mutex for creation and then the Conn's mutex for accessing etc. So thought it would show up as a race.

nats.go
```diff
@@ -2026,33 +2018,32 @@ func (nc *Conn) Request(subj string, data []byte, timeout time.Duration) (*Msg,
...
-	// Make sure scoped subscription is setup at least once on first
-	// call to Request(). Will handle duplicates in createRespMux.
-	if doSetup {
-		if err := nc.createRespMux(); err != nil {
-			return nil, err
-		}
+	// Make sure scoped subscription is setup only once on first call to
+	// Request().
+	var err error
+	nc.respSetup.Do(func() { err = nc.createRespMux() })
```
kozlovic: All calls will block until `createRespMux()` returns, right? That is, say "thread" 1 is first to go, is in the middle of executing `createRespMux()` and a second "thread" executes `nc.respSetup.Do()`, then the call will block, right? If not, then the code would not be correct.

The only drawback with this (and previous) approach is that we lock nc, use now this lock (I am sure it is implemented with locking/atomic), then again lock nc to create the respInbox. In the C client, I initialize map and respInbox in the lock above, then *only* when we need to create the sub there is extra locking required. So after initial creation of subscription, I require only 1 lock, here it is 3 per request. That's fine, we can iterate again later.

nats.go
```diff
@@ -1995,26 +1996,17 @@ func (nc *Conn) respHandler(m *Msg) {
...
-	if nc.respMux == nil {
-		nc.respSub = ginbox
-		nc.respMux = s
-		nc.respMap = make(map[string]chan *Msg)
-	} else {
-		// Discard duplicate, don't set others.
-		defer s.Unsubscribe()
-	}
-	nc.mu.Unlock()
+	nc.respSub = ginbox
```
kozlovic: Should we have locking around this? What if the connection is closed from another go-routine?

nats.go
derekcollison: The Once struct and hence the Do has both a mutex and atomic for fast path evaluation for when the function has already been called. But this again is mixing and matching the Once mutex with the Connection mutex. But that may be ok in this case.

tylertreat: I'm OK with reverting the `sync.Once` change if we don't want it. The important thing in this PR is the map cleanup on request timeout, though I don't like the current race where multiple resp subscriptions can get created and then are subsequently unsubscribed, which this change prevents.

derekcollison: Do with think their is a race using different locks to protect the same structure fields? Should we try a test that spins up some Go routines and then they all try to subscribe at the same time?

tylertreat: Potentially. I will write a test and see if I can trigger a race. Otherwise, we might just be better off doing what Ivan did in cnats.

kozlovic: I think sync.Once() is great, but we could leave the use of it only when we detect that the respMux is not yet created. So it could be a mix of what we do in c-nats and use of sync.Once. 

That is, when first grabbing nc' mutex, we could setup the respMap, etc.. if not yet initialized, and prepare the respInbox and put it in the map, then snapshot if we need to create the subsription: `createSub := nc.respMux == nil`.  When releasing the nc's mutex, check `createSub` and invoke sync.Once, but this time just to create the subscription (not the map, which would already be done), then in the function, setup respMux under nc.mu protection please.

Again, in this case that means that we would grab nc.mu only once per request, except at initialization time where we may call sync.Once from multiple go-routines (if there are parallel Requests).

(Alternatively, or if I am not making sense, you could simply fix the map leak on timeout condition - which I had addressed in the c client and wanted to report back but did not have time, and then have I could submit a PR for the rest if it makes it easier to understand - and see if what I say actually works :-)).

nats.go
```diff
@@ -1996,17 +1996,19 @@ func (nc *Conn) respHandler(m *Msg) {
...
 // new style responses. This will be on an _INBOX with an
 // additional terminal token. The subscription will be on
-// a wildcard. Caller is responsible for synchronization.
+// a wildcard. Caller is responsible for ensuring this is
+// only called once.
 func (nc *Conn) createRespMux() error {
-	// _INBOX wildcard
-	ginbox := fmt.Sprintf("%s.*", NewInbox())
+	nc.mu.Lock()
+	ginbox := nc.respSub
```
derekcollison: Don't need this tmp variable now it seems.

tylertreat:
@kozlovic implemented what you suggested. I realized there is a lot of duplication between `RequestWithContext` and `Request`. Would be good if we could somehow unify them or at least reduce duplicated code...

nats.go
```diff
@@ -1999,10 +1999,7 @@ func (nc *Conn) respHandler(m *Msg) {
 // a wildcard. Caller is responsible for ensuring this is
 // only called once.
 func (nc *Conn) createRespMux() error {
-	nc.mu.Lock()
-	ginbox := nc.respSub
-	nc.mu.Unlock()
-	s, err := nc.Subscribe(ginbox, nc.respHandler)
+	s, err := nc.Subscribe(nc.respSub, nc.respHandler)
```
derekcollison: Apologies, I now see that you were snapshotting the subject from the *Conn struct, which is probably needed since now line 2002 accesses it without a lock.

context.go
```diff
@@ -20,32 +21,36 @@ func (nc *Conn) RequestWithContext(ctx context.Context, subj string, data []byte
...
-	// call to Request(). Will handle duplicates in createRespMux.
-	if doSetup {
-		if err := nc.createRespMux(); err != nil {
-			return nil, err
-		}
+	// Do setup for the new style.
+	if nc.respMap == nil {
+		// _INBOX wildcard
+		nc.respSub = fmt.Sprintf("%s.*", NewInbox())
+		nc.respMap = make(map[string]chan *Msg)
```
kozlovic: I would then snapshot respSub here (like `wsub := nc.respSub`) and the pass it to `nc.createRespMux()`.

context.go
```diff
@@ -20,32 +21,36 @@ func (nc *Conn) RequestWithContext(ctx context.Context, subj string, data []byte
...
 
+	if createSub {
+		// Make sure scoped subscription is setup only once.
+		var err error
+		nc.respSetup.Do(func() { err = nc.createRespMux() })
+		if err != nil {
+			return nil, err
+		}
+	}
+
```
kozlovic: Do we need to remove from map on timeout here too?

PR #305, opened 2017-07-07
Switch to use empty struct for signaling in pongs channel

wallyqs:
Similar to https://github.com/nats-io/go-nats/pull/304 this also changes to use zero size empty struct for signaling in the pending pongs channel.


kozlovic:
LGTM. (we still use `chan bool` in all tests and have the `Wait()` accepting a `chan bool`. I would not change those, I don't think it is worth it ;-)

wallyqs:
@kozlovic yes agreed! thanks for checking :)

PR #320, opened 2017-10-01
Update Go versions for Travis.CI

msoap:
Synchronize versions with [gnatsd](https://github.com/nats-io/gnatsd/blob/master/.travis.yml)


kozlovic:
Thanks!

PR #326, opened 2017-10-20
[FIXED] ErrorHandler not always reporting proper error

### LINKED ISSUE #325: ErrorHandler not always reporting proper error
Say that your application has no permission to publish and subscribe on foo.
```
nc.subscribe('foo', cb)
nc.publish('foo', []byte("hello")
```
These 2 calls should result in getting permission violation errors presented
to the ErrorHandler, one for the subscribe, one for the publish.
However, the error handler will be invoked with 2 errors corresponding
to the publish permission error. If the application was doing a publish
then a subscribe, the error callback will likely get 2 errors about permission
violation related to subscription.

kozlovic:
The issue was that we were using `nc.err` as a parameter of
the closure function sent to `nc.ach`. The value of the error
would then change and be whatever last value is.
We now create an error so that this is unique per scheduled
async error.

Updated the test that checks not only that we get the permission
violation error, but that we get the expected one (one for sub
and for for pub).

Resolves #325

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -1682,9 +1682,11 @@ slowConsumer:
 // permissions violation on either publish or subscribe.
 func (nc *Conn) processPermissionsViolation(err string) {
 	nc.mu.Lock()
-	nc.err = errors.New("nats: " + err)
+	// create error here so we can pass it as a closure to the async cb dispatcher.
+	e := errors.New("nats: " + err)
+	nc.err = e
 	if nc.Opts.AsyncErrorCB != nil {
-		nc.ach <- func() { nc.Opts.AsyncErrorCB(nc, nil, nc.err) }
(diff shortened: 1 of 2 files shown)


nats.go
```diff
@@ -1682,9 +1682,11 @@ slowConsumer:
 // permissions violation on either publish or subscribe.
 func (nc *Conn) processPermissionsViolation(err string) {
 	nc.mu.Lock()
-	nc.err = errors.New("nats: " + err)
+	// create error here so we can pass it as a closure to the async cb dispatcher.
+	e := errors.New("nats: " + err)
```
tylertreat: Unrelated to this, but IMO we should consider using "typed" errors. Users should not have to resort to parsing error strings to handle different errors.

PR #327, opened 2017-10-22
Resize pending pongs slice when flush times out

wallyqs:
When a `Flush` call has timed out, the channel awaiting the pong reply is changed to be nil but it is not dropped from the pending pongs slice, so it can continue to grow for the same connection.  This can occur when the client is attempting to flush during the reconnection state for example.


derekcollison:
PONG should be FIFO and ordered.

wallyqs:
Thanks for the feedback. Right, this would affect the ordering so closing the PR

PR #334, opened 2017-12-13
[ADDED] CustomDialer interface/option setter

### LINKED ISSUE #329: question about customize dialer
Dialer option

func Dialer(dialer *net.Dialer) Option {
	return func(o *Options) error {
		o.Dialer = dialer
		return nil
	}
}

It required net.Dialer structure.  It is not an interface.  I like to provide the customized dialer, eg. proxy . 

The dial is used as
...
	dialer := nc.Opts.Dialer
	nc.conn, err = dialer.Dial("tcp", nc.url.Host)
	if err != nil {
		return err
	}

It called dialer.Dial function.  

http.Transport will use function to do customized
type Transport struct {
...
	Dial func(network, addr string) (net.Conn, error)
...
}

How could I override the Dial.  Is there any suggestions?

kozlovic:
We recently added a Dialer option, however, we made it a *net.Dialer,
which is very restrictive (since *net.Dialer is a struct, not an
interface).
One could imagine creating a Dialer that would create pipes between
the client and an embedded server. Since Dial() returns a net.Conn,
which is an interface, everything would work the same.

Resolves #329

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -123,6 +123,12 @@ type asyncCB func()
 // Option is a function on the options for a connection.
 type Option func(*Options) error
 
+// CustomDialer can be used to specify any dialer, not necessarily
+// a *net.Dialer.
+type CustomDialer interface {
+	Dial(network, address string) (net.Conn, error)
+}
+
 // Options can be used to create a customized connection.
 type Options struct {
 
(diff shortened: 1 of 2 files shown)


kozlovic:
This is also quite useful. For instance, one could create a custom dialer to return a connection that can trace Read (and/or Write), or introduce some delay, errors, etc... that could be useful in tests.

Here is an example of a CustomDialer to create a traceable connection:
```
type MyDialer struct{}

type TracedConn struct {
	net.Conn
}

func (tc *TracedConn) Read(b []byte) (n int, err error) {
	n, err = tc.Conn.Read(b)
	if err != nil {
		fmt.Printf("@@ conn err: %v\n", err)
	} else {
		fmt.Printf("@@ conn read: %s\n", b[:n])
	}
	return n, err
}

func (md *MyDialer) Dial(network, address string) (net.Conn, error) {
	d := &net.Dialer{}
	c, err := d.Dial(network, address)
	if err != nil {
		return nil, err
	}
	return &TracedConn{Conn: c}, nil
}

func main() {
	myDialer := &MyDialer{}
	nc, err := nats.Connect(*urls, nats.SetCustomDialer(myDialer))
       (...)
}
```

PR #336, opened 2017-12-15
Add Go 1.8 TLS config Clone method.

### LINKED ISSUE #335: CloneTLSConfig does not perform a complete copy in Go 1.8+
Since Go 1.8 `tls.Config` has a field called `GetClientCertificate`. The existing methods for cloning `tls.Configs` in the NATS codebase do not copy this field. Indeed, it should be deferring to  `conf.Clone()` in Go 1.8+ environments.

Related issues:
https://github.com/golang/go/issues/19264
https://github.com/gorilla/websocket/issues/215

johanbrandhorst:
Fixes #335


kozlovic:
@johanbrandhorst Thanks! Have you tried a `git mv util/tls.go util/tls_go17.go`? This would maintain the git log for this file. You would then need to add the build directive to work only for go 1.7 (as you did).

kozlovic:
That being said, it is no big deal since there was only 1 commit.. I wonder if we should then also get rid of the pre 1.7 file, since we no longer support that.

johanbrandhorst:
Sure, I can do that.

johanbrandhorst:
I did the `git mv` but it seems creating a new `tls.go` messes a bit with history anyway. Happy with this?

johanbrandhorst:
Any chance we could cut a new release with this bug fix?

kozlovic:
There is a pending PR that we may want to include first. Will discuss internally and see how soon we can cut one out. I'll keep you posted.

johanbrandhorst:
Thanks for the amazing response times @kozlovic, really impressed!

kozlovic:
@johanbrandhorst Done: https://github.com/nats-io/go-nats/releases/tag/v1.4.0. Thanks for the contribution!

PR #338, opened 2018-01-17
[ADDED] Conn.Barrier() API

kozlovic:
Suppose that you have a connection with two subscriptions, let's
call them "pub" and "close".
Let's say that you know that when receiving a message on "close",
no new message will arrive on "pub" (because the sender does
send "pub" and then a final message on "close"). But because
"pub" and "close" are two different subscriptions, there is no
guarantee that the message on "close" is not going to be dispatched
before the last message on "pub".

The Barrier() API can be used to solve this issue. In the example
above, if the message callback on "close" were to call that API,
this will ensure that all pending messages on "pub" have been
dipatched before the provided function in the Barrier() API gets
executed.

See `test/TestBarrier()` test for more details.

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -20,6 +20,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/nats-io/go-nats/util"
@@ -351,6 +352,12 @@ type Msg struct {
 	Data    []byte
 	Sub     *Subscription
 	next    *Msg
+	barrier *barrierInfo
+}
+
+type barrierInfo struct {
+	refs int64
+	f    func()
 }
 
 // Tracks various stats received and sent on this connection,
@@ -1571,6 +1578,13 @@ func (nc *Conn) waitForMsgs(s *Subscription) {
 			if s.pHead == nil {
 				s.pTail = nil
 			}
+			if m.barrier != nil {
+				s.mu.Unlock()
+				if atomic.AddInt64(&m.barrier.refs, -1) == 0 {
+					m.barrier.f()
+				}
+				continue
+			}
 			s.pMsgs--
 			s.pBytes -= len(m.Data)
 		}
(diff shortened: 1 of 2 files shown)


nats.go
```diff
@@ -3006,3 +3033,48 @@ func (nc *Conn) TLSRequired() bool {
...
+func (nc *Conn) Barrier(f func()) error {
+	nc.mu.Lock()
+	defer nc.mu.Unlock()
+	if nc.isClosed() {
+		return ErrConnectionClosed
+	}
+	nc.subsMu.Lock()
+	defer nc.subsMu.Unlock()
+	// Need to figure out how many non chan subscriptions there are
+	numSubs := 0
```
derekcollison: For small numbers should be fine, but should we consider tracking a variable on nc? nc.numAsyncSubs

nats.go
```diff
@@ -3006,3 +3033,48 @@ func (nc *Conn) TLSRequired() bool {
...
+		}
+	}
+	if numSubs == 0 {
+		f()
+		return nil
+	}
+	barrier := &barrierInfo{refs: int64(numSubs), f: f}
+	for _, sub := range nc.subs {
+		sub.mu.Lock()
+		if sub.mch == nil {
```
derekcollison: This should be consistent with above when counting, but if its not, does f() never get called or called too early? Should we care?

derekcollison: If barrier is address, could we just count in place here and add it to the barrier struct after this iteration?

kozlovic: comment 1: the `nc.subsMu`tex is held for the duration of this call so no subscription can be removed after counting it.
comment 2: no because we still do individual sub locking, so once we post the barrier message, it is possible that a subscription processes it (after releasing the sub.mu lock here) before we have post them all. The dispatcher will possibly decrement below 0 or fire too soon.

nats.go
kozlovic: I'd rather leave this for now, avoiding any issue with discrepancies between +1 and -1 when removing a subscription.

kozlovic:
@derekcollison Not sure if I follow, but if the idea would be to use single go channel to use for several `ChanSubscription`s it means that I would have a single go routine dequeueing and that routine would have to check message subject, then invoke corresponding function for processing. This limits the throughput. It is nice to have the ability to dispatch concurrently when we can, but have a way to have a barrier when we need.

derekcollison:
You could queue the closed over f() with message etc. and dispatch to a group of Go routines.

derekcollison:
LGTM

nats.go
kozlovic: Maybe I misunderstood comment 1. I do not count sync and chan subscriptions because we don't have real control on when messages get dequeued (for sure for chan subscription, I would have no place to check for that special message).

PR #340, opened 2018-01-25
[ADDED] ReconnectBufSize() Option function

ripienaar:
This creates a function one can use to set the ReconnectBufSize option,
it's similar in design to the other functions already here

Also makes small fixes to some doc strings that I noticed

### DIFF
--- a/test/conn_test.go
+++ b/test/conn_test.go
@@ -700,8 +700,7 @@ func TestCallbacksOrder(t *testing.T) {
 		nats.ClosedHandler(cch),
 		nats.ErrorHandler(ech),
 		nats.ReconnectWait(50*time.Millisecond),
-		nats.DontRandomize(),
-		nats.ReconnectBufSize(10*1024))
+		nats.DontRandomize())
 
 	if err != nil {
 		t.Fatalf("Unable to connect: %v\n", err)
--- a/test/conn_test.go
+++ b/test/conn_test.go
@@ -700,7 +700,9 @@ func TestCallbacksOrder(t *testing.T) {
 		nats.ClosedHandler(cch),
 		nats.ErrorHandler(ech),
 		nats.ReconnectWait(50*time.Millisecond),
-		nats.DontRandomize())
+		nats.DontRandomize(),
+		nats.ReconnectBufSize(10*1024))
+
 	if err != nil {
 		t.Fatalf("Unable to connect: %v\n", err)
 	}
--- a/nats.go
+++ b/nats.go
@@ -535,6 +535,14 @@ func MaxReconnects(max int) Option {
 	}
 }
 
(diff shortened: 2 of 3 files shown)


ripienaar:
test failure does not appear to be my doing :)

kozlovic:
I have recycled the test. Would you mind adding a test so that code coverage shows that this function was invoked? It could be as simple as creating a connection with that option and then checking that the internal reconnect buffer size matches what you have set.

ripienaar:
OK - I did check, none of these others are tested, but sure.

kozlovic:
From coverall, it seems that they are since they show in green. Keep in mind that there are tests in `test` directory and in `nats_test.go`

ripienaar:
indeed, I saw there's a `nats_test.go` and figured if its tested its in there, missed the `test` dir :)

kozlovic:
Thanks, but that is really to make coverall happy. Could you instead look at test `TestReconnectBufSize` and make it a table test where you run the same content but once with option set, the way it is right now, and once with the new option you added?
If you can't that's fine, we can add it after the PR is merged.

ripienaar:
erm, I guess, though not really experienced with go's built in test framework, will have to wait for another day as that's all i had time for today

derekcollison:
Thanks for the submission, this would be helpful. We do require tests though.

ripienaar:
yup, will look at test tomorrow

test/conn_test.go
```diff
@@ -700,7 +700,9 @@ func TestCallbacksOrder(t *testing.T) {
 		nats.ClosedHandler(cch),
 		nats.ErrorHandler(ech),
 		nats.ReconnectWait(50*time.Millisecond),
-		nats.DontRandomize())
+		nats.DontRandomize(),
+		nats.ReconnectBufSize(10*1024))
```
kozlovic: Since you added a specific test, you may want to revert this change.

ripienaar:
@kozlovic looks like reworking that test into a table test is quite an undertaking - but as per your earlier suggestion the thing to test is if the option gets set or not, since that other test makes sure that the rest of the code does with whatever was set.

So I reckon this is sufficient to test the functionality of the Option.

There appears to be some weirdness, maybe something in my environments but these tests panic for me on timeout - even before my changes.  But seems ok on travis :)

kozlovic:
@ripienaar You may have a server running outside of the test suite...

kozlovic:
LGTM

ripienaar:
ok, let me know if there's anything else, but that'll be for tomorrow, thanks!

derekcollison:
Thanks! @kozlovic will merge.

kozlovic:
Thanks for the contribution!

PR #342, opened 2018-02-18
nats: simplify error check in Conn.connect

charlievieth:
This commit replaces the regex used to check for the string literal
'connection refused' with strings.Contains in `Conn.connect`. Using
regexp.Match requires compiling the regex on each call - since the
string being matched is a literal strings.Contains is a simpler and
more performant option here.

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -15,7 +15,6 @@ import (
 	"math/rand"
 	"net"
 	"net/url"
-	"regexp"
 	"runtime"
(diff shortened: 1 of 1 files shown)


charlievieth:
I'm unable to reproduce the the CI error on go1.9 - is this a flake?

kozlovic:
Thanks for the contribution!

PR #344, opened 2018-02-27
[IMPROVED] Server pool is updated based on server notifications

kozlovic:
The server pool was only growing when new servers were discovered.
Now, the client library updates its server pool based on server's
INFO protocols (true for server 1.0.6+).
The DiscoveredServersCB is still invoked only when new servers
are added (as in never seen as opposed to added back after leaving
the cluster).

The code should work ok with older servers but will take advantage
of changes in the server (https://github.com/nats-io/gnatsd/pull/626)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
 language: go
 sudo: false
 go:
+- 1.10.x
 - 1.9.x
 - 1.8.x
-- 1.7.x
 install:
 - go get -t ./...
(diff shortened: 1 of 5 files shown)


kozlovic:
Note that one test checks that server version will be 1.0.6+ since test expects INFO content to be updated when servers leave the cluster, which does not happen with current version (1.0.4) of the server.

derekcollison:
LGTM

PR #346, opened 2018-03-04
Fixed locking in Barrier when there is no async subscription

### LINKED ISSUE #345: Synchronous Barrier call holds conn mutex
When [`Barrier`](https://godoc.org/github.com/nats-io/go-nats#Conn.Barrier) is called with no active asynchronous subscriptions, the call will be made synchronously with both `nc.mu` and `nc.subsMu` held. Calling any other method on the [`Conn`](https://godoc.org/github.com/nats-io/go-nats#Conn) in the callback will than block the whole connection.

E.g. calling

```go
conn, _ := nats.Connect("nats://127.0.0.1:4222")
conn.Barrier(func() {
	conn.TLSRequired()
})
```

will block since the callback is called synchronously in `Barrier`, which holds the two mutexes.

As long as there is at least one asynchronous subscription, the callback will be called later without any mutex held. So this will work:

```go
conn, _ := nats.Connect("nats://127.0.0.1:4222")
conn.Subscribe("nats_is_awesome", func(_ *nats.Msg) {})
conn.Barrier(func() {
	conn.TLSRequired()
})
```

kozlovic:
I originally started `f` as a go routine, but decided not to, however doing so I should have released the locks. This is done now.

Related to #338 
Resolves #345

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -3077,12 +3077,11 @@ func (nc *Conn) TLSRequired() bool {
 // the call.
 func (nc *Conn) Barrier(f func()) error {
 	nc.mu.Lock()
-	defer nc.mu.Unlock()
 	if nc.isClosed() {
+		nc.mu.Unlock()
 		return ErrConnectionClosed
 	}
 	nc.subsMu.Lock()
-	defer nc.subsMu.Unlock()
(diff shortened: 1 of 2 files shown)


derekcollison:
LGTM

PR #348, opened 2018-03-06
[FIXED] Protocols received right after first PONG may be processed

kozlovic:
This manifested with the recent changes where server may send
an async INFO protocol right after sending the initial PONG to
the client.
After the client sends the CONNECT protocol, it expects to receive
a PONG. The reading of the protocol is done in place (in the
sendConnect() function) but a buffer reader was used. This is wrong
since any data/protocol sent right after by the server may be
consumed when trying to read the PONG instead of leaving that to
the readLoop.
Since we are just looking for the PONG (or OK and PONG) here, perform
a byte-by-byte read there. Alternatively, if doing a buffered read,
the buffer will have to be passed to the readLoop so that whatever
is in the buffer is read prior to resuming reading from socket.

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -11,6 +11,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"math/rand"
 	"net"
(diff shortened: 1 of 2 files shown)


nats.go
```diff
@@ -1283,6 +1285,24 @@ func (nc *Conn) sendConnect() error {
 	return nil
 }
 
+// read byte-by-byte from the connection until the given 'c' character is found.
+func (nc *Conn) readStringUpTo(c byte) (string, error) {
```
derekcollison: I think their is a func in the stdlib that does this no?

kozlovic: I could not find. We cannot use a buffered reader because it does fill up the buffer in the background, thing that we don't want to happen. I looked at a buffered reader with `io.LimitedReader`, but then you need to know the number of bytes you want to read.

I did rename the function to `readProto` and don't pass the delimiter anymore. This function is not meant to be an all purpose function.

derekcollison: ok

PR #349, opened 2018-03-06
Added NATSPublisher interface

kirooha:
There is a useful interface to implement it in other packages and extend the default Conn struct.

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -264,8 +264,8 @@ const (
 	nuidSize = 22
 )
 
-// A NatsPublisher publishes a NATS message.
-type NATSPublisher interface {
+// A Publisher publishes a NATS message.
+type Publisher interface {
 	// Publish publishes the data argument to the given subject. The data
 	// argument is left untouched and needs to be correctly interpreted on
 	// the receiver.
@@ -281,7 +281,7 @@ type NATSPublisher interface {
 	PublishRequest(subj, reply string, data []byte) error
 }
 
-// A Conn implements NATSPublisher and represents a bare connection to a nats-server.
+// A Conn implements Publisher and represents a bare connection to a nats-server.
 // It can send and receive []byte payloads.
 type Conn struct {
 	// Keep all members for which we use atomic at the beginning of the
--- a/nats.go
+++ b/nats.go
@@ -265,12 +265,20 @@ const (
 )
 
 // A NatsPublisher publishes a NATS message.
-//
-// Publish publishes the data argument to the given subject. The data
-// argument is left untouched and needs to be correctly interpreted on
-// the receiver.
 type NATSPublisher interface {
+	// Publish publishes the data argument to the given subject. The data
+	// argument is left untouched and needs to be correctly interpreted on
+	// the receiver.
 	Publish(string, []byte) error
+
+	// PublishMsg publishes the Msg structure, which includes the
+	// Subject, an optional Reply and an optional Data field.
+	PublishMsg(m *Msg) error
+
+	// PublishRequest will perform a Publish() excpecting a response on the
+	// reply subject. Use Request() for automatically waiting for a response
+	// inline.
+	PublishRequest(subj, reply string, data []byte) error
 }
 
 // A Conn implements NATSPublisher and represents a bare connection to a nats-server.
--- a/nats.go
+++ b/nats.go
@@ -264,7 +264,16 @@ const (
 	nuidSize = 22
 )
 
-// A Conn represents a bare connection to a nats-server.
+// A NatsPublisher publishes a NATS message.
+//
+// Publish publishes the data argument to the given subject. The data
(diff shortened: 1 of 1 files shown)


kirooha:
@derekcollison could you review my pr? I want to add nats transport to go-kit and need this interface for it.

nats.go
```diff
@@ -264,7 +264,16 @@ const (
 	nuidSize = 22
 )
 
-// A Conn represents a bare connection to a nats-server.
+// A NatsPublisher publishes a NATS message.
+//
+// Publish publishes the data argument to the given subject. The data
+// argument is left untouched and needs to be correctly interpreted on
+// the receiver.
+type NATSPublisher interface {
```
kozlovic: Could be just `Publisher` since it is package `nats`.
Should we add the other `Publish` functions there too?

kirooha: @kozlovic added

kozlovic: I think the name would be better if it was `Publisher`, not `NATSPublisher`. Also, about adding the other ones, it was a question ;-)
Had a quick look at go-kit and don't find an equivalent interface defined, so why do you need go-nats to have this interface?
Finally, at one point there were discussion that we should have nats.Conn be an interface instead of a struct, but that would be a 2.0 type of release. This needs more internal discussion.

kirooha: if there was `Conn` as the interface it would decide my problem.
You are right, go-kit doesn't have an equivalent of `Publisher` interface. The problem is my decorator to extend `nats.Conn`. I'm trying to create nats transport like http transport and there is `interceptingWriter` implementing `http.ResponseWriter`. I wanted to create something like this.
If you are interested you can look at my fork https://github.com/kirooha/kit/blob/master/transport/nats/server.go#L195 

Maybe the better solution is creating nats transport without this decorator. I mean client code will depend on `nats.Conn` not some interface of publishing. And when there is `Conn` as the interface I will able to add decorator. What do you think?

nats.go
```diff
@@ -264,7 +264,24 @@ const (
...
+	// Subject, an optional Reply and an optional Data field.
+	PublishMsg(m *Msg) error
+
+	// PublishRequest will perform a Publish() excpecting a response on the
+	// reply subject. Use Request() for automatically waiting for a response
+	// inline.
+	PublishRequest(subj, reply string, data []byte) error
+}
+
+// A Conn implements Publisher and represents a bare connection to a nats-server.
```
wallyqs: `Conn` would implement the `Publisher` interface here, but an interface type is more helpful when declared outside of the package where the implementation exists (so that you can use the behavior of a type without having to import the package)

kirooha: I don't want to abstract from nats and use `Publisher` as abstract publishing to somewhere. I offer `Publisher` as the interface to encapsulate implementation. It can be helpful in case when you want to expand behavior of `Conn`. 

In my point of view, in this case `Publisher` should be in the package.

kozlovic: Actually, I do agree with Wally. I am not sure why the interface in nats package would help nats users. If I was working on a project that need to integrate with different messaging vendors, I would probably come up with a common interface, and since likely each vendor would have a different send/publish method, I would have to have encapsulate each vendor and implement the publisher interface. If you could give me a concrete example on how the Publisher interface in nats package would help, that may change my opinion. Thanks!

kirooha: I will try to describe my concrete example in go-kit.
There is http transport in that package and also there is `http.ResponseWriter` interface in stdlib. Client code of go-kit decides to what write by this interface. It means client code mustn't know about the implementation of `http.ResponseWriter`. It gives an ability to go-kit to redefine that writer on its own writer. But client code still works with the interface of the `http` package.
For nats, client code must work with `Conn` struct and go-kit doesn't have the ability to redefine publisher. One of the possible solutions to define the interface in go-kit. But it doesn't release from importing of the `nats`. Also, we can use the `http` package as an example. Because there is `ResponseWriter` interface there. For me, it looks similar.

The difference between two solutions looks like this (it's client code)
`func encodeResponse(_ context.Context, reply string, p kitnatstransport.Publisher, response interface{}) error {`
`func encodeResponse(_ context.Context, reply string, p nats.Publisher, response interface{}) error {`

Client code for http transport
`func encodeResponse(_ context.Context, w http.ResponseWriter, response interface{}) error {`

I understand that you have only one implementation and don't see sense to add interface. But maybe it's a good solution to add some abstract contracts like `Publisher`, `Subscriber` and so on?

kirooha: @kozlovic what do you think? if you don't agree with my changes I will try to solve my case another way.

kozlovic: @kirooha Sorry for the delay. After some internal discussions, we still feel that having the interface defined outside of the implementation makes more sense. So at this point we are not going to include this in the package. We do appreciate you using NATS and your contribution and feedback. If we later decide to have such interface, we will reference this PR and reach out to you. Thanks again!

PR #356, opened 2018-03-23
Added test ensuring version matches tag on new release

kozlovic:
Doing a new release by pushing a tag will allow a test to run and
verify that the library version matches the given tag.

```
git tag -a vX.Y.Z -m "Release vX.Y.Z"
git push --tags
```

If the test fails, then we would delete the newly created tag,
fix the version string and recreate the tag.

### DIFF
--- a/nats_test.go
+++ b/nats_test.go
@@ -23,6 +23,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"os"
 	"reflect"
 	"runtime"
 	"strings"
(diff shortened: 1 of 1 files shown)


nats_test.go
```diff
@@ -64,6 +65,24 @@ func stackFatalf(t *testing.T, f string, args ...interface{}) {
...
+		t.SkipNow()
+	}
+	// We expect a tag of the form vX.Y.Z. If that's not the case,
+	// we need someone to have a look. So fail if first letter is not
+	// a `v`
+	if tag[0] != 'v' {
+		t.Fatalf("Expect tag to start with `v`, tag is: %s", tag)
+	}
+	// Strip the `v` from the tag for the version comparison.
+	tag = tag[1:]
```
derekcollison: Could just inline into comparison below but keep the comment.

derekcollison:
LGTM, minor nit but you can take it or leave it.

nats_test.go
kozlovic: Done

PR #358, opened 2018-03-26
doc(readme): add license scan status

xizhao:
Adds FOSSA scan status to README for CNCF license scans.

Ref: @caniszczyk


caniszczyk:
LGTM

@derekcollison this is part of our license scanning support in CNCF and required for CNCF projects

PR #361, opened 2018-03-31
Create PingInterval helper function

karpovicham:
Related to https://github.com/nats-io/go-nats/issues/359 issue.

The new `PingInterval` function make it simpler to configure NATS connection. It's created similar way to `ReconnectWait` function. https://github.com/nats-io/go-nats/blob/master/nats.go#L534

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -546,6 +546,14 @@ func MaxReconnects(max int) Option {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


derekcollison:
LGTM - General comment is we should test that the correct values are set, not just that an error was not returned creating the connection.

PR #362, opened 2018-05-02
Added additional JSON tests,

aricart:
Validate encoding/decoding of arrays, empty arrays, null (nil), true (bool), empty strings on the go client.

### DIFF
--- a/test/json_test.go
+++ b/test/json_test.go
@@ -52,6 +52,26 @@ func TestEncBuiltinJsonMarshalString(t *testing.T) {
 	}
 }
 
+func TestEncBuiltinJsonMarshalEmptyString(t *testing.T) {
+	s := RunServerOnPort(TEST_PORT)
+	defer s.Shutdown()
+
+	ec := NewJsonEncodedConn(t)
+	defer ec.Close()
+	ch := make(chan bool)
+
+	ec.Subscribe("json_empty_string", func(s string) {
+		if s != "" {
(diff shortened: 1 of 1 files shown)


test/json_test.go
```diff
@@ -74,6 +94,96 @@ func TestEncBuiltinJsonMarshalInt(t *testing.T) {
...
+	defer ec.Close()
+
+	type TestType struct{}
+	ch := make(chan bool)
+
+	var testValue *TestType
+
+	ec.Subscribe("json_null", func(i interface{}) {
+		if i != nil {
+			t.Fatalf("Received test bool of '%t', wanted 'nil'\n", i)
```
kozlovic: copy/paste error

test/json_test.go
```diff
@@ -74,6 +94,96 @@ func TestEncBuiltinJsonMarshalInt(t *testing.T) {
...
+func TestEncBuiltinJsonMarshalEmptyArray(t *testing.T) {
+	s := RunServerOnPort(TEST_PORT)
+	defer s.Shutdown()
+
+	ec := NewJsonEncodedConn(t)
+	defer ec.Close()
+
+	ch := make(chan bool)
+
+	var a = make([]string, 0)
```
kozlovic: Would have done that too for the test above, but here seem more natural to have: `a := []string{}`, but that may just be personal preference ;-)

PR #364, opened 2018-05-10
run the mcb hook in parallel, instead of running one by one

asxalex:
from my point of view, the subscription handler should be run in parallel, like an http server, every time the 'api'(a publication in nats) is called, the handler should be called, instead of waiting for the last one to finish. Making the running of "mcb" as a golang coroutine satisfies this need. please make me informed if the change spoils something or it would cause some disaster :)

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -1657,7 +1657,7 @@ func (nc *Conn) waitForMsgs(s *Subscription) {
 
 		// Deliver the message.
 		if m != nil && (max == 0 || delivered <= max) {
-			mcb(m)
+			go mcb(m)
 		}
 		// If we have hit the max for delivered msgs, remove sub.
 		if max > 0 && delivered >= max {


asxalex:
oh, this makes some test failed~

asxalex:
As i'd like to get the outcome of the mcb, and reply some message to the publisher. If i make my handler a go routine, whether communicating with chan or not, I still have to wait for the handler to finish.

kozlovic:
@asxalex As @ripienaar mentioned, you should have your message callback do something like: 
```
func (m *nats.Msg) {
  go handleMsg(m)
}
```
As you have seen from the tests and some (lots ?) users rely on the ordering of invocation of the message callback for a given subscription.

Therefore, we are closing this PR since there is a very easy way and non disruptive way to the community to implement your desired behavior.

Thank you for using NATS!

asxalex:
@kozlovic Ah, I confused myself, sorry for that. That solved my problem, thanks a lot :)

PR #365, opened 2018-05-18
[FIXED] Deadlocks related to Conn's callbacks

teh-cmc:
Hey, a bit of context first:
- `nc.ach` is the channel used to push user-defined callbacks within a `Conn`.
- (Almost) All user-defined callbacks take a `Conn` as parameter.
- `Conn` has a single global writer-only mutex protecting (almost) all of its fields.
- When a `Conn` has to process an error, it does something akin to the following:
```Go
nc.mu.Lock()
if nc.Opts.UserDefinedCB != nil {
    nc.ach <- nc.Opts.UserDefinedCB(nc)
}
nc.mu.Unlock()
```
- Error callbacks are called asynchronously by a dedicated goroutine that loops over `nc.ach`, basically:
```Go
for cb := range nc.ach {
    cb()
}
```

You see where this is going: since `UserDefinedCB` receives the current `Conn`, and since the `Conn` only has a single lock protecting its fields, should the callback passed by the user ever call any method on the `Conn`, it will try to grab that lock. E.g.:
```Go
func mycallback(conn *nats.Conn) { log.Println(conn.LastError()) }
```

Now, most of the time, this doesn't cause any trouble.  
In an overloaded client generating a lot of errors, though, this will lead to a deadlock as soon as the rate of errors exceeds the rate at which callbacks are executed, causing `nc.ach`'s buffered channel to be full.  
When that happens, the goroutine that's blocking on `nc.ach <- nc.Opts.UserDefinedCB(nc)` is still holding the lock, preventing the user callbacks from going forward and thus preventing the asynchronous dispatcher from consuming the channel: deadlock.

This actually happened to me in a production setting; not a fun time, I must say.

This PR adds an integration test showing the issue and solves it by always releasing the global lock prior to pushing a user defined callback in the async queue.  
This assumes that the end-user knows that the `Opts` field of a `Conn` is *not* thread-safe; but since that field is already public and has no associated lock, this was already the case.


derekcollison:
Great catch. Thanks for the feedback and apologies for the production issues. Might I suggest we snapshot the callbacks from opts in a local variable prior to releasing the lock. Otherwise I think LGTM.

test/conn_test.go
```diff
@@ -28,6 +28,7 @@ import (
 	"sync/atomic"
 	"testing"
 	"time"
+	_ "unsafe" // linkname
```
derekcollison: Make package nats to avoid the use of unsafe and alias to processErr.

test/conn_test.go
```diff
@@ -1961,3 +1962,42 @@ func TestReceiveInfoWithEmptyConnectURLs(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Should have connected ok: %v", err)
+	}
+
+	wg := &sync.WaitGroup{}
+	for i := 0; i < 10000; i++ {
+		wg.Add(1)
+		go func() {
+			// overwhelm asyncCB with errors
+			processErr(nc, nats.AUTHORIZATION_ERR)
```
derekcollison: With package of nats can call directly on nc.

teh-cmc:
> Might I suggest we snapshot the callbacks from opts in a local variable prior to releasing the lock.

Sure, I'd have no problem with that; but then again with `nc.Opts` being public, that would just provide a false/misleading sense of safety, imo.  
If you confirm that's the behaviour you want, I'll do the changes.

> Make package nats to avoid the use of unsafe and alias to processErr.

I'm not sure what you mean by "Make package nats"?

derekcollison:
Understood regarding opts. Just reading and placing into a local variable before unlocking is good practice. I agree issues with write and that its public.

If you make the package statement at the top of the test file nats vs test, then you can access the private method nc.processErr without importing unsafe, etc.

teh-cmc:
> If you make the package statement at the top of the test file nats vs test, then you can access the private method nc.processErr without importing unsafe, etc.

Still not sure I understand, sorry. That file is in a subpackage dedicated to integration tests, I'd have to move it up to the main package in order to rename it to `nats`; that would be quite the migration with lots of breaking changes... Is this what you're asking? Or am I missing something?

derekcollison:
If the test wants to use private methods from the package, place the test under the same package. Look at nats_test.go.

nats_test.go
```diff
@@ -1130,3 +1131,39 @@ func TestProcessErrAuthorizationError(t *testing.T) {
...
+	}
+	nc, err := o.Connect()
+	if err != nil {
+		t.Fatalf("Should have connected ok: %v", err)
+	}
+
+	wg := &sync.WaitGroup{}
+	for i := 0; i < 10000; i++ {
+		wg.Add(1)
+		go func() {
```
derekcollison: nit but probably do not need 1k Go routines, just one that does the iteration loop to cause the channel to become full.

kozlovic: Actually would not even work if running with `-race`, which we do, because there is a limit of 8192 go routines in that mode.

nats_test.go
```diff
@@ -1130,3 +1131,39 @@ func TestProcessErrAuthorizationError(t *testing.T) {
...
+		// do something with nc that requires locking behind the scenes
+		_ = nc.LastError()
+	}
+	nc, err := o.Connect()
+	if err != nil {
+		t.Fatalf("Should have connected ok: %v", err)
+	}
+
+	wg := &sync.WaitGroup{}
+	for i := 0; i < 10000; i++ {
```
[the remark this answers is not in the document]
teh-cmc: Sure, might as well now that we're here! Exactly `cap(nc.ach)` won't be enough to make sure we meet the right timings every time though; `cap(nc.ach)*10` seems to do the trick no matter how many Ps I throw at it.

teh-cmc:
I've submitted the modification to avoid spawning too many goroutines.
This will fix the issue with `-race`, although running the test suite with the `-race` will probably never catch the issue anyway as it slows down the program too much to get the right timing.

I cannot allocate more time to this, unfortunately.
I'd suggest we merge this for now, so the issue is fixed on master while you guys find a better way to fix the root of the issue as @kozlovic suggests.

Thanks

kozlovic:
@teh-cmc Thanks a lot for your contribution and time! Could you just remove the `_unsafe` import that is no longer needed? Then we can merge.

derekcollison:
Good point Ivan on moving to LL and signaling since we already have that code in the client. Separate PR though IMO.

PR #369, opened 2018-06-12
[FIXED] Dispatching of asynchronous callbacks

kozlovic:
This is a follow up on #365. Working on another issue and running
tests, I still had a race on the original ach or a panic on send
to close channel.

Move from a channel to a linked list with dedicated lock.

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -761,7 +761,8 @@ func (o Options) Connect() (*Conn, error) {
 	}
 
 	// Create the async callback handler.
-	nc.ach = newAsyncCallbacksHandler()
+	nc.ach = &asyncCallbacksHandler{}
+	nc.ach.cond = sync.NewCond(&nc.ach.mu)
 
 	if err := nc.connect(); err != nil {
 		return nil, err
@@ -1542,13 +1543,6 @@ func (nc *Conn) processOpErr(err error) {
 	nc.err = err
 	nc.mu.Unlock()
 	nc.Close()
-}
-
-// Returns an initialized asyncCallbacksHandler object
-func newAsyncCallbacksHandler() *asyncCallbacksHandler {
-	ac := &asyncCallbacksHandler{}
-	ac.cond = sync.NewCond(&ac.mu)
-	return ac
 }
 
 // dispatch is responsible for calling any async callbacks
--- a/nats.go
+++ b/nats.go
@@ -1605,7 +1605,7 @@ func (ac *asyncCallbacksHandler) pushOrClose(f func(), close bool) {
 		ac.head = cb
 	}
 	ac.tail = cb
-	ac.cond.Signal()
+	ac.cond.Broadcast()
 }
 
 // readLoop() will sit on the socket reading and processing the
--- a/nats.go
+++ b/nats.go
@@ -130,7 +130,17 @@ type ConnHandler func(*Conn)
 type ErrHandler func(*Conn, *Subscription, error)
 
 // asyncCB is used to preserve order for async callbacks.
-type asyncCB func()
+type asyncCB struct {
+	f    func()
+	next *asyncCB
+}
+
+type asyncCallbacksHandler struct {
+	mu   sync.Mutex
+	cond *sync.Cond
+	head *asyncCB
+	tail *asyncCB
+}
 
 // Option is a function on the options for a connection.
 type Option func(*Options) error
@@ -269,9 +279,6 @@ const (
 	// Default server pool size
 	srvPoolSize = 4
 
-	// Channel size for the async callback handler.
-	asyncCBChanSize = 32
-
 	// NUID size
 	nuidSize = 22
 )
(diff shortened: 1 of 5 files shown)


nats.go
```diff
@@ -1542,35 +1544,70 @@ func (nc *Conn) processOpErr(err error) {
...
+
+// Add the given function to the tail of the list and
+// signals the dispatcher.
+func (ac *asyncCallbacksHandler) pushOrClose(f func(), close bool) {
+	ac.mu.Lock()
+	defer ac.mu.Unlock()
+	// Make sure that library is not calling push with nil function,
+	// since this is used to notify the dispatcher that it should stop.
+	if !close && f == nil {
+		panic("pushing a nil callback")
```
derekcollison: Do we have any other panics in our library?

nats.go
```diff
@@ -1542,35 +1544,70 @@ func (nc *Conn) processOpErr(err error) {
...
-	return func() {
-		nc.mu.Lock()
-		if nc.ach != nil {
-			close(nc.ach)
-			nc.ach = nil
-		}
-		nc.mu.Unlock()
-	}
+// Returns an initialized asyncCallbacksHandler object
+func newAsyncCallbacksHandler() *asyncCallbacksHandler {
```
derekcollison: Seems to be only called once and very short, maybe just inline?

nats.go
```diff
@@ -1542,35 +1544,70 @@ func (nc *Conn) processOpErr(err error) {
...
+// signals the dispatcher.
+func (ac *asyncCallbacksHandler) pushOrClose(f func(), close bool) {
+	ac.mu.Lock()
+	defer ac.mu.Unlock()
+	// Make sure that library is not calling push with nil function,
+	// since this is used to notify the dispatcher that it should stop.
+	if !close && f == nil {
+		panic("pushing a nil callback")
+	}
+	cb := &asyncCB{f: f}
```
derekcollison: if close is true does f have to be nil?

nats.go
```diff
@@ -1542,35 +1544,70 @@ func (nc *Conn) processOpErr(err error) {
...
+	if ac.tail != nil {
+		ac.tail.next = cb
+	} else {
+		ac.head = cb
+	}
+	ac.tail = cb
+	ac.cond.Signal()
```
derekcollison: For close scenarios I like this to be a Broadcast instead.

test/conn_test.go
```diff
@@ -817,6 +824,15 @@ func TestCallbacksOrder(t *testing.T) {
 	if rtime.Before(dtime1) || dtime2.Before(rtime) || atime2.Before(atime1) || ctime.Before(atime2) {
 		t.Fatalf("Wrong callback order:\n%v\n%v\n%v\n%v\n%v\n%v", dtime1, rtime, atime1, atime2, dtime2, ctime)
 	}
+
+	// Close the other connection
+	ncp.Close()
+
+	// Give a chance for the go routine to stop
+	time.Sleep(150 * time.Millisecond)
```
derekcollison: Be good if this was deterministic instead of time based.

nats.go
kozlovic: This is not user accessible. This is to protect us, writers of the library. That is, to catch a case where we would add some code that happen to call push() with a nil callback (if we were not checking nc.Opts.<callback name>).

nats.go
kozlovic: actually called in a test too.

nats.go
kozlovic: I originally used a `closed` boolean in the structure but thought that I could do without and pass nil as the last function to indicate close event. But yes, f will be nil if close is true. Again, this is not user accessible and only close() will call pushOrClose with `nil, true`

nats.go
kozlovic: Sorry, but there is only one go routine ever calling Wait() so Signal() is enough.

test/conn_test.go
kozlovic: I could add specific wait group in the struct and use that for the test, but I feel it is a bit too much. What I could do is repeat the test with a little delay in case of failure and fail the test after few attempts. I will do that.

nats.go
derekcollison: I know that is the desired state. But kind of like your panic thought best to kick out any and all IMO.

kozlovic: Ok, makes sense.

nats.go
```diff
@@ -1605,7 +1605,7 @@ func (ac *asyncCallbacksHandler) pushOrClose(f func(), close bool) {
 		ac.head = cb
 	}
 	ac.tail = cb
-	ac.cond.Signal()
+	ac.cond.Broadcast()
```
derekcollison: I meant only on close condition. Signal as default is fine.

derekcollison:
LGTM

nats.go
kozlovic: Done.. but to be honest, this is some code that one will scratch their head in few months/years and ask themselves: why is that? ;-)
If there is a chance that we have multiple go routines calling Wait() then it should always be Broadcast(), but that would defeat the purpose of this code anyway since it is meant to ensure that callbacks are dispatched in order.

PR #370, opened 2018-06-12
[FIXED] Close() may not release resources immediately

### LINKED ISSUE #368: Close() doesn't appear to free resources properly.
I noticed a memory issue with an application that periodically (every few mins) opens a nats connection, does some work and closes it again.

I reproduced the issue with the below code.

```
package main

import (
	"net/http"
	_ "net/http/pprof"
	"time"

	"github.com/nats-io/go-nats"
)

func main() {
	go func() {
		panic(http.ListenAndServe(":6060", nil))
	}()

	for {
		err := openAndClose()
		if err != nil {
			panic(err)
		}
		// Don't hurt things too quickly.
		time.Sleep(10 * time.Millisecond)
	}
}

func openAndClose() error {
	conn, err := nats.Connect("http://localhost:4222", nats.Name("test-client-id-2"))
	if err != nil {
		return err
	}
	conn.Close()
	return nil
}
```

pprof heap output suggests the memory usage is due to allocations in `github.com/nats-io/go-nats.(*Conn).createConn`  which suggests (to me at least) that something is retaining a reference to the connections and the associated resources are not being freed.

kozlovic:
When a connection is closed very quickly after being created, there
was a race condition that could lead to the connection object still
being referenced (and therefore holding resources) until the first
ping interval elapsed.
This was because the pinger was created in spinGoRoutines() which
was itself created as a go routine. When the Close() call was made,
the ping timer may still be nil (therefore no way to stop it), but
then created in the go routine (which caused the reference to the
connection object). By default, it means that the connection object
would be held for 30 seconds.
Doing connect/close in a tight loop would show the process size
growing.

This PR refactors where we create the pinger and the use of the
wait group. There had been a change in a past to attempt to fix
a WaitGroup panic. I believe this approach is more stable. We do
create the pinger and bump the wait group in processConnectInit,
under connection lock. The doReconnect() go routine waits for
this wait group on entry and then on iterations where a failure
occurs after we know we have started the go routines.
In case of failures in the reconect loop, some state has also been
properly reset (namely the connection's buffered writer).

Resolves #368

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -295,7 +295,7 @@ type Conn struct {
 	// Opts holds the configuration of the Conn.
 	// Modifying the configuration of a running Conn is a race.
 	Opts    Options
-	wg      *sync.WaitGroup
+	wg      sync.WaitGroup
 	url     *url.URL
 	conn    net.Conn
 	srvPool []*srv
@@ -992,7 +992,7 @@ func (nc *Conn) makeTLSConn() {
 
 // waitForExits will wait for all socket watcher Go routines to
 // be shutdown before proceeding.
-func (nc *Conn) waitForExits(wg *sync.WaitGroup) {
+func (nc *Conn) waitForExits() {
 	// Kick old flusher forcefully.
 	select {
 	case nc.fch <- struct{}{}:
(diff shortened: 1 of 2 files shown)


derekcollison:
LGTM

PR #371, opened 2018-06-20
Removed use of ssl_required in test

kozlovic:
Some tests did still reference ssl_required (which has been
removed, tls_required is the proper field name). This has no
impact since in these tests, the value was set to false.

### DIFF
--- a/test/conn_test.go
+++ b/test/conn_test.go
@@ -437,7 +437,7 @@ func TestMoreErrOnConnect(t *testing.T) {
 				// Stick around a bit
 				<-case1
 			case 2:
-				info := fmt.Sprintf("INFO {\"server_id\":\"foobar\",\"version\":\"0.7.3\",\"go\":\"go1.5.1\",\"host\":\"%s\",\"port\":%d,\"auth_required\":false,\"tls_required\":false,\"max_payload\":1048576}\r\n", addr.IP, addr.Port)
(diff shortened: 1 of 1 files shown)


test/conn_test.go
```diff
@@ -437,7 +437,7 @@ func TestMoreErrOnConnect(t *testing.T) {
 				// Stick around a bit
 				<-case1
 			case 2:
-				info := fmt.Sprintf("INFO {\"server_id\":\"foobar\",\"version\":\"0.7.3\",\"go\":\"go1.5.1\",\"host\":\"%s\",\"port\":%d,\"auth_required\":false,\"ssl_required\":false,\"max_payload\":1048576}\r\n", addr.IP, addr.Port)
+				info := fmt.Sprintf("INFO {\"server_id\":\"foobar\",\"version\":\"0.7.3\",\"go\":\"go1.5.1\",\"host\":\"%s\",\"port\":%d,\"auth_required\":false,\"tls_required\":false,\"max_payload\":1048576}\r\n", addr.IP, addr.Port)
```
derekcollison: Update gnatsd version and go version?

kozlovic: These are not used at all (especially on those tests), so I wonder if I should completely remove that from those hand crafted INFO? We are not going to update it every time there is an update ;-)

derekcollison: I like that idea..

PR #375, opened 2018-07-25
NoEcho Option Support

derekcollison:
Added support to pass in a NoEcho option. Old servers that do not support will error.

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -88,6 +88,7 @@ var (
 	ErrInvalidMsg           = errors.New("nats: invalid message or message nil")
 	ErrInvalidArg           = errors.New("nats: invalid argument")
 	ErrInvalidContext       = errors.New("nats: invalid context")
+	ErrNoEchoNotSupported   = errors.New("nats: no echo option not supported by this server")
 	ErrStaleConnection      = errors.New("nats: " + STALE_CONNECTION)
 )
 
@@ -166,6 +167,11 @@ type Options struct {
 	// NoRandomize configures whether we will randomize the
 	// server pool.
 	NoRandomize bool
+
+	// NoEcho configures whether the server will echo back messages
+	// that are sent on this connection if we also have matching subscriptions.
+	// Note this is supported on servers >= version 1.2. Proto 1 or greater.
+	NoEcho bool
 
 	// Name is an optional name label which will be sent to the server
 	// on CONNECT to identify the client.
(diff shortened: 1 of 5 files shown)


nats.go
```diff
@@ -1203,18 +1220,25 @@ func (nc *Conn) connectProto() (string, error) {
...
+
+	cinfo := connectInfo{o.Verbose, o.Pedantic, user, pass, token,
+		o.Secure, o.Name, LangString, Version, clientProtoInfo, !o.NoEcho}
+
 	b, err := json.Marshal(cinfo)
 	if err != nil {
 		return _EMPTY_, ErrJsonParse
 	}
+
+	// Check if NoEcho is set and we have a server that supports it.
```
kozlovic: Why not do this test in `processExpectedInfo()` instead?

nats_test.go
```diff
@@ -1164,3 +1165,60 @@ func TestPingTimerLeakedOnClose(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Error on subscribe: %v", err)
+	}
+
+	err = nc.Publish("foo", []byte("Hello World"))
+	if err != nil {
+		t.Fatalf("Error on publish: %v", err)
+	}
+	nc.Flush()
+	nc.Flush()
```
kozlovic: intentional to have 2?

kozlovic: Also, especially in Travis, this is not enough to guarantee that there is no bug and message is not going to be incorrectly delivered. Maybe you would need 1 flush, but using a different connection, publish an other message, that message should be received, but not the first one.

All that being said, I do believe that the feature is tested in the server. We should focus here in testing APIs related features, which is make sure that the "echo" field is set as expected in the INFO, etc...

nats.go
derekcollison: I could if you think that is a better place. When forming the connectProto we know we have received the server info and are processing the options to some degree, hence why I stuck it there.

nats_test.go
derekcollison: Yes the two was intentional instead of having to do a sleep or gosched() call since first may have the scheduler unblock the flush call before firing the async callback. I could make it sync maybe.. Or just comment as to why there are two.

nats.go
kozlovic: That's fine I guess, although you could do this check at top of function, since there is no point in marshalling etc.. if we are going to return an error.

derekcollison: I wanted a marshal error to take precedence.

kozlovic: I see.

kozlovic:
LGTM

PR #376, opened 2018-08-06
Context tests fixes

wallyqs:
There were some context usage related failures in Travis due to timing issues:

```
--- FAIL: TestOldContextRequestWithTimeoutCanceled (0.03s)
	context_test.go:141: Expected request with timeout context to fail: %!s(<nil>)
```
https://travis-ci.org/nats-io/go-nats/jobs/412394490#L693

In order to prevent this, added an extra wait to the context tests that assumed that the context cancellation was already propagated.

### DIFF
--- a/test/cluster_test.go
+++ b/test/cluster_test.go
@@ -17,7 +17,6 @@ import (
 	"fmt"
 	"math"
 	"net"
-	"regexp"
 	"runtime"
 	"strings"
 	"sync"
@@ -168,8 +167,9 @@ func TestAuthServers(t *testing.T) {
 		t.Fatalf("Expect Auth failure, got no error\n")
 	}
 
-	if matched, _ := regexp.Match(`authorization`, []byte(err.Error())); !matched {
-		t.Fatalf("Wrong error, wanted Auth failure, got '%s'\n", err)
+	expected := `authorization`
+	if !strings.Contains(err.Error(), expected) {
+		t.Fatalf("Wrong error, wanted Auth failure, got '%s'\n", err.Error())
 	}
 
 	// Test that we can connect to a subsequent correct server.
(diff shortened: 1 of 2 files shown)


test/context_test.go
```diff
@@ -136,6 +137,24 @@ func testContextRequestWithTimeoutCanceled(t *testing.T, nc *nats.Conn) {
 	// Cancel the context already so that rest of requests fail.
 	cancelCB()
 
+	// Wait a bit for context to be eventually canceled.
+	errCh := make(chan error)
+	tf := time.AfterFunc(50*time.Millisecond, func() {
```
derekcollison: Sometimes can't avoid sleeps but where possible I like for this to be more deterministic. I added a waitFor(condition, timeout) helper function in my drain branch.

test/cluster_test.go
```diff
@@ -168,8 +167,9 @@ func TestAuthServers(t *testing.T) {
 		t.Fatalf("Expect Auth failure, got no error\n")
 	}
 
-	if matched, _ := regexp.Match(`authorization`, []byte(err.Error())); !matched {
-		t.Fatalf("Wrong error, wanted Auth failure, got '%s'\n", err)
+	expected := `authorization`
```
derekcollison: I fixed this one too in drain branch ;)

test/context_test.go
wallyqs: great, there might be a few staticcheck related conflicts with the drain branch so thinking on branching from it and already adopt the new helper instead

derekcollison: ok I can merge your changes into the drain branch if you want. If so just cancel this PR.

test/cluster_test.go
wallyqs: PR to include fixes into the `drain` branch instead is here: https://github.com/nats-io/go-nats/pull/377 
closing this one for now

PR #378, opened 2018-08-06
Support for Drain Mode

derekcollison:
This PR allows clients to drain subscriptions or complete connections. This allows for graceful shutdown of clients, especially queue subscribers.

on a connection drain, any errors will fire the asyncErrCB. When all publish messages have been flushed and all subscriptions drained properly the connection will move to a closed state.

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -3341,7 +3341,7 @@ func (nc *Conn) isReconnecting() bool {
 
 // Test if Conn is connected or connecting.
 func (nc *Conn) isConnected() bool {
-	return nc.status == CONNECTED
+	return nc.status == CONNECTED || nc.isDraining()
 }
 
 // Test if Conn is in the draining state.
--- a/nats.go
+++ b/nats.go
@@ -1735,8 +1735,6 @@ func (nc *Conn) waitForMsgs(s *Subscription) {
 				continue
 			}
 			msgLen = len(m.Data)
-			//			s.pMsgs--
-			//			s.pBytes -= len(m.Data)
 		}
 		mcb := s.mcb
 		max = s.max
@@ -3218,9 +3216,8 @@ func (nc *Conn) drainConnection() {
 
 	// Do subs first
 	for _, s := range subs {
-		err := s.Drain()
-		// We will notify about these but continue.
-		if err != nil {
+		if err := s.Drain(); err != nil {
+			// We will notify about these but continue.
 			pushErr(err)
 		}
 	}
@@ -3256,10 +3253,11 @@ func (nc *Conn) drainConnection() {
 	nc.Close()
 }
 
-// Drain will put a connection into a drain state. Publishes after this call
-// will fail. All subscriptions will be put into a drain state. Upon completion
-// the connection will be closed. Use the ClosedCB() option to know when the
-// connection has moved from draining to closed.
+// Drain will put a connection into a drain state. All subscriptions will
+// immediately be put into a drain state. Upon completion, the publishers
+// will be drained and can not publish any additional messages. Upon draining
+// of the publishers, the connection will be closed. Use the ClosedCB()
+// option to know when the connection has moved from draining to closed.
 func (nc *Conn) Drain() error {
 	nc.mu.Lock()
 	defer nc.mu.Unlock()
--- a/nats.go
+++ b/nats.go
@@ -125,7 +125,8 @@ const (
 	CLOSED
 	RECONNECTING
 	CONNECTING
-	DRAINING
+	DRAINING_SUBS
+	DRAINING_PUBS
 )
 
 // ConnHandler is used for asynchronous events such as
@@ -1703,8 +1704,19 @@ func (nc *Conn) waitForMsgs(s *Subscription) {
 	var closed bool
 	var delivered, max uint64
 
+	// Used to account for adjustments to sub.pBytes when we wrap back around.
+	msgLen := -1
+
 	for {
 		s.mu.Lock()
+		// Do accounting for last msg delivered here so we only lock once
+		// and drain state trips after callback has returned.
+		if msgLen >= 0 {
+			s.pMsgs--
+			s.pBytes -= msgLen
+			msgLen = -1
+		}
+
 		if s.pHead == nil && !s.closed {
 			s.pCond.Wait()
 		}
@@ -1722,8 +1734,9 @@ func (nc *Conn) waitForMsgs(s *Subscription) {
 				}
 				continue
 			}
-			s.pMsgs--
-			s.pBytes -= len(m.Data)
+			msgLen = len(m.Data)
+			//			s.pMsgs--
+			//			s.pBytes -= len(m.Data)
 		}
 		mcb := s.mcb
 		max = s.max
@@ -2139,7 +2152,7 @@ func (nc *Conn) publish(subj, reply string, data []byte) error {
 		return ErrConnectionClosed
 	}
 
-	if nc.isDraining() {
+	if nc.isDrainingPubs() {
 		nc.mu.Unlock()
 		return ErrConnectionDraining
 	}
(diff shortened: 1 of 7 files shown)


derekcollison:
This should force a version bump for new functionality. I will add.

kozlovic:
@derekcollison This seems overly complex. As I see it works only for non channel subscriptions, why not Drain be a a _close() with a param that sets a subscription's drain boolean to true which then ensure that all messages are dispatched before the waitForMsgs loop exits?

kozlovic:
that is, _close() is already going through list of subs to notify them that they are closed. If a boolean `drain` was set to true in case of a Drain() then the waitForMsgs loop would not exit simply because `closed` is true if `drain` is also true. It will dispatch all messages that it has. I think we should still honor max delivered messages though.

derekcollison:
I can not control what a client will do with a chan of messages, I have no way of semantically understanding that. But agree this may not help those although it will make sure they are on the channel before moving to a closed state. So we could show an example to have a second barrier to closedCB that makes sure all channels are len() == 0.

derekcollison:
I do not want to close a connection until I know that all published messages are processed and all subscriptions have been drained.

kozlovic:
Drain() could be Flush() then _close() with new boolean asking to drain.

derekcollison:
Possibly, but I see _close as a func that tells server to close, that is why I did not overload it for Drain().

kozlovic:
I was not complaining that it won't work for chan subs, I understand we can't, that's why Barrier() works only for async subscriptions.

derekcollison:
Also thinking more about it, _close preps for releasing everybody without a response etc, like flushers. Drain should be more friendly until it times out or has completed.

test/drain_test.go
```diff
@@ -0,0 +1,286 @@
...
+	// Should happen quickly that we get to zero, so do not need to wait long.
+	waitFor(t, 2*time.Second, 10*time.Millisecond, func() error {
+		if numSubs := nc.NumSubscriptions(); numSubs != 0 {
+			return fmt.Errorf("Expected no subscriptions, got %d\n", numSubs)
+		}
+		return nil
+	})
+}
+
+func TestDrainSlowSubscriber(t *testing.T) {
```
kozlovic: Not sure what this test is doing. Don't see a Drain() in there...

nats.go
```diff
@@ -3089,6 +3181,91 @@ func (nc *Conn) IsConnected() bool {
...
 }
 
+// drainConnection will run in a separate Go routine and will
+// flush all publishes and drain all active subscriptions.
+func (nc *Conn) drainConnection() {
+	// Snapshot subs list.
+	nc.mu.Lock()
+	subs := make([]*Subscription, 0, len(nc.subs))
+	for _, s := range nc.subs {
+		subs = append(subs, s)
```
kozlovic: Since we can't Drain a non async subscription, should we select only async subscriptions here? But if that's the case, then checking on `NumSubscriptions() == 0` is kind of wrong.

derekcollison: No because this makes sure the messages are in the channel for sync and they can do what they want with those messages.

derekcollison:
I have a logic error that I thought of last night. We want the subscribers to be able to respond to requests, so publish needs to work until subscribers are drained. Then we can turn off publisher and flush. Will fix today.

nats.go
```diff
@@ -1705,8 +1734,9 @@ func (nc *Conn) waitForMsgs(s *Subscription) {
...
 				continue
 			}
-			s.pMsgs--
-			s.pBytes -= len(m.Data)
+			msgLen = len(m.Data)
+			//			s.pMsgs--
+			//			s.pBytes -= len(m.Data)
 		}
 		mcb := s.mcb
 		max = s.max
```
kozlovic: I believe there is a condition where with auto-unsub the code below may remove the sub but not update pMsgs which would cause Drain() to timeout since the value would not go to 0. Maybe in this case we need to artificially set it to 0 in code below that looks for a barrier message.

nats.go
```diff
@@ -3089,6 +3194,98 @@ func (nc *Conn) IsConnected() bool {
...
+		nc.mu.Lock()
+		if errCB != nil {
+			nc.ach.push(func() { errCB(nc, nil, err) })
+		}
+		nc.mu.Unlock()
+	}
+
+	// Do subs first
+	for _, s := range subs {
+		err := s.Drain()
```
kozlovic: Err handling could be inlined.

nats.go
```diff
@@ -3089,6 +3194,98 @@ func (nc *Conn) IsConnected() bool {
...
+		pushErr(err)
+		nc.Close()
+		return
+	}
+
+	// Move to closed state.
+	nc.Close()
+}
+
+// Drain will put a connection into a drain state. Publishes after this call
```
kozlovic: Should update comment here. Publishes calls do not really fail after this call, but some time after that ;-)
Also, I think it says it (with reference to "Upon completion ..."), but maybe we should make it clear that this call does not block waiting for drain to complete.

test/drain_test.go
```diff
@@ -0,0 +1,296 @@
...
+	defer nc.Close()
+
+	received := int32(0)
+	expected := int32(50)
+	sleep := 10 * time.Millisecond
+
+	// Create a slow subscriber
+	sub, err := nc.Subscribe("foo", func(_ *nats.Msg) {
+		time.Sleep(sleep)
+		atomic.AddInt32(&received, 1)
```
kozlovic: Could we add to this test a subscription on "bar" that this callback sends a message to? This way, we would check that publishes are ok and ensure that all messages on bar are received too.

nats.go
derekcollison: fixed.

nats.go
```diff
@@ -1705,8 +1734,9 @@ func (nc *Conn) waitForMsgs(s *Subscription) {
 				}
 				continue
 			}
-			s.pMsgs--
-			s.pBytes -= len(m.Data)
+			msgLen = len(m.Data)
+			//			s.pMsgs--
```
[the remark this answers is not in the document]
derekcollison: fixed

nats.go
derekcollison: fixed

test/drain_test.go
derekcollison: fixed

derekcollison:
Test added for auto-unsub. All issues addressed, please take another look. Thanks for all the feedback.

nats.go
kozlovic: Actually removeSub() will close the sub, and the drain checks on sub's closed boolean too so that is ok.

nats.go
```diff
@@ -3149,6 +3344,16 @@ func (nc *Conn) isConnected() bool {
 	return nc.status == CONNECTED
```
wallyqs: I wonder if we should consider the draining states as being connected as well, since looks like the flusher would bail once it is not in the CONNECTED state: https://github.com/nats-io/go-nats/blob/e04b7c4dc201c9e648595d08885d8fe50b5d0721/nats.go#L1904-L1908
Should the flusher be running while in the draining subs/pubs state?  Although once the subs/pubs have been drained the call to `Flush()` before closing does a `bw.Flush()` the internal buffer anyway: https://github.com/nats-io/go-nats/blob/e04b7c4dc201c9e648595d08885d8fe50b5d0721/nats.go#L2841

kozlovic: Very good catch... @derekcollison Maybe we should not use a connection status, but just booleans in connection object so we don't have this kind of side effects...

derekcollison: Yes good point. Since we use flush manually in the drain state it may be ok, but I will double check tomorrow and make sure. Thanks.

derekcollison:
ok should be good to go.

kozlovic:
LGTM

PR #381, opened 2018-08-13
[IMPROVED] URL parsing allowing to skip scheme and default port

kozlovic:
With this change, you could pass URL "host" and it will translate
to "nats://host:4222". Also, if attempting to connect to a server
that requires TLS, the Secure option will be automatically turned
on and an attempt to make a secure connection will be made.

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -297,6 +297,9 @@ const (
 
 	// NUID size
 	nuidSize = 22
+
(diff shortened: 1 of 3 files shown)


kozlovic:
@derekcollison Not sure if this is exactly what you had in mind.

PR #385, opened 2018-08-24
Allow nil options in the nats.Connect API

ColinSullivan1:
Users may want to setup a pattern using options that could be nil. 

An example would be:

```go
var ca, cert nats.Option

if (cafile != "") {
    ca = nats.RootCAs(cafile)
}
if (certfile != "") {
   cert = nats.ClientCert(certfile, keyfile)
}

nc, err := nats.Connect("nats://myhost:4222", ca, cert)
```

Today this would result in a fault if an option isn't set, e.g. if a CA wasn't used.  This PR allows this pattern.

Signed-off-by: Colin Sullivan <colin@synadia.com>

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -465,8 +465,10 @@ func Connect(url string, options ...Option) (*Conn, error) {
 	opts := GetDefaultOptions()
 	opts.Servers = processUrlString(url)
 	for _, opt := range options {
-		if err := opt(&opts); err != nil {
-			return nil, err
+		if opt != nil {
+			if err := opt(&opts); err != nil {
+				return nil, err
+			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


kozlovic:
LGTM

nats.go
```diff
@@ -465,8 +465,10 @@ func Connect(url string, options ...Option) (*Conn, error) {
 	opts := GetDefaultOptions()
 	opts.Servers = processUrlString(url)
 	for _, opt := range options {
-		if err := opt(&opts); err != nil {
-			return nil, err
+		if opt != nil {
```
derekcollison: Could be one line. 

```
if opt != nil && err := opt(&opts); err != nil
```

PR #387, opened 2018-08-26
[FIXED] Request or NextMsg with Context may return message after being canceled

wallyqs:
Currently behavior of context in the client is not deterministic when the context has already been canceled and is possible for a client to still receive messages from a request, as shown in the test occasionally failing in Travis:

```
--- FAIL: TestOldContextRequestWithTimeout (0.13s)
    context_test.go:91: Expected request with context to fail
```

In order to fix this, now we check first whether the context has been canceled already previous to waiting for the message to prevent successfully making requests using an invalid context.

### DIFF
--- a/context.go
+++ b/context.go
@@ -31,6 +31,13 @@ func (nc *Conn) RequestWithContext(ctx context.Context, subj string, data []byte
 	if nc == nil {
 		return nil, ErrInvalidConnection
 	}
+	// Check whether the context is done already before making
+	// the request.
+	select {
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	default:
+	}
 
 	nc.mu.Lock()
 	// If user wants the old style.
@@ -116,6 +123,11 @@ func (s *Subscription) NextMsgWithContext(ctx context.Context) (*Msg, error) {
 	if s == nil {
 		return nil, ErrBadSubscription
 	}
+	select {
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	default:
+	}
 
 	s.mu.Lock()
 	err := s.validateNextMsgState()
@@ -124,7 +136,6 @@ func (s *Subscription) NextMsgWithContext(ctx context.Context) (*Msg, error) {
 		return nil, err
 	}
 
-	// snapshot
 	mch := s.mch
 	s.mu.Unlock()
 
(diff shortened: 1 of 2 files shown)


context.go
```diff
@@ -31,6 +31,13 @@ func (nc *Conn) RequestWithContext(ctx context.Context, subj string, data []byte
 	if nc == nil {
 		return nil, ErrInvalidConnection
 	}
+	// Check whether the context is done already before making
+	// the request.
+	select {
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	default:
+	}
```
wallyqs: Without this check, if a message was received fast enough then it would have succeeded sometimes even if the `context.Done()` channel was already closed because of `select` randomly choosing to receive the channel:
https://github.com/nats-io/go-nats/blob/2f0e8883ae354dfdeb89680e43c8c6918d34b815/context.go#L66-L84

context.go
```diff
@@ -31,6 +31,13 @@ func (nc *Conn) RequestWithContext(ctx context.Context, subj string, data []byte
 	if nc == nil {
 		return nil, ErrInvalidConnection
 	}
+	// Check whether the context is done already before making
+	// the request.
+	select {
```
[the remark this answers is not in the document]
ripienaar: Godoc says

```
	// If Done is not yet closed, Err returns nil.
  	// If Done is closed, Err returns a non-nil error explaining why:
  	// Canceled if the context was canceled
  	// or DeadlineExceeded if the context's deadline passed.
  	// After Err returns a non-nil error, successive calls to Err return the same error.
  	Err() error
```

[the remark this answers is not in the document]
wallyqs: Right, no need for `select` here since there would be already an error when canceled. Thanks.

context.go
```diff
@@ -33,10 +33,8 @@ func (nc *Conn) RequestWithContext(ctx context.Context, subj string, data []byte
 	}
 	// Check whether the context is done already before making
 	// the request.
-	select {
-	case <-ctx.Done():
+	if ctx.Err() != nil {
```
derekcollison: Can this condition change after this call but before we evaluate ctx->Done() below? Leading to the same problem as before but with a slimmer chance of happening? /cc @kozlovic

derekcollison:
LGTM

context.go
derekcollison: Sounds good.

kozlovic:
LGTM

PR #389, opened 2018-08-30
Synchronous write mode

gwik:
In order to better support a more "realtime" use case, this adds a synchronous mode where publishing and subscribing writes to the socket immediately.
Obviously the throughput in this mode is much lower than with buffered/async flush. But the throughput for a single connection is largely enough for us (and probably anyone) and it matches our use case better.

This is still work in progress but I wanted to allow you to tell me what you think early.
This lacks test and probably the SubscribeSync option is badly named as it collides with a method name with a different meaning.

### DIFF
--- a/examples/nats-bench.go
+++ b/examples/nats-bench.go
@@ -48,6 +48,7 @@ func main() {
 	var numMsgs = flag.Int("n", DefaultNumMsgs, "Number of Messages to Publish")
 	var msgSize = flag.Int("ms", DefaultMessageSize, "Size of the message.")
 	var csvFile = flag.String("csv", "", "Save bench data to csv file")
+	var publishSync = flag.Bool("publishsync", false, "Flush to the socket on publish")
 
 	log.SetFlags(0)
 	flag.Usage = usage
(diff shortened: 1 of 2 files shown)


kozlovic:
Few comments: I do think that there is a value for being able to flush the socket buffer on each write. For instance, when writing the C Client, I did add an option for the connection to send as soon as possible (ok, I am bad at names and called it `natsOptions_SetSendAsap()`.

The point is that if we were to do that with the Go client, it could be a global connection option. Don't have to make it pub/sub specific. When that option is on, the current kickFlusher() could become flushOrKickFlusher() and either flush the buffer or notify the channel.

You have a good point on the fact that a write may happen and so we should have been using deadline there.

I don't think you need to check for `nc.conn == nil` since in the function there is a check if connection is closed. If not, nc.conn will not be nil.

For the subscription, you can't ignore the isReconnecting() even if the SubscribeSync is true. It means that the buffer is set to a mem buffer because the tcp connection is gone.

gwik:
Good to read that you find value in it.

> I don't think you need to check for nc.conn == nil since in the function there is a check if connection is closed. If not, nc.conn will not be nil.

I enter this branch while testing.

> For the subscription, you can't ignore the isReconnecting() even if the SubscribeSync is true. It means that the buffer is set to a mem buffer because the tcp connection is gone.

Good to know, that means it should return an error immediately right?

gwik:
I was thinking about having both flush and reconnect buffer. I’ll check the C client.

Issue #195, opened 2016-06-08
Should Connect return errors if it cannot connect?


tomwilkie:
Currently, the client 'constructor' (`nats.Connect`) will return `nil, err` if it can't connect to NATS.  This behaviour puts a strict ordering on the start up sequences of my services; the NATS client is effectively instantiated in my `main()` function, such that the service cannot start if it can't contact NATS.  However my service are able to tolerate NATS not being there, and degrade gracefully - as message delivery is at most once, I kinda don't care too much if it fails.

My preference would be for connection state to be completely internal; client creation should only fail it it is passed a malformed url, for example.  The client should try to connect asynchronously in the background, failing requests until it does - the service not being there is just another failure case to me. I believe this is just a special case of the NATS coming and going, which you already have code to deal with.  This seems like something the client library should do - I can't be the first person to hit this.

(As a side note, I actually don't have control over the order in which services are started; its not needed by anything else; the golang SQL client libraries and the AWS libraries all behave in my desired way)

derekcollison:
I will take a look. In general, yes, I felt you should need to connect to a server at least once before a connection could be returned. We buffer messages while disconnected for some period of time until we hit limits, and we could re-use this logic. But we also have others pointing out that they do not want this behavior, and that they want the ability to control this behavior, or at least have a ttl.

Zambiorix:
Is there any progress on this one?
I'm also want to leave the connection status internal.

I'm also using Aerospike DB and they have the same behaviour for client startup.
But they also have an option "FailIfNotConnected" that you can set/unset.

Can go-nats use the same approach? Offer an option to turn off connection failure at startup?
Once it is running, it'll connect/reconnect when servers are available ..

Should be fairly trivial to implement, not?

Thanks
Cheers

Gerd

mcqueary:
Hi @tomwilkie @Zambiorix We rarely view a fundamental behavior change as trivial :) At a minimum, I suspect it would introduce new failure scenarios that we'd need to test for. Additionally, for consistency, we'd want to implement it across all clients we support. 

We're undertaking a new round of roadmap planning and review over the next few weeks and will discuss further as part of that.

kekoav:
I would also like to see retries on `Connect()` to allow for some intermittent network failures to not fail the whole system(or have to implement my own retry logic on background). There is already support for "Reconnect" when servers become unreachable. This is configured through the options `MaxReconnect`, `ReconnectWait`, `ReconnectedCB` and `Timeout`.

[...]

PR #390, opened 2018-09-04
Option for background connection attempts

shaxbee:
Provides `Background(ctx context.Context)` option for asynchronous connection process, making `Options.Connect()` non-blocking and always return nil error.

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -17,6 +17,7 @@ package nats
 import (
 	"bufio"
 	"bytes"
+	"context"
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/json"
@@ -280,6 +281,10 @@ type Options struct {
 	// UseOldRequestStyle forces the old method of Requests that utilize
 	// a new Inbox and a new Subscription for each request.
 	UseOldRequestStyle bool
+
+	// Background specifies context that should be use for background connection attempt
+	// if no context is specified Connect() call will be blocking
+	Background context.Context
 }
 
 const (
@@ -706,6 +711,14 @@ func UseOldRequestStyle() Option {
 	}
 }
 
+// Background is an Option to specify Context for background connection attempt
+func Background(ctx context.Context) Option {
+	return func(o *Options) error {
+		o.Background = ctx
+		return nil
+	}
+}
+
 // Handler processing
 
 // SetDisconnectHandler will set the disconnect event handler.
@@ -804,12 +817,12 @@ func (o Options) Connect() (*Conn, error) {
 	nc.ach = &asyncCallbacksHandler{}
 	nc.ach.cond = sync.NewCond(&nc.ach.mu)
 
-	if err := nc.connect(); err != nil {
+	if o.Background != nil {
+		nc.bgConnect(o.Background)
+		return nc, nil
+	} else if err := nc.connect(); err != nil {
 		return nil, err
 	}
-
-	// Spin up the async cb dispatcher on success
-	go nc.ach.asyncCBDispatcher()
 
 	return nc, nil
 }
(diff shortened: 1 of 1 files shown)


wallyqs:
Thanks for the PR, I don't think that we should add a `Background` context state to the client right now for the connection attempts, I would expect a background state to mean for example that the client will be connected for as long as the context is valid and disconnect from NATS when the context is canceled.

Also we don't currently support reconnection attempts for when establishing the connection on the first attempt (see: https://github.com/nats-io/go-nats/issues/195)

On the other hand, good news is that you could still implement your own custom logic using a custom dialer, and also make use of APIs like `DialContext` part of the standard library which support connection tracing, sharing a couple of examples below:

```go
type customDialer struct {
	ctx context.Context
}

func (cd *customDialer) Dial(network, address string) (net.Conn, error) {
	d := &net.Dialer{}

	for lastAttempt := time.Now(); ; lastAttempt = time.Now() {
		select {
		case <-cd.ctx.Done():
			return nil, cd.ctx.Err()
		default:
			if conn, err := d.DialContext(cd.ctx, network, address); err == nil {
				return conn, nil
			}
		}

		sleepTime := time.Duration(0)
		if time.Since(lastAttempt) < 2*time.Second {
			sleepTime = 2*time.Second - time.Since(lastAttempt)
		}
		if sleepTime <= 0 {
			runtime.Gosched()
		} else {
			time.Sleep(sleepTime)
		}
	}
}

func main() {
	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(5*time.Second))
	defer cancel()

	cd := &customDialer{
		ctx: ctx,
	}
	var err error
	var nc *nats.Conn
	opts := []nats.Option{nats.SetCustomDialer(cd), nats.ReconnectWait(2 * time.Second)}
	go func() {
		nc, err = nats.Connect("127.0.0.1:4222", opts...)
	}()
	// ...
```

Or enabling trace borrowing `httptrace` package hooks:

```go
	trace := &httptrace.ClientTrace{
		ConnectStart: func(network, addr string) {
			log.Println("Connection attempt...")
		},
		ConnectDone: func(network, addr string, err error) {
			if err == nil {
				log.Println("Established TCP connection!")
				cancel()
			}
		},
	}
	ctx = httptrace.WithClientTrace(ctx, trace)
	cd := &customDialer{
		ctx: ctx,
	}

	var err error
	var nc *nats.Conn
	opts := []nats.Option{nats.SetCustomDialer(cd), nats.ReconnectWait(2 * time.Second)}
```

shaxbee:
Thanks a lot! Custom dialer solution looks much cleaner!

wallyqs:
no problem :) ok closing this one for now

wallyqs:
Added full example on a gist here: https://gist.github.com/wallyqs/2cf37ed3b19517e16e466461376fd53b

shaxbee:
@wallyqs The one problem i see is that it is not possible to connect in non-blocking way, is there any way to achieve this?

ripienaar:
@shaxbee I am curious what the use case is for non blocking backgrounded connect? If you want to do message processing in the background just start the worker code in a go routine?

shaxbee:
@ripienaar This is same as #195 - in cloud system i'd like service to start even if nats is unavailable momentarily for example due to restart or any other issue, similarly to what happens in database/sql - calling sql.Open() doesn't block even if destination host is unreachable.

ripienaar:
You can still start - just handle your routine that connects and process data in a go routine?

```
go thing.Run(ctx, wg)

// continue startup

wg.Wait()
```

this way your main loop that might potentially block on the network initial connection doesnt hold up the startup loop.  It's a very common pattern in Go and I think NATS not reinventing this wheel is just fine.

PR #393, opened 2018-09-11
[CHANGED] Honor write timeout on every conn operation

gwik:
Writing to the buffer can flush to the connection if the buffer is full so the
writes needs to be timed out if a FlusherTimeout is set.

As a preliminary to https://github.com/nats-io/go-nats/pull/389

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -229,8 +229,8 @@ type Options struct {
 	// DrainTimeout sets the timeout for a Drain Operation to complete.
 	DrainTimeout time.Duration
 
-	// FlusherTimeout is the maximum time to wait for the flusher loop
-	// to be able to finish writing to the underlying connection.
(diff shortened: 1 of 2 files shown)


gwik:
I skipped the `TestCustomFlusherTimeout` until you fix it, the CI failed for something else.

kozlovic:
Thanks. I am thinking hard on the new option FlusherTimeout(). This is setting the `FlusherTimeout` option, but this no longer affects just the flusher. So I think we should change the description of that option.

gwik:
Updated, let me know if it is suits you. Or rephrase and I'll merge.
Sorry for the long delay.

derekcollison:
@kozlovic Can you take a quick look at this? I think it went stale.

PR #395, opened 2018-09-26
[ADDED] Ability to retrieve CID from server currently connected to

### LINKED ISSUE #394: Add API to get the CID
Following https://github.com/nats-io/gnatsd/pull/687, need to add an API to get the client's ID from the currently connected server.

kozlovic:
Server 1.2.0 returns the client's ID (cid) through the INFO protocol.
An API is added to retrieve this value. This can be useful for
debugging and monitoring.

Resolves #394

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -428,6 +428,7 @@ type serverInfo struct {
 	MaxPayload   int64    `json:"max_payload"`
 	ConnectURLs  []string `json:"connect_urls,omitempty"`
 	Proto        int      `json:"proto,omitempty"`
+	CID          uint64   `json:"client_id,omitempty"`
 }
 
 const (
@@ -3475,4 +3476,18 @@ func (nc *Conn) Barrier(f func()) error {
 	nc.subsMu.Unlock()
 	nc.mu.Unlock()
 	return nil
+}
+
+// GetCID returns the client ID assigned by the server the client
+// is currently connected to. Note that the value would change if
+// client reconnects. Also, returned value may be 0 if connecting
+// to a server pre 1.2.0, which did not send the CID back to the
+// client.
+func (nc *Conn) GetCID() (uint64, error) {
+	nc.mu.Lock()
+	defer nc.mu.Unlock()
+	if nc.isClosed() {
+		return 0, ErrConnectionClosed
+	}
+	return nc.info.CID, nil
 }
--- a/test/conn_test.go
+++ b/test/conn_test.go
@@ -2051,3 +2051,66 @@ func TestNilOpts(t *testing.T) {
 		t.Fatal("Unexpected error: option not set.")
 	}
 }
+
+func TestGetCID(t *testing.T) {
+	optsA := test.DefaultTestOptions
+	optsA.Port = -1
+	optsA.Cluster.Port = -1
+	srvA := RunServerWithOptions(optsA)
+	defer srvA.Shutdown()
+
+	optsB := test.DefaultTestOptions
+	optsB.Port = -1
+	optsB.Cluster.Port = -1
+	optsB.Routes = server.RoutesFromStr(fmt.Sprintf("nats://127.0.0.1:%d", srvA.ClusterAddr().Port))
+	srvB := RunServerWithOptions(optsB)
+
+	ch := make(chan bool, 1)
+	nc1, err := nats.Connect(fmt.Sprintf("nats://127.0.0.1:%d", srvA.Addr().(*net.TCPAddr).Port),
+		nats.ReconnectHandler(func(_ *nats.Conn) {
+			ch <- true
+		}))
+	if err != nil {
+		t.Fatalf("Error on connect: %v", err)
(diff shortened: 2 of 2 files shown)


kozlovic:
@derekcollison Not sure of the naming `GetCID()` and also if we should return error or not. That is, 0 would be if connected to older server but also if not currently connected. Right now, the cid and error are returned.

nats.go
```diff
@@ -3476,3 +3477,17 @@ func (nc *Conn) Barrier(f func()) error {
 	nc.mu.Unlock()
 	return nil
 }
+
+// GetCID returns the client ID assigned by the server the client
+// is currently connected to. Note that the value would change if
+// client reconnects. Also, returned value may be 0 if connecting
+// to a server pre 1.2.0, which did not send the CID back to the
+// client.
+func (nc *Conn) GetCID() (uint64, error) {
```
derekcollison: GetClientID?

nats.go
```diff
@@ -3476,3 +3477,17 @@ func (nc *Conn) Barrier(f func()) error {
 	nc.mu.Unlock()
 	return nil
 }
+
+// GetCID returns the client ID assigned by the server the client
+// is currently connected to. Note that the value would change if
+// client reconnects. Also, returned value may be 0 if connecting
```
derekcollison: Should return an error if not provided.

nats.go
```diff
@@ -3476,3 +3477,17 @@ func (nc *Conn) Barrier(f func()) error {
...
+// to a server pre 1.2.0, which did not send the CID back to the
+// client.
+func (nc *Conn) GetCID() (uint64, error) {
+	nc.mu.Lock()
+	defer nc.mu.Unlock()
+	if nc.isClosed() {
```
derekcollison: If 0 return an error that it was not provided.

test/conn_test.go
```diff
@@ -2051,3 +2051,66 @@ func TestNilOpts(t *testing.T) {
...
+		t.Fatal("Unexpected cid value, make sure server is 1.2.0+")
+	}
+
+	nc2, err := nats.Connect(fmt.Sprintf("nats://127.0.0.1:%d", srvB.Addr().(*net.TCPAddr).Port))
+	if err != nil {
+		t.Fatalf("Error on connect: %v", err)
+	}
+	defer nc2.Close()
+
+	// Stop server A, nc1 will reconnect to B, and should have different CID
```
derekcollison: Could be the same since CID is scoped to the server.

nats.go
kozlovic: Done

nats.go
kozlovic: Done, please comment on error variable name and content

nats.go
kozlovic: Done

test/conn_test.go
kozlovic: No since I made sure that there was another client connected to server B prior to killing A and having nc1 reconnect to server B.

kozlovic:
@derekcollison Made some changes and also added a test with fake server returning no CID and making sure GetClientID() returns proper error.

nats.go
```diff
@@ -93,6 +93,7 @@ var (
 	ErrInvalidArg             = errors.New("nats: invalid argument")
 	ErrInvalidContext         = errors.New("nats: invalid context")
 	ErrNoEchoNotSupported     = errors.New("nats: no echo option not supported by this server")
+	ErrNoClientIDReturned     = errors.New("nats: client ID not returned by this server")
```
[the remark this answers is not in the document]
kozlovic: Ok will change.

nats.go
```diff
@@ -3478,16 +3479,19 @@ func (nc *Conn) Barrier(f func()) error {
...
+// the client reconnects.
+// This function returns ErrNoClientIDReturned if the server is of a
+// version prior to 1.2.0.
+func (nc *Conn) GetClientID() (uint64, error) {
 	nc.mu.Lock()
 	defer nc.mu.Unlock()
 	if nc.isClosed() {
 		return 0, ErrConnectionClosed
 	}
+	if nc.info.CID == 0 {
```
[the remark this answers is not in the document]
kozlovic: Then the server should make sure that if it wraps, then it start again to 1.
I guess we could search "client_id" in the info string we get before doing unmarshal and if not present have a flag in connection that says cid not set?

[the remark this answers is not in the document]
derekcollison: Maybe post an issue against the server.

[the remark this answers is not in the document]
kozlovic: Ok, is it really worth it? I mean at 10M new connections a second, it would take 58,494 years without restarting the server before it wraps (if my math is correct).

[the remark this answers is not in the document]
derekcollison: Agree.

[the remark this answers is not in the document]
kozlovic: Ok, so will change the error name/content and update

PR #399, opened 2018-10-26
Nkey support

derekcollison:
Base nkey support.

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -96,7 +96,7 @@ var (
 	ErrNoEchoNotSupported     = errors.New("nats: no echo option not supported by this server")
 	ErrClientIDNotSupported   = errors.New("nats: client ID not supported by this server")
 	ErrNkeyButNoSigCB         = errors.New("nats: Nkey defined without a signature handler")
-	ErrNkeysNoSupported       = errors.New("nats: Nkeys not supported by the server.")
+	ErrNkeysNoSupported       = errors.New("nats: Nkeys not supported by the server")
 	ErrStaleConnection        = errors.New("nats: " + STALE_CONNECTION)
 )
 
@@ -699,6 +699,19 @@ func Token(token string) Option {
 	}
 }
 
+// Nkey will set the public Nkey and the signature callback to
+// sign the server nonce.
+func Nkey(pubKey string, sigCB SignatureHandler) Option {
+	return func(o *Options) error {
+		o.Nkey = pubKey
+		o.SignatureCB = sigCB
+		if sigCB == nil {
+			return ErrNkeyButNoSigCB
+		}
+		return nil
+	}
+}
+
 // Dialer is an Option to set the dialer which will be used when
 // attempting to establish a connection.
 // DEPRECATED: Should use CustomDialer instead.
@@ -1227,16 +1240,6 @@ func (nc *Conn) checkForSecure() error {
 		o.Secure = true
 	}
 
-	if o.Nkey != "" && nc.info.Nonce == "" {
-		return ErrNkeysNoSupported
-	}
-
-	// Check if we have an nkey but no signature callback
-	// defined.
-	if o.Nkey != "" && o.SignatureCB == nil {
-		return ErrNkeyButNoSigCB
-	}
-
 	// Need to rewrap with bufio
 	if o.Secure {
 		nc.makeTLSConn()
@@ -1264,6 +1267,15 @@ func (nc *Conn) processExpectedInfo() error {
 	// Parse the protocol
 	if err := nc.processInfo(c.args); err != nil {
 		return err
+	}
+
+	if nc.Opts.Nkey != "" && nc.info.Nonce == "" {
+		return ErrNkeysNoSupported
+	}
+
+	// Check if we have an nkey but no signature callback defined.
+	if nc.Opts.Nkey != "" && nc.Opts.SignatureCB == nil {
+		return ErrNkeyButNoSigCB
 	}
 
 	return nc.checkForSecure()
@@ -1301,7 +1313,7 @@ func (nc *Conn) connectProto() (string, error) {
 	}
 
 	if nkey != "" {
-		sigraw := nc.Opts.SignatureCB([]byte(nc.info.Nonce))
+		sigraw := o.SignatureCB([]byte(nc.info.Nonce))
 		sig = base64.StdEncoding.EncodeToString(sigraw)
 	}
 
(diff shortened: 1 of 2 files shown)


nats.go
```diff
@@ -94,6 +95,8 @@ var (
 	ErrInvalidContext         = errors.New("nats: invalid context")
 	ErrNoEchoNotSupported     = errors.New("nats: no echo option not supported by this server")
 	ErrClientIDNotSupported   = errors.New("nats: client ID not supported by this server")
+	ErrNkeyButNoSigCB         = errors.New("nats: Nkey defined without a signature handler")
+	ErrNkeysNoSupported       = errors.New("nats: Nkeys not supported by the server.")
```
kozlovic: No punctuation.

nats.go
```diff
@@ -1208,6 +1227,16 @@ func (nc *Conn) checkForSecure() error {
 		o.Secure = true
 	}
 
+	if o.Nkey != "" && nc.info.Nonce == "" {
```
kozlovic: So if user defines NKey, then it can connect only to servers that support  it, is that ok?

nats.go
```diff
@@ -1208,6 +1227,16 @@ func (nc *Conn) checkForSecure() error {
...
+		return ErrNkeysNoSupported
+	}
+
+	// Check if we have an nkey but no signature callback
+	// defined.
+	if o.Nkey != "" && o.SignatureCB == nil {
```
kozlovic: Why put this test in here? Since we have a test above that basically says that a client that has NKey cannot connect to a server that does not support it, why not do this test early in Connect()?

nats.go
```diff
@@ -261,6 +269,14 @@ type Options struct {
...
 	SubChanLen int
 
+	// Nkey sets the public nkey that will be used to authenticate
+	// when connecting to the server
+	Nkey string
+
+	// SignatureCB designates the function used to sign the nonce
+	// presented from the server.
+	SignatureCB SignatureHandler
+
```
kozlovic: There is no option function to set those, forcing users to use old methods of connecting from Options. I would recommend adding one function that sets the NKey and SignatureCB at the same time, which could ensure that signature is set or otherwise return error. We would still need to do the check outside, though, in case users connection from Options.

nats.go
```diff
@@ -261,6 +269,14 @@ type Options struct {
 	// dictated by PendingLimits()
 	SubChanLen int
 
+	// Nkey sets the public nkey that will be used to authenticate
+	// when connecting to the server
+	Nkey string
+
+	// SignatureCB designates the function used to sign the nonce
+	// presented from the server.
+	SignatureCB SignatureHandler
```
wallyqs: Maybe we could use a small `keypair` interface for signing the nonces as an alternative to the callback, for example:

```
diff --git a/nats.go b/nats.go
index 9eaa71e..71b4d04 100644
--- a/nats.go
+++ b/nats.go
@@ -168,6 +168,11 @@ type CustomDialer interface {
 	Dial(network, address string) (net.Conn, error)
 }
 
+type keypair interface {
+	Sign(input []byte) ([]byte, error)
+	PublicKey() (string, error)
+}
+
 // Options can be used to create a customized connection.
 type Options struct {
 
@@ -277,6 +282,9 @@ type Options struct {
 	// presented from the server.
 	SignatureCB SignatureHandler
 
+	// Nkeys is the keypair that will be used to sign nonces sent on connect.
+	Nkeys keypair
+
```

That way we could support usage like this:

```go
	kp, err := nkeys.FromSeed("SUAEL6....")
	if err != nil {
		log.Fatal(err)
	}
	opts := nats.DefaultOptions
	opts.Nkeys = kp

	nc, err := opts.Connect()
	if err != nil {
		log.Fatal(err)
	}
```

Internally the client would be the one doing the signing, but since using the small interface does not require to import "github.com/nats-io/nkeys" to be able to do it.

```
@@ -1297,12 +1300,21 @@ func (nc *Conn) connectProto() (string, error) {
 		user = nc.Opts.User
 		pass = nc.Opts.Password
 		token = nc.Opts.Token
		nkey = nc.Opts.Nkey
+		nkeys = nc.Opts.Nkeys
 	}
 
+	if nkeys != nil {
+		nonce := []byte(nc.info.Nonce)
+		signed, err := nkeys.Sign(nonce)
+		if err != nil {
+			return _EMPTY_, err
+		}
+		sig = base64.StdEncoding.EncodeToString(signed)
+
+		nkey, err = nkeys.PublicKey()
+		if err != nil {
+			return _EMPTY_, err
+		}
```

sasbury:
I am not sure we should call the interface keypair, since that could be confused with the Nkey keypair. For java and TS we called it AuthHandler I think.

nats.go
derekcollison: Fixed.

nats.go
derekcollison: I believe so, if a user defines them that is what they expect and if the server does not support it that should error IMO.

nats.go
derekcollison: I can move it..

nats.go
derekcollison: Agree will add it.

nats.go
```diff
@@ -699,6 +699,19 @@ func Token(token string) Option {
 	}
 }
 
+// Nkey will set the public Nkey and the signature callback to
+// sign the server nonce.
+func Nkey(pubKey string, sigCB SignatureHandler) Option {
+	return func(o *Options) error {
+		o.Nkey = pubKey
+		o.SignatureCB = sigCB
+		if sigCB == nil {
```
kozlovic: Well, technically the error says that there is a key but no callback, so you should add `pubKey != "" && ..`

nats.go
```diff
@@ -1227,16 +1240,6 @@ func (nc *Conn) checkForSecure() error {
 		o.Secure = true
 	}
 
-	if o.Nkey != "" && nc.info.Nonce == "" {
```
kozlovic: That's not really the one I say we should be moving

nats.go
```diff
@@ -1266,6 +1269,15 @@ func (nc *Conn) processExpectedInfo() error {
 		return err
 	}
 
+	if nc.Opts.Nkey != "" && nc.info.Nonce == "" {
+		return ErrNkeysNoSupported
+	}
+
+	// Check if we have an nkey but no signature callback defined.
```
kozlovic: I meant this one should be moved in Options.Connect() itself.

nats.go
derekcollison: ok

nats.go
derekcollison: ok got it.

nats.go
derekcollison: I think that would actually end up making more work for the user even if our api looked simpler.

PR #401, opened 2018-11-03
Added test for NkeyOptionFromSeed

kozlovic:

### DIFF
--- a/nats_test.go
+++ b/nats_test.go
@@ -23,6 +23,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io/ioutil"
+	"net"
 	"os"
 	"reflect"
 	"runtime"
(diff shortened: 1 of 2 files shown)


derekcollison:
You could have a SU prefix but the rest be bad. Like SUPER

nats_test.go
```diff
@@ -1319,3 +1321,112 @@ func TestNkeyAuth(t *testing.T) {
...
+	return fName
+}
+
+func TestNKeyOptionFromSeed(t *testing.T) {
+	if _, err := NkeyOptionFromSeed("file_that_does_not_exist"); err == nil {
+		t.Fatal("Expected error got none")
+	}
+
+	seedFile := createTmpFile(t, []byte(`
+# No seed
```
derekcollison: Since parser can deal with spaces etc maybe just do inline since bytes and not have indentation.

nats_test.go
```diff
@@ -1319,3 +1321,112 @@ func TestNkeyAuth(t *testing.T) {
...
+THIS_NOT_A_NKEY_SEED
+	`))
+	defer os.Remove(seedFile)
+	if _, err := NkeyOptionFromSeed(seedFile); err == nil || !strings.Contains(err.Error(), "seed found") {
+		t.Fatalf("Expected error about seed not found, got %v", err)
+	}
+	os.Remove(seedFile)
+
+	seedFile = createTmpFile(t, []byte(`
```
derekcollison: Same here.

derekcollison:
Ah ok, need to trim in nk lib, got it. Thanks.

derekcollison:
Actually that code to load is in this lib, not nk. We can trim in nkeyPairFromSeedFile.

kozlovic:
> Actually that code to load is in this lib, not nk. We can trim in nkeyPairFromSeedFile.

Yes, it is in nkeyPairFromSeedFile() that trimming should be done. I can do that as part of this PR if you want.

derekcollison:
ok I just did it while looping the lines.
```go
func nkeyPairFromSeedFile(seedFile string) (nkeys.KeyPair, error) {
	var seed []byte
	contents, err := ioutil.ReadFile(seedFile)
	if err != nil {
		return nil, fmt.Errorf("nats: %v", err)
	}
	defer wipeSlice(contents)

	lines := bytes.Split(contents, []byte("\n"))
	for _, line := range lines {
		line = bytes.TrimLeft(line, " \t")
		if bytes.HasPrefix(line, []byte("SU")) {
			seed = line
			break
		}
	}
	if seed == nil {
		return nil, fmt.Errorf("nats: No nkey user seed found in %q", seedFile)
	}
	kp, err := nkeys.FromSeed(seed)
	if err != nil {
		return nil, err
	}
	return kp, nil
}
```

derekcollison:
Feel free to put it in as well.

kozlovic:
@derekcollison Updated with trimming spaces and ensure that test detect that we found the key, just that the key is bad.

PR #402, opened 2018-11-03
Faster and smaller reply suffix

derekcollison:
Also bumped to 1.7 due to nkeys support.

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -2364,10 +2364,7 @@ func (nc *Conn) Request(subj string, data []byte, timeout time.Duration) (*Msg, 
 
 	// Do setup for the new style.
 	if nc.respMap == nil {
-		// _INBOX wildcard
-		nc.respSub = fmt.Sprintf("%s.*", NewInbox())
-		nc.respMap = make(map[string]chan *Msg)
-		nc.respRand = rand.New(rand.NewSource(time.Now().UnixNano()))
+		nc.initNewResp()
 	}
 	// Create literal Inbox and map to a chan msg.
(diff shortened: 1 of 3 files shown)


nats.go
```diff
@@ -2448,17 +2455,29 @@ func NewInbox() string {
...
+		b[i] = rdigits[l%base]
+		l /= base
+	}
 	return string(b[:])
 }
 
+// NewRespInbox is the new format used for _INBOX.
+func (nc *Conn) NewRespInbox() string {
+	nc.mu.Lock()
+	s := nc.newRespInbox()
```
wallyqs: `nc.NewRespInbox()` will fail if it is attempted to be used without having called `Request` before, maybe should check here and initialize if undefined?

```diff
diff --git a/nats.go b/nats.go
index aaf4995..bb92cdc 100644
--- a/nats.go
+++ b/nats.go
@@ -2473,6 +2473,11 @@ func (nc *Conn) newRespInbox() string {
 // NewRespInbox is the new format used for _INBOX.
 func (nc *Conn) NewRespInbox() string {
 	nc.mu.Lock()
+	if nc.respMap == nil {
+		nc.respSub = fmt.Sprintf("%s.*", NewInbox())
+		nc.respMap = make(map[string]chan *Msg)
+		nc.respRand = rand.New(rand.NewSource(time.Now().UnixNano()))
+	}
 	s := nc.newRespInbox()
 	nc.mu.Unlock()
 	return s
```

derekcollison: Good point, will fix.

PR #406, opened 2018-11-08
Make all examples go install friendly

derekcollison:


derekcollison:
Yep I am slow. @aricart mentioned this awhile back as well.

PR #422, opened 2018-12-29
Add import stament to example in README

cemremengu:
For completeness


kozlovic:
Thank you for the contribution!

PR #427, opened 2019-02-06
Added support for -nkey flag to support nkey authentication in tools

aricart:


examples/nats-sub/main.go
```diff
@@ -52,11 +53,24 @@ func main() {
...
+	}
+
 	// Use UserCredentials
 	if *userCreds != "" {
 		opts = append(opts, nats.UserCredentials(*userCreds))
 	}
 
+	// Use Nkey authentication.
+	if *nkeyFile != "" {
+		opt, err := nats.NkeyOptionFromSeed(*nkeyFile)
```
derekcollison: Wonder if we can let err processing just happen in Connect to keep examples a bit shorter?

aricart: tbh,  I think that examples should do things exactly as we want folks to do them. If what you are saying is that setting an option shouldn't return an error until connect then we should change that api instead.

PR #432, opened 2019-02-21
Improve NextMsg() performance when msg available

derekcollison:
If we have a msg available return immediately without setting up timeouts etc.
On my machine took NextMsg(0) from >450ns to 55ns per call.

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -807,7 +807,7 @@ func Nkey(pubKey string, sigCB SignatureHandler) Option {
 }
 
 // SyncQueueLen will set the maximum queue len for the internal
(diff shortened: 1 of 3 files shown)


nats.go
```diff
@@ -806,6 +806,15 @@ func Nkey(pubKey string, sigCB SignatureHandler) Option {
 	}
 }
 
+// SyncQueueLen will set the maximum queue len for the internal
+// channel used for SubsribeSync().
```
wallyqs:
```suggestion
// channel used for SubscribeSync().
```

PR #435, opened 2019-02-22
Use RWMutex in some simple cases

MaruHyl:


nats.go
```diff
@@ -1283,38 +1283,47 @@ func (nc *Conn) ConnectedUrl() string {
 	if nc == nil {
 		return _EMPTY_
 	}
-	nc.mu.Lock()
-	defer nc.mu.Unlock()
```
derekcollison: Defers can be expensive but in non performance paths where the code between locks is not drop dead simple I prefer them for ease of reading the code later.

nats.go
```diff
@@ -350,7 +350,7 @@ type Conn struct {
 	// atomic.* functions crash on 32bit machines if operand is not aligned
 	// at 64bit. See https://github.com/golang/go/issues/599
 	Statistics
-	mu sync.Mutex
+	mu sync.RWMutex
```
derekcollison: These are more complex than simple mutexes and unless there is a large amount of contention where reads outweigh writes substantially this is usually not worth it.

nats.go
```diff
@@ -1283,38 +1283,47 @@ func (nc *Conn) ConnectedUrl() string {
 	if nc == nil {
 		return _EMPTY_
 	}
-	nc.mu.Lock()
-	defer nc.mu.Unlock()
+	var url string
+	nc.mu.RLock()
```
derekcollison: Defer code reads better IMO here.

nats.go
```diff
@@ -3519,23 +3529,26 @@ func (nc *Conn) Close() {
 
 // IsClosed tests if a Conn has been closed.
 func (nc *Conn) IsClosed() bool {
-	nc.mu.Lock()
-	defer nc.mu.Unlock()
-	return nc.isClosed()
+	nc.mu.RLock()
```
derekcollison: Again not performance critical so defer ok and I prefer the defer version to be honest.

MaruHyl:
@derekcollison  okay,  I changed to defer version someplace and committed.And also use RLock in `readLoop` now.

MaruHyl:
okay.

derekcollison:
FYI you could have done this on this PR. `git rebase -i HEAD~N` then `git push --force` since you are the only one on this one.

PR #438, opened 2019-02-22
Use RWMutex in some simple cases

MaruHyl:
#435 I need to reopen a PR because of force-pushed.

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -350,7 +350,7 @@ type Conn struct {
 	// atomic.* functions crash on 32bit machines if operand is not aligned
 	// at 64bit. See https://github.com/golang/go/issues/599
 	Statistics
-	mu sync.Mutex
+	mu sync.RWMutex
 	// Opts holds the configuration of the Conn.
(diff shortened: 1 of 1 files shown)


nats.go
```diff
@@ -2834,9 +2840,10 @@ func (nc *Conn) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg) (*Sub
 
 // NumSubscriptions returns active number of subscriptions.
 func (nc *Conn) NumSubscriptions() int {
-	nc.mu.Lock()
-	defer nc.mu.Unlock()
-	return len(nc.subs)
+	nc.mu.RLock()
```
derekcollison: Prefer the defer version, and all of these are not critical perf wise, so let's revert these.

MaruHyl:
@derekcollison okay, reverted.

nats.go
MaruHyl: done.

MaruHyl:
@derekcollison  Travis done.

PR #440, opened 2019-03-01
refactor examples

andyxning:
This PR will refactor examples in 
* Enhance `Usage` function to print default flag settings info in `-h` output
* Remove redundant address reference

### DIFF
--- a/examples/nats-bench/main.go
+++ b/examples/nats-bench/main.go
@@ -34,7 +34,8 @@ const (
 )
 
 func usage() {
-	log.Fatalf("Usage: nats-bench [-s server (%s)] [--tls] [-np NUM_PUBLISHERS] [-ns NUM_SUBSCRIBERS] [-n NUM_MSGS] [-ms MESSAGE_SIZE] [-csv csvfile] <subject>\n", nats.DefaultURL)
+	log.Printf("Usage: nats-bench [-s server (%s)] [--tls] [-np NUM_PUBLISHERS] [-ns NUM_SUBSCRIBERS] [-n NUM_MSGS] [-ms MESSAGE_SIZE] [-csv csvfile] <subject>\n", nats.DefaultURL)
+	flag.PrintDefaults()
 }
 
 var benchmark *bench.Benchmark
--- a/examples/nats-echo/main.go
+++ b/examples/nats-echo/main.go
@@ -34,7 +34,8 @@ import (
 // nats-echo -s demo.nats.io:4443 <subject> (TLS version)
 
 func usage() {
-	log.Fatalf("Usage: nats-echo [-s server] [-creds file] [-t] <subject>")
+	log.Printf("Usage: nats-echo [-s server] [-creds file] [-t] <subject>\n")
+	flag.PrintDefaults()
 }
 
 func printMsg(m *nats.Msg, i int) {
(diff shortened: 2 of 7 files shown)


andyxning:
Some thing like below:
```
 ➜ ~/Github/go/src/github.com/nats-io/go-nats/examples/nats-bench  git:(refactor_examples) ✗ go run main.go -h
Usage: nats-bench [-s server (nats://localhost:4222)] [--tls] [-np NUM_PUBLISHERS] [-ns NUM_SUBSCRIBERS] [-n NUM_MSGS] [-ms MESSAGE_SIZE] [-csv csvfile] <subject>
  -creds string
    	User Credentials File
  -csv string
    	Save bench data to csv file
  -ms int
    	Size of the message. (default 128)
  -n int
    	Number of Messages to Publish (default 100000)
  -np int
    	Number of Concurrent Publishers (default 1)
  -ns int
    	Number of Concurrent Subscribers
  -s string
    	The nats server URLs (separated by comma) (default "nats://localhost:4222")
  -tls
    	Use TLS Secure Connection
exit status 2
```

derekcollison:
I prefer the simple version, but I understand there could be a use for seeing the defaults. Could we make that another flag possibly?

andyxning:
> Could we make that another flag possibly?

`-v`?

derekcollison:
Can we do -h for simple version and --help for the one with defaults.

andyxning:
> Can we do -h for simple version and --help for the one with defaults.

This is hard to do with `flag` pkg, imo.

derekcollison:
How so, I think it is straight forward but I could be wrong.

andyxning:
@derekcollison PTAL.

examples/nats-bench/main.go
```diff
@@ -34,7 +36,12 @@ const (
 )
 
 func usage() {
-	log.Fatalf("Usage: nats-bench [-s server (%s)] [--tls] [-np NUM_PUBLISHERS] [-ns NUM_SUBSCRIBERS] [-n NUM_MSGS] [-ms MESSAGE_SIZE] [-csv csvfile] <subject>\n", nats.DefaultURL)
+	log.Printf("Usage: nats-bench [-s server (%s)] [--tls] [-np NUM_PUBLISHERS] [-ns NUM_SUBSCRIBERS] [-n NUM_MSGS] [-ms MESSAGE_SIZE] [-csv csvfile] <subject>\n", nats.DefaultURL)
+	for _, arg := range os.Args {
+		if strings.Contains(arg, "-help") {
```
derekcollison: Can we not just have a normal bool flag for `--help`?

andyxning: > Can we not just have a normal bool flag for --help?

Yes, we can. But there is no good way to check either `-h` or `--help` command line argument is specified nor register callbacks for `-h` and `--help`.

derekcollison: Ah ok. And docs say -h and --help or -help are all the same. So maybe we do go back to print usage followed by the defaults like you originally had. Apologies.

andyxning:
@derekcollison PTAL again.

andyxning:
> And lastly does PrintDefaults exit the program?

No. This is not what PrintDefaults should do. This is done out of `Usage` function which invoke the PrintDefaults function.

> flag.PrintDefaults I believe goes to stderr by default, and log.Printf to stdout

`flag.PrintDefaults` is really goes to stderr by default. `log.Printf` is also goes to stderr by default.

PR #444, opened 2019-03-05
Add Go 1.12 to the build

wallyqs:
- Adds Go 1.12 to Travis build and removes Go 1.10
- Fixes a static check error from test due to DualStack field now being deprecated


.travis.yml
```diff
@@ -1,8 +1,8 @@
 language: go
 sudo: false
 go:
+- 1.12.x
```
kozlovic: Could you then change line 21 to use `1.12` as the go version check?

wallyqs: sounds good, ok just updated but will wait for build result

PR #445, opened 2019-03-05
Properly randomize multiple adders for single host.

derekcollison:
We were using range but range seems to not be random from a distriution standpoint.

### DIFF
--- a/nats_test.go
+++ b/nats_test.go
@@ -1289,7 +1289,7 @@ eyJ0eXAiOiJqd3QiLCJhbGciOiJlZDI1NTE5In0.eyJqdGkiOiJBSFQzRzNXRElDS1FWQ1FUWFJUTldP
 ------END NATS USER JWT------
 
 ************************* IMPORTANT *************************
-NKEY Seed printed below can be used sign and prove identity.
+NKEY Seed printed below can be used to sign and prove identity.
 NKEYs are sensitive and should be treated as secrets.
 
 -----BEGIN USER NKEY SEED-----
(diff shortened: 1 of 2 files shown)


nats_test.go
```diff
@@ -1392,6 +1415,28 @@ func TestUserCredentialsChainedFile(t *testing.T) {
...
+	if server.VERSION[0] == '1' {
+		t.Skip()
+	}
+	ts := runTrustTLSServer()
+	defer ts.Shutdown()
+
+	chainedFile := createTmpFile(t, []byte(chained))
+	os.Remove(chainedFile)
+
+	nc, err := Connect(fmt.Sprintf("tls://127.0.0.1:%d", TEST_PORT),
```
kozlovic: This test does not seem to be related to host shuffling. Also, this test "works" because there is a single server. If you had 2 servers, then the initial connect will get an INFO from first server with second connect URL which means that library will then try to connect to the second IP which then would produce the x509 error message.
cc @wallyqs

derekcollison: Will remove and submit in second PR.

awootton:
FYI in go version 1.9.3 theres no .Shuffle in math/random so nats needs a more recent version of go.
...
go get -u github.com/gorilla/websocket
go get -u github.com/nats-io/go-nats
# github.com/nats-io/go-nats
../../github.com/nats-io/go-nats/nats.go:1222:3: undefined: "math/rand".Shuffle
Makefile:74: recipe for target 'get-deps' failed
make[1]: *** [get-deps] Error 2

derekcollison:
ok will discuss with @kozlovic on whether we should keep support for 1.9.x

awootton:
> ok will discuss with @kozlovic on whether we should keep support for 1.9.x
I'm ok if you decide not to. People are going to have to upgrade their go eventually.

derekcollison:
1.9 is not officially supported by the Go team so we will leave as is..

PR #447, opened 2019-03-14
Fix for race on sub sync type setting

derekcollison:
This can occur when a sync or queue sync subscription is being created at the same time the connection is being closed.

https://github.com/google/go-cloud/issues/1556

/cc @vangent

### DIFF
--- a/context.go
+++ b/context.go
@@ -93,7 +93,7 @@ func (nc *Conn) oldRequestWithContext(ctx context.Context, subj string, data []b
 	inbox := NewInbox()
 	ch := make(chan *Msg, RequestChanLen)
 
-	s, err := nc.subscribe(inbox, _EMPTY_, nil, ch)
+	s, err := nc.subscribe(inbox, _EMPTY_, nil, ch, true)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


nats.go
```diff
@@ -2754,7 +2754,9 @@ func (nc *Conn) SubscribeSync(subj string) (*Subscription, error) {
 	mch := make(chan *Msg, nc.Opts.SubChanLen)
 	s, e := nc.subscribe(subj, _EMPTY_, nil, mch)
 	if s != nil {
+		s.mu.Lock()
```
kozlovic: I would not do that. Instead, I would either let nc.subscribe() infer the type, or if not possible, pass the type as an argument. We need to have the type set before the subscription is added to the connection's subs map.

derekcollison: ok will take another look.

derekcollison: Take a look now..

context.go
```diff
@@ -93,7 +93,7 @@ func (nc *Conn) oldRequestWithContext(ctx context.Context, subj string, data []b
 	inbox := NewInbox()
 	ch := make(chan *Msg, RequestChanLen)
 
-	s, err := nc.subscribe(inbox, _EMPTY_, nil, ch)
+	s, err := nc.subscribe(inbox, _EMPTY_, nil, ch, true)
```
kozlovic: Are you sure? We did not set it to SyncSubscription for this one, so should it be `false` here?

derekcollison: If we create the chan I was setting it to sync, it would have been set to a chanSub type, but I think that was wrong. Has no effect.

kozlovic: I see..

PR #461, opened 2019-05-10
Updates for rename and go.mod support

derekcollison:


kozlovic:
@derekcollison This is not triggering Travis-ci because of the name change you think? chicken and egg problem?

wallyqs:
Also need to update import in this test: https://github.com/nats-io/nats.go/blob/ad2d0cc1a82bd042861b6a141009545fa2e07c17/test/json_test.go#L21

derekcollison:
Travis is running, not sure what you mean.

derekcollison:
Thanks Wally, will update.

derekcollison:
Updated Travis and Coveralls

PR #462, opened 2019-05-14
fixes #454. Added DisconnectedErrCB to client

### LINKED ISSUE #454: DisconnectedCB doesn't have err in API
Hi, `DisconnectedCB` callback API doesn't provide access to error caused disconnection. I wonder if you will accept PR fixing it in a backward compatible way, e.g. adding new callback?
```
// DisconnectErrHandler is used to process disconnect events.
type DisconnectErrHandler func(*Conn, error)

// DisconnectErrHandler is an Option to set the disconnected err handler.
func DisconnectErrHandler(cb DisconnectErrHandler) Option {
	return func(o *Options) error {
		o.DisconnectErrCB = cb
		return nil
	}
}
```
WDYT?

mkorolyov:


nats.go
```diff
@@ -266,8 +269,15 @@ type Options struct {
 
 	// DisconnectedCB sets the disconnected handler that is called
 	// whenever the connection is disconnected.
+	// Will not be called if DisconnectedErrCB was initiated.
```
kozlovic: should it be: `Will not be called if DisconnectedErrCB is set`

nats.go
```diff
@@ -266,8 +269,15 @@ type Options struct {
...
+	// DEPRECATED. use DisconnectedErrCB which passes initial err
```
kozlovic: `which passes error that caused the disconnect event`?

nats.go
```diff
@@ -266,8 +269,15 @@ type Options struct {
...
 	DisconnectedCB ConnHandler
 
+	// DisconnectedErrCB sets the disconnected error handler that is called
+	// whenever the connection is disconnected.
+	// DisconnectedCB will not be called if DisconnectedErrCB was initiated.
```
kozlovic: `DisconnectedCB will not be called if DisconnectedErrCB is set`?

nats.go
```diff
@@ -3511,8 +3543,12 @@ func (nc *Conn) close(status Status, doCBs bool) {
 
 	// Perform appropriate callback if needed for a disconnect.
 	if doCBs {
-		if nc.Opts.DisconnectedCB != nil && nc.conn != nil {
-			nc.ach.push(func() { nc.Opts.DisconnectedCB(nc) })
+		if nc.conn != nil {
+			if nc.Opts.DisconnectedErrCB != nil {
+				nc.ach.push(func() { nc.Opts.DisconnectedErrCB(nc, ErrConnectionClosed) })
```
wallyqs: I wonder if maybe here we should use `nil` for the error instead of `ErrConnectionClosed`? That would mean that when calling `nc.Close()`, the disconnected callback is called but there is no error since it was called by the user.

kozlovic: Right. We should then add this to the "doc" of DisconnectedErrCB that it is possible to get `nil` as an error, one case would be when the user closes the connection.

mkorolyov: Hi @wallyqs , thanks for review. Found possible missed case with err passing:
nats.go:1438 where we trying to initialize connected connection and got an err as result, if so we close connection with status `DISCONNECTED`. I think we should pass acquired error to new callback too, adding `err` as param to `close()` method? WDYT?
Adding err as param to `close()` will fix pointed issue when user closes connection, we will pass nil here with status `CLOSED`.

nats.go
mkorolyov: thanks, fixed docs

nats.go
mkorolyov: thanks, fixed docs

nats.go
mkorolyov: thanks, fixed docs

nats.go
wallyqs: I think that sounds good.

mkorolyov:
@kozlovic @wallyqs Thanks for fast review! I've updated docs and added logic with proper err passing for close/disconnect status. Looks like it could be merged.

One thing i want to mention that i noticed blinking test `TestRecvChanAsyncLeakGoRoutines`, but it became green after another travis build. I didn't found any relation between changes i've made and that blinking test, could you pls take a look to make sure we are not breaking something under the hood? Thanks a lot!

nats.go
```diff
@@ -266,8 +269,17 @@ type Options struct {
...
 	// DisconnectedCB sets the disconnected handler that is called
 	// whenever the connection is disconnected.
+	// Will not be called if DisconnectedErrCB is set
+	// DEPRECATED. Use DisconnectedErrCB which passes error that caused
+	// the disconnect event.
 	DisconnectedCB ConnHandler
 
+	// DisconnectedErrCB sets the disconnected error handler that is called
+	// whenever the connection is disconnected.
+	// Disconnected error could be nil in case if user closes connection.
```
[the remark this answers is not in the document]
mkorolyov: Thanks! sorry for my poor English :)

nats.go
```diff
@@ -687,6 +699,14 @@ func DrainTimeout(t time.Duration) Option {
 	}
 }
 
+// DisconnectErrHandler is an Option to set the disconnected err handler.
```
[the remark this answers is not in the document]
mkorolyov: replaced

nats.go
```diff
@@ -150,6 +150,9 @@ const (
 // disconnected and closed connections.
 type ConnHandler func(*Conn)
 
+// DisconnectErrHandler is used to process error which caused disconnected event.
```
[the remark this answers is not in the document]
mkorolyov: updated, thanks!

nats.go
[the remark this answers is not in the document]
kozlovic: No worries! Not sure what I propose is better (not my native tongue either)

kozlovic:
@mkorolyov I was getting ready to do a release and noticed that we missed some deprecated statements, but then I am having second thoughts. I wonder if we should not have named the handler `type ConnErrHandler func(*Conn, error)` instead of `DisconnectErrHandler`. The reason is that we could later on use that for other events that disconnect, so I would not want to limit ourselves (or have to add yet another handler). What do you think? I will submit a PR to see how that looks.

mkorolyov:
> @mkorolyov I was getting ready to do a release and noticed that we missed some deprecated statements, but then I am having second thoughts. I wonder if we should not have named the handler `type ConnErrHandler func(*Conn, error)` instead of `DisconnectErrHandler`. The reason is that we could later on use that for other events that disconnect, so I would not want to limit ourselves (or have to add yet another handler). What do you think? I will submit a PR to see how that looks.

@kozlovic i can rename myself in PR, have some time at the moment :)

kozlovic:
Ok great!

PR #465, opened 2019-05-17
Fixes to example tools

wallyqs:
Apply similar fix as https://github.com/nats-io/nats.go/pull/456 and call exit 1 after showing the tool usage/help message.

### DIFF
--- a/examples/nats-bench/main.go
+++ b/examples/nats-bench/main.go
@@ -37,7 +37,11 @@ const (
 func usage() {
 	log.Printf("Usage: nats-bench [-s server (%s)] [--tls] [-np NUM_PUBLISHERS] [-ns NUM_SUBSCRIBERS] [-n NUM_MSGS] [-ms MESSAGE_SIZE] [-csv csvfile] <subject>\n", nats.DefaultURL)
 	flag.PrintDefaults()
-	os.Exit(1)
+}
+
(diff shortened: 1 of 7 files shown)


wallyqs:
I think we should exit 0 at least in the case for when only showing help, then exit 1 on incorrect usage. I could add an `showUsageAndDie` function for that for example, then exit 0 when only showing help.

derekcollison:
ok I agree.

wallyqs:
Applied updates to the tools so that: 

- `-h`  is exit 0
- wrong usage, is exit 1
- wrong/unknown flags and arguments is exit 2

examples/nats-bench/main.go
```diff
@@ -37,6 +37,10 @@ const (
 func usage() {
 	log.Printf("Usage: nats-bench [-s server (%s)] [--tls] [-np NUM_PUBLISHERS] [-ns NUM_SUBSCRIBERS] [-n NUM_MSGS] [-ms MESSAGE_SIZE] [-csv csvfile] <subject>\n", nats.DefaultURL)
 	flag.PrintDefaults()
+}
+
+func showUsageAndDie() {
```
derekcollison: Maybe `ShowUsageAndExit(exitcode int)`?

wallyqs: sounds good to me,  just updated

PR #472, opened 2019-05-25
Add Respond()

derekcollison:
This will allow a user to respond to a subscription callback that has not closed over a connection by referencing the request message.

```go
func cb(m *nats.Msg) {
  m.Respond([]byte("I am here")
}
```

### DIFF
--- a/test/basic_test.go
+++ b/test/basic_test.go
@@ -362,6 +362,40 @@ func TestPubSubWithReply(t *testing.T) {
 	}
 }
 
+func TestMsgRespond(t *testing.T) {
+	s := RunDefaultServer()
+	defer s.Shutdown()
+	nc := NewDefaultConnection(t)
+	defer nc.Close()
+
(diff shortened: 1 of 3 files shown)


test/basic_test.go
```diff
@@ -362,6 +362,40 @@ func TestPubSubWithReply(t *testing.T) {
...
+	}
+
+	sub, err := nc.Subscribe("req", func(msg *nats.Msg) {
+		msg.Respond([]byte("42"))
+	})
+	if err != nil {
+		t.Fatal("Failed to subscribe: ", err)
+	}
+
+	// Fake the bound notion by assigning Sub directly to test no reply.
```
kozlovic: I would have simply created a sync sub on another subject, publish a message (not request) to that subject, get the message, then call respond. No need to fake it.

derekcollison: Its public so easy to do, but if you feel strongly I could make legit here.

PR #473, opened 2019-05-30
Prepare for next release

kozlovic:
Use lint directives for staticcheck to work


test/basic_test.go
```diff
@@ -639,6 +639,7 @@ func TestSimultaneousRequests(t *testing.T) {
 	var wg sync.WaitGroup
 	for i := 0; i < 50; i++ {
 		wg.Add(1)
+		//lint:ignore SA2002 t.Fatalf in go routine will happen only if test fails
```
derekcollison: Why not add to global ignore?

kozlovic: global ignore file seem to not be used anymore (actually should remove from repo, forgot to do that).

kozlovic: updated travis and removed the file

dominikh:
You can use `//lint:file-ignore` to ignore checks file-wide. 

Of course, ignoring SA2002 with the argument that "t.Fatalf in go routine will happen only if test fails" is absurd. Of course t.Fatalf only gets called when a test fails, that is its purpose. However, calling t.FailNow (which is what t.Fatalf does) in a goroutine other than the one running the actual test is incorrect and may lead to incorrect test execution, as the failing test will not terminate if t.FailNow is being called in another goroutine. There is of course also the undefined nature of doing things that the documentation explicitly forbids.

kozlovic:
Calling t.Fatalf() will cause a data race if it is in a go routine other than the go routine the test executes from. The point of my comment is to say that if we get the t.Fatalf() it means that the code is broken (or test has a timing issue that would need to be fixed). Point is I don't care if I get a race, because if we hit t.Fatal, we have a bigger problem. If test does not terminate for any reason, the overall suite will fail after test timeout.
I did not know about file-wide, so I could have used that indeed. And long term I would have replaced the use of t.Fatalf with sending to an err channel, but there were a bit too many since we wanted to get the release out quickly.

docmerlin:
@kozlovic it can throw an error in the wrong test, as a result.  It can also cause other strange problems, that make debugging the actual failing test hard.

kozlovic:
Noted. Again, plan is to not use the ignore and use error channel, etc..

PR #476, opened 2019-06-03
Moving to 1.8.0

kozlovic:
Since client did not have any backward compatible changes, we
should not have moved to 2.0.0. Also 2.0.0 would force users to
modify their import path (due to go.mod need to have v2).


kozlovic:
Got go ahead from Derek.

Issue #478, opened 2019-06-04
Cannot go get client


coffeefedora:
I have a project that is outside of the go src folder, I am running go 1.12.5 on windows  amd64 (win10)  This is my first attempt to use NATS, not a great start!  

Getting this error:

go get github.com/nats-io/nats.go/
go: finding github.com/nats-io/nats.go v1.8.0
go: finding github.com/nats-io/nkeys v0.0.2
go: finding github.com/nats-io/nuid v1.0.1
go: finding golang.org/x/crypto v0.0.0-20181203042331-505ab145d0a9
CreateFile github.com/nats-io/nats.go: The system cannot find the path specified.

jar3b:
I have a similar issue using linux (ubuntu 18.04)

```
user@host:~/work$ go version
go version go1.12 linux/amd64
user@host:~/work$ go get github.com/nats-io/nats.go/
stat github.com/nats-io/nats.go: no such file or directory
```

Removing `~/go/pkg/mod/github.com/nats-io` folder doesn't solve the issue.

Also with vgo (import was specified as `"github.com/nats-io/nats.go"`)

```
user@host:~/work$go mod tidy
...
go: github.com/nats-io/go-nats@v1.8.0: parsing go.mod: unexpected module path "github.com/nats-io/nats.go"
go: error loading module requirements
```

kozlovic:
I think the issue now is with nats.go and stan.go libs not modified to change the nats-server `v2` import as suggested [here](https://github.com/nats-io/nats.go/issues/477#issuecomment-498620833). I can work on that, but I am not 100% sure, so to test it means that I need to create a new tag to test the theory, which if prove wrong means I will have to delete the tag. What do you think?

coffeefedora:
Since you are doing a rename, may I ask why nats.go if it the trailing .go means having to end it with a /?  I've never seen a go module do that, and it kinda ... smells.

Why not name it nats-client or something more indicative of what it is?  

Either way though, I am stuck which sucks because I was hoping to test nats today.  :(

jar3b:
@kozlovic, I think this is a good idea, i can test new tag as soon as possible. In the current situation build with 1.8.0 (latest) is still impossible with default configuration.

UPD: I tried to change the imports in go tests in fork and it all worked (with nats.io, but i think stan.io fix must be similar)

kozlovic:
@coffeefedora About the naming, it is not my decision...

jar3b:
[...]

PR #481, opened 2019-06-07
Add note to README on using go get and Go modules

wallyqs:
`go get` needs special handling with trailing slashes when using Go modules so adding to the readme (more info at https://github.com/nats-io/nats.go/issues/478)

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,17 @@ go get github.com/nats-io/nats.go/
 go get github.com/nats-io/nats-server
 ```
 
+When using Go modules (`GO111MODULE=on`):
+
+```bash
+# Go client latest and explicit version
+go get github.com/nats-io/nats.go/@latest
+go get github.com/nats-io/nats.go/@v1.8.1
+
+# Server
+go get github.com/nats-io/nats-server/v2
+```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -15,6 +15,17 @@ go get github.com/nats-io/nats.go/
 go get github.com/nats-io/nats-server
 ```
 
+When using Go modules (`GO111MODULE=on`):
```
derekcollison: Go 1.12 makes them on by default I think, so maybe say "When using or transitioning to Go Modules support, ..."

README.md
```diff
@@ -15,6 +15,17 @@ go get github.com/nats-io/nats.go/
...
+
+```bash
+# Go client latest and explicit version
+go get github.com/nats-io/nats.go/@latest
+go get github.com/nats-io/nats.go/@v1.8.1
+
+# Server
+go get github.com/nats-io/nats-server/v2
```
derekcollison: They may need or want the 1.x version so we should show an option for getting that version too, but comment it out and give a comment.

wallyqs: thanks, ok made the updates

Issue #459, opened 2019-05-09
Document that this library is thread-safe


JensRantil:
I don't have a reference for this (can't remember where I read it), but in Go libraries should be assumed to not be thread-safe unless otherwise stated. As such, since [this library is thread-safe](https://natsio.slack.com/archives/C069GSYFP/p1557411778233400?thread_ts=1557395155.230600&cid=C069GSYFP) it should probably be documented on [package](https://godoc.org/github.com/nats-io/go-nats) or [`Conn`](https://godoc.org/github.com/nats-io/go-nats#Conn) level.

Issue #457, opened 2019-04-25
Document ErrTimeout is returned from Subscription.NextMsg(...)


JensRantil:
I had to have a look at source code to figure it out. Good practise to document in GoDoc.

URL: https://godoc.org/github.com/nats-io/go-nats#Subscription.NextMsg

PR #482, opened 2019-06-07
fix documentation for issues #457 & #459

gcolliso:


nats.go
```diff
@@ -363,6 +363,7 @@ const (
 
 // A Conn represents a bare connection to a nats-server.
 // It can send and receive []byte payloads.
+// The connection is thread-safe.
```
derekcollison: "The connection is safe to use in multiple Go routines concurrently."

PR #491, opened 2019-07-11
Updated go.mod

derekcollison:


derekcollison:
I was looking at that too, no way to do that except by hand it seems. I could remove them though, at least the server ones etc.

kozlovic:
If you undo the changes and do just a `go mod build ./...` it should not include the server deps, I think.

derekcollison:
Check now.

PR #492, opened 2019-07-12
[FIXED] Wrong error returned in NextMsg and NextMsgWithContext

### LINKED ISSUE #404: NextMsgWithContext() returns ErrConnectionClosed, but underlying connection is fine.
`NextMsgWithContext()` returns a different error, depending when Unsubscribe() is called:

`ErrConnectionClosed` is returned if `NextMsgWithContext()` waits while `Unsubscribe()` is called.
`ErrBadSubscription` is returned if `Unsubscribe()` is called before `NextMsgWithContext()`.

`ErrConnectionClosed` seems wrong here, as it hint to an underlying connection problem, but instead it was just a normal `Unsubscribe()`.

example:
```
sub, _ := natsCon.SubscribeSync("topic")
// func() { // results in ErrBadSubscription
go func() { // results in ErrConnectionClosed
          time.Sleep(time.Second)
          sub.Unsubscribe()
}()
_, err = sub.NextMsgWithContext(ctx)
log.Print(err)
```
version: github.com/nats-io/go-nats v1.6.0

kozlovic:
If the application is in NextMsg() (or NextMsgWithContext()) and
the subscription is unsubscribed from a different go routine, the
NextMsg() call would return that the connection is closed, instead
of the subscription.

Resolves #404

### DIFF
--- a/context.go
+++ b/context.go
@@ -140,10 +140,7 @@ func (s *Subscription) NextMsgWithContext(ctx context.Context) (*Msg, error) {
 	select {
 	case msg, ok = <-mch:
 		if !ok {
-			s.mu.Lock()
-			err = s.validateNextMsgState()
-			s.mu.Unlock()
-			return nil, err
+			return nil, s.getNextMsgErr()
 		}
 		if err := s.processNextMsgDelivered(msg); err != nil {
 			return nil, err
@@ -156,10 +153,7 @@ func (s *Subscription) NextMsgWithContext(ctx context.Context) (*Msg, error) {
 	select {
 	case msg, ok = <-mch:
 		if !ok {
-			s.mu.Lock()
-			err = s.validateNextMsgState()
-			s.mu.Unlock()
-			return nil, err
+			return nil, s.getNextMsgErr()
 		}
 		if err := s.processNextMsgDelivered(msg); err != nil {
 			return nil, err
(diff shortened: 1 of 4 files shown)


context.go
```diff
@@ -153,7 +156,10 @@ func (s *Subscription) NextMsgWithContext(ctx context.Context) (*Msg, error) {
 	select {
 	case msg, ok = <-mch:
 		if !ok {
-			return nil, ErrConnectionClosed
+			s.mu.Lock()
```
derekcollison: Since we repeat this four times, let's make it its own func.

kozlovic: I see, will do.

nats.go
```diff
@@ -3162,6 +3156,18 @@ func (s *Subscription) validateNextMsgState() error {
 	return nil
 }
 
+// This is called when the sync channel has been closed.
+// The error returned will be either connection or subscription
+// closed depending on what caused NextMsg() to fail.
+func (s *Subscription) getNextMsgErr() error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
```
derekcollison: This is in fast (kinda) path, can we not use defer?

kozlovic: I was going to, but realize that this is when the subscription (or connection) is closed. I would not consider this fast path.

kozlovic: Meaning that this is a single event per subscription. After that call, subscription is dead.

derekcollison: Good point.

PR #495, opened 2019-07-22
exposed creds parsing of a creds file

aricart:
Thinking that we should have a common parsing for the creds. Likely since nats.go won't have other dependencies, the parsing code for a creds file should be hosted here.

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -793,7 +793,7 @@ func TokenHandler(cb AuthTokenHandler) Option {
 // for a user's JWT and a filename for the user's private Nkey seed.
 func UserCredentials(userOrChainedFile string, seedFiles ...string) Option {
 	userCB := func() (string, error) {
-		return userFromFile(userOrChainedFile)
+		return UserFromFile(userOrChainedFile)
 	}
 	var keyFile string
 	if len(seedFiles) > 0 {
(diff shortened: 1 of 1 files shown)


aricart:
> I was thinking the same thing for the OSCON demo but did not pull the trigger, wanted to think more about it and decide how to surface vs just making current functions public.

I thought of an util package, but usually those end up being catch-all. Maybe push to the JWT package?

derekcollison:
Let me look back at the OSCON demo later today. Working on another item this am.

derekcollison:
Looping in @kozlovic for his opinion as well.

kozlovic:
I understand the want to make them public and no objection per-se. But as always, we have to be careful what we make public since then we are stuck with it. As for the location for these utils, not sure. If you are in a situation where you would import nats.go just for those but are not using NATS api, then this is the wrong location.

derekcollison:
I needed them for the OSCON demo, but was using NATS. They could be on jwt pkg as well possibly.. But feels a bit better on nats.go IMO.

derekcollison:
I think having them in JWT makes sense until we know more about usage patterns of signing and encrypting connection payloads.

PR #500, opened 2019-07-31
Check bad subjects and queues on subscribe

derekcollison:
https://github.com/nats-io/nats-server/issues/84


derekcollison:
Fixed.

PR #503, opened 2019-08-03
Clear lastErr on reconnect success

derekcollison:

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -1623,7 +1623,6 @@ func normalizeErr(line string) string {
 // applicable. Will wait for a flush to return from the server for error
 // processing.
 func (nc *Conn) sendConnect() error {
-
 	// Construct the CONNECT protocol string
 	cProto, err := nc.connectProto()
 	if err != nil {
@@ -1850,6 +1849,17 @@ func (nc *Conn) doReconnect(err error) {
 
 		// Process connect logic
 		if nc.err = nc.processConnectInit(); nc.err != nil {
+			// If we have a lastErr recorded for this server
+			// do the normal processing here. We might get closed.
+			if nc.current.lastErr != nil {
+				nc.mu.Unlock()
+				nc.processErr(nc.err.Error())
+				nc.mu.Lock()
+				if nc.isClosed() {
+					break
+				}
+			}
+
 			nc.status = RECONNECTING
 			// Reset the buffered writer to the pending buffer
 			// (was set to a buffered writer on nc.conn in createConn)
--- a/nats.go
+++ b/nats.go
@@ -1453,6 +1453,7 @@ func (nc *Conn) connect() error {
 			if err == nil {
 				nc.srvPool[i].didConnect = true
 				nc.srvPool[i].reconnects = 0
+				nc.current.lastErr = nil
 				returnedErr = nil
 				break
 			} else {
@@ -1891,11 +1892,18 @@ func (nc *Conn) doReconnect(err error) {
 		if nc.Opts.ReconnectedCB != nil {
 			nc.ach.push(func() { nc.Opts.ReconnectedCB(nc) })
 		}
+
+		lastErr := nc.current.lastErr
+
 		// Release lock here, we will return below.
 		nc.mu.Unlock()
 
 		// Make sure to flush everything
 		nc.Flush()
+
+		if lastErr != nil && !nc.IsClosed() {
+			nc.clearCurrentLastErr()
+		}
 
 		return
 	}
@@ -2275,6 +2283,14 @@ func (nc *Conn) processAuthError(err error) {
 	} else {
 		nc.current.lastErr = err
 	}
+	nc.mu.Unlock()
+}
+
+// clearCurrentLastErr will clear the last error when we know we have
+// successfully connected after a flush.
+func (nc *Conn) clearCurrentLastErr() {
+	nc.mu.Lock()
+	nc.current.lastErr = nil
 	nc.mu.Unlock()
 }
 
--- a/nats.go
+++ b/nats.go
@@ -1852,8 +1852,9 @@ func (nc *Conn) doReconnect(err error) {
 			// If we have a lastErr recorded for this server
 			// do the normal processing here. We might get closed.
 			if nc.current.lastErr != nil {
+				err := nc.err
 				nc.mu.Unlock()
-				nc.processErr(nc.err.Error())
+				nc.processErr(err.Error())
 				nc.mu.Lock()
 				if nc.isClosed() {
 					break
--- a/nats_test.go
+++ b/nats_test.go
@@ -1454,6 +1454,9 @@ func TestExpiredUserCredentials(t *testing.T) {
 	if err := WaitTime(ch, 2*time.Second); err != nil {
 		t.Fatal("Should have closed after multiple failed attempts.")
 	}
+	if stats := nc.Stats(); stats.Reconnects > 2 {
+		t.Fatalf("Expected at most 2 reconnects, got %d", stats.Reconnects)
+	}
 }
 
 func TestExpiredUserCredentialsRenewal(t *testing.T) {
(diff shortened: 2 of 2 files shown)


nats.go
```diff
@@ -1891,12 +1892,19 @@ func (nc *Conn) doReconnect(err error) {
...
+
+		lastErr := nc.current.lastErr
+
 		// Release lock here, we will return below.
 		nc.mu.Unlock()
 
 		// Make sure to flush everything
 		nc.Flush()
 
+		if lastErr != nil && !nc.IsClosed() {
```
kozlovic: As seen in some of the Travis builds, I am not sure this would work. Looking at server code, I did not find a place where a client would be rejected right at the processing of the connect, therefore, I believe that the initial client PING (Flush here) could be successful and still have the auth expired right after. Which if that were to happen would have the effect to clear the error, get expired right after, reconnect, clear error, expire, etc..

derekcollison: The server should queue the error on CONNECT iff the credentials have expired when the CONNECT arrives. I am seeing the flappers. Ran it locally for >1k times for both tests and saw no problems.

derekcollison: I wonder if we should not fire the reconnect CB until we get the results of the flush back.

kozlovic: Can you point me to the server code where the -ERR is sent when receiving the CONNECT? I don't see that. If that is not happening, then the CONNECT followed by PING could result in PONG being received before any -ERR is received.

derekcollison: https://github.com/nats-io/nats-server/blob/master/server/auth.go#L321

IIRC CONNECT triggers a process CONNECT which will check auth and if false is returned will send ERR and close the connection.

https://github.com/nats-io/nats-server/blob/master/server/client.go#L1239

kozlovic: Then if that is the case, test should *always* fail since if we set reconnect attempts to -1, then the client will try to reconnect for ever. The connect process is synchronous. Client lib sends CONNECT + PING and wait in place for PONG or (ERR) (https://github.com/nats-io/nats.go/blob/master/nats.go#L1679). Note that it could fail at any point before that if connection is closed (so Flush() in sendConnect() could fail, so is readProto()). Still sendConnect() would then return an error, which will cause processConnectInit() to return an error. Therefore, it should loop trying to reconnect: https://github.com/nats-io/nats.go/blob/master/nats.go#L1851

derekcollison: Not sure I am following. If we get same error for same server that is auth error of some sort we bail and close the connection.

derekcollison: Happy to jump on GH if I am not following your train of thought correctly.

kozlovic: Say that the client connects OK (jwt still valid), when the JWT expires, the server will send -ERR and close connection. The client receives the -ERR from readLoop and call processErr which does not close connection and allow a reconnect.
Now we are in doReconnect(). The createConn() calls sendConnect() and once the JWT has expired, from what you said, the server should have sent an ERR right away, so the client lib should receive this in readProto() BEFORE readLoop is even started. So processErr is not in play here.

derekcollison: Let's jump on GH when you have a few minutes. I think I am following you but want to chat.

derekcollison: ok pushed a change that should handle correctly. Take a look when you have a minute.

nats.go
```diff
@@ -1850,6 +1849,17 @@ func (nc *Conn) doReconnect(err error) {
 
 		// Process connect logic
 		if nc.err = nc.processConnectInit(); nc.err != nil {
+			// If we have a lastErr recorded for this server
+			// do the normal processing here. We might get closed.
+			if nc.current.lastErr != nil {
+				nc.mu.Unlock()
+				nc.processErr(nc.err.Error())
```
kozlovic: I don't think this is correct. Suppose that the lib reconnects to a server that does not send an -ERR, but processConnectInit() returns an error (could be a TCP error after initial connect, etc..) then processErr() will cause the connection to be closed, therefore aborting the reconnect attempts.

Let me ask you this: why do we store the lastErr in the server object in the pool? Imagine a client with 10 server urls. It is connected to 1st, then its JWT expires and first server close the connection, client will reconnect to server2, which will return JWT expiration error on connect, so it will move to server 3, etc.. Only when reaching server 1 again the logic of twice the same auth expiration error would kick in.
If we consider that the User Authentication Expired error is a function of the client JWT, not the server, then there is no reason to store the last error in `srv` struct, should be in `Conn`. We actually don't need to store `lastErr` but simply a boolean that indicates that we got that an auth error. Then another boolean that indicates that we should abort (re)connect attempts if we get an auth error and the first boolean was set. 
A simple test could be done in processConnectInit() when we know we got an -ERR and check for the 2 possible auth errors and set those flags. Here in doReconnect we would check the abort flag and simply break if set.
I can try to work on that if you prefer or we could discuss more.

The only caveat is that you treat the AUTHORIZATION_ERR and AUTHENTICATION_EXPIRED_ERR with regard to lastErr and failing after 2nd error. Unlike AUTHENTICATION_EXPIRED_ERR which I argue is a function of the client, not the server, a client could get AUTHORIZATION_ERR on one server and not on another from its pool (which would mean that the auth config is not the same in the cluster).

derekcollison: I am trying to group the auths together in some fashion. The enclosing account could expire, be revoked, our user could be revoked etc.

Yes I fail when I "loop" back around and get the same error again. That is by design.

kozlovic: Point is that processErr() will close the connection if given -ERR is none of the one we allow for a reconnect. Here, you may not even get a -ERR passed to processErr() since it could be that we got a error such as socket closed, etc..

derekcollison: We only call it if we have a lastErr recorded.

kozlovic: Please re-read at least this:
```
I don't think this is correct. Suppose that the lib reconnects to a server that does not send an -ERR, but processConnectInit() returns an error (could be a TCP error after initial connect, etc..) then processErr() will cause the connection to be closed, therefore aborting the reconnect attempts.
```

derekcollison: The connection will be closed iff the connection had been established and a lastErr (which right now is auth only) was recorded.

kozlovic: 1- Client has S1, S2 in URLs and connects to S1 with valid JWT.
2- JWT about to expire so user updates JWT so that it is valid for next reconnect
3- JWT expires, server sends -ERR auth expired and closes conn
4- client processes async ERR and stores lastErr = AuthExpired
5- client start reconnect logic due to connection closed by server
6- client creates TCP connection to S2 but while reading from socket, server dies, socket closes. processConnectInit() returns an error (nc.err is NOT set at this point).
7- since err != nil and lastErr is set, you call processErr(nil), that will close the connection.

So client stops reconnecting while it should have go back to S1 and since JWT had been updated should succeed.

derekcollison: S2 will not have a lastErr set, so it should not close and will skip that conditional and go back to S1.

kozlovic: But processErr() will fall into the final `else` statement and close the connection because there is no -ERR passed to processErr()..

kozlovic: actually you will even get a panic() because you call `processErr(nc.err.Error())` whil nc.err could be nil.

derekcollison: S2 will not call processErr since lastErr will not be set. Maybe I am missing something and its time to jump on a GH.

derekcollison: Conditional is on nc.current.lastErr, not nc.err.

kozlovic: But we could have been back to S1 here where there was a recorded lastErr and still not be able to successfully return from processConnectInit() and it will close the connection and likely panic since nc.err may be cleared in this doReconnect() loop!

derekcollison: nc.err is set but could get cleared after we release the lock so will snapshot.

kozlovic: Apology for the ranting.. I was wrong. Yes, you set nc.err when calling nc.processConnectInit() so it won't be nil and we discussed and you are ok with connection being closed when we are back to say S1 that previously had lastErr, even if the actual nc.err is not an auth error.

PR #505, opened 2019-08-06
Trim "nats: " before invoking processErr in doReconnect

kozlovic:
Original test was flaky (and likely not produced proper sequence) on Travis and slower local Linux VM. Adding tracing we would see that the client keeps getting async -ERR auth expired errors. This is because the JWT is considered valid on each (re)CONNECT - even though it has expired - but server will then "immediately" send the expired error from timer's callback. The issue I believe is that JWT lib considers the JWT expired only if expires time > "now" with a mix of Unix() 1 sec granularity.
Removing the NATS Server/JWT from the test makes the test deterministic and actually test the library, not the server handling of expired JWTs...

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -1852,9 +1852,10 @@ func (nc *Conn) doReconnect(err error) {
 			// If we have a lastErr recorded for this server
 			// do the normal processing here. We might get closed.
 			if nc.current.lastErr != nil {
-				err := nc.err
+				// Remove possible "nats: " prefix
+				errStr := strings.TrimPrefix(nc.err.Error(), "nats: ")
 				nc.mu.Unlock()
-				nc.processErr(err.Error())
+				nc.processErr(errStr)
 				nc.mu.Lock()
 				if nc.isClosed() {
 					break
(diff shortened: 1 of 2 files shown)


nats.go
```diff
@@ -1904,18 +1909,12 @@ func (nc *Conn) doReconnect(err error) {
...
 
-		lastErr := nc.current.lastErr
-
 		// Release lock here, we will return below.
 		nc.mu.Unlock()
 
 		// Make sure to flush everything
 		nc.Flush()
 
-		if lastErr != nil && !nc.IsClosed() {
```
derekcollison: Why was this removed? Not needed? It was originally.

kozlovic: I moved the clearing up to line 1874. We need to clear under the lock after processConnectInit() did return OK and before the readLoop has a chance to start. Otherwise you would take the risk of clearing an error that could have been already sent by the server.
This was actually happening with the trimming fix but original test in a Linux VM. I could see the server sending async -ERR but going into reconnect loop because the error was cleared at the wrong place.

PR #511, opened 2019-08-22
Check-in missing test deps

variadico:
Currently, there are required dependencies that are missing from go.mod and
go.sum files. The files get updated when you run `go test`, we just haven't
checked them in.

Edit: module files get updated when you do a `go test`, not `go build`.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,9 @@
 module github.com/nats-io/nats.go
 
 require (
-	github.com/nats-io/jwt v0.2.12
+	github.com/golang/protobuf v1.3.2
+	github.com/nats-io/jwt v0.2.14
+	github.com/nats-io/nats-server/v2 v2.0.4
 	github.com/nats-io/nkeys v0.1.0
 	github.com/nats-io/nuid v1.0.1
 )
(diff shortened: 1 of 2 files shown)


kozlovic:
Are you sure? We don't need the server to build (but I may be wrong). We need to test. In my opinion we should have the dependencies only to build the library, not to test it.

variadico:
Aah, you're totally correct, @kozlovic!

I think we still may want to lock down the deps, though. Would it be possible for someone to run `go test` and get different versions of the server than when CI runs `go test`? Possibly leading to different test results?

kozlovic:
Oh sure, testing - if not locking the deps - would pull/use whatever if local/latest and may have different results, but I think a test should work regardless of the server version, and if a test requires a specific one (say version 2.0+, it usually has a conditional test at the beginning).
My main concern with pulling the server in the lib as a dep is that then we have Streaming server for instance that has a dep on both nats server and nats client, and so it looks weird. Also if we do lock deps because of testing, we should then do the same for the nats server and server needs lib for tests, so you can imagine the circular dependency. It may still work, but it makes my head spin thinking of that say we have a new server version and updates the deps on the client, but then client is a dep in the server, etc...

variadico:
Aaah, okay. Yeah, we can't do this otherwise, we get an import cycle. :+1:

Issue #509, opened 2019-08-20
nats.UserCredentials doesn't support tilde character


variadico:
I was reading the tutorial at https://synadia.com/ngs/signup. One of the steps in that tutorial is to call `UserCredentials` like this.

```
nats.UserCredentials("~/.nkeys/synadia/accounts/ngs/users/ngs.creds")
```

However, `UserCredentials` doesn't expand `~/` to the users home directory. So running that code gives you this error.

```
nats: open ~/.nkeys/synadia/accounts/ngs/users/ngs.creds: no such file or directory
```

We could add support to expand `~/` here https://github.com/nats-io/nats.go/blob/master/nats.go#L4018-L4025.

Or we could just update the code example on the signup page. Kinda leaning towards just updating the example.

derekcollison:
I like adding in expand support myself.

sasbury:
We use a go lib in ncs and ngs to help with that, Alberto will recall the
name it eludes me while on my phone

aricart:
https://github.com/mitchellh/go-homedir is the library you are looking for.
Also, that is not sufficient, usually the path has to be converted to an
absolute path after ~ expansions, and to eliminate other '.' or '..'.

variadico:
What do you guys think of updating the example?

```
nats.UserCredentials(homedir.Expand("~/.nkeys/synadia/accounts/ngs/users/ngs.creds"))

or even just

nats.UserCredentials("ngs.creds")
```

Reasoning:
* nats.go doesn't have to depend on go-homedir, 1 less dependency, less download, less code
* not sure if tilde paths are common for this project, maybe it's not worth it to users?
* library users can decide if they want to use go-homedir or not

sasbury:
ok as an example to me

derekcollison:
We can't just make the path correct with the tilde? Seems like should be quite doable, no?

variadico:
Well, looks like we already have a number of deps nats pulls in. So we could just add homedir. homedir seems to handle a bunch of edge cases. So seems like it would be easiest and least surprising to users.

PR #512, opened 2019-08-23
Expand paths with tilde

variadico:
Currently, if you pass a path with a tilde to nats.UserCredentials it fails
because it doesn't expand the tilde to mean the user's home directory.

This change adds support to expand paths with a tilde, such as "~/.nkeys".

Resolves https://github.com/nats-io/nats.go/issues/509

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -28,6 +28,8 @@ import (
 	"math/rand"
 	"net"
 	"net/url"
+	"os"
+	"path/filepath"
 	"runtime"
 	"strconv"
 	"strings"
@@ -4016,12 +4018,57 @@ func wipeSlice(buf []byte) {
 }
 
 func userFromFile(userFile string) (string, error) {
-	contents, err := ioutil.ReadFile(userFile)
+	path, err := expandPath(userFile)
+	if err != nil {
+		return _EMPTY_, fmt.Errorf("nats: %v", err)
+	}
+
+	contents, err := ioutil.ReadFile(path)
 	if err != nil {
 		return _EMPTY_, fmt.Errorf("nats: %v", err)
 	}
 	defer wipeSlice(contents)
 	return jwt.ParseDecoratedJWT(contents)
+}
+
+func homeDir() (string, error) {
+	if runtime.GOOS == "windows" {
+		homeDrive, homePath := os.Getenv("HOMEDRIVE"), os.Getenv("HOMEPATH")
+		userProfile := os.Getenv("USERPROFILE")
+
+		var home string
+		if homeDrive == "" || homePath == "" {
+			if userProfile == "" {
+				return _EMPTY_, errors.New("nats: failed to get home dir, require %HOMEDRIVE% and %HOMEPATH% or %USERPROFILE%")
+			}
+			home = userProfile
+		} else {
+			home = filepath.Join(homeDrive, homePath)
+		}
+
+		return home, nil
+	}
+
+	home := os.Getenv("HOME")
+	if home == "" {
+		return _EMPTY_, errors.New("nats: failed to get home dir, require $HOME")
+	}
+	return home, nil
+}
+
+func expandPath(p string) (string, error) {
+	p = os.ExpandEnv(p)
+
+	if !strings.HasPrefix(p, "~") {
+		return p, nil
+	}
+
+	home, err := homeDir()
+	if err != nil {
+		return _EMPTY_, err
+	}
+
+	return filepath.Join(home, p[1:]), nil
 }
 
 func nkeyPairFromSeedFile(seedFile string) (nkeys.KeyPair, error) {
--- a/nats_test.go
+++ b/nats_test.go
@@ -98,6 +98,103 @@ func TestVersionMatchesTag(t *testing.T) {
 	}
 }
 
+func TestExpandPath(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		origUserProfile := os.Getenv("USERPROFILE")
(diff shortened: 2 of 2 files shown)


variadico:
No, the logic isn't complex. However, the dep seems to handle a lot of edge cases.
https://github.com/mitchellh/go-homedir/blob/master/homedir.go

If we only want to solve the simple case, then we could probably just do this.
```
if strings.HasPrefix(path, "~") {
    path = filepath.Join(os.Getenv("HOME"), path[1:])
}
```

Maybe at some point we'll get an issue saying that there is a bug in the expansion logic. Or maybe not.

Another option is that we could just copy the file over into the nats repo (with all the license stuff).

wallyqs:
IMO checking whether $HOME or $home are being defined in env when `~` is the first char from the path would good enough, then can return an error saying explicitly saying that $HOME has to be defined for it to work.

ripienaar:
In my experience this handles the bulk of cases, hashi one is undeniably better but this is good enough for me. 


```
func homeDir() (string, error) {
	if runtime.GOOS == "windows" {
		drive := os.Getenv("HOMEDRIVE")
		home := os.Getenv("HOMEDIR")

		if home == "" || drive == "" {
			return "", fmt.Errorf("Cannot determine home dir, ensure HOMEDRIVE and HOMEDIR is set")
		}

		return filepath.Join(os.Getenv("HOMEDRIVE"), os.Getenv("HOMEDIR")), nil
	}

	home := os.Getenv("HOME")

	if home == "" {
		return "", fmt.Errorf("Cannot determine home dir, ensure HOME is set")
	}

	return home, nil
}
```

derekcollison:
I prefer to limit the dependencies unless we absolutely have to have them.

variadico:
All right. Implemented `homeDir`. Added tests.
https://github.com/nats-io/nats.go/pull/512/commits/80090aff5e243283fc924b23fdace3d3fd57c0eb

Wasn't sure if I should add tests for windows as well... I probably should, right?

variadico:
Oookay, so turns out this homedir function is already in the stdlib.
https://golang.org/pkg/os/#UserHomeDir

The other one in `os/user` is bad because it uses cgo. This one in `os`, basically does what we were doing, i.e. just checking env vars.

So, I deleted the original function we had. I also cross compiled from Linux to Windows to verify `os.UserHomeDir` worked. It correctly printed the home dir in command prompt, powershell, and the linux subsystem.

One weird thing is I'm not sure how to expect Windows users to give us a path. I added some Windows tests too, hopefully @ColinSullivan1 can take a look at those to see if they make sense.

nats_test.go
```diff
@@ -98,6 +98,94 @@ func TestVersionMatchesTag(t *testing.T) {
...
+			wantPath string
+			wantErr  bool
+		}{
+			{path: "/Foo/Bar", home: `C:\Foo\Bar`, wantPath: "/Foo/Bar", wantErr: false},
+			{path: "Foo/Bar", home: `C:\Foo\Bar`, wantPath: "Foo/Bar", wantErr: false},
+			{path: "~/Fizz", home: `C:\Foo\Bar`, wantPath: `C:\Foo\Bar\Fizz`, wantErr: false},
+
+			// missing USERPROFILE env var
+			{path: "~/fizz", wantErr: true},
+		}
```
variadico: Hey, @ColinSullivan1. Are these paths normally how Windows folks would give us a path?

nats.go
```diff
@@ -4016,14 +4018,32 @@ func wipeSlice(buf []byte) {
...
 	return jwt.ParseDecoratedJWT(contents)
 }
 
+func expandTildePath(p string) (string, error) {
+	if !strings.HasPrefix(p, "~") {
+		return p, nil
+	}
+
+	home, err := os.UserHomeDir()
+	if err != nil {
```
derekcollison: Should this just have ~/ -> /?

nats_test.go
```diff
@@ -98,6 +98,94 @@ func TestVersionMatchesTag(t *testing.T) {
 	}
 }
 
+func TestExpandTildePath(t *testing.T) {
+	t.Run("unix", func(t *testing.T) {
+		if runtime.GOOS == "windows" {
+			t.Skip()
```
derekcollison: Let's see if we can get one of these from @ColinSullivan1

nats.go
variadico: Hm? Sorry, don't follow.

derekcollison: if the homedir return an err should we just remove the ~?

nats_test.go
ColinSullivan1: Not normally, unless running in some type of non-standard shell.  paths would be \\foo\\bar, wtih backslashes (escaped for the test).

nats_test.go
ColinSullivan1: Will do...

nats.go
```diff
@@ -4016,14 +4018,32 @@ func wipeSlice(buf []byte) {
...
 	defer wipeSlice(contents)
 	return jwt.ParseDecoratedJWT(contents)
 }
 
+func expandTildePath(p string) (string, error) {
+	if !strings.HasPrefix(p, "~") {
+		return p, nil
+	}
+
+	home, err := os.UserHomeDir()
```
ColinSullivan1: Golang doesn't check for `%HOMEDRIVE%%HOMEPATH%` which is also used.  It'd be more complicated but great for users if env var expansion in general was supported.

derekcollison:
Congrats! First PR merge!

PR #514, opened 2019-08-30
[ADDED] Allow client to choose whether to execute callbacks on Close()

### LINKED ISSUE #513: Client calling `Close()` should not execute callbacks
On go the close/disconnect handlers are called when the client performs a close(). This is non-sensical as the application called `Close()` so it knows that it called it itself.

On other conditions, the notification is completely valid - as the application is not expecting the connection to have gone away. 

The main issue is:
```
// Close will close the connection to the server. This call will release
// all blocking calls, such as Flush() and NextMsg()
func (nc *Conn) Close() {
    if nc != nil {
        nc.close(CLOSED, true, nil)
    }
}
```
This should have the `doCBs` argument set to `false`, and internal uses by `doReconnect()`, `drainConnectino()`, `processErr()`, `processOpErr()` should be calling the internal `close()` with the correct arguments for the situation.

aricart:
Added connection option `NoCallbacksAfterClientClose` which disables the invocation of callbacks if the client code directly calls `Close()`. In case of a library initiated close, callbacks will be
invoked as they currently are.

### DIFF
--- a/test/reconnect_test.go
+++ b/test/reconnect_test.go
@@ -727,61 +727,45 @@ func TestConnCloseNoCallback(t *testing.T) {
 	ts := startReconnectServer(t)
 	defer ts.Shutdown()
 
-	cch := make(chan bool)
+	// create a connection that manually sets the options
+	var conns []*nats.Conn
+	cch := make(chan string, 2)
 	opts := reconnectOpts
 	opts.ClosedCB = func(_ *nats.Conn) {
-		cch <- true
+		cch <- "manual"
 	}
 	opts.NoCallbacksAfterClientClose = true
-
 	nc, err := opts.Connect()
 	if err != nil {
 		t.Fatalf("Should have connected ok: %v", err)
 	}
-	defer nc.Close()
+	conns = append(conns, nc)
 
-	nc.Flush()
-	// Close the connection, we don't expect to get a notification
-	nc.Close()
-	// Shutdown the server
-	ts.Shutdown()
-
-	// Even on Windows (where a createConn takes more than a second)
-	// we should be able to break the reconnect loop with the following
-	// timeout.
-	if err := WaitTime(cch, 3*time.Second); err != nil {
-		// yay no callback
-	} else {
-		t.Fatal("Got a closed callback, but shouldn't have")
-	}
-}
-
-func TestConnCloseNoCallbackFromOptionsFunc(t *testing.T) {
-	ts := startReconnectServer(t)
-	defer ts.Shutdown()
-
-	cch := make(chan bool)
-	nc, err := nats.Connect(reconnectOpts.Url, nats.NoCallbacksAfterClientClose(),
+	// and another connection that uses the option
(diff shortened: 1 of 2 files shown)


test/reconnect_test.go
```diff
@@ -722,3 +722,66 @@ func TestReconnectTLSHostNoIP(t *testing.T) {
...
+	nc, err := opts.Connect()
+	if err != nil {
+		t.Fatalf("Should have connected ok: %v", err)
+	}
+	defer nc.Close()
+
+	nc.Flush()
+	// Close the connection, we don't expect to get a notification
+	nc.Close()
+	// Shutdown the server
```
kozlovic: I would not shutdown the server here, not necessary to test that the callback is not invoked.

test/reconnect_test.go
```diff
@@ -722,3 +722,66 @@ func TestReconnectTLSHostNoIP(t *testing.T) {
...
+	ts.Shutdown()
+
+	// Even on Windows (where a createConn takes more than a second)
```
kozlovic: Waiting 3 secs is too much. A 500ms is more than enough for the callback to be invoked if it was.

test/reconnect_test.go
```diff
@@ -722,3 +722,66 @@ func TestReconnectTLSHostNoIP(t *testing.T) {
...
+		}))
+	if err != nil {
+		t.Fatalf("Should have connected ok: %v", err)
+	}
+	defer nc.Close()
+
+	nc.Flush()
+	// Close the connection, we don't expect to get a notification
+	nc.Close()
+	// Shutdown the server
```
kozlovic: Same comments for the rest of this test. Actually you should have combined them in a single test and have nc1 and nc2 (one with options one with func) and close them both and have a single select. (I would then have created cch as a buffered channel of size 2).

go.mod
```diff
@@ -1,7 +1,9 @@
 module github.com/nats-io/nats.go
 
 require (
-	github.com/nats-io/jwt v0.2.12
+	github.com/golang/protobuf v1.3.2
```
kozlovic: Please revert changes here. We don't want the go.mod to include deps for tests.

go.sum
```diff
@@ -1,5 +1,12 @@
+github.com/golang/protobuf v1.3.2 h1:6nsPYzhq5kReh6QImI3k5qWzO4PEbvbIW2cwSfR/6xs=
```
kozlovic: Revert changes to this file.

test/reconnect_test.go
```diff
@@ -722,3 +722,50 @@ func TestReconnectTLSHostNoIP(t *testing.T) {
...
+		c.Close()
+	}
+
+	// if the timeout happens we didn't get data from the channel
+	// if we get a value from the channel that connection type failed.
+	select {
+	case <-time.After(500 * time.Millisecond):
+		// test passed - we timed so no callback was called
+	case what := <-cch:
+		t.Fatal(fmt.Sprintf("%s issued a callback and it shouldn't have", what))
```
kozlovic: Could be replaced with simple `t.Fatalf("%s issued a ...`

PR #521, opened 2019-09-16
[FIXED] Data race between processMsg() and Stats()

### LINKED ISSUE #520: Race Condition between processMsg() and Stats()
Client version: 1.8.1

The following race condition occurred while trying to call the `Stats` function while a message was being processed. Note the `Stats` function call in the test that produced this, was being called from a go routine to periodically scrap the NATS client stats.
```
WARNING: DATA RACE
Write at 0x00c000254000 by goroutine 48:
  github.com/******/******/vendor/github.com/nats-io/nats%2ego.(*Conn).processMsg()
      /go/src/******/******/vendor/github.com/nats-io/nats.go/nats.go:2083 +0xb3
  github.com/******/******/vendor/github.com/nats-io/nats%2ego.(*Conn).parse()
      /go/src/github.com/******/******/vendor/github.com/nats-io/nats.go/parser.go:177 +0x2c05
  github.com/******/******/vendor/github.com/nats-io/nats%2ego.(*Conn).readLoop()
      /go/src/github.com/******/******/vendor/github.com/nats-io/nats.go/nats.go:1983 +0x59e

Previous read at 0x00c000254000 by goroutine 60:
  github.com/******/******/vendor/github.com/nats-io/nats%2ego.(*Conn).Stats()
      /go/src/github.com/******/******/vendor/github.com/nats-io/nats.go/nats.go:3712 +0xbf
  github.com/******/******/vendor/github.com/******/******/******.(*Client).natsStatsMonitor()
      /go/src/github.com/******/******/vendor/github.com/******/******/******/conn.go:43 +0x209
```

I believe https://github.com/nats-io/nats.go/blob/master/nats.go#L2164-L2168 is using a Read only lock while it is doing an update to `nc.InBytes` and `nc.InMsgs`, when it should be a Write lock as well.

kozlovic:
Resolves #520

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -2164,8 +2164,8 @@ func (nc *Conn) processMsg(data []byte) {
 	nc.subsMu.RLock()
 
 	// Stats
-	nc.InMsgs++
-	nc.InBytes += uint64(len(data))
+	atomic.AddUint64(&nc.InMsgs, 1)
+	atomic.AddUint64(&nc.InBytes, uint64(len(data)))
 
 	sub := nc.subs[nc.ps.ma.sid]
 	if sub == nil {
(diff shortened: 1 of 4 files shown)


nats.go
```diff
@@ -3884,15 +3884,22 @@ func (nc *Conn) Stats() Statistics {
 	// Stats are updated either under connection's mu or subsMu mutexes.
 	// Lock both to safely get them.
 	nc.mu.Lock()
-	nc.subsMu.RLock()
+	// Note: It would be more logical to get subsMu's read lock here since we are
+	// "reading" the stats. However, in processMsg() we don't lock the connection
+	// (unless we detect a slow consumer) but we do get subsMu read lock to fetch
+	// the subscription the incoming message is for. We do update InMsgs/Bytes under
+	// that read lock. Swithching processMsg()'s subsMu to write lock would decrease
```
derekcollison: Spelling on Switching.

derekcollison: We do an update under a read lock? That sounds wrong.

kozlovic: Replaced all that with updates of InMsgs/InBytes in processMsg() using atomic add and atomic load in Stats().

derekcollison: How does that affect performance? I think that or a simple lock depending on which performs better.

kozlovic: Getting subsMu.Lock() (instead of RLock) dropped perf by 5%. Replacing with atomic dropped by 2%, this is why I changed.
Note that although unconventional, the original approach of getting the write lock in Stats() as opposed to processMsg() was not affecting performance (obviously since there was no change in processMsg()).

derekcollison: I like perf more than anyone, but this is the correct way IMO. Thanks.

PR #529, opened 2019-11-12
PubSub from same connection performance improvement

derekcollison:
PubSub from same connection was contending on lock checking closed state.


norace_test.go
```diff
@@ -0,0 +1,97 @@
...
+// +build !race
+
+package nats
+
+import (
+	"testing"
+	"time"
+)
+
+func TestParseStateReconnectFunctionality(t *testing.T) {
```
kozlovic: If you move this to a new "norace_test.go" file, I would recommend naming this test TestNoRace... so that from travis you add a new command to run all the "no race" tests like: `go test -v -run=TestNoRace -p=1 ./...`

go.mod
```diff
@@ -1,7 +1,9 @@
 module github.com/nats-io/nats.go
 
 require (
+	github.com/golang/protobuf v1.3.2
 	github.com/nats-io/jwt v0.3.0
+	github.com/nats-io/nats-server/v2 v2.1.0
```
kozlovic: So far we have avoided adding the server deps because it is just for test. If we start to do the same on the server side, they will both depend on each other and so a project that pull nats.go say at 1.9.2 would need server v2.1.0 but if later we have server v2.1.1 that pulls nats.go v1.9.1, ... I don't know, my head spins already..

derekcollison: Sorry, last commit pulled it in.

norace_test.go
derekcollison: Sounds reasonable.

go.mod
derekcollison: Will revert.

go.mod
```diff
@@ -3,7 +3,6 @@ module github.com/nats-io/nats.go
 require (
 	github.com/golang/protobuf v1.3.2
```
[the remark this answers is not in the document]
derekcollison: yes

PR #536, opened 2019-12-31
support acknowledging messages received from jetstream

ripienaar:
Small helper to assist with push based observable testing on JS

### DIFF
--- a/examples/nats-sub/main.go
+++ b/examples/nats-sub/main.go
@@ -18,6 +18,7 @@ import (
 	"log"
 	"os"
 	"runtime"
+	"strings"
 	"time"
 
 	"github.com/nats-io/nats.go"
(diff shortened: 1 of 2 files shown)


ripienaar:
staticheck moaning about 

```
$ staticcheck ./...

test/conn_test.go:478:13: unnecessary use of fmt.Sprintf (S1039)
test/conn_test.go:1968:20: unnecessary use of fmt.Sprintf (S1039)
```

those problems are preexisting

derekcollison:
Yes we use head for staticcheck on travis so we see this from time to time.

ripienaar:
OK, `jsm obs next` aka `jsm obs sub` when subscribing to a push based observable will now do something like `nats-sub` and supports acknowledging messages etc

PR #539, opened 2020-01-10
support setting a reply subject when publishing messages

ripienaar:


ripienaar:
nats-req doesn't support listening for more than 1 reply.  I wanted a easy way to test js next with a batch of 10.  But in general being able to publish with this set is useful in case the listener is a different process

derekcollison:
Should we add optional way for nats-req to wait for more responses?

ripienaar:
Yeah I want to do that too - I dont think Request() support that so would need to set up a inbox and all sorts, but yeh i want to do that

derekcollison:
But you still want option for reply for nats-pub yes?

I would probably keep nc.Publish() and add conditional if reply != "" use nc.PublishReply().

ripienaar:
Yeah as a general toy to play with patterns its nice to be able to put arbitrary replies :) Ok - How'd I miss`PublishReply` :P will rework

derekcollison:
No worries.

ripienaar:
Updated and updated for latest staticcheck, remaining issues not related to this PR

PR #543, opened 2020-02-12
publish not checking reconnecting status after closing.

theduman:
After successful initialization if nats server fails Conn status is set to Reconnecting but publish function is not checking and not returning error.

image

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -2558,6 +2558,11 @@ func (nc *Conn) publish(subj, reply string, data []byte) error {
 		return ErrConnectionDraining
 	}
 
+	if nc.isReconnecting() {
+		nc.mu.Unlock()
+		return ErrConnectionReconnecting
+	}
+
 	// Proactively reject payloads over the threshold set by server.
 	msgSize := int64(len(data))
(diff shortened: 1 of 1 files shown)


ripienaar:
I think that's intended behaviour because messages published during reconnect goes into the buffer set by ReconnectBufSize and will be published when reconnected. If the buffer is full you'll get ErrReconnectBufExceeded

theduman:
> I think that's intended behaviour because messages published during reconnect goes into the buffer set by ReconnectBufSize and will be published when reconnected. If the buffer is full you'll get ErrReconnectBufExceeded

If I fail publishing and reconnecting state is fail for me in my case, I need to know right in time. Is there any connection check function?

ripienaar:
There is IsConnected() but stil some buffering, set reconnect buffer to like 1 byte and i guess you'll always be notified via errors?

theduman:
> There is IsConnected() but stil some buffering, set reconnect buffer to like 1 byte and i guess you'll always be notified via errors?

reconnect buffer solved my problem. thanks mate

PR #560, opened 2020-04-27
Added generic header support

derekcollison:
This is meant to mirror HTTP headers as much as possible for interop not only between our clients but HTTP clients in general. I actually used the http.Header directly, so we get canonical names, etc.

I thought about adding in a way to add the data field to a message such that you could detect mime-type and content-length etc. But Go's builtin type detection is not that good, so stuck with basics.

```go
sub, err := nc.SubscribeSync("headers")

m := nats.NewMsg("headers")
m.Header.Add("Accept-Encoding", "json")
m.Header.Add("Authorization", "s3cr3t")
m.Data = []byte("Hello Headers!")

nc.PublishMsg(m)
rmsg, _ := sub.NextMsg(time.Second)

auth := rmsg.Header.Get("Authorization") // auth is "s3cr3t"

```


derekcollison:
On the wire the above looks like the following..

```
"⚡NATS⚡/0.1\r\nAccept-Encoding: json\r\nAuthorization: s3cr3t\r\n\r\nHello Headers!"
```

nats.go
```diff
@@ -2518,13 +2544,71 @@ func (nc *Conn) Publish(subj string, data []byte) error {
...
 
+// Used to create a new message for publishing that will use headers.
+func NewMsg(subject string) *Msg {
+	return &Msg{
+		Subject: subject,
+		Header:  make(http.Header),
+	}
+}
+
+const hdrMagic = "⚡NATS⚡/0.1\r\n"
```
ripienaar, line 2555: this seems pretty good in general except for the little lightnings :) telnet protocol demos nicely, this makes it hard

derekcollison: We need something unique to determine that this message has headers and we are text based so needed something like this. Have seen others doing it as well.

You going to put headers into a message via telnet?

ripienaar: Probably not, but would we want to make it impossible?

derekcollison: Not sure if telnet can send utf8

ripienaar: It's a minor nit, just remember reading the protocol spec and it made a point of being clear text and easy to play with over something like telnet - in reality thats useless, but its a good philosophy about the protocol as it has implications for how light weight clients can be etc

derekcollison: Yea I am playing with it, does not work yet but I think if I use escape codes instead of utf8 directly it may. Will report back.

derekcollison: Sounds like clients will need to indicate they can handle headers via CONNECT anyway, meaning telnet sessions etc will most likely not send or receive headers.

derekcollison: We moved this to new proto. Will update this branch to new protos.

derekcollison:
Folks, changed this over to the protocol based version, please take another look if you can.

test/headers_test.go
```diff
@@ -0,0 +1,80 @@
...
+	sub, err := nc.SubscribeSync(subject)
+	if err != nil {
+		t.Fatalf("Could not subscribe to %q: %v", subject, err)
+	}
+	defer sub.Unsubscribe()
+
+	m := nats.NewMsg(subject)
+	m.Header.Add("Accept-Encoding", "json")
+	m.Header.Add("Authorization", "s3cr3t")
+	m.Data = []byte("Hello Headers!")
```
ripienaar, line 45: not a huge fan of accessing these struct keys directly, but I guess that's what we do already so more of the same is ok.  Would have been really nice if nats.Msg could be a interface

go.mod
```diff
@@ -1,7 +1,12 @@
 module github.com/nats-io/nats.go
 
+go 1.14
```
ripienaar, line 3: intentional? think we also need `go mod tidy` here

test/headers_test.go
derekcollison: Use this directly - https://golang.org/pkg/net/http/#Header

go.mod
derekcollison: Agree will fix.

test/headers_test.go
ripienaar: Not sure how that relates, my point is I wish the Msg being sent around was a interface and not a struct, it would make it much easier to support things like different encodings etc.

But not something we can realistically fix, was just a observation

derekcollison: We are already working on a v2 generic client that is more focused on streams and services and will be JetStream aware, so that would be the place to do so.

Not following the different encodings though.

ripienaar: > We are already working on a v2 generic client that is more focused on streams and services and will be JetStream aware, so that would be the place to do so.

News to me :)

derekcollison: Just has a brainstorming meeting and some code by Wally and I a couple of weeks ago. Still very early, so plenty of time for review and input.

nats.go
```diff
@@ -2264,8 +2273,27 @@ func (nc *Conn) processMsg(data []byte) {
...
+	// Check if we have headers encoded here.
+	var h http.Header
+	var err error
+
+	if nc.ps.ma.hdr > 0 {
+		hbuf := msgPayload[:nc.ps.ma.hdr]
+		msgPayload = msgPayload[nc.ps.ma.hdr:]
+		h, err = decodeHeadersMsg(hbuf)
+		if err != nil {
+			// We will pass the message through but send async error.
```
kozlovic, line 2285: Should we though?

nats.go
```diff
@@ -2610,14 +2672,26 @@ func (nc *Conn) PublishMsg(m *Msg) error {
 	if m == nil {
 		return ErrInvalidMsg
 	}
-	return nc.publish(m.Subject, m.Reply, m.Data)
+	var hdr []byte
+	if len(m.Header) > 0 {
+		if !nc.info.Headers {
```
kozlovic, line 2677: Seeing this, what happens to messages that are buffered during a disconnect and flushed on a reconnect, possibly to a server that does not support them. Will have to document that every server in a cluster is expected to be at same level.

nats.go
```diff
@@ -2664,23 +2738,46 @@ func (nc *Conn) publish(subj, reply string, data []byte) error {
...
-		for l := len(data); l > 0; l /= 10 {
+
+	if hdr != nil {
+		if len(hdr) > 0 {
+			for l := len(hdr); l > 0; l /= 10 {
+				i -= 1
+				b[i] = digits[l%10]
+			}
+		} else {
+			i -= 1
```
kozlovic, line 2770: Same

nats.go
```diff
@@ -2664,23 +2738,46 @@ func (nc *Conn) publish(subj, reply string, data []byte) error {
...
 
 	var b [12]byte
 	var i = len(b)
-	if len(data) > 0 {
-		for l := len(data); l > 0; l /= 10 {
+
+	if hdr != nil {
+		if len(hdr) > 0 {
+			for l := len(hdr); l > 0; l /= 10 {
+				i -= 1
```
kozlovic, line 2766: I know this was there before, but linters flag `i -= 1`, replace with `i--`

nats.go
```diff
@@ -2664,23 +2738,46 @@ func (nc *Conn) publish(subj, reply string, data []byte) error {
...
+		}
+		mh = append(mh, b[i:]...)
+		mh = append(mh, ' ')
+		// reset for below.
+		i = len(b)
+	}
+
+	if msgSize > 0 {
+		for l := msgSize; l > 0; l /= 10 {
 			i -= 1
```
kozlovic, line 2781: While at it, and the one down below :-)

nats.go
```diff
@@ -2601,7 +2629,41 @@ func (nc *Conn) kickFlusher() {
...
+// Used to create a new message for publishing that will use headers.
+func NewMsg(subject string) *Msg {
+	return &Msg{
+		Subject: subject,
+		Header:  make(http.Header),
+	}
+}
+
+const (
+	hdrLine   = "NATS/1.0\r\n"
```
matthiashanel, line 2644: the .0 needed?

nats.go
```diff
@@ -1660,7 +1669,7 @@ func (nc *Conn) connectProto() (string, error) {
 	}
 
 	cinfo := connectInfo{o.Verbose, o.Pedantic, ujwt, nkey, sig, user, pass, token,
-		o.Secure, o.Name, LangString, Version, clientProtoInfo, !o.NoEcho}
+		o.Secure, o.Name, LangString, Version, clientProtoInfo, !o.NoEcho, true}
```
matthiashanel, line 1672: I wonder if true should become the header version 1.
And a corresponding check on the server client has header enabled if server supported header version >= client header version.

A client talking header version 2 should probably be downgraded to one without header when the server only speaks version 1.

nats.go
derekcollison: I think so in case someone or some other client language is messing up the headers, body should still pass thru IMO.

nats.go
derekcollison: If we queued messages that had headers and we reconnect to a server that does not we will try to send them and get disconnected I believe.

nats.go
derekcollison: Will fix.

nats.go
```diff
@@ -2601,7 +2629,41 @@ func (nc *Conn) kickFlusher() {
...
+	}
+	// FIXME(dlc) - pull status and description?
+	if !strings.HasPrefix(hl, hdrLine[:hdrPreEnd]) {
+		return nil, ErrBadHeaderMsg
+	}
+	tp := textproto.NewReader(br)
+	mh, err := tp.ReadMIMEHeader()
+	if err != nil {
+		return nil, ErrBadHeaderMsg
+	}
```
matthiashanel, line 2665: Any chance we can get at how many bytes the reader consumed and check to see if that matches the header/msg boundary

nats.go
derekcollison: Following convention from HTTP.

nats.go
derekcollison: Not sure what the ReadMIMEHeader() does but like I said may unpack or at least at a check for the trailing `\r\n\r\n`

nats.go
derekcollison: So change it to an into or string? I think I would prefer to add later when we cross that bridge, and if its needed.

matthiashanel: I had an int in mind, but yeah. 
The problem is if you change to a different version in the future, the then "old" server code, already needs to handle unknown versions.

derekcollison: It will since it will be looking for the first line to be NATS/1.0

matthiashanel: A message from a client sending a v2 header will pass through an old server and can be received by a v1 client who then trips.

compared to.

When you communicate your version to the server it can enable/disable header based on its understanding of header. Essentially what we do today when adding header. If you don't understand them we strip them.

derekcollison: I understand, probably will not need this for many years and would be a bunch of work. Also clients should check the version inlined in the header so they don't trip.

derekcollison: Meaning we need client support regardless, not sure servers need to be involved.

matthiashanel: Server need to understand versions as soon as they start reading them. (tracing)
They can provide a migration path by stripping the header if it is beyond what the client/other server can understand.

derekcollison: The server will be acting like a client here as well, same logic. Read first line of header and check for `NATS/V.v`

nats.go
derekcollison: I cleaned it up a bit.

derekcollison:
I think I am ready to land this.

derekcollison:
Point noted, but server will need to treat all payloads as foreign anyway so even with this not much benefit. And same with clients TBH.

PR #564, opened 2020-05-01
[FIXED] Added jitter in the reconnect logic

### LINKED ISSUE #563: Add jitter during reconnect
## Defect

Some jitter needs to be added during reconnect.  One option would be waiting for a duration that is a percentage of Reconnect wait - it'd be easy for users to configure.

kozlovic:
When library would wait for the reconnect wait interval, it will
now wait for a random time bounded by this value.
This prevents thundering herd reconnect issue, especially with
TLS.

Resolves #563

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -517,13 +517,12 @@ type Statistics struct {
 
 // Tracks individual backend servers.
 type srv struct {
-	url         *url.URL
-	didConnect  bool
-	reconnects  int
-	lastAttempt time.Time
-	lastErr     error
-	isImplicit  bool
-	tlsName     string
+	url        *url.URL
+	didConnect bool
+	reconnects int
+	lastErr    error
+	isImplicit bool
+	tlsName    string
 }
 
 type serverInfo struct {
@@ -1299,8 +1298,6 @@ func (nc *Conn) createConn() (err error) {
 	}
 	if _, cur := nc.currentServer(); cur == nil {
 		return ErrNoServers
-	} else {
-		cur.lastAttempt = time.Now()
 	}
 
 	// We will auto-expand host names if they resolve to multiple IPs
--- a/nats.go
+++ b/nats.go
@@ -45,19 +45,21 @@ import (
 
 // Default Constants
 const (
-	Version                 = "1.9.2"
-	DefaultURL              = "nats://127.0.0.1:4222"
-	DefaultPort             = 4222
-	DefaultMaxReconnect     = 60
-	DefaultReconnectWait    = 2 * time.Second
-	DefaultTimeout          = 2 * time.Second
-	DefaultPingInterval     = 2 * time.Minute
-	DefaultMaxPingOut       = 2
-	DefaultMaxChanLen       = 8192            // 8k
-	DefaultReconnectBufSize = 8 * 1024 * 1024 // 8MB
-	RequestChanLen          = 8
-	DefaultDrainTimeout     = 30 * time.Second
-	LangString              = "go"
+	Version                   = "1.9.3"
+	DefaultURL                = "nats://127.0.0.1:4222"
+	DefaultPort               = 4222
+	DefaultMaxReconnect       = 60
+	DefaultReconnectWait      = 2 * time.Second
+	DefaultReconnectJitter    = 100 * time.Millisecond
+	DefaultReconnectJitterTLS = time.Second
+	DefaultTimeout            = 2 * time.Second
+	DefaultPingInterval       = 2 * time.Minute
+	DefaultMaxPingOut         = 2
+	DefaultMaxChanLen         = 8192            // 8k
+	DefaultReconnectBufSize   = 8 * 1024 * 1024 // 8MB
+	RequestChanLen            = 8
+	DefaultDrainTimeout       = 30 * time.Second
+	LangString                = "go"
 )
 
 const (
@@ -127,15 +129,17 @@ func init() {
 // GetDefaultOptions returns default configuration options for the client.
 func GetDefaultOptions() Options {
 	return Options{
-		AllowReconnect:   true,
-		MaxReconnect:     DefaultMaxReconnect,
-		ReconnectWait:    DefaultReconnectWait,
-		Timeout:          DefaultTimeout,
-		PingInterval:     DefaultPingInterval,
-		MaxPingsOut:      DefaultMaxPingOut,
-		SubChanLen:       DefaultMaxChanLen,
-		ReconnectBufSize: DefaultReconnectBufSize,
-		DrainTimeout:     DefaultDrainTimeout,
+		AllowReconnect:     true,
+		MaxReconnect:       DefaultMaxReconnect,
+		ReconnectWait:      DefaultReconnectWait,
+		ReconnectJitter:    DefaultReconnectJitter,
+		ReconnectJitterTLS: DefaultReconnectJitterTLS,
+		Timeout:            DefaultTimeout,
+		PingInterval:       DefaultPingInterval,
+		MaxPingsOut:        DefaultMaxPingOut,
+		SubChanLen:         DefaultMaxChanLen,
+		ReconnectBufSize:   DefaultReconnectBufSize,
+		DrainTimeout:       DefaultDrainTimeout,
 	}
 }
 
@@ -181,6 +185,12 @@ type SignatureHandler func([]byte) ([]byte, error)
 
 // AuthTokenHandler is used to generate a new token.
 type AuthTokenHandler func() string
+
+// ReconnectDelayHandler is used to get from the user the desired
+// delay the library should pause before attempting to reconnect
+// again. Note that this is invoked after the library tried the
+// whole list of URLs and failed to reconnect.
+type ReconnectDelayHandler func(attempts int) time.Duration
 
 // asyncCB is used to preserve order for async callbacks.
 type asyncCB struct {
@@ -257,6 +267,24 @@ type Options struct {
 	// ReconnectWait sets the time to backoff after attempting a reconnect
 	// to a server that we were already connected to previously.
 	ReconnectWait time.Duration
+
+	// CustomReconnectDelayCB is invoked after the library tried every
+	// URL in the server list and failed to reconnect. It passes to the
+	// user the current number of attempts. This function returns the
+	// amount of time the library will sleep before attempting to reconnect
+	// again. It is strongly recommended that this value contains some
+	// jitter to prevent all connections to attempt reconnecting at the same time.
+	CustomReconnectDelayCB ReconnectDelayHandler
+
+	// ReconnectJitter sets the upper bound for a random delay added to
+	// ReconnectWait during a reconnect when no TLS is used.
+	// Note that any jitter is capped with ReconnectJitterMax.
+	ReconnectJitter time.Duration
+
+	// ReconnectJitterTLS sets the upper bound for a random delay added to
+	// ReconnectWait during a reconnect when TLS is used.
+	// Note that any jitter is capped with ReconnectJitterMax.
+	ReconnectJitterTLS time.Duration
 
 	// Timeout sets the timeout for a Dial operation on a connection.
 	Timeout time.Duration
@@ -411,6 +439,7 @@ type Conn struct {
 	ptmr    *time.Timer
 	pout    int
 	ar      bool // abort reconnect
+	rqch    chan struct{}
 
 	// New style response handler
 	respSub   string               // The wildcard subject
@@ -668,6 +697,24 @@ func ReconnectWait(t time.Duration) Option {
 func MaxReconnects(max int) Option {
 	return func(o *Options) error {
 		o.MaxReconnect = max
+		return nil
+	}
+}
+
+// ReconnectJitter is an Option to set the upper bound of a random delay added ReconnectWait.
+func ReconnectJitter(jitter, jitterForTLS time.Duration) Option {
+	return func(o *Options) error {
+		o.ReconnectJitter = jitter
+		o.ReconnectJitterTLS = jitterForTLS
+		return nil
+	}
+}
+
+// CustomReconnectDelay is an Option to set the CustomReconnectDelayCB option.
+// See CustomReconnectDelayCB Option for more details.
+func CustomReconnectDelay(cb ReconnectDelayHandler) Option {
+	return func(o *Options) error {
+		o.CustomReconnectDelayCB = cb
 		return nil
 	}
 }
@@ -1396,6 +1443,7 @@ func (nc *Conn) setup() {
 	nc.pongs = make([]chan struct{}, 0, 8)
 
 	nc.fch = make(chan struct{}, flushChanSize)
+	nc.rqch = make(chan struct{})
 
 	// Setup scratch outbound buffer for PUB
 	pub := nc.scratch[:len(_PUB_P_)]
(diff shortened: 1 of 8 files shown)


kozlovic:
I added a random delay capped to the reconnect wait. Also made sure that we kick out the doReconnect go routine (was previously possibly waiting for reconnectWait even when connection was closed.

Should we have a minimum wait? For instance use random but for [reconnectWait/2:reconnectWait]?

kozlovic:
Some tests will now sometimes fail because they expect the connection to be reconnecting based on the ReconnectWait duration. Waiting for decision on wait time before fixing those (for instance: TestReconnectAllowedFlags
    TestReconnectAllowedFlags: reconnect_test.go:104: Expected to be in a reconnecting state)

go.mod
```diff
@@ -1,7 +1,12 @@
 module github.com/nats-io/nats.go
 
 require (
+	github.com/golang/protobuf v1.4.0
 	github.com/nats-io/jwt v0.3.2
+	github.com/nats-io/nats-server/v2 v2.1.6
```
wallyqs, line 6: I know that go mod tooling added this, but should we continue to remove it to avoid depending on server?

go.mod
```diff
@@ -1,7 +1,12 @@
...
 	github.com/nats-io/nkeys v0.1.4
 	github.com/nats-io/nuid v1.0.1
+	google.golang.org/protobuf v1.21.0
 )
+
+go 1.13
```
wallyqs, line 12: Not completely sure of the impact of adding this one yet too...

go.mod
kozlovic: Arg.. it's my local update of staticcheck that did that likely and I forgot to remove. Will update.

go.mod
kozlovic: I will reset the go.mod/sum changes, but will be easier for me to do a git push force, so be aware.

nats.go
```diff
@@ -1844,7 +1848,11 @@ func (nc *Conn) doReconnect(err error) {
 		if sleepTime <= 0 {
 			runtime.Gosched()
 		} else {
-			time.Sleep(time.Duration(sleepTime))
+			rt.Reset(time.Duration(rand.Int63n(sleepTime)))
```
derekcollison, line 1851: Reconnect Wait time should still hold same semantics. Do not attempt reconnect for some period of time. What I think we need to add is a Jitter option which has a sane default, I might say 2 secs or so that is added onto reconnectWait.

kozlovic: So random between [0,2sec] added to the base, ok got it. Will update.

kozlovic: Argg.. quite a bit of tests seem to expect no more than reconnectWait, so we may get some flappers pop up, will have to fix those along the way... unless you think that added jitter should be an option?

derekcollison: I don't think it should be an option, but should be low as default. For tests we can set to 0 which should do old behavior.

kozlovic: expect that some of those tests are in /test package so I cannot play that trick without exposing an option or function to set that value to zero..

derekcollison: Jitter was always meant to be a public option, just wanted a sane default. For the case we were dealing with today they would probably set it to 5-10 secs.. I think that is alot for default cases where you have live backup servers and want quick reconnect. Maybe 1 second is plenty in that case, so [0,1] is default.

kozlovic: Ok so we do make it an Option in that it is settable. Still, it means that all tests in /test that relied on no more than reconnectWait will have to be altered, but that's ok. Just that I may not find them all at once.

ripienaar:
In my ideal world I want this to be an exponential back off. Or at least the ability to pass in a delay function to achieve that. 

Ie. Jitter by default but make it possible to do expo back offs via user supplied sleeper. Pass the sleeper a reconnect counter. 

At large site I had nats using all cpu for several minutes on restarts due to combos of slow TLS handshakes, timeouts and synced reconnects.

nats.go
derekcollison: Yes they will need to set Jitter to 0.. ;)

derekcollison:
@ripienaar Yes we were discussing about the serves and clients interacting. This is mostly to avoid thundering herd but if org controls both ends they can configure there clients to flatten the curve of reconnect rates, especially for high-demanding TLS handshakes, like 4K RSA keys ;)

ripienaar:
Sure, I am saying I want more control than this pr provides. 

This high cpu and literal outage for minutes was even with the fastest ciphers.

derekcollison:
ok gotcha..

kozlovic:
@ripienaar @derekcollison I can certainly add a `CustomReconnectJitter()` option. Here is what I propose, let me know if it matches your thoughts:
- An CustomReconnectJitter option that is given the URL the library is going to try to connect to along with the number of reconnect attempts for this URL. The user will return the additional jitter.
- If such option is specified, the default jitter is ignored

Question: Should the value returned by the invocation of the user's custom reconnect jitter callback be used as-is (in addition to the reconnect wait), or should the library still use a random value between 0 and that value?

derekcollison:
I think jitter should be handled as we discussed. What sounds like something RI might need is a reconnect failed handler such that he can possibly change reconnect time.

What I think he wants is a simple option (or default behavior) that just does exponential backoff.

@ripienaar is this more for the client app's benefit or the servers?

ripienaar:
Yeah no URL - this is for me to use in a go client. If I know I am working on a system with >50k clients, I will do the extra work to use write a custom jitter.

Using the stateless backoff here https://blog.gopheracademy.com/advent-2014/backoff/

```go
CustomReconnectJitter(func(n int) time.Duration {
  return backoff.Default.Duration(n)
})
```

And that's it, if I specify this you ignore the default you are proposing and people with complex needs can implement their crazy however they like.

When you're doing TLS at scale with NATS you have to be VERY careful - that's fine it's normal its not a NATS problem - but if I am at scale this is a key knob needed.

derekcollison:
Yes definitely needed for large scale TLS (not even large TBH, user we are working with was dealing with 1200).

Do we have a callback that exists now that says I tried everyone and no one picked up?

Feels like this callback would be where we reset the reconnect time (Jitter should probably not be tweaked here IMO).

The on connect again we can reset it back to normal, etc.

ripienaar:
Yeah, this is why my example takes a `n` input - it's the amount of reconnects in this cycle.  So you dont need to remember anything other than reconnect count.  The Jitter provider gives you the right sleep based on reconnect count.  Very little on the client library side to track

ColinSullivan1:
IMO, this can be more general than jitter, maybe label it ReconnectAttemptFailedCB.  Could be very useful in debugging.  Couldn't attempts be tracked via closure?  Along with custom, we could provide a few predefined callbacks that users could choose - jitter only, backoff w/ jitter, etc...

nats.go
```diff
@@ -260,9 +267,27 @@ type Options struct {
...
+	// ReconnectJitterMax caps the maximum jitter.
+	ReconnectJitterMax time.Duration
+
+	// ReconnectJitterCB sets the handler that is invoked after the library went
+	// through the whole list of URLs and failed to reconnect. The number of
+	// attempts and current jitter is passed as arguments to this handler.
+	// It can return a jitter value based on those metrics.
+	// The library will take a random value between 0 and the returned value
+	// and add that to the ReconnectWait duration.
+	// Note that any jitter is capped with ReconnectJitterMax.
```
ripienaar, line 289: so with this design library users still cant use their own backoff logic, we can only influence the jitter which not what I was suggesting earlier.  It's not nothing and will help, just pointing out that's not what I was suggesting at all.

nats.go
```diff
@@ -260,9 +267,27 @@ type Options struct {
...
+	// ReconnectWait during a reconnect when no TLS is used.
+	// Note that any jitter is capped with ReconnectJitterMax.
+	ReconnectJitterNonTLS time.Duration
+
+	// ReconnectJitterTLS sets the upper bound for a random delay added to
+	// ReconnectWait during a reconnect when TLS is used.
+	// Note that any jitter is capped with ReconnectJitterMax.
+	ReconnectJitterTLS time.Duration
+
+	// ReconnectJitterMax caps the maximum jitter.
```
ripienaar, line 280: worth pointing out here that setting to 0 will disable the capping

nats.go
derekcollison: So you want to control both then yes? The most important is to spread out jitter. What the min wait is stops the attempts from happening quickly but changing that and not jitter (unless you already have randomness in how folks got disconnected) will not help smooth reconnect. So for the grpc example above what are the achieving besides the client just not trying as often?

ripienaar: It's a true backoff:

 * not as often
 * over time less often
 * capped at a max duration
 * with jitter always

derekcollison: Point I was trying to make was that jitter is more important then the delay. I understand we could have both.

derekcollison: In example you provided above can you influence jitter?

ripienaar: They have a config that lets you influence all the things, max, default, jitter etc

derekcollison: ok gotcha. We should do the same although for us tls and jitter should have more of an influence as we discussed.

ripienaar: I think your proposed jitter only approach is fine for most and a good default.

I just dont want to make it impossible to get actual full featured backoff if desired, but dont think you need to go and implement one tbh (I'd still want to use my own that my own system configures consistently for all the places I backoff)

derekcollison: Understood. And I think we all agree sane defaults is the high level goal but I also want your use case handled well. Maybe we have the following.

`SetJitter` or `SetJitterDelta`

but we also have a combined one for that callback.
`SetReconnectWaitAndJitter(rw, j)`

derekcollison: As long as they are reset when we do reconnect etc.

ripienaar: sound like a ever growing list of confusing options, what was wrong with 

```go
CustomReconnectDelay(func(n int) time.Duration {
  return backoff.Default.Duration(n)
})```

that simply trusts that duration for the delay?

Comment (derekcollison) [reply nats.go:289]: That is fair, use some library for wait and jitter combined. And just make sane defaults for most scenarios.

In your case is your algorithm TLS aware?

Comment (ripienaar) [reply nats.go:289]: no - but your defaults could be of course and would be fine for most.

choria cant be run without tls unless you go and recompile it at which point i dont care :)

Comment (kozlovic) [reply nats.go:289]: I vote for that simple function. If set, supersede any ReconnectWait option. So we keep the current jitter options (nonTLS and TLS) for cases where the custom option is not set. Doc for CustomReconnectDelay() will clearly state that the library will sleep for given delay and it is recommended that the function returns a value that include some jitter.

Comment (kozlovic) [reply nats.go:289]: I am ready to submit updates. Do you guys prefer that I squash first or want to see just the difference (and I will only squash prior to merge)?

Comment (derekcollison) [reply nats.go:289]: Squash for me.

File: nats.go [hunk 6]
```diff
@@ -258,6 +268,24 @@ type Options struct {
 	// to a server that we were already connected to previously.
 	ReconnectWait time.Duration
 
+	// CustomReconnectDelayCB is invoked after the library tried every
+	// URL in the server list and failed to reconnect. It passes to the
+	// user the current number of attempts. This function returns the
+	// amount of time the library will sleep before attempting to reconnect
+	// again. It is strongly recommended that this value contains some
+	// jitter to prevent all connections to attempt reconnecting at the same time.
```
ripienaar, line 276: sweet 👍

go.mod
```diff
@@ -1,5 +1,7 @@
 module github.com/nats-io/nats.go
 
+go 1.14
```
derekcollison, line 3: We want this here?

nats.go
```diff
@@ -45,19 +45,21 @@ import (
...
-	DefaultReconnectWait    = 2 * time.Second
-	DefaultTimeout          = 2 * time.Second
-	DefaultPingInterval     = 2 * time.Minute
-	DefaultMaxPingOut       = 2
-	DefaultMaxChanLen       = 8192            // 8k
-	DefaultReconnectBufSize = 8 * 1024 * 1024 // 8MB
-	RequestChanLen          = 8
-	DefaultDrainTimeout     = 30 * time.Second
-	LangString              = "go"
+	Version                      = "1.9.2"
```
derekcollison, line 48: Probably bump to 1.9.3

nats.go
```diff
@@ -45,19 +45,21 @@ import (
...
+	DefaultURL                   = "nats://127.0.0.1:4222"
+	DefaultPort                  = 4222
+	DefaultMaxReconnect          = 60
+	DefaultReconnectWait         = 2 * time.Second
+	DefaultReconnectJitterNonTLS = 100 * time.Millisecond
```
derekcollison, line 53: Just `DefaultReconnectJitter` imo

nats.go
```diff
@@ -127,15 +129,17 @@ func init() {
...
-		Timeout:          DefaultTimeout,
-		PingInterval:     DefaultPingInterval,
-		MaxPingsOut:      DefaultMaxPingOut,
-		SubChanLen:       DefaultMaxChanLen,
-		ReconnectBufSize: DefaultReconnectBufSize,
-		DrainTimeout:     DefaultDrainTimeout,
+		AllowReconnect:        true,
+		MaxReconnect:          DefaultMaxReconnect,
+		ReconnectWait:         DefaultReconnectWait,
+		ReconnectJitterNonTLS: DefaultReconnectJitterNonTLS,
```
derekcollison, line 135: Same here. `ReconnectJitter` and `ReconnectJitterTLS`

nats.go
aricart: Just going to say that this is uber complex - 2 jitter values, and a function that tells them that it failed when the close should do that. My question is if that is the case, why don't we just have one jitter that is a random value added to reconnect time wait, and also an exponential backoff which starts at reconnect time wait + jitter.

aricart: Also with the advent of host:port knowing the type of server is more difficult - and in most cases the client doesn't care, as the server will be whatever they have deployed and likely will only be either tls or not.

kozlovic: The callback is for users that want to implement their own backoff, in which case the lib will use whatever value is returned. For the default jitters, it is true that TLS connections have usually more issues reconnecting due to the cost of tls handshake, so this is why there are 2 different values. Note that in both cases (plain and TLS) the lib still takes a random value between 0 and that value that is added to the reconnect wait.

aricart: One more - I think the correct thing would be to have a 'maxReconnectWait' that is analogous to reconnectWait, but that is random between 0-maxReconnectWait. That would give a bigger jitter window without the complexity of the multipliers.

aricart: > The callback is for users that want to implement their own backoff, in which case the lib will use whatever value is returned. For the default jitters, it is true that TLS connections have usually more issues reconnecting due to the cost of tls handshake, so this is why there are 2 different values. Note that in both cases (plain and TLS) the lib still takes a random value between 0 and that value that is added to the reconnect wait.

So why not just say have `jitter` as a boolean and that randomizes reconnectTimeWait. The jitter callback can be a dynamic thing - in case of node the property can be a boolean  or a function, and it's done.

derekcollison: We had long discussions on all of these, base connect and jitter are separate things in almost all other systems. We tried to simplify and had quite a few back and forths.

kozlovic:
@derekcollison @ripienaar Please consider the changes we have done here. Since we now sleep only at the end of the list, it changes the fact that we use to check the lastAttempt from each server we are trying to reconnect to. We have changed that (and I have removed the lastAttempt variable that was no longer used in https://github.com/nats-io/nats.go/pull/564/commits/b2be4bbb43e54388f4fdbabd1f23617390bfff96). Let's make sure this is correct before others port to respective client libraries.

nats.go
```diff
@@ -1818,33 +1866,63 @@ func (nc *Conn) doReconnect(err error) {
...
+		if !doSleep {
+			i++
+			// Release the lock to give a chance to a concurrent nc.Close() to break the loop.
 			runtime.Gosched()
 		} else {
-			time.Sleep(time.Duration(sleepTime))
+			i = 0
+			var st time.Duration
+			if crd != nil {
+				wlf++
```
derekcollison, line 1905: Should we count wlf regardless of crd? Noop now but we may use it.

kozlovic: I think we will move it out of the condition when we actually need it, will be easy to do.

PR #566, opened 2020-05-06
[FIXED] Randomize discovered server URLs if allowed

### LINKED ISSUE #565: load balancing: add discovered server in a random position of the server list
## Feature Request
Saw this through code inspection.
The connect list is shuffled. 
New server are added to the end.
This makes server, the client learns about, less likely to be picked during a re-connect.

#### Use Case:
Any situation where we have 2 server running, then start a third, then kill one of the former server.
Clients are equally distributed between server (a,b)
We start server c.
Because server c gets added to the end of the list it will result in the client having one of these two lists:
a, b, c (current is a)
b, a, c (current is b)

if now we restart either a or b, all clients will connect to the other one, not c.

#### Proposed Change:
insert the new server at a random location in the list.

This is useful to reduce thundering heard and better load balancing.

kozlovic:
After adding discovered URLs from the INFO, reshuffle the pool
if option allows

Resolves #565

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -2431,8 +2431,15 @@ func (nc *Conn) processInfo(info string) error {
 		}
 		nc.addURLToPool(fmt.Sprintf("%s://%s", nc.connScheme(), curl), true, saveTLS)
 	}
-	if hasNew && !nc.initc && nc.Opts.DiscoveredServersCB != nil {
-		nc.ach.push(func() { nc.Opts.DiscoveredServersCB(nc) })
+	if hasNew {
+		// If randomization is allowed, randomize the pool now that we have
+		// added all new URLs.
+		if !nc.Opts.NoRandomize {
+			nc.shufflePool()
+		}
+		if !nc.initc && nc.Opts.DiscoveredServersCB != nil {
+			nc.ach.push(func() { nc.Opts.DiscoveredServersCB(nc) })
+		}
 	}
 
 	return nil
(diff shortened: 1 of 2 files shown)


nats.go
```diff
@@ -2431,8 +2431,15 @@ func (nc *Conn) processInfo(info string) error {
 		}
 		nc.addURLToPool(fmt.Sprintf("%s://%s", nc.connScheme(), curl), true, saveTLS)
 	}
-	if hasNew && !nc.initc && nc.Opts.DiscoveredServersCB != nil {
-		nc.ach.push(func() { nc.Opts.DiscoveredServersCB(nc) })
+	if hasNew {
+		// If randomization is allowed, randomize the pool now that we have
+		// added all new URLs.
+		if !nc.Opts.NoRandomize {
+			nc.shufflePool()
```
derekcollison, line 2438: Should we shuffle with the new one added?

kozlovic: That's what I do. Above we add all new ones (as long as they are new). The `hasNew` boolean will be true if at least 1 URL was added. If that is the case then we shuffle the whole pool.

matthiashanel, line 2438: What happens to the server we are currently connected to?
Is it possible that this reshuffle, on next disconnect, causes us to try connecting from the server just disconnected?

kozlovic: That means I guess that we should shuffle but only [1:] and leave at pos 0 the one we are currently connected to?

derekcollison: I think so.. We should never try to reconnect to the server we were connected to unless its the only one.

kozlovic: @derekcollison @matthiashanel Have another look please.

PR #569, opened 2020-05-13
Release v1.10.0

kozlovic:

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ When using or transitioning to Go modules support:
 ```bash
 # Go client latest or explicit version
 go get github.com/nats-io/nats.go/@latest
-go get github.com/nats-io/nats.go/@v1.9.2
+go get github.com/nats-io/nats.go/@v1.10.0
 
 # For latest NATS Server, add /v2 at the end
 go get github.com/nats-io/nats-server/v2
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -325,6 +325,17 @@ nc, err := nats.Connect(servers)
 // This example means 10 seconds total per backend.
 nc, err = nats.Connect(servers, nats.MaxReconnects(5), nats.ReconnectWait(2 * time.Second))
 
+// You can also add some jitter for the reconnection.
+// This call will add up to 100 milliseconds for non TLS connections and 1 second for TLS connections.
+nc, err = nats.Connect(servers, nats.ReconnectJitter(100 * time.Millisecond, time.Second))
```
derekcollison, line 330: That is the default though now yes? Maybe show different values and call out the defaults here that you get with nothing added.

kozlovic: You are right. I did change to:
```
// You can also add some jitter for the reconnection.
// This call will add up to 500 milliseconds for non TLS connections and 2 seconds for TLS connections.
// If not specified, the library defaults to 100 milliseconds and 1 second, respectively.
nc, err = nats.Connect(servers, nats.ReconnectJitter(500*time.Millisecond, 2*time.Second))
```

derekcollison: LGTM

PR #572, opened 2020-05-27
[ADDED] LameDuckModeHandler option/setter

kozlovic:
This will be invoked when the server enters LDM mode.

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -386,6 +386,12 @@ type Options struct {
 	// callbacks after Close() is called. Client won't receive notifications
 	// when Close is invoked by user code. Default is to invoke the callbacks.
 	NoCallbacksAfterClientClose bool
+
+	// LameDuckModeHandler sets the callback to invoke when the server notifies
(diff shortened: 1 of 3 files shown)


nats.go
```diff
@@ -2457,6 +2475,9 @@ func (nc *Conn) processInfo(info string) error {
 	// did not include themselves in the async INFO protocol.
 	// If empty, do not remove the implicit servers from the pool.
 	if len(ncInfo.ConnectURLs) == 0 {
```
derekcollison, line 2477: The connectURLs is 0 here? Shouldn't it be everyone but that server?

kozlovic: This check was there before. You have to think of the case say of single server or where client advertise has been disabled. There is still value/need to react to an LDM notification.

derekcollison: Do we do the LDM callback twice then?

kozlovic: No, this `if` used to simply `return nil`, now I just check for LDM and handler and invoke, but still `return nil` at line 2481.

PR #573, opened 2020-06-01
Jwtv2

matthiashanel:
as discussed

TestExpiredUserCredentialsRenewal is moved to the server.


go.mod
```diff
@@ -1,8 +1,11 @@
 module github.com/nats-io/nats.go
 
+go 1.14
+
 require (
-	github.com/nats-io/jwt v0.3.3-0.20200519195258-f2bf5ce574c7
-	github.com/nats-io/nats-server/v2 v2.1.8-0.20200524125952-51ebd92a9093
-	github.com/nats-io/nkeys v0.1.4
+	github.com/golang/protobuf v1.4.2
+	github.com/nats-io/nats-server/v2 v2.1.7
```
kozlovic, line 7: I have always debated that I did not want to have NATS Server dep in this go.mod because this is used for tests only. In other words, starting from an empty go.mod, I would have only what `GO111MODULE=on go install` requires. However, since we started to have references to the server, we need to have `master` here because the header test requires NATS Server that is option aware (we are accessing an option that is not in v2.1.7). So could you update this and regenerate the go.mod/sum?

matthiashanel: done

PR #575, opened 2020-06-03
Add go_test.mod to separate test dependencies from client

wallyqs:
This uses the `-modfile` feature that was added to [Go 1.14](https://golang.org/doc/go1.14) to use a different Go modules file for when running the tests.  This helps avoid having the main `go.mod`  include any references to the server and makes it more flexible to test against arbitrary versions in master from the server.


wallyqs:
Opening as draft since this currently means we'd stop running tests for Go 1.13 and still some weeks before Go 1.15 is released.

wallyqs:
Since Go 1.15 is out now, this is ok to add now to keep the dependencies on the client small and still being able to run the tests with latest version of the server using: `go test ./... -v -modfile=go_test.mod`

wallyqs:
Closing this one for now since it would end up in having to use a Makefile to run tests correctly for dev which needs some discussion, will revisit before we tag a client release.

PR #576, opened 2020-06-18
Support for no responders error on requests

derekcollison:
If connected to a new server and a request is sent when there are no responders the request will return as soon as a response from the server is received. The response will have an empty payload and a 503 status.


derekcollison:
Just to be clear it is not opt in at this time, its always on if the server supports it since we can use the header to distinguish between valid empty responses. Want to make sure that was clear in case we feel we still need to have a way to opt out.

ripienaar:
Sure it’s opt in as in it doesn’t affect existing code. We don’t ascribe special meaning to empty reply and the API behaves as always. 

Early on we said we will always do empty reply because we think that won’t wise issues. The way it is now is much safer and those with older clients can be sure they won’t be affected

PR #581, opened 2020-07-07
[ADDED] RetryOnFailedConnect option

### LINKED ISSUE #195: Should Connect return errors if it cannot connect?

kozlovic:
Normally, nats.Connect() would fail if there is no server available
when this call is executed. With this new option, if no connection
can be made, this call will return no error and will trigger code
similar to the reconnect code. Therefore, MaxReconnect and ReconnectWait
options are used as if the library had been disconnected and is trying
to reconnect.
Note that subscription and publish calls will also behave as if the
library was in reconnection mode, which means that the calls are
buffered and produce no error until the reconnect buffer size is
full.
Obviously, since the connection is not connected, Flush or Request/Reply
calls would timeout.

If the ReconnectHandler is set, it will be invoked if the library
connects asynchronously.

Unrelated: fixed a test that had a t.skip()...

Resolves #195

### DIFF
--- a/test/conn_test.go
+++ b/test/conn_test.go
@@ -1470,41 +1470,48 @@ func TestDefaultOptionsDialer(t *testing.T) {
 	}
 }
 
-func TestCustomFlusherTimeout(t *testing.T) {
-	t.Skip("broken test")
+type lowWriteBufferDialer struct{}
 
+func (d *lowWriteBufferDialer) Dial(network, address string) (net.Conn, error) {
+	c, err := net.Dial(network, address)
+	if err != nil {
+		return nil, err
+	}
+	c.(*net.TCPConn).SetWriteBuffer(100)
+	return c, nil
+}
+
+func TestCustomFlusherTimeout(t *testing.T) {
 	s := RunDefaultServer()
 	defer s.Shutdown()
 
-	opts := &nats.Options{
-		Servers: []string{nats.DefaultURL},
-
-		// Reasonably large flusher timeout will not induce errors
-		// when we can flush fast
-		FlusherTimeout: 10 * time.Second,
-	}
-	nc1, err := opts.Connect()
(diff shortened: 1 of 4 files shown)


test/conn_test.go
```diff
@@ -1470,90 +1470,183 @@ func TestDefaultOptionsDialer(t *testing.T) {
...
-	payload := ""
-	for i := 0; i < 8192; i++ {
-		payload += "A"
-	}
-	payloadBytes := []byte(payload)
+	doneCh := make(chan struct{}, 1)
+	// We want to have a payload size that is big enough so that after
+	// few publish, the socket buffer will be full and produce the timeout.
+	// Since we try to produce the error in the flusher and not the publish
+	// call itslef, use a size that is a bit less than the internal
```
wallyqs, line 1498:
```suggestion
	// call itself, use a size that is a bit less than the internal
```

kozlovic: Thanks, I fixed and push-force the changes.

PR #585, opened 2020-08-14
add JetStream ack helpers

ripienaar:

### DIFF
--- a/jetstream.go
+++ b/jetstream.go
@@ -1,0 +1,356 @@
+package nats
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"strconv"
+	"strings"
+	"time"
+)
+
+var (
+	AckAck      = []byte("+ACK")
+	AckNak      = []byte("-NAK")
+	AckProgress = []byte("+WPI")
+	AckNext     = []byte("+NXT")
+	AckTerm     = []byte("+TERM")
+)
+
+// JetStreamMsgMetaData is metadata related to a JetStream originated message
+type JetStreamMsgMetaData struct {
+	Stream      string
+	Consumer    string
+	Parsed      bool
+	Delivered   int
+	StreamSeq   int
+	ConsumerSeq int
+	TimeStamp   time.Time
+}
+
+func (m *Msg) JetStreamMetaData() (*JetStreamMsgMetaData, error) {
+	var err error
+
+	if m.jsMeta != nil && m.jsMeta.Parsed {
+		return m.jsMeta, nil
+	}
+
+	m.jsMeta, err = m.parseJSMsgMetadata()
+
+	return m.jsMeta, err
+}
+
+func (m *Msg) parseJSMsgMetadata() (*JetStreamMsgMetaData, error) {
+	if m.jsMeta != nil {
+		return m.jsMeta, nil
+	}
+
+	if len(m.Reply) == 0 {
+		return nil, ErrNotJSMessage
+	}
+
+	meta := &JetStreamMsgMetaData{}
+
+	tsa := [32]string{}
+	parts := tsa[:0]
+	start := 0
+	btsep := byte('.')
+	for i := 0; i < len(m.Reply); i++ {
+		if m.Reply[i] == btsep {
+			parts = append(parts, m.Reply[start:i])
+			start = i + 1
+		}
+	}
+	parts = append(parts, m.Reply[start:])
+
+	if len(parts) != 8 || parts[0] != "$JS" || parts[1] != "ACK" {
+		return nil, ErrNotJSMessage
+	}
+
+	var err error
+
+	meta.Stream = parts[2]
+	meta.Consumer = parts[3]
+	meta.Delivered, err = strconv.Atoi(parts[4])
+	if err != nil {
+		return nil, ErrNotJSMessage
+	}
+
+	meta.StreamSeq, err = strconv.Atoi(parts[5])
+	if err != nil {
+		return nil, ErrNotJSMessage
+	}
+
+	meta.ConsumerSeq, err = strconv.Atoi(parts[6])
+	if err != nil {
+		return nil, ErrNotJSMessage
+	}
+
+	tsi, err := strconv.Atoi(parts[7])
+	if err != nil {
+		return nil, ErrNotJSMessage
+	}
+	meta.TimeStamp = time.Unix(0, int64(tsi))
+
+	meta.Parsed = true
+
+	return meta, nil
+}
+
+const jsStreamUnspecified = "not.set"
+
+type jsAcKOpts struct {
+	str string // stream to expect a ack from
+}
+
+type jsOpts struct {
+	timeout time.Duration
+	ctx     context.Context
+
+	ack jsAcKOpts
+}
+
+func newJsOpts() *jsOpts {
+	return &jsOpts{ack: jsAcKOpts{str: jsStreamUnspecified}}
+}
+
+func (j *jsOpts) context(dftl time.Duration) (context.Context, context.CancelFunc) {
+	if j.ctx != nil {
+		return context.WithCancel(j.ctx)
+	}
+
+	if j.timeout == 0 {
+		j.timeout = dftl
+	}
+
+	return context.WithTimeout(context.Background(), j.timeout)
+}
+
+// AckOption configures the various JetStream message acknowledgement helpers
+type AckOption func(opts *jsOpts) error
+
+// PublishOption configures publishing messages
+type PublishOption func(opts *jsOpts) error
+
+// PublishExpectsStream waits for an ack after publishing and ensure it's from a specific stream, empty arguments waits for any valid acknowledgement
+func PublishExpectsStream(stream ...string) PublishOption {
+	return func(opts *jsOpts) error {
+		switch len(stream) {
+		case 0:
+			opts.ack.str = ""
+		case 1:
+			opts.ack.str = stream[0]
+			if !isValidJSName(opts.ack.str) {
+				return ErrInvalidStreamName
+			}
+		default:
+			return ErrMultiStreamUnsupported
+		}
+
+		return nil
+	}
+}
+
+// PublishStreamTimeout sets the period of time to wait for JetStream to acknowledge receipt, defaults to JetStreamTimeout option
+func PublishStreamTimeout(t time.Duration) PublishOption {
+	return func(opts *jsOpts) error {
+		opts.timeout = t
+		return nil
+	}
+}
+
+// PublishCtx sets an interrupt context for waiting on a stream to reply
+func PublishCtx(ctx context.Context) PublishOption {
+	return func(opts *jsOpts) error {
+		opts.ctx = ctx
+		return nil
+	}
+}
+
+// AckWaitDuration waits for confirmation from the JetStream server
+func AckWaitDuration(d time.Duration) AckOption {
+	return func(opts *jsOpts) error {
+		opts.timeout = d
+		return nil
+	}
+}
+
+func (m *Msg) jsAck(body []byte, opts ...AckOption) error {
+	if m.Reply == "" {
+		return ErrMsgNoReply
+	}
+
+	if m == nil || m.Sub == nil {
+		return ErrMsgNotBound
+	}
+
+	m.Sub.mu.Lock()
+	nc := m.Sub.conn
+	m.Sub.mu.Unlock()
+
+	var err error
+	var aopts *jsOpts
+
+	if len(opts) > 0 {
+		aopts = newJsOpts()
+		for _, f := range opts {
+			if err = f(aopts); err != nil {
+				return err
+			}
+		}
+	}
+
+	if aopts == nil || aopts.timeout == 0 {
+		return m.Respond(body)
+	}
+
+	_, err = nc.Request(m.Reply, body, aopts.timeout)
+
+	return err
+}
+
+// Ack acknowledges a JetStream messages received from a Consumer, indicating the message
+// should not be received again later
+func (m *Msg) Ack(opts ...AckOption) error {
+	return m.jsAck(AckAck, opts...)
+}
+
+// Nak acknowledges a JetStream message received from a Consumer, indicating that the message
+// is not completely processed and should be sent again later
+func (m *Msg) Nak(opts ...AckOption) error {
+	return m.jsAck(AckNak, opts...)
+}
+
+// AckProgress acknowledges a Jetstream message received from a Consumer, indicating that work is
+// ongoing and further processing time is required equal to the configured AckWait of the Consumer
+func (m *Msg) AckProgress(opts ...AckOption) error {
+	return m.jsAck(AckProgress, opts...)
+}
+
+// AckNext performs an Ack() and request that the next message be sent to subject ib
+func (m *Msg) AckNext(ib string) error {
+	return m.RespondMsg(&Msg{Subject: m.Reply, Reply: ib, Data: AckNext})
+}
+
+// AckAndFetch performs an AckNext() and returns the next message from the stream
+func (m *Msg) AckAndFetch(opts ...AckOption) (*Msg, error) {
+	if m.Reply == "" {
+		return nil, ErrMsgNoReply
+	}
+
+	if m == nil || m.Sub == nil {
+		return nil, ErrMsgNotBound
+	}
+
+	m.Sub.mu.Lock()
+	nc := m.Sub.conn
+	m.Sub.mu.Unlock()
+
+	var err error
+
+	aopts := newJsOpts()
+	for _, f := range opts {
+		if err = f(aopts); err != nil {
+			return nil, err
+		}
+	}
+
+	ctx, cancel := aopts.context(nc.Opts.JetStreamTimeout)
+	defer cancel()
+
+	sub, err := nc.SubscribeSync(NewInbox())
+	if err != nil {
+		return nil, err
+	}
+	sub.AutoUnsubscribe(1)
+	defer sub.Unsubscribe()
+
+	err = m.RespondMsg(&Msg{Reply: sub.Subject, Data: AckNext, Subject: m.Reply})
+	if err != nil {
+		return nil, err
+	}
+	nc.Flush()
+
+	return sub.NextMsgWithContext(ctx)
+}
+
+// AckTerm acknowledges a message received from JetStream indicating the message will not be processed
+// and should not be sent to another consumer
+func (m *Msg) AckTerm(opts ...AckOption) error {
+	return m.jsAck(AckTerm, opts...)
+}
+
+// JetStreamPublishAck metadata received from JetStream when publishing messages
+type JetStreamPublishAck struct {
+	Stream   string `json:"stream"`
+	Sequence int    `json:"seq"`
+}
+
+// ParsePublishAck parses the publish acknowledgement sent by JetStream
+func ParsePublishAck(m []byte) (*JetStreamPublishAck, error) {
+	if bytes.HasPrefix([]byte("-ERR"), m) {
+		if len(m) > 7 {
+			return nil, fmt.Errorf(string(m[6 : len(m)-1]))
+		}
+
+		return nil, fmt.Errorf(string(m))
+	}
+
+	if !bytes.HasPrefix(m, []byte("+OK {")) {
+		return nil, fmt.Errorf("invalid JetStream Ack: %v", string(m))
+	}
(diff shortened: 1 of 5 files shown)


ripienaar:
For discussion, once we agree on a direction I'll add tests etc.

nats.go
```diff
@@ -3790,6 +3790,89 @@ func (m *Msg) RespondMsg(msg *Msg) error {
...
+	return m.Respond(AckAck)
+}
+
+// AckConfirm performs an Ack() and requests a confirmation from JetStream that the ack was received
+func (m *Msg) AckConfirm(nc *Conn, timeout time.Duration) (*Msg, error) {
+	if !m.isJSMsg() {
+		return nil, fmt.Errorf("not a JetStream originated message")
+	}
+
+	if nc == nil {
```
ripienaar, line 3830: I should probably do the same m.sub.conn thing here as Respond()

nats.go
```diff
@@ -3790,6 +3790,89 @@ func (m *Msg) RespondMsg(msg *Msg) error {
...
+	AckNak      = []byte("-NAK")
+	AckProgress = []byte("+WPI")
+	AckNext     = []byte("+NXT")
+	AckTerm     = []byte("+TERM")
+)
+
+// Ack acknowledges a JetStream messages received from a Consumer, indicating the message
+// should not be received again later
+func (m *Msg) Ack() error {
+	if !m.isJSMsg() {
```
derekcollison, line 3817: Not sure we care, if app thinks its acking we should just send reply unless there is no reply subject. Ack subject may change etc and this could be brittle.

nats.go
```diff
@@ -3790,6 +3790,89 @@ func (m *Msg) RespondMsg(msg *Msg) error {
...
+		return fmt.Errorf("not a JetStream originated message")
+	}
+
+	return m.Respond(AckAck)
+}
+
+// AckConfirm performs an Ack() and requests a confirmation from JetStream that the ack was received
+func (m *Msg) AckConfirm(nc *Conn, timeout time.Duration) (*Msg, error) {
```
derekcollison, line 3825: I think this just waits for the ack to the ack. Does not need to return it etc. This may be a variable option to plain Ack()?

nats.go
```diff
@@ -3790,6 +3790,89 @@ func (m *Msg) RespondMsg(msg *Msg) error {
...
+	if nc == nil {
+		return nil, fmt.Errorf("invalid connection supplied")
+	}
+
+	return nc.Request(m.Reply, AckAck, timeout)
+}
+
+// AckNak acknowledges a JetStream message received from a Consumer, indicating that the message
+// is not completely processed and should be sent again later
+func (m *Msg) AckNak() error {
```
derekcollison, line 3839: Just Nak() I think. Also same variable option to wait on ack of ack.

nats.go
ripienaar: Yeah today its a nil message but I think we might later want to include some additional metadata perhaps? Even just a stream name/id for correlation so I didnt want to force a API change if we decided to do that

nats.go
ripienaar: I made em all Ack* so they are easy to find in IDEs and obviously related, but sure msg has few methods so we could do that.

nats.go
ripienaar: so would you suggest a message signature of `func (m *Msg) Ack(confirm bool, timeout time.Duration) error` or something different?

derekcollison: `Ack(opts ...AckOption)`

ripienaar: Updated with options, AckNext is different and doesnt really fit that pattern by its nature, so not sure I am a fan of this but for sure it will work and better than adding Ack() and AckConfirm() or something

nats.go
ripienaar: @derekcollison how do these look? if good I'll add tests etc

derekcollison:
Will take a look tomorrow.

jetstream_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+	err = nc.Publish("js.in.test", []byte("hello"), ExpectAckFromStream("OTHER"))
+	if err == nil {
+		t.Fatalf("expected an error but got none")
+	}
+}
+
+func TestMsg_ParseJSMsgMetadata(t *testing.T) {
+	msg := &Msg{Reply: "$JS.ACK.ORDERS.NEW.1.2.3.1587466354254920000"}
+	err := msg.ParseJSMsgMetadata()
```
ripienaar, line 58: @derekcollison following up from our call, we thought parsing these into headers on demand in the client would be nice - it's ok, I guess but things like `JS-Delivered` being a string is very annoying.

Do you think we should do this anyway and then add some accessors on `msg` to fetch and convert as a helper?

derekcollison: ok let's add helpers then instead.

ripienaar: OK, done that, seems better.

Next to see what we can do for consumers

ripienaar:
updated per last nights call, no consumer stuff yet

derekcollison:
Do we have a test or example that it would be easy to quickly look at to get a sense of the API for publishing, acking, receiving etc?

ripienaar:
I didnt do receiving yet - I like small PRs - so this is ack related.

All tests here https://github.com/nats-io/nats.go/pull/585/files#diff-5ad60403adccb2da7f1010ef809aaa9a

```
// publish just wait for any  ack
err := nc.Publish("js.in.test", []byte("hello"), PublishExpectsStream("TEST"), PublishStreamTimeout(time.Second))

// publish wait for a specific stream - expects only one ack for now
err = nc.Publish("js.in.test", []byte("hello"), PublishExpectsStream("OTHER"), PublishStreamTimeout(time.Second))
```

metadata:

```go
msg := &Msg{Reply: "$JS.ACK.ORDERS.NEW.1.2.3.1587466354254920000"}
meta, err := msg.JetStreamMetaData()

// meta.Stream() etc
```

Acks, arguments optional, without means fire and forget:

```go
err = msg.Ack(AckWaitDuration(time.Second))
err = msg.Nak(AckWaitDuration(time.Second)) 
err = msg.AckTerm(AckWaitDuration(time.Second))
err = msg.AckProgress(AckWaitDuration(time.Second))
err = msg.AckNext(sub.Subject)
```

derekcollison:
ok looks pretty good. Can we have default timeout for publish of 2 seconds but allow override. Also need a Context so they can cancel with that etc.

ripienaar:
It defaults to JetStreamTimeout which defaults to 2 seconds and is set table when creating the client. But only if one then specified that a specific stream is requested, guess I can make it expect ack from anything if someone set PublishExpectsStream(“”) which right now would imply don’t wait

derekcollison:
Make the stream name a vararg and then you can have none, 1 or more..

ripienaar:
> Make the stream name a vararg and then you can have none, 1 or more..

I had considered that but decided not to since this would imply we support publishing to streams - we dont.

We spoke about this in the past and support more than 1 would put users in a position that they get an error they have no way to recover from, so best avoided and left to replication features to resolve multiple synced streams.

ripienaar:
OK, made it a vararg but it fails for > 1, also added `PublishCtx(ctx)`

jetstream.go
```diff
@@ -0,0 +1,334 @@
...
+	AckAck      = []byte("+ACK")
+	AckNak      = []byte("-NAK")
+	AckProgress = []byte("+WPI")
+	AckNext     = []byte("+NXT")
+	AckTerm     = []byte("+TERM")
+)
+
+// JetStreamMsgMetaData is metadata related to a JetStream originated message
+type JetStreamMsgMetaData struct {
+	stream    string
```
derekcollison, line 23: Why not just publicly accessible struct properties? I think that may be the preferred way here.

jetstream.go
```diff
@@ -0,0 +1,334 @@
...
+		return m.jsMeta, nil
+	}
+
+	if len(m.Reply) == 0 {
+		return nil, ErrNotJSMessage
+	}
+
+	meta := &JetStreamMsgMetaData{}
+
+	parts := strings.Split(m.Reply, ".")
```
derekcollison, line 83: Could handroll this one as well for performance. We have this for the proto parser and subjects I believe.

jetstream.go
```diff
@@ -0,0 +1,334 @@
...
+// PublishOption configures publishing messages
+type PublishOption func(opts *jsOpts) error
+
+// PublishExpectsStream waits for an ack after publishing and ensure it's from a specific stream, empty arguments waits for any valid acknowledgement
+func PublishExpectsStream(stream ...string) PublishOption {
+	return func(opts *jsOpts) error {
+		switch len(stream) {
+		case 0:
+			opts.ack.str = ""
+
```
derekcollison, line 160: Remove line (nit)

jetstream.go
```diff
@@ -0,0 +1,334 @@
...
+		case 1:
+			opts.ack.str = stream[0]
+			if !isValidJSName(opts.ack.str) {
+				return ErrInvalidStreamName
+			}
+
+		default:
+			return ErrInvalidStreamName
```
derekcollison, line 168: Would say not supported atm..

jetstream.go
```diff
@@ -0,0 +1,334 @@
...
+}
+
+// AckProgress acknowledges a Jetstream message received from a Consumer, indicating that work is
+// ongoing and further processing time is required equal to the configured AckWait of the Consumer
+func (m *Msg) AckProgress(opts ...AckOption) error {
+	return m.jsAck(AckProgress, opts...)
+}
+
+// AckNext performs an Ack() and request that the next message be sent to subject ib
+func (m *Msg) AckNext(ib string) error {
```
derekcollison, line 250: I think we need option here to for batch size as well. Also need to figure out overload for ack the ack and the next message if present.

jetstream.go
```diff
@@ -0,0 +1,334 @@
...
+		return ErrMsgNotBound
+	}
+
+	m.Sub.mu.Lock()
+	nc := m.Sub.conn
+	m.Sub.mu.Unlock()
+
+	var err error
+
+	aopts := newJsOpts()
```
derekcollison, line 214: We do not want to do this unless we know we have options below, otherwise just creating garbage.

jetstream.go
```diff
@@ -0,0 +1,334 @@
...
+	for _, f := range opts {
+		err = f(aopts)
```
derekcollison, line 216: Can make one line.

jetstream.go
```diff
@@ -0,0 +1,334 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	if aopts.timeout == 0 {
```
derekcollison, line 222: Wondering if this low level function should explicitly take an optional reply subject etc for ack of ack.

jetstream.go
```diff
@@ -0,0 +1,334 @@
...
+}
+
+// JetStreamPublishAck metadata received from JetStream when publishing messages
+type JetStreamPublishAck struct {
+	Stream   string `json:"stream"`
+	Sequence int    `json:"seq"`
+}
+
+// ParsePublishAck parses the publish acknowledgement sent by JetStream
+func ParsePublishAck(m []byte) (*JetStreamPublishAck, error) {
```
derekcollison, line 267: Why does this need to be public?

jetstream.go
```diff
@@ -0,0 +1,334 @@
...
+		return nil, fmt.Errorf("invalid JetStream Ack: %v", string(m))
+	}
+
+	ack := &JetStreamPublishAck{}
+	err := json.Unmarshal(m[3:], ack)
+	return ack, err
+}
+
+func isValidJSName(n string) bool {
+	if n == "" {
```
derekcollison, line 286: maybe just make one return line with both conditionals.

jetstream.go
```diff
@@ -0,0 +1,334 @@
...
+		return false
+	}
+
+	return !strings.ContainsAny(n, ">*. ")
+}
+
+func (nc *Conn) jsPublish(subj string, data []byte, opts []PublishOption) error {
+	aopts := newJsOpts()
```
derekcollison, line 294: Same here, only create if needed.

derekcollison: Probably can just be on the stack also if you want.

jetstream.go
```diff
@@ -0,0 +1,334 @@
...
+	var err error
+
+	for _, f := range opts {
+		err = f(aopts)
+		if err != nil {
+			return err
+		}
+	}
+
+	if aopts.timeout == 0 && aopts.ctx == nil && aopts.ack.str == jsStreamUnspecified {
```
derekcollison, line 304: `if aopts == nil || ...`

jetstream.go
```diff
@@ -0,0 +1,334 @@
...
+		return nc.publish(subj, _EMPTY_, nil, data)
+	}
+
+	ctx, cancel := aopts.context(nc.Opts.JetStreamTimeout)
```
derekcollison, line 308: So you are overloading context for timeout as well? If so should we use aopts.timeout if it was set?

jetstream.go
```diff
@@ -0,0 +1,334 @@
...
+	defer cancel()
+
+	resp, err := nc.RequestWithContext(ctx, subj, data)
+	if err != nil {
+		return err
+	}
+
+	ack, err := ParsePublishAck(resp.Data)
```
derekcollison, line 316: We should only do this if the pubAck option included a stream name, otherwise just check prefix for +OK or -ERR IMO.

nats.go
```diff
@@ -2723,7 +2741,11 @@ func (nc *Conn) kickFlusher() {
 // Publish publishes the data argument to the given subject. The data
 // argument is left untouched and needs to be correctly interpreted on
 // the receiver.
-func (nc *Conn) Publish(subj string, data []byte) error {
+func (nc *Conn) Publish(subj string, data []byte, opts ...PublishOption) error {
+	if opts != nil {
```
derekcollison, line 2745: `if len(opts) > 0`

derekcollison (changes requested):
Added in some comments, looks pretty good. Might want to have @kozlovic and @wallyqs review as well IMO.

jetstream.go
ripienaar: Because idiomatic go is to return concrete types with accessors so people can easily test etc, maybe we make `Stream` and `GetStream()`, I'd like to do the same on Msg tbh

jetstream.go
ripienaar: Yeah I thought of that, but we dont actually process batch size on AckNext

```go
case bytes.Equal(msg, AckNext):
		o.ackMsg(sseq, dseq, dcount)
		o.processNextMsgReq(nil, nil, subject, reply, nil)
```

while the signature for `processNextMsgReq` is `func (o *Consumer) processNextMsgReq(_ *subscription, _ *client, _, reply string, msg []byte) {` so we never send the message on, no pulling out of payload size.  This may well be a bug in JetStream but that's why I didnt add it here since its not supportd.

jetstream.go
ripienaar: not needed really since we are on m and have access to m.Reply as below? This does ack of ack if reply is set and options indicate we should wait for any reason

jetstream.go
ripienaar: so people can get access to metadata about the publish if they feel they need it, the would do a `Request()` and parse the response.

jetstream.go
ripienaar: will do, but this reads a lot better than combining it - compiler combines it anyway

jetstream.go
ripienaar: this passes the default timeout set using `nats.Connect(url, nats.JetStreamTimeout(5*time.Second)`, so that if someone sets no timeout and no context it would use the default timeout set for all API interactions. The context will be either based on the supplied ctx or on the supplied timeout or on default timeout as set above

jetstream.go
ripienaar: The first thing ParsePublishAck does is checks for -ERR and if its there pulls out the error message and returns that as `err` here, this way when someone does nc.Publish() the err they get back holds the jetstream supplied error message.

jetstream.go
ripienaar: hmm, cant find that example, found some in `processMsgArgs` but basing on those performed worse than string.Split, mind pointing at the one you had in mind?

jetstream.go
derekcollison: I would just make the structs members public and not have any accessors etc.

jetstream.go
derekcollison: Can't find it for reference, but here it is..

```go
	// Tokenize the subject. This should always be a literal subject.
	tsa := [32]string{}
	tts := tsa[:0]
	start := 0
	for i := 0; i < len(subject); i++ {
		if subject[i] == btsep {
			tts = append(tts, subject[start:i])
			start = i + 1
		}
	}
	tts = append(tts, subject[start:])
```

jetstream.go
derekcollison: Ah might be a not implemented feature..

derekcollison: What is `ib string` for here?

jetstream.go
derekcollison: Right but if I just care that it said OK I am parsing and creating alot of stuff for nothing.

jetstream.go
ripienaar: where to send the data, `subject` be better name

derekcollison: But you return a msg from the call itself no?

derekcollison: from `AckNext`

jetstream.go
ripienaar: OK , will do. Curious though what we're optimising this so heavily for, this is after a `Request()` that did network round trips, flush  etc, seems like a distraction, this isnt high performance code paths

jetstream.go
ripienaar: not following at all, lets zoom later

jetstream.go
derekcollison: Was suggestion, you don't have to put it in.

jetstream.go
derekcollison: Ignore that, too early in am for me. Apologies.

derekcollison: Should ackNext return the next message or an err if not available to match `Request`?

ripienaar: I dont think so, I believe typical case would be to feed a buffer or existing subscription.

derekcollison: I would have thought main use case was workqueue and you just pull next work item, process, ack and get next one and repeat.

ripienaar:
```go
nc.Subscribe(ib, func(m *nats.Msg) {
   // do stuff
   m.AckNext(ib)
})
```

more typically you might have a worker pull consuming messages off a queue, and the AckNext() would request the next message into the queue.

I'd rather add an additional function here if you want it to be more like a request for the next message

derekcollison: I think you are right, what would that look like? For the `Request()` like `AckNext()`

ripienaar: `msg, err := m.AckFetchNext(opts...)` or `s/Fetch/Request/`

derekcollison: `AckAndFetch()`?

ripienaar: k, will implement and we can see how it looks

ripienaar: Added `AckAndFetch()` https://github.com/nats-io/nats.go/pull/585/files#diff-5ad60403adccb2da7f1010ef809aaa9aR278

jetstream.go
```diff
@@ -0,0 +1,332 @@
...
+
+func (m *JetStreamMsgMetaData) StreamSequence() int {
+	return m.sSeq
+}
+
+func (m *JetStreamMsgMetaData) ConsumerSequence() int {
+	return m.cSeq
+}
+
+func (m *JetStreamMsgMetaData) PublishTime() time.Time {
```
derekcollison, line 56: Still think we remove these and just make the meta struct a data struct and make them public.

jetstream.go
derekcollison: ok what do you think?

ripienaar: seems fine

ripienaar:
@derekcollison reminder about this one

derekcollison (approved):
LGTM - Let's merge and get some working feedback. Thanks for your patience.

ripienaar:
Thanks, I am 1/2 through the next PR to bring consumer creation stuff and with this merged @ColinSullivan1 can start planning out how the java stuff will look.

Also ping @aricart

PR #587, opened 2020-08-17
[FIXED] RetryOnFailedConnect failing if getting protocol error

kozlovic:
If the library was able to create a TCP connection to a server but
failed for any reason (TLS timeout, etc..) RetryOnFailedConnect
would fail.


cbrake:
tested here -- thanks!

PR #596, opened 2020-10-06
Allow the Subscribe functions to access JetStream consumers

ripienaar:

### DIFF
--- a/jetstream.go
+++ b/jetstream.go
@@ -1,3 +1,16 @@
+// Copyright 2020 The NATS Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package nats
 
 import (
@@ -6,16 +19,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"strconv"
-	"strings"
 	"time"
-)
-
-var (
-	AckAck      = []byte("+ACK")
-	AckNak      = []byte("-NAK")
-	AckProgress = []byte("+WPI")
-	AckNext     = []byte("+NXT")
-	AckTerm     = []byte("+TERM")
 )
 
 // JetStreamMsgMetaData is metadata related to a JetStream originated message
@@ -100,7 +104,7 @@ func (m *Msg) parseJSMsgMetadata() (*JetStreamMsgMetaData, error) {
 
 const jsStreamUnspecified = "not.set"
 
-type jsAcKOpts struct {
+type jsAckOpts struct {
 	str string // stream to expect a ack from
 }
 
@@ -108,11 +112,13 @@ type jsOpts struct {
 	timeout time.Duration
 	ctx     context.Context
 
-	ack jsAcKOpts
+	ack        jsAckOpts
+	consumer   *ConsumerConfig
+	streamName string
 }
 
 func newJsOpts() *jsOpts {
-	return &jsOpts{ack: jsAcKOpts{str: jsStreamUnspecified}}
+	return &jsOpts{ack: jsAckOpts{str: jsStreamUnspecified}}
 }
 
 func (j *jsOpts) context(dftl time.Duration) (context.Context, context.CancelFunc) {
@@ -132,6 +138,18 @@ type AckOption func(opts *jsOpts) error
 
 // PublishOption configures publishing messages
 type PublishOption func(opts *jsOpts) error
+
+// ConsumerOption configures JetStream consumer behavior
+type ConsumerOption func(opts *jsOpts) error
+
+// Consumer creates a JetStream Consumer on a Stream
+func Consumer(stream string, cfg *ConsumerConfig) ConsumerOption {
+	return func(jopts *jsOpts) error {
+		jopts.consumer = cfg
+		jopts.streamName = stream
+		return nil
+	}
+}
 
 // PublishExpectsStream waits for an ack after publishing and ensure it's from a specific stream, empty arguments waits for any valid acknowledgement
 func PublishExpectsStream(stream ...string) PublishOption {
@@ -304,10 +322,6 @@ func ParsePublishAck(m []byte) (*JetStreamPublishAck, error) {
 	ack := &JetStreamPublishAck{}
 	err := json.Unmarshal(m[3:], ack)
 	return ack, err
-}
-
-func isValidJSName(n string) bool {
-	return !(n == "" || strings.ContainsAny(n, ">*. "))
 }
 
 func (nc *Conn) jsPublish(subj string, data []byte, opts []PublishOption) error {
--- a/jetstream.go
+++ b/jetstream.go
@@ -114,7 +114,7 @@ type jsOpts struct {
 }
 
 func newJsOpts() *jsOpts {
-	return &jsOpts{ackstr: jsStreamUnspecified}
+	return &jsOpts{ackstr: jsStreamUnspecified, consumer: &ConsumerConfig{}}
 }
 
 func (j *jsOpts) context(dftl time.Duration) (context.Context, context.CancelFunc) {
--- a/jetstream.go
+++ b/jetstream.go
@@ -104,21 +104,17 @@ func (m *Msg) parseJSMsgMetadata() (*JetStreamMsgMetaData, error) {
 
 const jsStreamUnspecified = "not.set"
 
-type jsAckOpts struct {
-	str string // stream to expect a ack from
-}
-
 type jsOpts struct {
 	timeout time.Duration
 	ctx     context.Context
 
-	ack        jsAckOpts
+	ackstr     string
 	consumer   *ConsumerConfig
 	streamName string
 }
 
 func newJsOpts() *jsOpts {
-	return &jsOpts{ack: jsAckOpts{str: jsStreamUnspecified}}
+	return &jsOpts{ackstr: jsStreamUnspecified}
 }
 
 func (j *jsOpts) context(dftl time.Duration) (context.Context, context.CancelFunc) {
@@ -139,13 +135,13 @@ type AckOption func(opts *jsOpts) error
 // PublishOption configures publishing messages
 type PublishOption func(opts *jsOpts) error
 
-// ConsumerOption configures JetStream consumer behavior
-type ConsumerOption func(opts *jsOpts) error
+// SubscribeOption configures JetStream consumer behavior
+type SubscribeOption func(opts *jsOpts) error
 
 // Consumer creates a JetStream Consumer on a Stream
-func Consumer(stream string, cfg *ConsumerConfig) ConsumerOption {
+func Consumer(stream string, cfg ConsumerConfig) SubscribeOption {
 	return func(jopts *jsOpts) error {
-		jopts.consumer = cfg
+		jopts.consumer = &cfg
 		jopts.streamName = stream
 		return nil
 	}
@@ -156,10 +152,10 @@ func PublishExpectsStream(stream ...string) PublishOption {
 	return func(opts *jsOpts) error {
 		switch len(stream) {
 		case 0:
-			opts.ack.str = ""
+			opts.ackstr = ""
 		case 1:
-			opts.ack.str = stream[0]
-			if !isValidJSName(opts.ack.str) {
+			opts.ackstr = stream[0]
+			if !isValidJSName(opts.ackstr) {
 				return ErrInvalidStreamName
 			}
 		default:
@@ -337,7 +333,7 @@ func (nc *Conn) jsPublish(subj string, data []byte, opts []PublishOption) error 
 		}
 	}
 
-	if aopts == nil || aopts.timeout == 0 && aopts.ctx == nil && aopts.ack.str == jsStreamUnspecified {
+	if aopts == nil || aopts.timeout == 0 && aopts.ctx == nil && aopts.ackstr == jsStreamUnspecified {
 		return nc.publish(subj, _EMPTY_, nil, data)
 	}
 
@@ -358,11 +354,11 @@ func (nc *Conn) jsPublish(subj string, data []byte, opts []PublishOption) error 
 		return ErrInvalidJSAck
 	}
 
-	if aopts.ack.str == jsStreamUnspecified || aopts.ack.str == "" {
+	if aopts.ackstr == jsStreamUnspecified || aopts.ackstr == "" {
 		return nil
 	}
 
-	if ack.Stream == aopts.ack.str {
+	if ack.Stream == aopts.ackstr {
 		return nil
 	}
 
--- a/jetstream_consumer.go
+++ b/jetstream_consumer.go
@@ -1,0 +1,333 @@
+// Copyright 2020 The NATS Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package nats
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+	"time"
+)
+
+func (nc *Conn) createOrUpdateConsumer(opts *jsOpts) (*ConsumerInfo, error) {
+	if opts.streamName == "" {
+		return nil, ErrStreamNameRequired
+	}
+	if opts.consumer == nil {
+		return nil, ErrConsumerConfigRequired
+	}
+
+	crj, err := json.Marshal(&JSApiConsumerCreateRequest{
+		Stream: opts.streamName,
+		Config: *opts.consumer,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	ctx, cancel := opts.context(nc.Opts.JetStreamTimeout)
+	defer cancel()
+
+	var subj string
+	switch len(opts.consumer.Durable) {
+	case 0:
+		subj = fmt.Sprintf(JSApiConsumerCreateT, opts.streamName)
+	default:
+		subj = fmt.Sprintf(JSApiDurableCreateT, opts.streamName, opts.consumer.Durable)
+	}
+
+	resp, err := nc.RequestWithContext(ctx, subj, crj)
+	if err != nil {
+		return nil, err
+	}
+
+	cresp := &JSApiConsumerCreateResponse{}
+	err = json.Unmarshal(resp.Data, cresp)
+	if err != nil {
+		return nil, err
+	}
+
+	if cresp.IsError() {
+		return nil, cresp.ToError()
+	}
+
+	return cresp.ConsumerInfo, nil
+}
+
+const (
+	JSApiConsumerCreateT = "$JS.API.CONSUMER.CREATE.%s"
+	JSApiDurableCreateT  = "$JS.API.CONSUMER.DURABLE.CREATE.%s.%s"
+)
+
+type ApiError struct {
+	Code        int    `json:"code"`
+	Description string `json:"description,omitempty"`
+}
+
+// Error implements error
+func (e ApiError) Error() string {
+	switch {
+	case e.Description == "" && e.Code == 0:
+		return "unknown JetStream Error"
+	case e.Description == "" && e.Code > 0:
+		return fmt.Sprintf("unknown JetStream %d Error", e.Code)
+	default:
+		return e.Description
+	}
+}
+
+// NotFoundError is true when the error is one about a resource not found
+func (e ApiError) NotFoundError() bool { return e.Code == 404 }
+
+// ServerError is true when the server returns a 5xx error code
+func (e ApiError) ServerError() bool { return e.Code >= 500 && e.Code < 600 }
+
+// UserError is true when the server returns a 4xx error code
+func (e ApiError) UserError() bool { return e.Code >= 400 && e.Code < 500 }
+
+// ErrorCode is the JetStream error code
+func (e ApiError) ErrorCode() int { return e.Code }
+
+type JSApiResponse struct {
+	Type  string    `json:"type"`
+	Error *ApiError `json:"error,omitempty"`
+}
+
+// ToError extracts a standard error from a JetStream response
+func (r JSApiResponse) ToError() error {
+	if r.Error == nil {
+		return nil
+	}
+
+	return *r.Error
+}
+
+// IsError determines if a standard JetStream API response is a error
+func (r JSApiResponse) IsError() bool {
+	return r.Error != nil
+}
+
+// io.nats.jetstream.api.v1.consumer_create_request
+type JSApiConsumerCreateRequest struct {
+	Stream string         `json:"stream_name"`
+	Config ConsumerConfig `json:"config"`
+}
+
+// io.nats.jetstream.api.v1.consumer_create_response
+type JSApiConsumerCreateResponse struct {
+	JSApiResponse
+	*ConsumerInfo
+}
+
+type AckPolicy int
+
+const (
+	AckNone AckPolicy = iota
+	AckAll
+	AckExplicit
+)
+
+func (p AckPolicy) String() string {
+	switch p {
+	case AckNone:
+		return "None"
+	case AckAll:
+		return "All"
+	case AckExplicit:
+		return "Explicit"
+	default:
+		return "Unknown Acknowledgement Policy"
+	}
+}
+
+func (p *AckPolicy) UnmarshalJSON(data []byte) error {
+	switch string(data) {
+	case jsonString("none"):
+		*p = AckNone
+	case jsonString("all"):
+		*p = AckAll
+	case jsonString("explicit"):
+		*p = AckExplicit
+	default:
+		return fmt.Errorf("can not unmarshal %q", data)
+	}
+
+	return nil
+}
+
+func (p AckPolicy) MarshalJSON() ([]byte, error) {
+	switch p {
+	case AckNone:
+		return json.Marshal("none")
+	case AckAll:
+		return json.Marshal("all")
+	case AckExplicit:
+		return json.Marshal("explicit")
+	default:
+		return nil, fmt.Errorf("unknown acknowlegement policy %v", p)
+	}
+}
+
+type ReplayPolicy int
+
+const (
+	ReplayInstant ReplayPolicy = iota
+	ReplayOriginal
+)
+
+func (p ReplayPolicy) String() string {
+	switch p {
+	case ReplayInstant:
+		return "Instant"
+	case ReplayOriginal:
+		return "Original"
+	default:
+		return "Unknown Replay Policy"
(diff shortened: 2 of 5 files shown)


derekcollison:
I will get to this today. Thanks.

derekcollison:
You have a spelling mistake you want to fix.

`./jetstream.go:142:51: "behaviour" is a misspelling of "behavior"`

enc.go
```diff
@@ -182,7 +182,7 @@ func (c *EncodedConn) QueueSubscribe(subject, queue string, cb Handler) (*Subscr
 }
 
 // Internal implementation that all public functions will use.
-func (c *EncodedConn) subscribe(subject, queue string, cb Handler) (*Subscription, error) {
+func (c *EncodedConn) subscribe(subject, queue string, cb Handler, opts ...SubscribeOption) (*Subscription, error) {
```
derekcollison, line 185: For now I am ok not upgraded EncodedConn. We need a better direction for it anyway.

jetstream.go
```diff
@@ -1,3 +1,16 @@
+// Copyright 2012-2020 The NATS Authors
```
derekcollison, line 1: This is new file so just 2020

jetstream.go
```diff
@@ -133,6 +139,31 @@ type AckOption func(opts *jsOpts) error
 // PublishOption configures publishing messages
 type PublishOption func(opts *jsOpts) error
 
+// SubscribeOption configures JetStream subscriber behaviour
+type SubscribeOption func(opts *jsOpts) error
+
+// SubscribeConsumer creates a JetStream consumer on stream based on DefaultConsumer and modified by opts
+func SubscribeConsumer(stream string, opts ...ConsumerConfigOption) SubscribeOption {
```
derekcollison, line 146: Just Consumer I think..

jetstream.go
```diff
@@ -133,6 +139,31 @@ type AckOption func(opts *jsOpts) error
 // PublishOption configures publishing messages
 type PublishOption func(opts *jsOpts) error
 
+// SubscribeOption configures JetStream subscriber behaviour
+type SubscribeOption func(opts *jsOpts) error
```
derekcollison, line 143: ConsumerOption IMO.

ripienaar:
> You have a spelling mistake you want to fix.
> 
> `./jetstream.go:142:51: "behaviour" is a misspelling of "behavior"`

Yeah noticed but this PR isn’t about correctness it’s about looking for feedback about approach and general shape of things

jetstream_consumer.go
```diff
@@ -0,0 +1,611 @@
...
+	ReplayPolicy:  ReplayInstant,
+}
+
+// SampledDefaultConsumer is the configuration that will be used to create new Consumers in NewConsumer
+var SampledDefaultConsumer = ConsumerConfig{
+	DeliverPolicy:   DeliverAll,
+	AckPolicy:       AckExplicit,
+	AckWait:         30 * time.Second,
+	ReplayPolicy:    ReplayInstant,
+	SampleFrequency: "100%",
```
derekcollison, line 40: Should this just be uint8?

jetstream_consumer.go
```diff
@@ -0,0 +1,611 @@
...
+	if cresp.IsError() {
+		return nil, cresp.ToError()
+	}
+
+	return cresp.ConsumerInfo, nil
+}
+
+const (
+	JSApiConsumerCreateT                   = "$JS.API.CONSUMER.CREATE.%s"
+	JSApiDurableCreateT                    = "$JS.API.CONSUMER.DURABLE.CREATE.%s.%s"
```
derekcollison, line 283: This feels like management lib/module which should be separate no?

derekcollison (changes requested):
Few comments and I think we may have pulled in too much of the API/Management layers for the base lib. WDYT?

derekcollison:
Since consumer config is public should we just make them fill one of those out and pass that in as only Consumer option for Subscribe?

jetstream.go
ripienaar: The Ack related ones start with Ack, publish ones with Publish etc. Makes them easy to find in IDE completion. I suspect we would add a good few more Subscribe related options in time

jetstream_consumer.go
ripienaar: JS expects a string iirc

jetstream_consumer.go
ripienaar: Yes did consider trimming them but having these here will help people who want to use Request() to interact with the API for things we don’t have here. So kept them but also happy to say people should call into jsm.go for that

ripienaar:
> Since consumer config is public should we just make them fill one of those out and pass that in as only Consumer option for Subscribe?

Yeah, that’s pretty much what we have all been saying would be the better approach in recent calls. So yes I would be happy to strip this down to that. Leaving the more verbose API in jsm

derekcollison:
Let's do that, we can always add in the additional helpers if requested.

enc.go
ColinSullivan1: I agree... most clients don't support encoded conns and didn't see it used much.  NKIT may be better suited.

ripienaar:
Simpliefied down to 

```go
	cfg := &nats.ConsumerConfig{
		Durable:        "nats",
		DeliverSubject: nats.NewInbox(),
		DeliverPolicy:  nats.DeliverAll,
		AckPolicy:      nats.AckExplicit,
		AckWait:        5 * time.Second,
		ReplayPolicy:   nats.ReplayInstant,
	}

	sub, err := nc.Subscribe("", cb, nats.SubscribeConsumer("TEST", cfg))
	if err != nil {
		t.Fatalf("create failed: %s", err)
	}
```

derekcollison:
We should have proper defaults and defaults processing so you don't have to fill them all in IMO.. Or have sane default configs but I like the default processing if they are not filled in..

jetstream.go
```diff
@@ -139,11 +139,11 @@ type AckOption func(opts *jsOpts) error
...
-type SubscribeOption func(opts *jsOpts) error
+// ConsumerOption configures JetStream consumer behavior
+type ConsumerOption func(opts *jsOpts) error
 
-// SubscribeConsumer creates a JetStream consumer on stream based on DefaultConsumer and modified by opts
-func SubscribeConsumer(stream string, cfg *ConsumerConfig) SubscribeOption {
+// Consumer creates a JetStream Consumer on stream based on DefaultConsumer and modified by opts
+func Consumer(stream string, cfg *ConsumerConfig) ConsumerOption {
 	return func(jopts *jsOpts) error {
 		jopts.consumer = cfg
```
derekcollison, line 148: Should we process and fill in sane defaults here so we do not force users to fill in all of the fields?

ripienaar:
Defaults to `{
  "deliver_policy": "all",
  "ack_policy": "none",
  "replay_policy": "instant"
}` which seems fair, rest of the stuff is fairly optional?

Previously we had some default starting points, but without modifier functions its not clear how a default would be a starting point in the model where the user makes a new config? Got an example of what you have in mind?

derekcollison:
I think that is reasonable sans maybe ack none ;)

derekcollison (approved):
LGTM

ripienaar:
> I think that is reasonable sans maybe ack none ;)

Yeah unfortunately since these are numbers underneath I have no way to say it’s unset acknone vs specifically requested acknone. 

This was copy pasted from the server basically. So think it’s just how it is. 

I will do some more tests etc. 

Given we are taking a very light touch here it would be easy to do a stream add function, think we should?

jetstream.go
ColinSullivan1: An API could allocate the consumer config struct and populate defaults  e.g. `NewConsumerConfig`.  It'd avoid issues around cases where an empty string or zero is a valid value but not necessarily the default or a good one.

derekcollison:
We can change values of enums if needed to get the correct behavior. Would be breaking though most likely.

derekcollison:
Re: stream add, not for now IMO.. We can add, but can't take away ;)

Interested in what @kozlovic thinks.

kozlovic:
> Re: stream add, not for now IMO.. We can add, but can't take away ;)
> 
> Interested in what @kozlovic thinks.

I would really like to move all config out of the client libraries (I am having cold sweat thinking of doing all that in the C client ;-)). But I understand that for consumer we may not have much of a choice. However, for stream I think it can be done from admin or with template (if I am not mistaken).

ripienaar:
Yeah for go we have jsm.go for that so it’s fine.

ColinSullivan1 (approved):
LGTM

kozlovic (approved):
LGTM

jetstream_consumer.go
```diff
@@ -0,0 +1,335 @@
...
+type SequencePair struct {
+	ConsumerSeq uint64 `json:"consumer_seq"`
+	StreamSeq   uint64 `json:"stream_seq"`
+}
+
+type ConsumerInfo struct {
+	Stream         string         `json:"stream_name"`
+	Name           string         `json:"name"`
+	Config         ConsumerConfig `json:"config"`
+	Created        time.Time      `json:"created"`
```
aricart, line 322: Same, here we have a very large number that is not friendly to anything other than Go.

nats.go
```diff
@@ -492,6 +493,10 @@ type Subscription struct {
 	// only be processed by one member of the group.
 	Queue string
 
+	// ConsumerConfig is the configuration for the JetStream consumer if one was created
+	// or updated using the subscription options
+	ConsumerConfig *ConsumerConfig
```
aricart, line 498: When we have some other feature we don't know about, that is not 'consumer' specific this will require a breaking change in the signature. I think the function signatures should have a "SubscriptionOption" that has ConsumerOptions.

nats.go
```diff
@@ -3268,17 +3273,53 @@ func badQueue(qname string) bool {
 }
 
 // subscribe is the internal subscribe function that indicates interest in a subject.
-func (nc *Conn) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync bool) (*Subscription, error) {
+func (nc *Conn) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync bool, opts ...ConsumerOption) (*Subscription, error) {
```
aricart, line 3276: Here's an example - if we had the subscribe option, the sig would be stable.

jetstream.go
```diff
@@ -108,7 +112,9 @@ type jsOpts struct {
 	timeout time.Duration
 	ctx     context.Context
 
-	ack jsAcKOpts
+	ack        jsAcKOpts
```
aricart, line 115: `jsAcKOpts`? There's a capital K and also stutter with 'js'

jetstream.go
aricart: Yes, should have all the defaults there.

jetstream_consumer.go
```diff
@@ -0,0 +1,335 @@
...
+
+	return cresp.ConsumerInfo, nil
+}
+
+const (
+	JSApiConsumerCreateT = "$JS.API.CONSUMER.CREATE.%s"
+	JSApiDurableCreateT  = "$JS.API.CONSUMER.DURABLE.CREATE.%s.%s"
+)
+
+type ApiError struct {
```
aricart, line 75: Is it an API Error or a RequestError - `ApiError` means that they called it wrong, but likely if there's some other situation it is not an `ApiError`.

jetstream_consumer.go
```diff
@@ -0,0 +1,335 @@
...
+// ServerError is true when the server returns a 5xx error code
+func (e ApiError) ServerError() bool { return e.Code >= 500 && e.Code < 600 }
+
+// UserError is true when the server returns a 4xx error code
+func (e ApiError) UserError() bool { return e.Code >= 400 && e.Code < 500 }
+
+// ErrorCode is the JetStream error code
+func (e ApiError) ErrorCode() int { return e.Code }
+
+type JSApiResponse struct {
```
aricart, line 104: JSResponse?

jetstream_consumer.go
```diff
@@ -0,0 +1,335 @@
...
+
+const (
+	ReplayInstant ReplayPolicy = iota
+	ReplayOriginal
+)
+
+func (p ReplayPolicy) String() string {
+	switch p {
+	case ReplayInstant:
+		return "Instant"
```
aricart, line 194: should all these magic strings that change case, etc be constants?

jetstream_consumer.go
```diff
@@ -0,0 +1,335 @@
...
+		return json.Marshal("original")
+	case ReplayInstant:
+		return json.Marshal("instant")
+	default:
+		return nil, fmt.Errorf("unknown replay policy %v", p)
+	}
+}
+
+var (
+	AckAck      = []byte("+ACK")
```
aricart, line 227: "+ACK" came on my radar a couple of days ago. IMHO the regular Ack should just be the empty payload. The other values are options to the response which alter the meaning.

aricart:
Sorry to have this count of comments, but if we could change something, now is the time to get it right.

jetstream.go
ripienaar: It’s not needed the basic structure defaults inherently.

jetstream_consumer.go
```diff
@@ -0,0 +1,335 @@
...
+//
+// NATS Schema Type io.nats.jetstream.api.v1.consumer_configuration
+type ConsumerConfig struct {
+	Durable         string        `json:"durable_name,omitempty"`
+	DeliverSubject  string        `json:"deliver_subject,omitempty"`
+	DeliverPolicy   DeliverPolicy `json:"deliver_policy"`
+	OptStartSeq     uint64        `json:"opt_start_seq,omitempty"`
+	OptStartTime    *time.Time    `json:"opt_start_time,omitempty"`
+	AckPolicy       AckPolicy     `json:"ack_policy"`
+	AckWait         time.Duration `json:"ack_wait,omitempty"`
```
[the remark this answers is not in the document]
ripienaar: That would be a server change and be breaking. So up to Derek really

jetstream_consumer.go
ripienaar: These are generic errors we use all over in many different messages from the server responses. Only one here but it’s used many places. I kept it the same so it’s a copy paste when we update the server

jetstream_consumer.go
ripienaar: Go times in json is a string in rfc3339 format not a big number

aricart: That is possibly even worse, depending on the language difficult to parse. Unix time is always super easy.

nats.go
ripienaar: Not true here, this ConsumerConfig is the actual JetStream Consumer configuration as returned by the create API, it doesnt have extra functionality it is what it is, if something else will be needed in the future it would be a different struct key

nats.go
ripienaar: Yeah, agree

jetstream.go
ripienaar: typo, will fix - am happy with the stutter though.

jetstream_consumer.go
ripienaar: Also copy and pasted from the server, as the server evolve I want to make it easy to sync this - by just copy and pasting, so will keep as is, I think I agree probably it would be better as constants - I suspect these are entirely unused in this code though so meh

jetstream_consumer.go
ripienaar: Used to be, Derek prefers +ACK both are accepted.  Regardless, copied from the server so will resolve.

ripienaar:
OK I flattened and extended the tests a bit, so this is good to go.

jetstream.go
```diff
@@ -100,19 +104,21 @@ func (m *Msg) parseJSMsgMetadata() (*JetStreamMsgMetaData, error) {
 
 const jsStreamUnspecified = "not.set"
 
-type jsAcKOpts struct {
+type jsAckOpts struct {
```
derekcollison, line 107: This need to be a struct? Just embed the string into jsOpts?

jetstream_consumer.go
```diff
@@ -0,0 +1,333 @@
...
+		return "unknown JetStream Error"
+	case e.Description == "" && e.Code > 0:
+		return fmt.Sprintf("unknown JetStream %d Error", e.Code)
+	default:
+		return e.Description
+	}
+}
+
+// NotFoundError is true when the error is one about a resource not found
+func (e ApiError) NotFoundError() bool { return e.Code == 404 }
```
derekcollison, line 91: I still think in clients we absorb this and don't surface, just do new Error for each one where it may surface. This feels like jsm not base client lib.

jetstream_consumer.go
```diff
@@ -0,0 +1,333 @@
...
+	return *r.Error
+}
+
+// IsError determines if a standard JetStream API response is a error
+func (r JSApiResponse) IsError() bool {
+	return r.Error != nil
+}
+
+// io.nats.jetstream.api.v1.consumer_create_request
+type JSApiConsumerCreateRequest struct {
```
derekcollison, line 122: Again would hide.

nats.go
```diff
@@ -3187,15 +3194,15 @@ func (nc *Conn) respToken(respInbox string) string {
...
+func (nc *Conn) Subscribe(subj string, cb MsgHandler, opts ...ConsumerOption) (*Subscription, error) {
+	return nc.subscribe(subj, _EMPTY_, cb, nil, false, opts...)
 }
 
 // ChanSubscribe will express interest in the given subject and place
 // all messages received on the channel.
 // You should not close the channel until sub.Unsubscribe() has been called.
-func (nc *Conn) ChanSubscribe(subj string, ch chan *Msg) (*Subscription, error) {
-	return nc.subscribe(subj, _EMPTY_, nil, ch, false)
+func (nc *Conn) ChanSubscribe(subj string, ch chan *Msg, opts ...ConsumerOption) (*Subscription, error) {
```
derekcollison, line 3204: Should we move ConsumerOption to just SubscribeOption and possibly use it to overload stuff (options) we already have?

nats_test.go
```diff
@@ -2540,3 +2541,76 @@ func TestLameDuckMode(t *testing.T) {
...
+		m.Ack()
+		seen++
+		if seen == 20 {
+			cancel()
+		}
+	}
+
+	cfg := &ConsumerConfig{
+		Durable:        "nats",
+		DeliverSubject: NewInbox(),
```
derekcollison, line 2565: I feel we should hide this by default and handle cleanly. We can expose if they really want it but by default when we subscribe should just do the right thing. We do not need to know about the inbox etc.

nats_test.go
```diff
@@ -2540,3 +2541,76 @@ func TestLameDuckMode(t *testing.T) {
...
+	if err.Error() != "nats: JetStream consumer creation failed: consumer already exists" {
+		t.Fatalf("expected already exist error got %q", err.Error())
+	}
+
+	sub.Unsubscribe()
+
+	// should be able to create an ephemeral which requires interest to exist first
+	// before creating the ephemeral
+	cfg.Durable = ""
+	sub, err = nc.Subscribe("test", cb, Consumer("TEST", cfg))
```
derekcollison, line 2606: For instance, here we have a `cb` so we know we need push, we can do the inbox thing for the user and not make them do all the boilerplate from above.

derekcollison (changes requested):

nats.go
ripienaar: It was `SubscribeOption`, you had me change it to this.

nats_test.go
ripienaar: this only makes sense as long as we only support 1 client per push consumer, if someone wanted another we need to give them the option to set it, so i could if say sub=="" AND DeliverySubject=="" then make an Inbox but leave it as is if a subject is given.

nats.go
derekcollison: I was just wondering if we could use the options for generic stuff too like timeouts vs what we have now.

nats_test.go
derekcollison: We get sub on normal Subscribe() call, but yes if we are doing push based subscriber vs SubscribeSync() then spin up an inbox delivery subject if its not present.

jetstream.go
ripienaar: doesnt need to be now no, used to have a few more ack related things there

nats.go
ripienaar: Yes, thats why it was more generically named before :) I will change back

nats_test.go
ripienaar: As things stand `SubscribeSync` is also a push under the covers for the moment, so will handle the case where no subject is given at all by setting inbox.

ripienaar: The problem with this is is subj on Subscribe is empty and the DeliverySubject is empty - we dont know the users intention, they could be trying to make a pull consumer.  We abuse DeliverySubject to convey the intent to make a push/pull.

So I cant use a Inbox if `subj==""&&DeliverySubject==""` because the user might be trying to make a pull consumer but mistakenly giving it to a normal subscribe, so for now we fail in that case.  Later when we have a way to do fast pull consumers here we can revisit this perhaps?

jetstream_consumer.go
ripienaar: not used here so just removed them, we only use the thing that makes a err from a response.  I copy/pasted all the code rather than cherry pick since while we are still changing how things work it will be a pain to keep yet-another-package up to date, but fine, will do

nats_test.go
ripienaar: Would leave as is for the moment until we know how pull handling will actually look, easier to add the bit of magic than take it away if it seems weird later

nats.go
```diff
@@ -3268,17 +3275,58 @@ func badQueue(qname string) bool {
...
 
 // subscribe is the internal subscribe function that indicates interest in a subject.
-func (nc *Conn) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync bool) (*Subscription, error) {
+func (nc *Conn) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync bool, opts ...SubscribeOption) (*Subscription, error) {
 	if nc == nil {
 		return nil, ErrInvalidConnection
 	}
+
+	var aopts *jsOpts
+	if len(opts) > 0 {
```
ripienaar, line 3284: Should I fail here if old style subscriptions are not used?

nats_test.go
derekcollison: I think we do pull for SubscribeSync..

nats.go
derekcollison: Not following, fail why?

nats_test.go
ripienaar: I can drop the subscribe options till we have pull support from SubscribeSync() or leave as is and it works at least? We dont have pull support so not sure what the action you want here

nats.go
ripienaar: Because as discussed, much of JS will not work with it.

nats_test.go
derekcollison: Subscribe should always be push based, so if they do not supply deliver subject in config create inbox etc.

nats.go
derekcollison: This is just low level subscribe. We aren't doing the mux inbox trick here, just on Request().

derekcollison: JS works with new style unless multiple concurrent requests in flight.

nats_test.go
ripienaar: I dont think a API should change the users intent ever, if the user asks the impossible tell him so is a much better behavour, users could be pre-creating a pull consumer and try to use it via Subscribe, thats an error that we can give a nice friendly error for rather than the opaque "consumer already exist" he'd get from the server

nats.go
ripienaar: If a user specify a NewInbox() on a consumer as in the tests, we really cant says "unless multiple concurrent" we have no way of knowing when that is happening.  The consequence is literally messages go missing and never get delivered, we cant have that.

nats_test.go
derekcollison: Not following. I am just saying that if a consumer cfg option is passed to subscribe without a delivery subject create one and do not force the user to do all the boilerplate. If it is set use it as is..

nats.go
derekcollison: `NewInbox()` just creates a subject that you still need to subscribe to IIRC. So I think we are fine there.

nats_test.go
ripienaar: Without a delivery subject == pull consumer.  How do I know he isnt intending to access a pull consumer? Do I just override his wish and make a pull one, or educate him about this mistake. 

For queue and sync based subscribes it does it for the user

derekcollison: Subscribe forces user to provide a callback, that has always been pushed based. IMO, pull based for JetStream is equivalent to SubscribeSync where the user does NextMsg().

ripienaar: I understand that! but people make mistakes.  Whatever this is not helpful, will put in the behaviour you want.

ripienaar: They dont have to make a low level subscribe. This code does it for them.  We're arguing over one line that would make the behaviour clear and inform a user when their intent doesnt match what the underlying code can implement for them.

derekcollison: I think if its empty create one and do low level subscribe etc. If deliver subject is set, do nothing..

What are you exactly recommending? Apologies, 50 balls in the air already this am.

nats.go
```diff
@@ -3289,14 +3288,17 @@ func (nc *Conn) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSyn
 			}
 		}
 
-		// for subscribe sync case we make it so users don't need to supply a inbox that they never use
-		if aopts.consumer.DeliverSubject == "" && subj == "" && (cb == nil || ch == nil) {
+		// when no DeliverySubject is given, while we don't have any support for pull mode
```
derekcollison, line 3291: I wonder if we remove deliver subject all together for the consumer config for the Go client?

ripienaar: I have to send it to the server, and not all consumers are going to be unique inboxes based, I might want to do a queue subscribe on a js consumer between 2 clients - but want the first client that start to create the consumer rather than do so from cli/etc. then I have to do so on a well known delivery subject.

derekcollison: Feels like that would fit under QueueSubscribe no? Trying to keep what folks already know top of mind.

ripienaar: Regardless we can’t remove DeliverySubject even if we only do it in one place, I was responding specifically to you asking why not remove it.

derekcollison: We need it sent to the server, but the public cfg struct that a user would utilize for the consumer config option for Subsc ribe() does not have to have it.

ripienaar: So you suggest I make ConsumerConfig without that key and yet another type that, I guess, embeds it and has the subject that I then set, this 2nd shadow type is set to the server?

Tbh if I were to try learn how JS works this would just be all too magical if we have public structures that doesn’t match the documented APIs or schemas and where people need to read the nats.go code to figure out how the JS API works and what magical under the cover things happen- since it would be completely non obvious how you go from this API behaviour to how JetStream works - it would just be frustrating and confusing. 

As it is we are doing the absolute minimum here and by making it magic and non standard - the server API being standard - we are jus adding  a bunch of wtf behaviour and strange magic if we keep going this way. 

But that’s the last comment I will make so will just go ahead and do whatever you want so will do this next week.

derekcollison: I think what we are trying to do is make a straightforward way for Go programmers to consume messages from a JetStream stream.

We both are pushing for the same thing I believe just approaching it differently. I would like JetStream users to see consumers as just another NATS subscription vs something totally different like we did with STAN. Under the covers they are different, but I believe at the base client lib these should just look like other subscribers.

ripienaar:
removed DeliverySubject.

derekcollison (approved):
LGTM - Let's start writing some sample JS apps to get a feel if we are headed in the right direction. WDYT?

ripienaar:
OK, can we merge this one? We're on a bit of a tight timeline and other languages have to follow the same design.

derekcollison:
Yes feel free to merge.

PR #598, opened 2020-10-21
Add support for status line plus description

derekcollison:
Not sure if this is the right approach. The server will use inline status and descriptions for efficiency and originally we promote these to Status. Added in support for Description. Note the Go does not do this per se on HTTP.

https://golang.org/pkg/net/http/#Response

### DIFF
--- a/nats_test.go
+++ b/nats_test.go
@@ -31,6 +31,7 @@ import (
 	"reflect"
 	"regexp"
 	"runtime"
+	"strconv"
 	"strings"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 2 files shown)


nats_test.go
```diff
@@ -2446,6 +2447,22 @@ func TestHeaderParser(t *testing.T) {
...
+	// Check that we can do inline status and descriptions
+	checkStatus := func(hdr string, status int, description string) {
+		t.Helper()
+		hdrs, err := decodeHeadersMsg([]byte(hdr + "\r\n\r\n"))
+		if err != nil {
+			t.Fatalf("Unexpected error: %v", err)
+		}
+		if code, err := strconv.Atoi(hdrs.Get(statusHdr)); err != nil || code != status {
+			t.Fatalf("Expected status of %d, got %s", status, hdrs.Get(statusHdr))
+		}
```
kozlovic, line 2460: You are not checking the expected `description` here.

kozlovic (changes requested):
Test is not checking expected description.. Also, afraid that server adding description here is going to break already released clients that support headers (C, .NET, etc..). Looping @ColinSullivan1 so he is aware, maybe need to notify others.

nats_test.go
derekcollison: Thanks added.

kozlovic (approved):
LGTM. But wanted to confirm if server will now return "503 No Responders" or stick with "503"? If the former, then it will break some clients (meaning that libs won't match status == "503").

derekcollison:
For no no responders is still plain 503 but JetStream has added ina bunch of signalling back to user via headers for core operations etc.

PR #599, opened 2020-10-28
support consumer pending counts

ripienaar:


jetstream.go
```diff
@@ -97,6 +99,14 @@ func (m *Msg) parseJSMsgMetadata() (*JetStreamMsgMetaData, error) {
 	}
 	meta.TimeStamp = time.Unix(0, int64(tsi))
 
+	meta.Pending = -1
```
derekcollison, line 102: We should do scanf like server, I think that will be generally faster and I may do a very low level fast version which we could borrow here but for now I would do Sscanf like server.

derekcollison (changes requested):

jetstream.go
ripienaar: yeah I spent a bunch of time wasting on Sscanf and its no good - in server you know the Stream and Consumer, here we have to also extract the Stream and Consumer.  fmt.Sscanf only works with strings if they are space seperated not `.` as here.

derekcollison: ok so keep as is and I can do a high speed one at some point in Go and you can take it too.

derekcollison (approved):
LGTM

PR #601, opened 2020-10-30
support requesting multiple messages in next 

ripienaar:


jetstream.go
```diff
@@ -252,9 +252,14 @@ func (m *Msg) AckProgress(opts ...AckOption) error {
...
-// AckNext performs an Ack() and request that the next message be sent to subject ib
-func (m *Msg) AckNext(ib string) error {
-	return m.RespondMsg(&Msg{Subject: m.Reply, Reply: ib, Data: AckNext})
+// AckNext performs an Ack() and request that the next cnt message be sent to subject inbox
+func (m *Msg) AckNext(inbox string, cnt uint) error {
+	if cnt == 0 {
+		cnt = 1
+	}
+
+	data := append(AckNext, []byte(" "+strconv.Itoa(int(cnt)))...)
```
derekcollison, line 261: Would only do this if != 1

derekcollison (changes requested):

ripienaar:
updated

jetstream.go
```diff
@@ -252,9 +252,19 @@ func (m *Msg) AckProgress(opts ...AckOption) error {
 	return m.jsAck(AckProgress, opts...)
 }
 
-// AckNext performs an Ack() and request that the next message be sent to subject ib
-func (m *Msg) AckNext(ib string) error {
-	return m.RespondMsg(&Msg{Subject: m.Reply, Reply: ib, Data: AckNext})
+// AckNextMulti performs an Ack() and request that the next count message be sent to subject inbox
+func (m *Msg) AckNextMulti(inbox string, count uint) error {
```
ripienaar, line 256: @derekcollison did this before realising we also want to support the full request structure, so what API do you suggest? `AckNextMulti(inbox, count, timeout)`? where timeout is a duration.

derekcollison: Need to think about this one a bit more.. Will try to do so a bit later today.

ripienaar: gentle reminder about this one :)

derekcollison: Since neither of these will return a message or series of messages, meaning they just communicate where to deliver the next batch, we just need default and full blow request.

`AckNext(*nextReq)` where if nextReq is nil then assume same delivery subject and batch of 1.
Otherwise they can fill in a formal request?

Or we do this.

`AckNext(batchSize int)` and `AckNextRequest(nextReq)`

ripienaar: we really want them to create that structure? Not do it for them by just asking the 2 possible values? Seems a lot easier this way than expect people to make a whole structure to me.

derekcollison: We have more than 2 values and in this case the reply subject is probably same as from the message we are using to do AckNext. So really only batchSize is a common use case but could also set "don't wait flag" etc. 

If you think the request is fixed and we will not add any fields in we could try to explode it out instead.

ripienaar: updated

jetstream.go
```diff
@@ -252,9 +252,29 @@ func (m *Msg) AckProgress(opts ...AckOption) error {
...
+	Expires time.Time `json:"expires,omitempty"`
+	// Batch is how many messages to request
+	Batch int `json:"batch,omitempty"`
+	// NoWait indicates that if the Consumer has consumed all messages an Msg with Status header set to 404
+	NoWait bool `json:"no_wait,omitempty"`
+}
+
+// AckNextRequest performs an acknowledgement of a message and request the next messages based on req
+func (m *Msg) AckNextRequest(inbox string, req *AckNextRequest) error {
+	rj, err := json.Marshal(req)
```
derekcollison, line 267: Need to check if its nil I think.

jetstream.go
```diff
@@ -252,9 +252,29 @@ func (m *Msg) AckProgress(opts ...AckOption) error {
...
+	// Expires is the time when the server will stop honoring this request
+	Expires time.Time `json:"expires,omitempty"`
+	// Batch is how many messages to request
+	Batch int `json:"batch,omitempty"`
+	// NoWait indicates that if the Consumer has consumed all messages an Msg with Status header set to 404
+	NoWait bool `json:"no_wait,omitempty"`
+}
+
+// AckNextRequest performs an acknowledgement of a message and request the next messages based on req
+func (m *Msg) AckNextRequest(inbox string, req *AckNextRequest) error {
```
derekcollison, line 266: Why pass in inbox? In this version we will have already established the subscriber for the messages themselves since AckNext does not return new messages. I don't think we need it TBH looking at how I used it in testing.

jetstream.go
```diff
@@ -252,9 +252,29 @@ func (m *Msg) AckProgress(opts ...AckOption) error {
...
+	rj, err := json.Marshal(req)
+	if err != nil {
+		return err
+	}
+
+	return m.RespondMsg(&Msg{Subject: m.Reply, Reply: inbox, Data: append(AckNext, append([]byte{' '}, rj...)...)})
+}
+
+// AckNext performs an Ack() and request that the next message be sent to subject inbox, to request multiple messages use AckNextRequest()
+func (m *Msg) AckNext(inbox string) error {
```
derekcollison, line 276: I think just `AckNext()` no inbox needed.

derekcollison (changes requested):

ripienaar:
updated

derekcollison (approved):
LGTM

PR #603, opened 2020-11-11
expose additional info line information to the client

ripienaar:

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -590,19 +590,20 @@ type srv struct {
 // The INFO block received from the server.
 type serverInfo struct {
 	ID           string   `json:"server_id"`
+	Name         string   `json:"server_name"`
+	Proto        int      `json:"proto"`
 	Host         string   `json:"host"`
-	Port         uint     `json:"port"`
-	Version      string   `json:"version"`
-	AuthRequired bool     `json:"auth_required"`
-	TLSRequired  bool     `json:"tls_required"`
-	TLSAvailable bool     `json:"tls_available"`
+	Port         int      `json:"port"`
 	Headers      bool     `json:"headers"`
+	AuthRequired bool     `json:"auth_required,omitempty"`
+	TLSRequired  bool     `json:"tls_required,omitempty"`
+	TLSAvailable bool     `json:"tls_available,omitempty"`
 	MaxPayload   int64    `json:"max_payload"`
-	ConnectURLs  []string `json:"connect_urls,omitempty"`
-	Proto        int      `json:"proto,omitempty"`
 	CID          uint64   `json:"client_id,omitempty"`
 	ClientIP     string   `json:"client_ip,omitempty"`
 	Nonce        string   `json:"nonce,omitempty"`
+	Cluster      string   `json:"cluster,omitempty"`
+	ConnectURLs  []string `json:"connect_urls,omitempty"`
 	LameDuckMode bool     `json:"ldm,omitempty"`
 }
 
@@ -1489,7 +1490,7 @@ func (nc *Conn) waitForExits() {
 	nc.wg.Wait()
 }
 
-// Report the connected server's Url
+// ConnectedUrl reports the connected server's URL
 func (nc *Conn) ConnectedUrl() string {
 	if nc == nil {
 		return _EMPTY_
@@ -1519,7 +1520,7 @@ func (nc *Conn) ConnectedAddr() string {
 	return nc.conn.RemoteAddr().String()
 }
 
-// Report the connected server's Id
+// ConnectedServerId reports the connected server's Id
 func (nc *Conn) ConnectedServerId() string {
 	if nc == nil {
 		return _EMPTY_
@@ -1532,6 +1533,36 @@ func (nc *Conn) ConnectedServerId() string {
 		return _EMPTY_
 	}
 	return nc.info.ID
+}
+
+// ConnectedServerName reports the connected server's name
+func (nc *Conn) ConnectedServerName() string {
+	if nc == nil {
+		return _EMPTY_
+	}
+
+	nc.mu.RLock()
+	defer nc.mu.RUnlock()
+
+	if nc.status != CONNECTED {
+		return _EMPTY_
+	}
+	return nc.info.Name
+}
+
+// ConnectedClusterName reports the connected server's cluster name if any
+func (nc *Conn) ConnectedClusterName() string {
+	if nc == nil {
+		return _EMPTY_
+	}
+
+	nc.mu.RLock()
+	defer nc.mu.RUnlock()
+
+	if nc.status != CONNECTED {
+		return _EMPTY_
+	}
+	return nc.info.Cluster
 }
 
 // Low level setup for structs, etc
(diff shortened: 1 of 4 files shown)


nats.go
```diff
@@ -589,21 +589,30 @@ type srv struct {
...
+	MaxPayload        int64    `json:"max_payload"`
+	Name              string   `json:"server_name"`
+	Nonce             string   `json:"nonce,omitempty"`
+	Port              int      `json:"port"`
+	Proto             int      `json:"proto"`
+	TLSAvailable      bool     `json:"tls_available,omitempty"`
+	TLSRequired       bool     `json:"tls_required,omitempty"`
+	TLSVerify         bool     `json:"tls_verify,omitempty"`
+	Version           string   `json:"version"`
+	WSConnectURLs     []string `json:"ws_connect_urls,omitempty"` // Contains URLs a ws client can connect to.
```
ripienaar, line 615: might not need all these, i mainly updated it with the server

derekcollison: I see your editor sorted them too 👎

nats.go
```diff
@@ -589,21 +589,30 @@ type srv struct {
...
+	ClientIP          string   `json:"client_ip,omitempty"`
+	Cluster           string   `json:"cluster,omitempty"`
+	Dynamic           bool     `json:"cluster_dynamic,omitempty"`
+	GitCommit         string   `json:"git_commit,omitempty"`
+	GoVersion         string   `json:"go"`
+	Headers           bool     `json:"headers"`
+	Host              string   `json:"host"`
+	ID                string   `json:"server_id"`
+	IP                string   `json:"ip,omitempty"`
+	JetStream         bool     `json:"jetstream,omitempty"`
```
derekcollison, line 604: Would remove JetStream and LDM since we already process that IIRC.

nats.go
```diff
@@ -1489,7 +1498,7 @@ func (nc *Conn) waitForExits() {
 	nc.wg.Wait()
 }
 
-// Report the connected server's Url
+// ConnectedUrl report the connected server's Url
```
derekcollison, line 1501: reports and Url -> URL

nats.go
```diff
@@ -1519,7 +1528,7 @@ func (nc *Conn) ConnectedAddr() string {
 	return nc.conn.RemoteAddr().String()
 }
 
-// Report the connected server's Id
+// ConnectedServerId report the connected server's Id
```
derekcollison, line 1531: reports

nats.go
ripienaar: Hehe, can undo, there's some weird things like this client has a int64 for MaxPayload but the server reports int32, I guess we canr really change either at this point so will keep casting here differently from the server

derekcollison (changes requested):
Looks pretty good, minor cleanup.

nats.go
derekcollison: You can probably make it int if you want..

ripienaar: I thought we wouldnt want to change this sig `func (nc *Conn) MaxPayload() int64`

nats.go
ripienaar: Can't change this, its an existing function, I just fixed the comment to match reality

nats.go
ripienaar: We dont process JetStream and LDM was here already - I considered adding a function to expose the JetStream property but I dont think it's useful, in typical real world clusters the individual servers might not be JS enabled only the storage node etc, do you think exposing it will be useful?

nats.go
derekcollison: Ah yes your right..

nats.go
derekcollison: Correct, would remove JS for now..

nats.go
derekcollison: I meant at end of the comment.

nats.go
ripienaar: ok, remove JS and put this in the sme order as in the server rather than sorted

nats.go
ripienaar: doh, ok, fixed.

derekcollison (approved):
Just remove Dynamic and we should be good to go.

derekcollison (approved):
LGTM

kozlovic:
Please do not merge, give me a chance to review. The ConnectClientURLs, etc.. are normally not sent to client, this is between servers.

nats.go
kozlovic: Remove `WSConnectURLs`, it is not set by the server and instead uses `connect_urls`. I would also not have renamed `ConnectURLs` to `ClientConnectURLs` since from a client perspective, this is the connect urls :-)

nats.go
```diff
@@ -2599,7 +2636,7 @@ func (nc *Conn) processInfo(info string) error {
 	// if advertise is disabled on that server, or servers that
 	// did not include themselves in the async INFO protocol.
 	// If empty, do not remove the implicit servers from the pool.
-	if len(ncInfo.ConnectURLs) == 0 {
+	if len(nc.info.ClientConnectURLs) == 0 {
```
kozlovic, line 2639: Rename here would not be necessary if you do not rename the field in the Info.

kozlovic (changes requested):
I understand the pain of maintaining this, adding only the things that we need in the client, but the server's Info contains things that are either not for the clients (for instance sending to gateways, or route specific, etc..), so I would really "port" only those that are relevant and want to expose.
Having some fields in here that are not related to the client, etc.. may make it even more confusing in the future.

nats.go
```diff
@@ -589,21 +589,28 @@ type srv struct {
...
-	CID          uint64   `json:"client_id,omitempty"`
-	ClientIP     string   `json:"client_ip,omitempty"`
-	Nonce        string   `json:"nonce,omitempty"`
-	LameDuckMode bool     `json:"ldm,omitempty"`
+	ID                string   `json:"server_id"`
+	Name              string   `json:"server_name"`
+	Version           string   `json:"version"`
+	Proto             int      `json:"proto"`
+	GitCommit         string   `json:"git_commit,omitempty"`
+	GoVersion         string   `json:"go"`
```
kozlovic, line 597: GitCommit and GoVersion not exposed, so no need here.

nats.go
```diff
@@ -589,21 +589,28 @@ type srv struct {
...
+	Host              string   `json:"host"`
+	Port              int      `json:"port"`
+	Headers           bool     `json:"headers"`
+	AuthRequired      bool     `json:"auth_required,omitempty"`
+	TLSRequired       bool     `json:"tls_required,omitempty"`
+	TLSVerify         bool     `json:"tls_verify,omitempty"`
```
kozlovic, line 603: Not used.

nats.go
```diff
@@ -589,21 +589,28 @@ type srv struct {
...
+	TLSAvailable      bool     `json:"tls_available,omitempty"`
+	MaxPayload        int64    `json:"max_payload"`
+	IP                string   `json:"ip,omitempty"`
```
kozlovic, line 606: Not used (and not relevant to client I think).

kozlovic (changes requested):

kozlovic (approved):
LGTM

PR #604, opened 2020-11-12
Fix echo example to be more robust

philpennock:
The geolocation service can have ratelimits.  Support using
`$GEOLOCATION_JSON_URL` from environment to override the URL.  As long as the
schema of what's returned is the same, this will work.  This has been tested to
work well with a paid license for the same service we use by default, which
adds `?key=SECRET` to the URL.

Add handling for in-response errors, for improved diagnostics.
Increase timeout for the API service, based on past observed behavior.
Log our geolocation handling at startup (while censoring as needed).

### DIFF
--- a/examples/nats-echo/main.go
+++ b/examples/nats-echo/main.go
@@ -1,4 +1,4 @@
-// Copyright 2018-2019 The NATS Authors
+// Copyright 2018-2020 The NATS Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
@@ -23,6 +23,7 @@ import (
 	"os"
 	"os/signal"
 	"runtime"
+	"strings"
 	"syscall"
 	"time"
 
@@ -33,8 +34,14 @@ import (
 // nats-echo -s demo.nats.io <subject>
 // nats-echo -s demo.nats.io:4443 <subject> (TLS version)
 
+const (
+	DefaultGeolocationURL = "https://ipapi.co/json"
+	GEOLOC_ENV_VARNAME    = "GEOLOCATION_JSON_URL"
+	TimeoutGeolocation    = 10 * time.Second // only used at startup
+)
+
 func usage() {
-	log.Printf("Usage: nats-echo [-s server] [-creds file] [-t] <subject>\n")
+	log.Printf("Usage: nats-echo [-s server] [-creds file] [-t] [-geo] <subject>\n")
 	flag.PrintDefaults()
 }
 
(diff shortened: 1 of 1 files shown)


examples/nats-echo/main.go
```diff
@@ -33,8 +34,14 @@ import (
 // nats-echo -s demo.nats.io <subject>
 // nats-echo -s demo.nats.io:4443 <subject> (TLS version)
 
+const (
+	DefaultGeolocationURL = "https://ipapi.co/json"
+	GEOLOC_ENV_VARNAME    = "GEOLOCATION_JSON_URL"
```
derekcollison, line 39: Would make it more of a Go name, GeolocationEnv

examples/nats-echo/main.go
```diff
@@ -172,5 +188,18 @@ func lookupGeo() string {
...
+	if g.Error {
+		log.Printf("IP Geo data retrieval failed, reason %q message %q", g.Reason, g.Message)
+		return ""
+	}
+	result := g.Region + ", " + g.Country
+
+	// If the env var is used, then there might be something like ?key=... in the URL.
+	// Censor that.
+	logURL := url
+	if offset := strings.IndexByte(logURL, '?'); offset >= 0 {
```
derekcollison, line 200: The std lib I am sure has something that does this already.

derekcollison (changes requested):

examples/nats-echo/main.go
kozlovic: From the url object, you could use a printf with `"%s://%s", u.Scheme, u.Host`

philpennock: We're not making a URL object here, and the stdlib `net/url` package got a `.Redacted() string` method fairly recently (Go 1.15, we still need to support Go 1.14, I believe?), but it only covers a password in the `user:pass@` part (userinfo within the authority), not a key in a query param.

The spec requires `"?"` before the `query` part, and as `?` is in gen-delims for the URI syntax, it has to be percent-encoded if it appears elsewhere.  So for "nuke all query params before logging", this fix is correct.

If we _also_ want to round-trip through a URL object, to redact a password there, we could, but I'm not aware of any geolocation services using user:pass in URLs for access, instead of either a query param or a header-borne token.  So I think this fix is correct and sufficient, without breaking compatibility with Go 1.14.

What I have changed is to add `"?[...]"` to the string when this happens, to make it clearer that a non-default URL which happens to be the same service provider is used.

examples/nats-echo/main.go
philpennock: Good call.  Fixed.

examples/nats-echo/main.go
derekcollison: Why do we need the query params at all for printing out?

philpennock: To indicate that we did remove something.  When using the code with simple truncation (as in the first commit of this PR) I found myself having to double-check that I really had set the env override correctly.  Using a paid account for this geolocation API, the only difference in URL was that `?key=secretstuff` got added, which I was removing to avoid logging secrets.  So I couldn't tell that I really had deployed the env var fix.

Logging with `?[...]` in there will make it clear that an environment override really has taken effect.

derekcollison (approved):
LGTM

PR #605, opened 2020-11-12
Fixed NATS Server dep

kozlovic:
Since we don't vendor in this repo, we need to use go.mod, so removing GO111MODULE=off and the go get for the nats-server (which would pull tip)


derekcollison (approved):
LGTM

PR #607, opened 2020-11-15
Changes to default pending and default error handler.

derekcollison:
Slow consumer state still seems off, this is a first step trying to improve.
Increased the default number of pending messages limit but kept bytes the same.

Also introduced a default ErrHandler that will print out to stderr in case none has been set.

Someone needs to look deeper into why we are so imbalanced when receiving messages like queued up JetStream messages for a consumer.

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -1184,7 +1184,7 @@ func defaultErrHandler(nc *Conn, sub *Subscription, err error) {
 	if nc != nil {
 		nc.mu.RLock()
 		cid = nc.info.CID
-		defer nc.mu.RUnlock()
+		nc.mu.RUnlock()
 	}
 	var errStr string
 	if sub != nil {
@@ -2464,6 +2464,12 @@ func (nc *Conn) processMsg(data []byte) {
 
 	sub.mu.Lock()
 
+	// Check if closed.
+	if sub.closed {
+		sub.mu.Unlock()
+		return
+	}
+
 	// Subscription internal stats (applicable only for non ChanSubscription's)
 	if sub.typ != ChanSubscription {
 		sub.pMsgs++
@@ -2495,7 +2501,9 @@ func (nc *Conn) processMsg(data []byte) {
 		if sub.pHead == nil {
 			sub.pHead = m
 			sub.pTail = m
-			sub.pCond.Signal()
+			if sub.pCond != nil {
+				sub.pCond.Signal()
+			}
 		} else {
 			sub.pTail.next = m
 			sub.pTail = m
--- a/nats.go
+++ b/nats.go
@@ -57,7 +57,7 @@ const (
 	DefaultJetStreamTimeout   = 2 * time.Second
 	DefaultPingInterval       = 2 * time.Minute
 	DefaultMaxPingOut         = 2
-	DefaultMaxChanLen         = 8192            // 8k
+	DefaultMaxChanLen         = 8 * 1024        // 8k
 	DefaultReconnectBufSize   = 8 * 1024 * 1024 // 8MB
 	RequestChanLen            = 8
 	DefaultDrainTimeout       = 30 * time.Second
@@ -1164,6 +1164,11 @@ func (o Options) Connect() (*Conn, error) {
 	nc.ach = &asyncCallbacksHandler{}
 	nc.ach.cond = sync.NewCond(&nc.ach.mu)
 
+	// Set a default error handler that will print to stderr.
+	if nc.Opts.AsyncErrorCB == nil {
+		nc.Opts.AsyncErrorCB = defaultErrHandler
+	}
+
 	if err := nc.connect(); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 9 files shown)


nats.go
```diff
@@ -1174,6 +1179,22 @@ func (o Options) Connect() (*Conn, error) {
 	return nc, nil
 }
 
+func defaultErrHandler(nc *Conn, sub *Subscription, err error) {
+	var cid uint64
+	if nc != nil {
+		nc.mu.RLock()
+		cid = nc.info.CID
+		defer nc.mu.RUnlock()
```
kozlovic, line 1187: Is it intentional that you keep the connection's lock until the end of this function? Error handlers are already serialized, if that was the concern.

nats.go
```diff
@@ -2393,18 +2414,18 @@ func (nc *Conn) waitForMsgs(s *Subscription) {
...
 	// Stats
 	atomic.AddUint64(&nc.InMsgs, 1)
 	atomic.AddUint64(&nc.InBytes, uint64(len(data)))
 
+	// Don't lock the connection to avoid server cutting us off if the
+	// flusher is holding the connection lock, trying to send to the server
+	// that is itself trying to send data to us.
+	nc.subsMu.RLock()
 	sub := nc.subs[nc.ps.ma.sid]
+	nc.subsMu.RUnlock()
```
kozlovic, line 2426: This function was protected by the subsMu so that remove could not happen while processing a message. I believe this is the reason we now can get the panic:
```
=== RUN   TestCloseChanOnSubscriber
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x4bcb30]
goroutine 6453 [running]:
sync.(*Cond).Signal(0x0)
	/home/travis/.gimme/versions/go1.15.5.linux.amd64/src/sync/cond.go:65 +0x30
github.com/nats-io/nats%2ego.(*Conn).processMsg(0xc000702600, 0xc0005a000d, 0x5, 0x7ff3)
	/home/travis/gopath/src/github.com/nats-io/nats.go/nats.go:2498 +0xc7a
github.com/nats-io/nats%2ego.(*Conn).parse(0xc000702600, 0xc0005a0000, 0x398, 0x8000, 0x398, 0x0)
	/home/travis/gopath/src/github.com/nats-io/nats.go/parser.go:192 +0x26e5
github.com/nats-io/nats%2ego.(*Conn).readLoop(0xc000702600)
	/home/travis/gopath/src/github.com/nats-io/nats.go/nats.go:2323 +0x279
created by github.com/nats-io/nats%2ego.(*Conn).processConnectInit
	/home/travis/gopath/src/github.com/nats-io/nats.go/nats.go:1639 +0x32b
FAIL	github.com/nats-io/nats.go/test	42.511s
```

You need to check for `sub.closed` line 2466, after lock the sub. If closed, return.

norace_test.go
```diff
@@ -97,3 +100,77 @@ func TestNoRaceParseStateReconnectFunctionality(t *testing.T) {
...
+	nc.Flush()
+
+	if nm := str.State().Msgs; nm != toSend {
+		t.Fatalf("Expected to have stored all %d msgs, got only %d", toSend, nm)
+	}
+
+	var received uint64
+	done := make(chan bool, 1)
+
+	nc.Subscribe("d", func(m *Msg) {
```
kozlovic, line 142: Get the sub here and call sub.SetPendingLimits(-1,-1).

test/sub_test.go
```diff
@@ -908,10 +918,19 @@ func TestChanSubscriberPendingLimits(t *testing.T) {
 			switch typeSubs {
 			case 0:
 				sub, err = nc.ChanSubscribe("foo", ch)
+				if err := sub.SetPendingLimits(pending, -1); err == nil {
```
kozlovic, line 921: I thought limits did not apply to Channel subscribe, because the limit is really whatever the user sets for the provided channel?

nats.go
kozlovic: By the way, for this test: TestUnsubscribeChanOnSubscriber, I also see the slow consumer error reported in some cases, so may have to set dummy error handler there too.

kozlovic (changes requested):

nats.go
derekcollison: No, will fix.

nats.go
derekcollison: ok will take a deeper look.

derekcollison: ok fixed.

norace_test.go
derekcollison: Yes I know that one but this test is trying to test default behaviors such that we can receive messages in bursts without having to set pending to unlimited.

test/sub_test.go
derekcollison: The test expects an error since its checking for `err == nil` but the t.Fatalf descriptions are wrong, will fix.

kozlovic (approved):
LGTM

ripienaar (approved):

PR #609, opened 2020-11-16
ensure response message goes to the reply subject in RespondMsg

### LINKED ISSUE #606: RespondMsg doesn't have the right context for the response subject
```
// RespondMsg allows a convenient way to respond to requests in service based subscriptions that might include headers
func (m *Msg) RespondMsg(msg *Msg) error {
	if m == nil || m.Sub == nil {
		return ErrMsgNotBound
	}
	if m.Reply == "" {
		return ErrMsgNoReply <---- this check is good
	}
	m.Sub.mu.Lock()
	nc := m.Sub.conn
	m.Sub.mu.Unlock()
	// No need to check the connection here since the call to publish will do all the checking.
	return nc.PublishMsg(msg) <---- msg is not set for any subject or wrong subject
}
```

The correct implementation should be something like:

```
// RespondMsg allows a convenient way to respond to requests in service based subscriptions that might include headers
func (m *Msg) RespondMsg(msg *Msg) error {
	if m == nil || m.Sub == nil {
		return ErrMsgNotBound
	}
	if m.Reply == "" {
		return ErrMsgNoReply 
	}
        msg.Subject = m.Reply
	m.Sub.mu.Lock()
	nc := m.Sub.conn
	m.Sub.mu.Unlock()
	// No need to check the connection here since the call to publish will do all the checking.
	return nc.PublishMsg(msg)
}
```

ripienaar:
Resolves #606 

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -3888,6 +3888,7 @@ func (m *Msg) RespondMsg(msg *Msg) error {
 	if m.Reply == "" {
 		return ErrMsgNoReply
 	}
+	msg.Subject = m.Reply
 	m.Sub.mu.Lock()
 	nc := m.Sub.conn
 	m.Sub.mu.Unlock()
--- a/nats_test.go
+++ b/nats_test.go
@@ -2645,3 +2645,40 @@ func TestJSSubscribe(t *testing.T) {
 		t.Fatalf("subscriber not subscribed to the delivery subject")
 	}
 }
+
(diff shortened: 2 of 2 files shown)


derekcollison (approved):
LGTM

kozlovic (approved):
LGTM

aricart (approved):
LGTM

PR #611, opened 2020-11-28
JetStream API updates

derekcollison:
This is the updated API that we have been discussing that builds off of @ripienaar's work.

Has a JetStream context, matches on subjects. Built for multiple JetStream contexts, etc.
Supports ephemeral and durable consumers as well as push and pull based.
Has the beginning of some management stuff like AddStream and AddConsumer, need more filled in.

```go
nc, err := nats.Connect(s.ClientURL())
js, err := nc.JetStream()

// Create a stream
_, err = js.AddStream(&nats.StreamConfig{Name: "foo"})

// Publish to a stream.
js.Publish("foo", []byte("Hello JS!"))

// Subscribe
sub, err := js.SubscribeSync("foo")
sub, err := js.Subscribe("foo", func(m *nats.Msg) {
	q <- m
})

// ChanSubscribe
msgs := make(chan *nats.Msg, 10)
sub, err := js.ChanSubscribe("foo", msgs, nats.Durable("dlc"), nats.Pull(10))
```

This should be a decent starting point for @wallyqs and @variadico to work from.

I am still playing with some ideas on stalled pull based consumers but wanted this posted. Will keep looking at that issue.


js.go
```diff
@@ -0,0 +1,1175 @@
...
+	// JSApiConsumerInfoT is used to create consumers.
+	JSApiConsumerInfoT = "$JS.API.CONSUMER.INFO.%s.%s"
+	// JSApiRequestNextT is the prefix for the request next message(s) for a consumer in worker/pull mode.
+	JSApiRequestNextT = "$JS.API.CONSUMER.MSG.NEXT.%s.%s"
+	// JSApiStreamCreate is the endpoint to create new streams.
+	JSApiStreamCreateT = "$JS.API.STREAM.CREATE.%s"
+)
+
+// JetStream returns a JetStream context for pub/sub interactions.
+func (nc *Conn) JetStream(optApiPre ...string) (JetStream, error) {
```
ripienaar, line 109: seems we'd want to pass other options in here like default timeout etc, this should probably be a list of functions

js.go
```diff
@@ -0,0 +1,1175 @@
...
+			js.pre = optApiPre[0]
+			if !strings.HasSuffix(js.pre, tokSep) {
+				js.pre = js.pre + tokSep
+			}
+		} else {
+			return nil, ErrJetStreamBadPre
+		}
+	}
+
+	resp, err := nc.Request(js.apiSubj(JSApiAccountInfo), nil, defaultRequestWait)
```
ripienaar, line 124: should be able to skip this for cases where I dont have the full API only some imported subjects

js.go
```diff
@@ -0,0 +1,1175 @@
...
+	return js, nil
+}
+
+func (js *js) apiSubj(subj string) string {
+	if js.pre == _EMPTY_ {
+		return subj
+	}
+	var b strings.Builder
+	b.WriteString(js.pre)
+	b.WriteString(subj)
```
ripienaar, line 147: Is the intention that people would import like `js.derek.$JS.API.....` since thats all this supports.

I'd have thought thats not how one would do it, ie. you export `$JS.API.>` and I import that into `js.derek` leaving subjects like `js.derek.INFO`

js.go
```diff
@@ -0,0 +1,1175 @@
...
+	return func(opts *subOpts) error {
+		if batchSize == 0 {
+			return errors.New("nats: batch size of 0 not valid")
+		}
+		opts.pull = batchSize
+		return nil
+	}
+}
+
+func ManualAck() SubOpt {
```
ripienaar, line 582: if not in manual  mode what happens if I ack it anyway? I dont see special handling for that here

js.go
```diff
@@ -0,0 +1,1175 @@
...
+			start = i + 1
+		}
+	}
+	tokens = append(tokens, subject[start:])
+	if len(tokens) != expectedTokens || tokens[0] != "$JS" || tokens[1] != "ACK" {
+		return nil, ErrNotJSMessage
+	}
+
+	meta := &MsgMetaData{
+		Delivered: uint64(parseNum(tokens[4])),
```
ripienaar, line 744: parseNum can return -1 which will panic when cast to uint64

ripienaar (changes requested):
generally good,I assume this is going to someone else to flesh out andclean up etc so not too intense review just some comments

js.go
derekcollison: Yes I agree with you, was trying to set the stage quickly for multiple contexts but server needs a few more minor mods for it to fully work.

js.go
derekcollison: I have this on my list  to think through, essentially attaching to consumers etc. An idea could be for the above JetStream() context options. We could have AttachOnly() as an option that would skip that.

js.go
derekcollison: That would involve full on mapping vs just prefix prepend. But I think we want to match what folks would want to do here from the application space.

js.go
derekcollison: Server does not care, I could try to detect and suppress in client but for now thought it was fine as is.

js.go
derekcollison: Actually just makes it a big number.

https://play.golang.org/p/Rxyg7lYmDv6

ripienaar: Hmm swear I did the same play and it panicked lol. Don’t work in tbe same room as a 3 year old :P

derekcollison: lol no worries, I knew it was ok b/c I hit that issue in JetStream's filestore ;)

ColinSullivan1 (commented):
Looks good, very straightforward.  The Poll() API looks clean.  No other issues/comments at the moment.

js.go
derekcollison: Actually I misspoke, that is possible. Will make that change and the func options for the JetStream context itself.

js.go
derekcollison: ok made this change and squashed.

js.go
derekcollison: ok this should work and the import test works now properly.

js.go
derekcollison: ok I gave this one a first pass as a separate change below.

derekcollison:
I am going to squash these all down to one commit.

js.go
```diff
@@ -0,0 +1,1246 @@
...
+	sub.jsi = &jsSub{js: js}
+
+	// If we are creating or updating let's process that request.
+	if shouldCreate {
+		// If not set default to ack explicit.
+		if cfg.AckPolicy == ackPolicyNotSet {
+			cfg.AckPolicy = AckExplicit
+		}
+		// If we have acks at all and the MaxAckPending is not set go ahead
+		// and set to the internal max.
```
ColinSullivan1, line 507: In NATS streaming IIRC we found better performance and stability with a much lower number - e.g. 32 or 64.  In java, the default pending limit is 64k - imo that'd put a lot of memory pressure on the client.  /cc @kozlovic

kozlovic: Right. The default in NATS Streaming is 1024. Having this number high will cause the server to load/resend a lot of messages when not ack'ed in a timely fashion, or when the server has not yet detected the subject interest is gone (right now it is local, so not an issue, but may become with clustering in place).

js.go
```diff
@@ -0,0 +1,1246 @@
...
+	ExpectedLastMsgIdHdr = "Nats-Expected-Last-Msg-Id"
+)
+
+func (js *js) PublishMsg(m *Msg, opts ...PubOpt) (*PubAck, error) {
+	var o pubOpts
+	if len(opts) > 0 {
+		if m.Header == nil {
+			m.Header = http.Header{}
+		}
+		for _, f := range opts {
```
kozlovic, line 218: Have we evaluated the performance effect of variadic publish options as opposed to passing a pointer to a structure? It may be fine for a connect, or subscribe, but I wonder if this is a good approach for publish due to the fact that it can be called many times a second ;-).

js.go
derekcollison: We have two limits, one for number of messages and one for byte size pending which is still applicable.

js.go
derekcollison: Publish is an RPC, none of that matters TBH. I care about costs of processing in the server though for sure.

wallyqs (approved):
LGTM

js.go
```diff
@@ -0,0 +1,1246 @@
...
+	start, tokens := 0, tsa[:0]
+	subject := m.Reply
+	for i := 0; i < len(subject); i++ {
+		if subject[i] == btsep {
+			tokens = append(tokens, subject[start:i])
+			start = i + 1
+		}
+	}
+	tokens = append(tokens, subject[start:])
+	if len(tokens) != expectedTokens || tokens[0] != "$JS" || tokens[1] != "ACK" {
```
ColinSullivan1, line 809: Can the response be prefixed?

PR #612, opened 2020-12-03
jetstream: Fix AckSync test using import

wallyqs:
Fixes imported stream and adds export to test ack of ack when using consumers with direct mode/no api access.


derekcollison (approved):
LGTM

PR #613, opened 2020-12-03
jetstream: Updates for JetStream DirectOnly support

wallyqs:
In case of using JetStream without API access, when calling subscribe  it is needed to use PushDirect or PullDirect options.

In this PR, we add support so that instead of using explicit options on subscribe, the subject can be used instead.  

For example, given a set of imports like:

```hcl
# Push based
{ stream:  { subject: "p.d", account: JS } }

# Pull based
{ service: { subject: "$JS.API.CONSUMER.MSG.NEXT.ORDERS.dA1", account: JS }, to: "next.order" }
```

We can now create the subscriptions as follows:

```go
js, err := nc.JetStream(nats.DirectOnly())

// Push based
js.SubscribeSync("p.d")
// instead of: js.SubscribeSync("ORDERS", nats.PushDirect("p.d"))

// Pull based to subject
js.SubscribeSync("next.order", nats.Pull(batch))

// Pull using attach
sub, err = js.SubscribeSync("ORDERS", nats.Attach("ORDERS", "d1"), nats.Pull(batch))
// 
// can replace => js.SubscribeSync("ORDERS", nats.PullDirect("ORDERS", "d1", batch))
```

This also removes the PushDirect and PullDirect options since just calling subscribe would try to do the right thing.

### DIFF
--- a/js.go
+++ b/js.go
@@ -398,7 +398,7 @@ func (js *js) QueueSubscribe(subj, queue string, cb MsgHandler, opts ...SubOpt) 
 	return js.subscribe(subj, queue, cb, nil, opts)
 }
 
-// Subscribe will create a subscription to the appropriate stream and consumer.
+// ChanSubscribe will create a subscription to the appropriate stream and consumer using the given channel.
 func (js *js) ChanSubscribe(subj string, ch chan *Msg, opts ...SubOpt) (*Subscription, error) {
 	return js.subscribe(subj, _EMPTY_, nil, ch, opts)
 }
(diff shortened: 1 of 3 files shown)


derekcollison:
`next.order` is not the subject though to match on the stream, is the subject to get the next message from the system.

So if stream ORDERS has subjects `foo` and `bar`, those subjects are what we need in the subscribe, not the next message subject.

Also the comments show both push and pull but I believe both are pull, and attach needs to access the API to know the delivery subject for a push based consumer unless supplied directly.

kozlovic:
Non related to this PR, but as seen with data races in the PR travis execution, the way the subscription is created is wrong:
https://github.com/nats-io/nats.go/pull/613/files#diff-176a2342c55218b70f165bed6a7731dc83d901f9513382f3c98bbd3d08ce0678L494-L498

We can't create the low-level NATS subscription and then modify the subscription object. As soon as nc.subscribe() returns, then there is a chance that a message is delivered to this subscription. Even locking sub after this call to assign the jsi, etc. is wrong because there is still a window where the callback will be invoked while sub.jsi is not yet set, which means that the message may be treated as a non JS message (while it is), etc..

derekcollison:
I will fix that, you are right.

PR #614, opened 2020-12-04
Assign jsSub during low level subscription

derekcollison:
Ivan pointed out that setting the jsSub portion of the low level sub after we called subscribe could cause some issues.


kozlovic (approved):
LGTM

PR #615, opened 2020-12-04
[ADDED] tls in example clients

egodigitus:
Added 3 new optional flags to the pub & sub examples. In order to easily test a client connection with a tls enabled nats-server:
- tlscert <file>
- tlskey <file>
- tlsrootca <file>

### DIFF
--- a/examples/nats-pub/main.go
+++ b/examples/nats-pub/main.go
@@ -1,4 +1,4 @@
-// Copyright 2012-2019 The NATS Authors
+// Copyright 2012-2020 The NATS Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
--- a/examples/nats-pub/main.go
+++ b/examples/nats-pub/main.go
@@ -26,7 +26,7 @@ import (
 // nats-pub -s demo.nats.io:4443 <subject> <msg> (TLS version)
 
 func usage() {
-	log.Printf("Usage: nats-pub [-s server] [-creds file] [-tlscert file] [-tlskey file] [-tlsrootca file] <subject> <msg>\n")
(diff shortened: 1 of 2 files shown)


examples/nats-pub/main.go
```diff
@@ -38,8 +38,11 @@ func showUsageAndExit(exitcode int) {
 func main() {
 	var urls = flag.String("s", nats.DefaultURL, "The nats server URLs (separated by comma)")
 	var userCreds = flag.String("creds", "", "User Credentials File")
-	var showHelp = flag.Bool("h", false, "Show help message")
+	var tlsClientCert = flag.String("tlscert", "", "TLS client certificate file")
+	var tlsClientKey = flag.String("tlskey", "", "Private key file for client certificate")
+	var tlsRootCA = flag.String("tlsrootca", "", "Root CA file for server verification")
```
wallyqs, line 43: How about this one being `cacert` so that it is more like `curl`?

```
     --cacert <file> CA certificate to verify peer against
```

egodigitus: Fine with me. Was trying to be as close to the existing naming conventions as possible. A client user would potentially stumble upon this doc while setting up tls: https://docs.nats.io/developing-with-nats/security/tls
There it's called RootCA, just like the function handling the cert

wallyqs: Thanks for the link, I think you are right it should be like the options from the server where the options are:

```
> nats-server --tls --tlscert=server-cert.pem --tlskey=server-key.pem --tlscacert rootCA.pem --tlsverify
```

So using `--tlscacert` would be bettter in this case I think.

egodigitus: Aligned the naming to the nats-server flags.
tlsrootca is now called tlscacert
The two nats-server flags "--tls" & "--tlsverify" are rather server specific and are therefore omitted.

wallyqs (approved):
LGTM, thanks!

PR #617, opened 2020-12-07
Validate setting JetStream ApiPrefix

variadico:
This adds a check to `nats.ApiPrefix` that prevents the prefix from containing wildcards. When @wallyqs and I were exploring the code, we found that we could end up with invalid API subjects like `foo.>.STREAM.NAMES`.

Also, this removes what appears to be outdated prefix TODOs.

### DIFF
--- a/js.go
+++ b/js.go
@@ -183,6 +183,22 @@ func APIPrefix(pre string) JSOpt {
 		if !strings.HasSuffix(js.pre, ".") {
 			js.pre = js.pre + "."
 		}
+
+		toks := strings.Split(pre, ".")
+		for i, tok := range toks {
+			if tok == "" || tok == "*" {
+				return ErrJetStreamBadPre
+			}
+
+			if tok == ">" && i < len(toks)-1 {
+				return ErrBadSubject
+			} else if tok == ">" && i == len(toks)-1 {
+				js.pre = pre
+				return nil
+			}
+		}
+
+		js.pre = fmt.Sprintf("%s.", pre)
 		return nil
 	})
 }
--- a/nats_test.go
+++ b/nats_test.go
@@ -2599,3 +2599,61 @@ func TestMsg_RespondMsg(t *testing.T) {
 		t.Fatalf("did not get correct response: %q", resp.Data)
 	}
 }
+
+func TestJetStreamAPIPrefix(t *testing.T) {
+	cases := []struct {
+		prefix     string
+		wantPrefix string
+		wantErr    bool
+	}{
+		{
+			prefix:     "foo",
+			wantPrefix: "foo.",
+		},
+		{
+			prefix:     "foo.",
+			wantPrefix: "foo.",
+		},
+		{
+			prefix:     "foo.>",
+			wantPrefix: "foo.>",
+		},
+		{
+			prefix:     "foo.b*r.baz",
+			wantPrefix: "foo.b*r.baz.",
+		},
+		{
+			prefix:  "foo.*",
+			wantErr: true,
+		},
+		{
+			prefix:  "foo.*.bar",
+			wantErr: true,
+		},
+		{
+			prefix:  "foo.>.bar",
+			wantErr: true,
+		},
+		{
+			prefix:  ">",
+			wantErr: true,
+		},
+	}
+	for _, c := range cases {
+		t.Run(c.prefix, func(t *testing.T) {
+			jsOpt := APIPrefix(c.prefix)
+
+			js := new(js)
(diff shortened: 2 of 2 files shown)


js.go
```diff
@@ -150,6 +150,14 @@ func ApiPrefix(pre string) JSOpt {
 		if !strings.HasSuffix(js.pre, ".") {
 			js.pre = js.pre + "."
 		}
+
+		for _, ch := range js.pre {
```
derekcollison, line 154: Need to tokenize first. So do `strings.Split(js.pre, ".")` and check against the tokens. Although not common this is technically legal..

`foo.*>*'

derekcollison (changes requested):

js.go
variadico: Ah, okay. Just pushed some more tests and changed the validation logic a bit.

Added these cases:
```js
{
	prefix:     "foo",
	wantPrefix: "foo.",
},
{
	prefix:     "foo.",
	wantPrefix: "foo.",
},
{
	prefix:     "foo.>",
	wantPrefix: "foo.>",
},
{
	prefix:  "foo.*",
	wantErr: true,
},
{
	prefix:  "foo.>.bar",
	wantErr: true,
},
```
https://github.com/nats-io/nats.go/pull/617/commits/b2987b5e47db95f4631943d12446360d9f2271be#diff-6277b08b1b6714cf4cd8c16c727011ff2c99d3a8d1b41a0d3fc2d81a49b1f3f5R2609

nats_test.go
```diff
@@ -2599,3 +2599,49 @@ func TestMsg_RespondMsg(t *testing.T) {
...
+			wantPrefix: "foo.>",
+		},
+		{
+			prefix:  "foo.*",
+			wantErr: true,
+		},
+		{
+			prefix:  "foo.>.bar",
+			wantErr: true,
+		},
```
wallyqs, line 2628: Also an error: 
```
		{
			prefix:  ">",
			wantErr: true,
		},
```

variadico: Good catch. :+1:

nats_test.go
```diff
@@ -2599,3 +2599,53 @@ func TestMsg_RespondMsg(t *testing.T) {
 		t.Fatalf("did not get correct response: %q", resp.Data)
 	}
 }
+
+func TestApiPrefix(t *testing.T) {
```
wallyqs, line 2603:
```suggestion
func TestJetStreamApiPrefix(t *testing.T) {
```

wallyqs (approved):
LGTM

js.go
```diff
@@ -146,10 +146,30 @@ type JSOpt func(opts *js) error
 
 func ApiPrefix(pre string) JSOpt {
 	return func(js *js) error {
-		js.pre = pre
-		if !strings.HasSuffix(js.pre, ".") {
-			js.pre = js.pre + "."
+		if strings.ContainsAny(pre, " \t\r\n*") {
```
derekcollison, line 149: So this prevent a subject like `foo.b*r.baz`?

variadico: Yeah. I added that case now.
So we prevent these:
```go
{
	prefix:  "foo.*",
	wantErr: true,
},
{
	prefix:  "foo.b*r.baz",
	wantErr: true,
},
```

js.go
```diff
@@ -165,10 +165,27 @@ func (opt jsOptFn) configureJSContext(opts *js) error {
...
+		toks := strings.Split(pre, ".")
+		for i, tok := range toks {
+			if tok == "" || tok == "*" {
+				return ErrJetStreamBadPre
+			}
+
+			if tok == ">" && i < len(toks)-1 {
+				return ErrBadSubject
+			} else if tok == ">" && i == len(toks)-1 {
+				js.pre = pre
```
kozlovic, line 183: I am confused. If it is a prefix, why are we allowing something like `foo.>` here? being a prefix it means that it would then possibly added to something and then become `foo.>.bar`? But maybe I don't understand what that is.

nats_test.go
```diff
@@ -2599,3 +2599,61 @@ func TestMsg_RespondMsg(t *testing.T) {
...
+			prefix:     "foo",
+			wantPrefix: "foo.",
+		},
+		{
+			prefix:     "foo.",
+			wantPrefix: "foo.",
+		},
+		{
+			prefix:     "foo.>",
+			wantPrefix: "foo.>",
```
kozlovic, line 2619: This is strange that we allow this to be a "prefix".

variadico:
Closing for now. Will revisit.

PR #618, opened 2020-12-08
Trying to ack a non JS msg would panic and hang

derekcollison:

### DIFF
--- a/js.go
+++ b/js.go
@@ -718,6 +718,10 @@ func (m *Msg) checkReply() (*js, bool, error) {
 	}
 	sub := m.Sub
 	sub.mu.Lock()
+	if sub.jsi == nil {
+		sub.mu.Unlock()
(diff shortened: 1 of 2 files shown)


wallyqs (approved):
LGTM

js.go
```diff
@@ -718,6 +718,10 @@ func (m *Msg) checkReply() (*js, bool, error) {
 	}
 	sub := m.Sub
 	sub.mu.Lock()
+	if sub.jsi == nil {
+		sub.mu.Unlock()
+		return nil, false, ErrNotJSMessage
+	}
```
kozlovic, line 724: Let's remove the check for `js == nil` down there..

kozlovic (changes requested):

js.go
derekcollison: ok just saw this and already merged but will do so.

PR #621, opened 2020-12-10
Separate JetStream into smaller interfaces

wallyqs:
This separates the JetStream interface into two smaller interfaces:

A `JetStream` interface used for producing and consuming messages, and a `JetStreamManager` interface for creating streams/consumers. This also adds a new interface that is the compound of both called JetStreamContext and is the one that is being returned when calling `nc.JetStream()`.

This change allows to opt-in to the behaviors provided by the `JetStreamContext` as needed, for example:

```go
// Can be used to both create streams and publish
js, err := nc.JetStream()

// Can be used to produce/consume messages, but not for creating streams
var js nats.JetStream
js, err = nc.JetStream()

// Can be used for managing streams/consumers
var jsm nats.JetStreamManager
js, err = nc.JetStream()
```

### DIFF
--- a/js.go
+++ b/js.go
@@ -25,7 +25,7 @@ import (
 	"time"
 )
 
-// JetStream is the public interface for the JetStream context.
+// JetStream is the public interface for JetStream.
 type JetStream interface {
 	// Publishing messages to JetStream.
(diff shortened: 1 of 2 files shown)


kozlovic (approved):
LGTM

PR #622, opened 2020-12-11
js: Fix a few golint errors

wallyqs:
Fixes a few golint errors related to names:

```
js.go:54:6: type ApiError should be APIError
js.go:60:6: type ApiResponse should be APIResponse
js.go:99:2: const JSDefaultApiPrefix should be JSDefaultAPIPrefix
js.go:154:6: func ApiPrefix should be APIPrefix
js.go:164:6: func ApiRequestWait should be APIRequestWait
js.go:414:6: type ApiPaged should be APIPaged
```

### DIFF
--- a/js.go
+++ b/js.go
@@ -50,20 +50,20 @@ type JetStream interface {
 	// TODO(dlc) - add more
 }
 
-// ApiError is included in all API responses if there was an error.
-type ApiError struct {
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM - We should do server at some point too.

PR #625, opened 2020-12-12
js: Treat JS functional options as interfaces

wallyqs:
This changes some of the internals of how functional options are used so that it gives more flexibility when growing the number of options in the client.

The API stays the same but instead of options being concrete types, `PubOpt`, `SubOpt` and `JSOpt` are interfaces so that the behavior changes depending on where an option is used.     

This has a few benefits I think, for example `nats.MaxWait` currently looks like this:

```go
func MaxWait(ttl time.Duration) PubOpt { ... }
```

it can only be applied to publishing, and we also have a similar option for waiting for responses from the API:

```go
func APIRequestWait(wait time.Duration) JSOpt { ... }
```

These are both very similar options about setting a timeout, so with the changes in this PR we can now do:

```go
// MaxWait for responses against API in general including (Publish)
js, _ := nc.JetStream(nats.MaxWait(4*time.Second))

// Custom timeout for a single publish call
ack, _ := js.Publish("baz", msg, nats.MaxWait(2*time.Second))
```

Also, this allows `nats.MaxWait` to itself just be a `time.Duration`, so the same option can be converted into other types easily if needed:

```go
ttl := nats.MaxWait(2*time.Second)
nc.Request("hello", []byte("regular request"), time.Duration(ttl))
_, err = js.Publish("foo", msg, ttl)
```

Another place where this can be helpful is to give more flexibility in how to apply a `context.Context`. Currently we have this option which can only be used for a Publish:

```go
func Context(ctx context.Context) PubOpt { ... }
```

One possible problem with this is that it takes over the `Context` name from the `nats` package so in case we want to set a cancellation context for Subscribe or the JetStream context we will need extra options like `SubscribeContext`.

In this PR, instead we make `nats.Context` to be a type which is the same type as `context.Context`, but that can also be used as an option since it implements the `PubOpt` interface.

The API does not change so this will still work:

```go
js.Publish("baz", msg, nats.Context(ctx))
```

But `nats.Context` is just a type so it can be converted into a regular context again and be reused:

```go
natsCtx := nats.Context(ctx)
ctx2, cancel2 := context.WithTimeout(natsCtx, 2*time.Second)
defer cancel2()
_, err = js.Publish("foo", msg, ttl, natsCtx)

// Can be reused in other apis again if needed.
sub.NextMsgWithContext(ctx2)
```

This also leaves the door open to use `nats.Context` to be able to become a functional option to be used for subscriptions or passed to the JetStream context as well.

### DIFF
--- a/js.go
+++ b/js.go
@@ -121,8 +121,8 @@ func (nc *Conn) JetStream(opts ...JSOpt) (JetStream, error) {
 
 	js := &js{nc: nc, pre: JSDefaultAPIPrefix, wait: defaultRequestWait}
 
-	for _, f := range opts {
-		if err := f(js); err != nil {
+	for _, opt := range opts {
+		if err := opt.configureJSContext(js); err != nil {
 			return nil, err
 		}
 	}
@@ -148,31 +148,33 @@ func (nc *Conn) JetStream(opts ...JSOpt) (JetStream, error) {
 	return js, nil
 }
 
-// JSOpt configures options for the jetstream context.
-type JSOpt func(opts *js) error
+// JSOpt configures a JetStream context.
+type JSOpt interface {
+	configureJSContext(opts *js) error
+}
+
+// jsOptFn configures an option for the JetStream context.
+type jsOptFn func(opts *js) error
+
+func (opt jsOptFn) configureJSContext(opts *js) error {
+	return opt(opts)
+}
 
 func APIPrefix(pre string) JSOpt {
-	return func(js *js) error {
+	return jsOptFn(func(js *js) error {
 		js.pre = pre
 		if !strings.HasSuffix(js.pre, ".") {
 			js.pre = js.pre + "."
 		}
 		return nil
-	}
-}
-
-func APIRequestWait(wait time.Duration) JSOpt {
-	return func(js *js) error {
-		js.wait = wait
-		return nil
-	}
+	})
 }
 
 func DirectOnly() JSOpt {
-	return func(js *js) error {
+	return jsOptFn(func(js *js) error {
 		js.direct = true
 		return nil
-	}
+	})
 }
 
 func (js *js) apiSubj(subj string) string {
@@ -185,8 +187,17 @@ func (js *js) apiSubj(subj string) string {
 	return b.String()
 }
 
-// PubOpt configures options for publishing jetstream messages.
-type PubOpt func(opts *pubOpts) error
+// PubOpt configures options for publishing JetStream messages.
+type PubOpt interface {
+	configurePublish(opts *pubOpts) error
+}
+
+// pubOptFn is a function option used to configure JetStream Publish.
+type pubOptFn func(opts *pubOpts) error
+
+func (opt pubOptFn) configurePublish(opts *pubOpts) error {
+	return opt(opts)
+}
 
 type pubOpts struct {
 	ctx context.Context
@@ -222,8 +233,8 @@ func (js *js) PublishMsg(m *Msg, opts ...PubOpt) (*PubAck, error) {
 		if m.Header == nil {
 			m.Header = http.Header{}
 		}
-		for _, f := range opts {
-			if err := f(&o); err != nil {
+		for _, opt := range opts {
+			if err := opt.configurePublish(&o); err != nil {
 				return nil, err
 			}
 		}
@@ -253,7 +264,7 @@ func (js *js) PublishMsg(m *Msg, opts ...PubOpt) (*PubAck, error) {
 	var err error
 
 	if o.ttl > 0 {
-		resp, err = js.nc.RequestMsg(m, o.ttl)
+		resp, err = js.nc.RequestMsg(m, time.Duration(o.ttl))
 	} else {
 		resp, err = js.nc.RequestMsgWithContext(o.ctx, m)
 	}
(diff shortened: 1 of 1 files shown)


js.go
```diff
@@ -148,31 +148,33 @@ func (nc *Conn) JetStream(opts ...JSOpt) (JetStream, error) {
 	return js, nil
 }
 
-// JSOpt configures options for the jetstream context.
-type JSOpt func(opts *js) error
+// JSOpt configures a JetStream context.
+type JSOpt interface {
+	configureJSContext(opts *js) error
```
derekcollison, line 153: Does this mean folks can't write their own since it private?

js.go
```diff
@@ -148,31 +148,33 @@ func (nc *Conn) JetStream(opts ...JSOpt) (JetStream, error) {
...
+// jsOptFn configures an option for the JetStream context.
+type jsOptFn func(opts *js) error
+
+func (opt jsOptFn) configureJSContext(opts *js) error {
+	return opt(opts)
+}
 
 func APIPrefix(pre string) JSOpt {
-	return func(js *js) error {
+	return jsOptFn(func(js *js) error {
```
derekcollison, line 164: Again does this limit it to just us or direct PRs to this codebase?

derekcollison (commented):
I think this is a great direction, I really like some of the ideas here.

I made a few comments and will defer to @kozlovic for final sign off.

js.go
wallyqs: That's right, only direct PRs to this package would be able to create functions that can be applied as an option.

kozlovic: As you know, the way we currently the function options, if an option is missing the user can create a function in their own application and would not be blocked waiting on us to provide it (with a new library release). Being OSS, the user can always have a local copy/fork, but with go.mod that becomes a bit more tricky.

So we need to really evaluate the benefit of this approach vs the "risk" it can cause (users blocked and us having to quickly issue a release with the new option). That is a low risk if the library is stable (we have all the options we need), but not sure if that will be the case at this stage with JS.

I like the approach that something becomes generic, such as "max wait", but if we end-up with only that option that benefits from this change, it may not be worth it.

For the record: I am personally not a fan of the "options as functions" approach, I prefer the simple approach of passing a reference to a option struct (and library cloning or making copies of what is needed to not be affected by changes done to the options after the call). But it seems that this is the direction we have taken on several projects, so that's that.

wallyqs: Currently all the options for JetStream are private (`pubOpts`, `subOpts`,`*js`), but with the approach in this PR and by making a public options type for example, we actually can have that kind of usage of using a single struct to define all the settings at once.

For example in this [commit](https://github.com/wallyqs/nats.go/commit/8ddd21f183c2df3597957ff43ccdfa26d478a389), I have made a `PubOptions` type that it itself also implements the `PubOpt` interface so it can be used to configure a publish as well, so in that branch all of these would work:

```go
// Publish with default options
js.Publish("baz", msg)

// Publish with custom timeout and expected stream using single options struct.
js.Publish("baz", msg, &nats.PubOptions{MaxWait: 2*time.Second, Stream: "FOO"})

// Publish with custom timeout and expected stream
js.Publish("baz", msg, nats.MaxWait(2*time.Second), nats.ExpectStream("FOO"))
```

In that commit, I made another type `PubOptions` that is basically the same as `pubOptions` which is private, but with longer names and have it as public, but they could be refactored into being a single type later since all the changes are internal to the library.

Also the `PubOpt` interface methods are private in this PR, and defined as:

```go
type PubOpt interface {
	configurePublish(opts *pubOpts) error
}
```

so what users cannot do is to create structs that implement the `PubOpt` interface.  To increase the flexibility, what we can do is instead have the interface be defined in terms of the public type like:

```go
type PubOpt interface {
	configurePublish(opts *PubOptions) error
}
```

Then, if we want users to be able to create their own functional options we can make the following type public in the `nats` package:

```go
// PubOptFn is a function option used to configure a JetStream Publish.
type PubOptFn func(opts *PubOptions) error

func (opt PubOptFn) configurePublish(opts *PubOptions) error {
	return opt(opts)
}
```

With that approach users would be able to create their own functional options  if needed, without having to make `configurePublish` public in the `PubOpt` interface (sample implementation of this approach in this [commit](https://github.com/wallyqs/nats.go/commit/e9e14944baf9397babab96826802cc4e3ff83db2))

```go
	withNameAndTimeout := func(name string, ttl time.Duration) nats.PubOpt {
		return nats.PubOptFn(func(opts *nats.PubOptions) error {
			t.Logf("Configuring my own JS NATS publish...")
			opts.Stream = name
			opts.MaxWait = ttl
			return nil
		})
	}
	if _, err = js.Publish("foo", msg, withNameAndTimeout("TEST", 2*time.Second)); err != nil {
		t.Fatalf("Unexpected publish error: %v", err)
	}
```

js.go
wallyqs: Since `configureJSContext(opts *js)` is private then other packages would not be able to implement the their own `JSOpt`.  To open it up, first we need to change add a public type for the options of `*js` (for example, `*JSOptions`), then  we could change this to be defined as:

```go
configureJSContext(opts *JSOptions) error
```

and we can also do something similar for `*subOpts` and have a public `*SubOptions` type.  Though since we control the interface within the same package as private and others cannot implement, we can also defer this to be done later without breaking backwards compatibility if we want and "unseal" the interface, then change it into `ConfigureJSContext(opts *JSOptions)` so that other packages can implement an option.

kozlovic (approved):
LGTM. I trust that you have put way more thought on this than I have, and that you understood the "concern" about being able to cover options that user may need that are not be available in the library (which should not be the case).

wallyqs:
Thanks everyone for the review! I think this will increase some of the flexibility with the names in the client and we still can open up some of the types to let users configure as they want as a separate change.

PR #626, opened 2020-12-16
Add more Stream and Consumer management APIs

variadico:
This adds the follow APIs for Streams.
* UpdateStream
* DeleteStream
* SnapshotStream
* RestoreStream
* PurgeStream
* StreamLister

This adds the follow APIs for Consumers.
* ConsumerInfo
* ConsumerLister
* DeleteConsumer


js.go
```diff
@@ -1155,6 +1272,222 @@ type StreamState struct {
...
+}
+
+func (js *js) SnapshotStream(name string, cfg *StreamSnapshotConfig) (io.Reader, error) {
+	req, err := json.Marshal(cfg)
+	if err != nil {
+		return nil, err
+	}
+
+	ssSubj := js.apiSubj(fmt.Sprintf(JSApiStreamSnapshotT, name))
+	r, err := js.nc.Request(ssSubj, req, time.Second)
```
wallyqs, line 1356:
```suggestion
	r, err := js.nc.Request(ssSubj, req, js.wait)
```

js.go
```diff
@@ -1155,6 +1272,222 @@ type StreamState struct {
...
+// JSApiStreamRestoreResponse is the direct response to the restore request.
+type JSApiStreamRestoreResponse struct {
+	APIResponse
+	// Subject to deliver the chunks to for the snapshot restore.
+	DeliverSubject string `json:"deliver_subject"`
+}
+
+func (js *js) RestoreStream(name string, snapshot io.Reader) error {
+	rsSubj := js.apiSubj(fmt.Sprintf(JSApiStreamRestoreT, name))
+	r, err := js.nc.Request(rsSubj, nil, time.Second)
```
wallyqs, line 1405:
```suggestion
		r, err := js.nc.Request(rsSubj, nil, js.wait)
```

js.go
```diff
@@ -1155,6 +1272,222 @@ type StreamState struct {
...
+// JSApiStreamPurgeResponse.
+type JSApiStreamPurgeResponse struct {
+	APIResponse
+	Success bool   `json:"success,omitempty"`
+	Purged  uint64 `json:"purged"`
+}
+
+func (js *js) PurgeStream(name string) error {
+	psSubj := js.apiSubj(fmt.Sprintf(JSApiStreamPurgeT, name))
+	r, err := js.nc.Request(psSubj, nil, time.Second)
```
wallyqs, line 1439:
```suggestion
	r, err := js.nc.Request(psSubj, nil, js.wait)
```

js.go
```diff
@@ -1155,6 +1272,222 @@ type StreamState struct {
...
+func (js *js) ListStreams(offset int) ([]*StreamInfo, error) {
+	req, err := json.Marshal(JSApiStreamNamesRequest{
+		ApiPagedRequest: ApiPagedRequest{Offset: offset},
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	slSubj := js.apiSubj(JSApiStreamList)
+	r, err := js.nc.Request(slSubj, req, time.Second)
```
wallyqs, line 1476:
```suggestion
	r, err := js.nc.Request(slSubj, req, js.wait)
```

js.go
```diff
@@ -44,10 +45,29 @@ type JetStream interface {
...
+	PurgeStream(name string) error
+	// List streams.
+	ListStreams(offset int) ([]*StreamInfo, error)
+
+	// Create a consumer.
+	AddConsumer(stream string, cfg *ConsumerConfig) (*ConsumerInfo, error)
+	// Delete a consumer.
+	DeleteConsumer(stream, consumer string) error
+	// Consumer information.
+	ConsumerInfo(stream, durable string) (*ConsumerInfo, error)
```
derekcollison, line 68: This also comes off a subscription. `sub.ConsumerInfo()` which is how I use it most.

Also does not have to be durable, so just `name`

js.go
derekcollison: That's cool did not know you could suggest changes!

js.go
```diff
@@ -1068,6 +1106,85 @@ func (js *js) AddConsumer(stream string, cfg *ConsumerConfig) (*ConsumerInfo, er
...
+}
+
+// JSApiConsumerListResponse.
+type JSApiConsumerListResponse struct {
+	APIResponse
+	ApiPaged
+	Consumers []*ConsumerInfo `json:"consumers"`
+}
+
+func (js *js) ListConsumers(stream string, offset int) ([]*ConsumerInfo, error) {
```
derekcollison, line 1162: Be good if we could hide the paging at this layer or undo it below the covers and just make it a stream of messages with EOF.

variadico: All right. I added a `StreamLister` and `ConsumerLister` to iterate through pages.

This is what the calling code looks like.
```go
sl := js.NewStreamLister()
for sl.Next() {
	// sl.Page() returns a []*StreamInfo.
	// Each iteration can return a maximum of 256 items.
	fmt.Println(sl.Page())
}
if err := sl.Err(); err != nil {
	return err
}
```

Here's what it looks like in tests.
https://github.com/nats-io/nats.go/pull/626/files#diff-9e6bdee4051a6e36d6d507dcef63b5f9a39b281e421c08f02ec74a0fb6639e3aR592-R626

js.go
```diff
@@ -1127,6 +1162,121 @@ func (js *js) AddConsumer(stream string, cfg *ConsumerConfig) (*ConsumerInfo, er
...
+		return errors.New(resp.Error.Description)
+	}
+	return nil
+}
+
+func (js *js) ConsumerInfo(stream, durable string) (*ConsumerInfo, error) {
+	return js.getConsumerInfo(stream, durable)
+}
+
+// APIPagedRequest includes parameters allowing specific pages to be requests from APIs responding with APIPaged
```
kozlovic, line 1195: typo: `to be requested`?

js.go
```diff
@@ -1127,6 +1162,121 @@ func (js *js) AddConsumer(stream string, cfg *ConsumerConfig) (*ConsumerInfo, er
...
+// Next fetches the next ConsumerInfo page.
+func (c *ConsumerLister) Next() bool {
+	if c.err != nil {
+		return false
+	}
+	if c.pageInfo != nil && c.offset >= c.pageInfo.Total {
+		return false
+	}
+
+	if c.stream == _EMPTY_ {
```
kozlovic, line 1231: Should this be moved before the pageInfo test?

js.go
```diff
@@ -1127,6 +1162,121 @@ func (js *js) AddConsumer(stream string, cfg *ConsumerConfig) (*ConsumerInfo, er
...
+
+// JSApiConsumerListResponse.
+type JSApiConsumerListResponse struct {
+	APIResponse
+	APIPaged
+	Consumers []*ConsumerInfo `json:"consumers"`
+}
+
+// ConsumerLister fetches pages of ConsumerInfo objects.
+type ConsumerLister struct {
```
kozlovic, line 1212: Looking at the rest of the code, it seem that we don't have any lock here, which is fine, but should we then mention that this is not "thread safe"?

js.go
```diff
@@ -1127,6 +1162,121 @@ func (js *js) AddConsumer(stream string, cfg *ConsumerConfig) (*ConsumerInfo, er
 	return info.ConsumerInfo, nil
 }
 
+// JSApiConsumerDeleteResponse.
+type JSApiConsumerDeleteResponse struct {
+	APIResponse
+	Success bool `json:"success,omitempty"`
+}
+
+func (js *js) DeleteConsumer(stream, durable string) error {
```
kozlovic, line 1171: There is no "doc" of an exported function. Linters used to complain about this, not sure that is still the case. You could copy back what you have defined in the interface.

js.go
```diff
@@ -1214,6 +1364,262 @@ type StreamState struct {
...
+// JSApiStreamSnapshotResponse is the direct response to the snapshot request.
+type JSApiStreamSnapshotResponse struct {
+	APIResponse
+	// Estimate of number of blocks for the messages.
+	NumBlks int `json:"num_blks"`
+	// Block size limit as specified by the stream.
+	BlkSize int `json:"blk_size"`
+}
+
+func (js *js) SnapshotStream(name string, cfg *StreamSnapshotConfig) (io.Reader, error) {
```
kozlovic, line 1441: Doc?

js.go
```diff
@@ -1214,6 +1364,262 @@ type StreamState struct {
...
+	if err := json.Unmarshal(r.Data, &resp); err != nil {
+		return nil, err
+	}
+	if resp.Error != nil {
+		return nil, errors.New(resp.Error.Description)
+	}
+
+	// Setup to process snapshot chunks.
+	var snapshot []byte
+	done := make(chan bool)
```
kozlovic, line 1462: Make it a size 1: `make(chan bool, 1)`. Actually, make it a `chan struct{}`.
The reason for buffered is that if your code gives up in waiting and later message (of size 0) comes in, the callback will be stuck trying to send to this channel.

js.go
```diff
@@ -1214,6 +1364,262 @@ type StreamState struct {
...
+		m.Respond(nil)
+	})
+	if err != nil {
+		return nil, err
+	}
+	defer sub.Unsubscribe()
+	// Wait to receive the snapshot.
+	select {
+	case <-done:
+	case <-time.After(5 * time.Second):
```
kozlovic, line 1481: Should this be configurable? (as in an option/param to SnapshotStream)

js.go
```diff
@@ -1214,6 +1364,262 @@ type StreamState struct {
...
+	}
+
+	// Can be any size message.
+	var chunk [512]byte
+	for r := snapshot; ; {
+		n, err := r.Read(chunk[:])
+		if err != nil {
+			break
+		}
+		js.nc.Request(resp.DeliverSubject, chunk[:n], time.Second)
```
kozlovic, line 1516: Same, should this be configurable? Or use the `js.wait`? I know @wallyqs added a wait to make the wait being generic and used in various places.

js.go
```diff
@@ -1214,6 +1364,262 @@ type StreamState struct {
...
+	}
+	js.nc.Request(resp.DeliverSubject, nil, time.Second)
```
kozlovic, line 1518: Same than above.

js.go
```diff
@@ -1214,6 +1364,262 @@ type StreamState struct {
...
+// JSApiStreamListResponse list of detailed stream information.
+// A nil request is valid and means all streams.
+type JSApiStreamListResponse struct {
+	APIResponse
+	APIPaged
+	Streams []*StreamInfo `json:"streams"`
+}
+
+// StreamLister fetches pages of StreamInfo objects.
+type StreamLister struct {
```
kozlovic, line 1560: Same comment than for the consumer lister. Maybe add a note that this is not thread safe?

kozlovic (changes requested):
Some little changes needed I believe.

variadico:
Okay, @kozlovic. I think I resolved all of the comments you had. Thanks for the review. :+1:

js.go
```diff
@@ -1214,6 +1368,276 @@ type StreamState struct {
...
+	// Do not include consumers in the snapshot.
+	NoConsumers bool `json:"no_consumers,omitempty"`
+	// Optional chunk size preference.
+	// Best to just let server select.
+	ChunkSize int `json:"chunk_size,omitempty"`
+	// Check all message's checksums prior to snapshot.
+	CheckMsgs bool `json:"jsck,omitempty"`
+	// Timeout is the amount of time to wait for the snapshot to stream to the
+	// client.
+	Timeout time.Duration `json:"-"`
```
kozlovic, line 1437: My first thought was to do just that, but then I realized that this is a config that we "share" with the server, and I was just worried that adding something here could possibly conflict later on with what is in the server. This is why I suggested to be passed as a parameter to SnaphotStream() API, the same way we have a timeout for sub's NextMsg() for instance.

But I am interested in other people thought here, namely, do people think it is ok to add the Timeout to the configuration, without exporting the JSON tag so that this is not sent to the server. Any chance that this becomes a conflict in the future? @derekcollison @ripienaar

kozlovic (commented):
I would be ok with the change, but I have asked other to chime in with regards to the addition of the timeout in the stream snapshot config struct.

js.go
variadico: Aah, whoops. Missed that part. Makes sense to me. I changed the signature to

```
SnapshotStream(name string, timeout time.Duration, cfg *StreamSnapshotConfig
```

js.go
```diff
@@ -1214,6 +1368,269 @@ type StreamState struct {
...
+	APIResponse
+	// Estimate of number of blocks for the messages.
+	NumBlks int `json:"num_blks"`
+	// Block size limit as specified by the stream.
+	BlkSize int `json:"blk_size"`
+}
+
+// SnapshotStream creates a snapshot of a Stream. It returns a copy of the
+// snapshot data.
+func (js *js) SnapshotStream(name string, timeout time.Duration, cfg *StreamSnapshotConfig) (io.Reader, error) {
```
kozlovic, line 1448: It may have been fine leaving it in the config, was just asking for people view on that.

But, if we add the timeout to the API, I would add it last personally.

js.go
```diff
@@ -1214,6 +1368,269 @@ type StreamState struct {
...
+	req, err := json.Marshal(cfg)
```
kozlovic, line 1449: You don't check for `nil` config. Which is ok here, meaning marshal won't fail and sending that to the server is OK... but you then access cfg.DeliverSubject, which then would panic.
So maybe you should check for not `nil` and `DeliverSubject` being set?

kozlovic (changes requested):
Check snapshot config for nil and deliver subject.

js.go
variadico: I added it second so the calling code can look like this.

```go
js.SnapshotStream("foo", 5*time.Second, &nats.StreamSnapshotConfig{
	DeliverSubject: inboxSubj,
	ChunkSize:      1024,
})
```

If we add it to the end, then it'll look like this.
```go
js.SnapshotStream("foo", &nats.StreamSnapshotConfig{
	DeliverSubject: inboxSubj,
	ChunkSize:      1024,
}, 5*time.Second)
```

It's just style though. Personally, I like the way the first block looks like. But, do we have a preferred NATS style?

js.go
variadico: Good catch again! I added error checks here. :+1:

js.go
kozlovic: Agreed that the first looks better. I was just trying to be consistent with other APIs like NextMsg() that has the timeout at the end. But let's wait anyway to see if other think that Timeout in the config itself is better.

js.go
kozlovic: @ripienaar Any opinion on the question of having the Timeout in the config (with a risk of a conflict later on if that were to be added in the server side stream snapshot config) or as an argument of the function?

ripienaar: Personally I'd avoid it

kozlovic (approved):
LGTM

kozlovic:
@variadico Not sure why you removed snapshot? I think @ripienaar comment was about Timeout in config, but I may be wrong. Clearly, we can always add features later, so I don't have a problem removing, just curious as to why you remove it.

variadico:
@wallyqs wanted some more time to think about the snapshotting logic, so I removed that from this PR. That'll be added later. Additionally, he's going to add a few more management APIs.

derekcollison:
We should have DeleteMsg and EraseMsg for the Stream Management API..

wallyqs (approved):
LGTM, also adding DeleteMsg as a different PR here: Adding `DeleteMsg` as a separate PR here https://github.com/nats-io/nats.go/pull/630

PR #627, opened 2020-12-17
js: Add option to configure deliver policy from consumer

wallyqs:
Adds the following options to configure the DeliverPolicy from a consumer.

- `nats.DeliverAllAvailable()`
- `nats.DeliverLastReceived()`
- `nats.DeliverNew()`
- `nats.DeliverByStartTime(time.Time)`
- `nats.DeliverByStartSequence(uint64)`

### DIFF
--- a/js.go
+++ b/js.go
@@ -34,6 +34,7 @@ type JetStream interface {
 	// Subscribing to messages in JetStream.
 	Subscribe(subj string, cb MsgHandler, opts ...SubOpt) (*Subscription, error)
 	SubscribeSync(subj string, opts ...SubOpt) (*Subscription, error)
+
 	// Channel versions.
 	ChanSubscribe(subj string, ch chan *Msg, opts ...SubOpt) (*Subscription, error)
 	// QueueSubscribe.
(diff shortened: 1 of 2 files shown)


js.go
```diff
@@ -989,41 +1036,41 @@ var (
 type DeliverPolicy int
 
 const (
-	DeliverAll DeliverPolicy = iota
-	DeliverLast
-	DeliverNew
-	DeliverByStartSequence
-	DeliverByStartTime
+	DeliverAllPolicy DeliverPolicy = iota
```
wallyqs, line 1039: Renamed a few of these to be able to use `nats.DeliverAll()` as an option for Subscribe instead.

derekcollison: I prefer the shorter versions TBH.

```go
nats.DeliverAll
nats.DeliverNew
nats.DeliverLast
nats.StartTime
nats.StartSequence
```

derekcollison: Could also change server to match..

wallyqs: Changed to use the order names, this is the current mapping of functional option to delivery policy:

- nats.DeliverAllAvailable(): nats.DeliverAll
- nats.DeliverLastReceived(): nats.DeliverLast
- nats.DeliverNextReceived(): nats.DeliverNew
- nats.DeliverFromTime(publishTime): nats.StartTime
- nats.DeliverFromSequence(6): nats.StartSequence

derekcollison: I meant the function names such that things like `js.Subscribe("foo", nats.StartSequence(100))`

That is what I was thinking we should make shorter.. WDYT?

wallyqs: Sounds good to me! Just pushed an update using the shorter names for the options.

derekcollison (approved):
LGTM

PR #631, opened 2021-01-13
Add DeleteMsg to JetStream management

wallyqs:
Adds `DeleteMsg(name string, seq uint64)` to JetStream management to be able to [erase a message](https://github.com/nats-io/nats-server/blob/b048b6b3dea67ba00a36b713226a3db7dcb48bfb/server/jetstream_api.go#L1025-L1056).


derekcollison (approved):
LGTM

PR #633, opened 2021-01-13
Add some validation for durable names

wallyqs:
In case a Durable name includes a dot, then it would not be able to be created since it will not match against the `$JS.API.CONSUMER.CREATE.*` subject from the JetStream API and return a `jetstream not enabled` error.
With this change, we try to detect this in the client first and error in case an invalid durable name has been used.

### DIFF
--- a/js.go
+++ b/js.go
@@ -689,6 +689,10 @@ type subOpts struct {
 
 func Durable(name string) SubOpt {
 	return subOptFn(func(opts *subOpts) error {
+		if strings.Contains(name, ".") {
+			return ErrInvalidDurableName
+		}
+
 		opts.cfg.Durable = name
 		return nil
 	})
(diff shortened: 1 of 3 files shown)


js.go
```diff
@@ -689,6 +689,10 @@ type subOpts struct {
 
 func Durable(name string) SubOpt {
 	return subOptFn(func(opts *subOpts) error {
+		if strings.Contains(name, ".") {
+			return ErrInvalidDurableName
+		}
+
```
kozlovic, line 695: Seem you have some spaces here that are failing "go fmt". Remove the line.

test/js_test.go
```diff
@@ -570,6 +576,13 @@ func TestJetStreamManagement(t *testing.T) {
 		t.Fatalf("ConsumerInfo is not correct %+v", ci)
 	}
 
+	_, err := js.AddConsumer("foo", &nats.ConsumerConfig{
```
kozlovic, line 579: Make it a one line: `if _, err := js.AddConsumer(...); err != nats.ErrInvalidDurableName {`.
Otherwise, change `:=` to `=` since err is already defined.

test/js_test.go
```diff
@@ -475,6 +475,12 @@ func TestJetStreamSubscribe(t *testing.T) {
 			t.Fatalf("Timeout waiting for messages")
 		}
 	}
+
+	// Prevent invalid durable names
+	sub, err = js.SubscribeSync("baz", nats.Durable("test.durable"))
```
kozlovic, line 480: Since unless you have a bug, sub should not be returned, I would do this:
```
if _, err := js.SubscribeSync(..); err != nats.ErrInvalidDurableName {
 ...
```

kozlovic (changes requested):
Some minor changes.

kozlovic (approved):
LGTM

PR #635, opened 2021-01-14
Disable manual acks in case auto ack is enabled in async subs

### LINKED ISSUE #634: m.Ack() sends duplicate when auto-ack enabled
If we have async JS sub and call `m.Ack()` on message received it sends ack again. This should probably return an error.

wallyqs:
Adds a check so that `m.Ack()` cannot be used within an async callback unless the `nats.ManualAck()` option was used.

Fixes #634 

### DIFF
--- a/js.go
+++ b/js.go
@@ -568,12 +568,17 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, opts []
 	var sub *Subscription
 
 	// Check if we are manual ack.
+	var autoAck bool
 	if cb != nil && !o.mack {
+		autoAck = true
 		ocb := cb
-		cb = func(m *Msg) { ocb(m); m.Ack() }
+		cb = func(m *Msg) {
+			ocb(m)
+			m.autoAck()
+		}
 	}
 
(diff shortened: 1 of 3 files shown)


js.go
```diff
@@ -570,7 +570,12 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, opts []
 	// Check if we are manual ack.
 	if cb != nil && !o.mack {
 		ocb := cb
-		cb = func(m *Msg) { ocb(m); m.Ack() }
+		cb = func(m *Msg) {
+			m.Sub.jsi.autoAck = true
```
derekcollison, line 574: Can't we just set something once instead of flipping during the callback?

wallyqs: Ok changed the implementation a bit so that we only check once in case of using auto ack and async subs.

derekcollison (commented):
Apologies, thinking about it more maybe we should just flag the message as ack'd so you can call `m.Ack()` as many times as you want but we only send it once?

wallyqs:
I like that idea of Ack being idempotent instead, will take a look.

wallyqs:
Closing in favor of https://github.com/nats-io/nats.go/pull/637/files

PR #636, opened 2021-01-15
Make most of JS API types private, separate js and management parts

wallyqs:
This splits `js.go` into two files to keep the management parts separated, and also changes many of the helper types to being private since many not useful outside of this package.
Also includes some Go doc fixes.

### DIFF
--- a/context.go
+++ b/context.go
@@ -11,7 +11,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// A Go client for the NATS messaging system (https://nats.io).
 package nats
 
 import (
--- a/jsm.go
+++ b/jsm.go
@@ -1,0 +1,541 @@
+// Copyright 2021 The NATS Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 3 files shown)


context.go
```diff
@@ -11,7 +11,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// A Go client for the NATS messaging system (https://nats.io).
```
wallyqs, line 14: Removing this since was becoming duplicated in the Godoc.

jsm.go
```diff
@@ -0,0 +1,541 @@
+// Copyright 2020 The NATS Authors
```
kozlovic, line 1: Copyright 2021

kozlovic (commented):
Copyright date, otherwise, I like the idea of making things private at first. We will make things public only when needed. This is a big change, but I assume this is just moving things around.
I would check that it does not break NATS Server tests too much (if anything). I think that Derek has been using the JS changes in nats.go already quite a bit. Would you mind checking that everything still compiles, etc???

derekcollison:
Yes thank you, last update broke a few things for me. Nothing major but nice to not have extra work when upgrading to get some new functionality.

variadico (approved):
I edited `nats-server/go.mod` with `replace github.com/nats-io/nats.go => ../nats.go` and then ran `go test ./...`. Seems like this change is backwards compatible. :+1:

wallyqs:
Updated, and confirmed that this should not break the tests from the NATS Server right now.

kozlovic (approved):
LGTM

PR #637, opened 2021-01-20
js: Prevent sending duplicated Acks

### LINKED ISSUE #634: m.Ack() sends duplicate when auto-ack enabled

wallyqs:
Add suppression so that a msg is only attempted to be acked a single time (other than InProgress acks).

Fixes #634 

### DIFF
--- a/js.go
+++ b/js.go
@@ -22,6 +22,7 @@ import (
 	"net/http"
 	"strconv"
 	"strings"
+	"sync/atomic"
 	"time"
 )
 
@@ -827,10 +828,15 @@ func (m *Msg) checkReply() (*js, bool, error) {
 }
 
 // ackReply handles all acks. Will do the right thing for pull and sync mode.
+// It ensures that an ack is only sent a single time, regardless of
+// how many times it is being called to avoid duplicated acks.
 func (m *Msg) ackReply(ackType []byte, sync bool) error {
 	js, isPullMode, err := m.checkReply()
 	if err != nil {
 		return err
+	}
+	if atomic.LoadUint32(&m.ackd) == 1 {
+		return ErrInvalidJSAck
 	}
 	if isPullMode {
 		if bytes.Equal(ackType, AckAck) {
@@ -846,6 +852,13 @@ func (m *Msg) ackReply(ackType []byte, sync bool) error {
 	} else {
 		err = js.nc.Publish(m.Reply, ackType)
 	}
+
+	// Mark that the message has been acked unless it is AckProgress
+	// which can be sent many times.
+	if err == nil && !bytes.Equal(ackType, AckProgress) {
+		atomic.StoreUint32(&m.ackd, 1)
+	}
+
 	return err
 }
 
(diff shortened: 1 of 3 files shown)


js.go
```diff
@@ -827,26 +827,33 @@ func (m *Msg) checkReply() (*js, bool, error) {
...
-	}
-	if isPullMode {
-		if bytes.Equal(ackType, AckAck) {
-			err = js.nc.PublishRequest(m.Reply, m.Sub.Subject, AckNext)
-		} else if bytes.Equal(ackType, AckNak) || bytes.Equal(ackType, AckTerm) {
-			err = js.nc.PublishRequest(m.Reply, m.Sub.Subject, []byte("+NXT {\"batch\":1}"))
+// It ensures that an ack is only sent a single time, regardless of
+// how many times it is being called to avoid duplicated acks.
+func (m *Msg) ackReply(ackType []byte, sync bool) (err error) {
+	m.ackOnce.Do(func() {
```
derekcollison, line 833: This feels heavy to me..

js.go
```diff
@@ -866,12 +876,13 @@ func (m *Msg) Nak() error {
...
+// MsgMetaData is metadata associated with received messages.
 func (m *Msg) Term() error {
 	return m.ackReply(AckTerm, false)
 }
 
-// Indicate that this message is being worked on and reset redelkivery timer in the server.
+// InProgress indicates that this message is being worked on
+// and reset the redelivery timer in the server.
 func (m *Msg) InProgress() error {
 	return m.ackReply(AckProgress, false)
```
derekcollison, line 887: We could issue lots of these, the suppression code would suppress so we may need to look at ack type and the .ackd var.

derekcollison (changes requested):

js.go
wallyqs: Updated to skip suppressing acks in case they are to progress report.

js.go
```diff
@@ -827,11 +828,16 @@ func (m *Msg) checkReply() (*js, bool, error) {
...
 
 // ackReply handles all acks. Will do the right thing for pull and sync mode.
+// It ensures that an ack is only sent a single time, regardless of
+// how many times it is being called to avoid duplicated acks.
 func (m *Msg) ackReply(ackType []byte, sync bool) error {
 	js, isPullMode, err := m.checkReply()
 	if err != nil {
 		return err
 	}
+	if atomic.LoadUint32(&m.ackd) == 1 {
```
derekcollison, line 838: If the first ack I do is a progress type I think this will fail if I send anything after that..

wallyqs: The test `TestJetStreamSubscribe_AckDupInProgress` think should cover this, but this is what would happen right now:

```
m.InProgress() // Sends Ack to update progress
m.InProgress() // Sends Ack as well
m.Ack()             // Also sends Ack, after this no more acks are sent
m.Ack()             // Does not send Ack
m.InProgress() // Would not send Ack either
```

https://github.com/nats-io/nats.go/pull/637/files#diff-9e6bdee4051a6e36d6d507dcef63b5f9a39b281e421c08f02ec74a0fb6639e3aR1465-R1470

derekcollison: Apologies you are right. Was reading that wrong..

derekcollison (approved):
LGTM

PR #638, opened 2021-01-22
Add in support for optional cluster info for streams and consumers.

derekcollison:
Once this lands will update server tests to use this vs full blown JS cluster setup here.


wallyqs (approved):
LGTM

PR #639, opened 2021-01-22
js: Add AccountInfo to JetStreamManagement

wallyqs:

### DIFF
--- a/jsm.go
+++ b/jsm.go
@@ -55,6 +55,9 @@ type JetStreamManager interface {
 
 	// NewConsumerLister is used to return pages of ConsumerInfo objects.
 	NewConsumerLister(stream string) *ConsumerLister
+
+	// AccountInfo retrieves info about the JetStream usage from an account.
+	AccountInfo() (*AccountInfo, error)
 }
 
 // StreamConfig will determine the properties for a stream.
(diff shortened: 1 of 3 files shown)


jsm.go
```diff
@@ -102,22 +105,45 @@ type apiPagedRequest struct {
...
+
+// AccountInfo retrieves info about the JetStream usage from the current account.
+func (js *js) AccountInfo() (*AccountInfo, error) {
+	resp, err := js.nc.Request(js.apiSubj(apiAccountInfo), nil, js.wait)
+	if err != nil {
+		if err == ErrNoResponders {
+			err = ErrJetStreamNotEnabled
+		}
+		return nil, err
+	}
```
variadico, line 137: Should we overwrite `ErrNoResponders` with `ErrJetStreamNotEnabled` here or should we do that in the `JetStream` construction function above?

https://github.com/nats-io/nats.go/pull/639/files#diff-176a2342c55218b70f165bed6a7731dc83d901f9513382f3c98bbd3d08ce0678R134?

I'm thinking about the case where someone calls `AccountInfo` after the constructor function. Not sure which would be more informative. Just a thought.

variadico (approved):

jsm.go
wallyqs: Good point, will add it back to be done when calling `JetStream()` instead.

wallyqs: done

jsm.go
```diff
@@ -102,22 +105,45 @@ type apiPagedRequest struct {
...
 }
 
 type accountInfoResponse struct {
 	apiResponse
-	accountStats
+	AccountInfo
+}
+
+// AccountInfo retrieves info about the JetStream usage from the current account.
+func (js *js) AccountInfo() (*AccountInfo, error) {
```
wallyqs, line 130: btw the type that represents this in the server is called `JetStreamAccountStats`
https://github.com/nats-io/nats-server/blob/00db4ba7a1c7520641805ca14e6aae608d37b1d6/server/jetstream.go#L52

derekcollison (approved):
LGTM

PR #640, opened 2021-01-24
js: Add cleanup of consumers on unsubscribe

wallyqs:
In case Subscribe creates a consumer, then calling Unsubscribe will now make a DeleteConsumer request to remove from the server.

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -542,6 +542,30 @@ type jsSub struct {
 	stream   string
 	deliver  string
 	pull     int
+	attached bool
+	durable  bool
+}
+
+// unsubscribe deletes an ephemeral consumer from a JetStream subscription
+// unless it is attached/direct only.
+func (jsi *jsSub) unsubscribe() {
+	if jsi.attached || jsi.durable {
+		return
+	}
+
+	// Skip if in direct mode as well.
+	js := jsi.js
+	if js.direct {
+		return
+	}
+
+	err := js.DeleteConsumer(jsi.stream, jsi.consumer)
+	if err != nil {
+		nc := js.nc
+		nc.mu.Lock()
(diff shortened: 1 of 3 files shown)


nats.go
```diff
@@ -3525,6 +3527,22 @@ func (s *Subscription) Unsubscribe() error {
 	if conn.IsDraining() {
 		return ErrConnectionDraining
 	}
+
+	if jsi != nil && !jsi.attached {
```
derekcollison, line 3531: OK so we are saying same things, iff the sub created the consumer will we attempt to delete it when you unsubscribe.

Does this sound correct with Durables? I think it does but thinking through it if the app fails then you need to attach not create and that would mean you can not delete the durable using unsubscribe on app restart.

Maybe this just for ephemerals?

wallyqs: Agree, updated so that it only applies to ephemerals now.

derekcollison: ok want to think more about this I think.. For instance I may very well want the unsubscribe to kill the durable. Or, if unsubscribe nevers does how does the program clean it up? Also for consumer info from a sub might be good to understand if we created it somehow.

WDYT?

wallyqs: Opened a new PR trying to address some of this here: https://github.com/nats-io/nats.go/pull/644
Maybe we could change `Unsubscribe` to always delete the consumer even durables, which would be a behavior similar to the one in STAN: https://github.com/nats-io/stan.go/blob/master/sub.go#L44-L54

But instead of having both `Close` as in the STAN interface, we could change `Drain()` instead to mean gracefully process all the messages from a consumer but without deleting durable consumers.

wallyqs:
Closing in favor of this one https://github.com/nats-io/nats.go/pull/644

PR #644, opened 2021-01-28
js: Cleanup of JS consumers on unsub/drain

wallyqs:
This adds logic to the client so that some type of consumers can be cleaned up depending on whether `Unsubscribe` or `Drain` are called:

```go
// Deletes a consumer (both ephemeral and durables)
sub.Unsubscribe()

// Does not delete durable consumers, but ephemerals are deleted.
sub.Drain()
```

This also means that both `nc.Drain` would also clean up any ephemeral consumer that was created, but not doing anything to durable consumers that were created or attached.

This way in case a durable is created and then an application restarts when it reattaches to the same durable consumer, it can still delete the consumer that it created with `Unsubscribe` or gracefully close using `Drain` without deleting durable consumers.

### DIFF
--- a/js.go
+++ b/js.go
@@ -399,6 +399,33 @@ type NextRequest struct {
 	NoWait  bool       `json:"no_wait,omitempty"`
 }
 
+// jsSub includes JetStream subscription info.
+type jsSub struct {
+	js       *js
+	consumer string
+	stream   string
+	deliver  string
+	pull     int
+	durable  bool
+	attached bool
+}
+
+func (jsi *jsSub) unsubscribe(drainMode bool) error {
+	if drainMode && (jsi.durable || jsi.attached) {
+		// Skip deleting consumer for durables/attached
+		// consumers when using drain mode.
+		return nil
+	}
+
+	// Skip if in direct mode as well.
+	js := jsi.js
+	if js.direct {
+		return nil
+	}
+
+	return js.DeleteConsumer(jsi.stream, jsi.consumer)
+}
+
 // SubOpt configures options for subscribing to JetStream consumers.
 type SubOpt interface {
 	configureSubscribe(opts *subOpts) error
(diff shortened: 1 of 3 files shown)


test/js_test.go
```diff
@@ -1507,3 +1510,403 @@ func TestJetStreamSubscribe_AckDupInProgress(t *testing.T) {
...
+		got := string(resp.Data)
+		expected := "B.1"
+		if got != expected {
+			t.Errorf("Expected %v, got: %v", expected, got)
+		}
+		fetchConsumers(t, 2)
+
+		// There will be still all consumers since nc.Close
+		// does not delete ephemeral consumers.
+		nc.Close()
```
wallyqs, line 1761: Maybe we should be attempting to delete ephemeral consumers as well during `Close()`?

test/js_test.go
```diff
@@ -793,13 +796,13 @@ func TestJetStreamImport(t *testing.T) {
 			JS: {
 				jetstream: enabled
 				users: [ {user: dlc, password: foo} ]
-				exports [ { service: "$JS.API.>" },  { service: "foo" }]
+				exports [ { service: "$JS.API.>" },  { service: "test" }]
```
wallyqs, line 799: Temporarily changed since this test failing sometimes locally with a `stream create failed: subjects overlap with an existing stream` error.

test/js_test.go
derekcollison: We could, but server will clean them up on its on so not necessary.

test/js_test.go
derekcollison: We need to make sure if we are running server with JS that we always blow away and persisted store directories.

js.go
```diff
@@ -399,6 +399,33 @@ type NextRequest struct {
...
+		return nil
+	}
+
+	// Skip if in direct mode as well.
+	js := jsi.js
+	if js.direct {
+		return nil
+	}
+
+	return js.DeleteConsumer(jsi.stream, jsi.consumer)
```
wallyqs, line 426: This call fails with timeout during unsubscribe and `nc.Drain()` sometimes because in case the request mux handler got drained first then it will stop being able to receive responses. Trying to fix in another PR.

test/js_test.go
wallyqs: OK now should be fixed.

derekcollison (approved):
LGTM

PR #647, opened 2021-02-01
js: Remove JS timeout which got moved to JS context

wallyqs:

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -1,4 +1,4 @@
-// Copyright 2012-2020 The NATS Authors
+// Copyright 2012-2021 The NATS Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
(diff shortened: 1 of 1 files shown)


nats.go
```diff
@@ -839,14 +834,6 @@ func Timeout(t time.Duration) Option {
 	}
 }
 
-// JetStreamTimeout is an Option to set the timeout for access to the JetStream API
-func JetStreamTimeout(t time.Duration) Option {
```
derekcollison, line 843: I use this one, what is it replaced with?

wallyqs: Right now to be able to update the default JS API timeout, it has to be specified when creating the context, not when creating the connection for example:

```go
nc.JetStream(nats.MaxWait(5*time.Second))
```

derekcollison: That is the way I use it.. So that is good.

derekcollison (approved):
LGTM

PR #648, opened 2021-02-02
js: Make consumer attach based on durable name, remove Attach

wallyqs:
This makes the client do an extra lookup for consumer info to check whether it should attach or create automatically to avoid using `Attach`.


derekcollison:
Did you want me to take a look at this?

wallyqs:
Yes this should be ok to review now, also /cc @ColinSullivan1 @scottf as an approach to remove `Attach` using more JS API calls.

derekcollison (approved):
LGTM

ColinSullivan1 (approved):
LGTM.

PR #649, opened 2021-02-03
js: Allow Acking of JS messages without JS context

wallyqs:
This would allow regular subscribers to `Ack()` messages with a reply from JetStream.

### DIFF
--- a/test/js_test.go
+++ b/test/js_test.go
@@ -498,8 +498,8 @@ func TestAckForNonJetStream(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Unexpected error: %v", err)
 	}
-	if err := m.Ack(); err != nats.ErrNotJSMessage {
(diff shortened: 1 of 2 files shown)


test/js_test.go
```diff
@@ -970,6 +986,27 @@ func TestJetStreamImportDirectOnly(t *testing.T) {
 		}
 	}
 
+	// Do push based consumer using a regular NATS subscription on the import subject.
+	sub, err = nc.SubscribeSync("p.d3")
```
wallyqs, line 990: Regular subscribe here is the same as this:
```
sub, err = js.SubscribeSync("ORDERS", nats.PushDirect("p.d3"))
```
so with this change we could remove `PushDirect` as well potentially.

derekcollison: I think we should.

derekcollison (approved):
LGTM

ColinSullivan1 (approved):
LGTM.

PR #650, opened 2021-02-03
Request/Response handler should drain last

wallyqs:
Sometimes async handlers will still be using the request handler to receive responses, but these can go missing if we drain the respMux first.


derekcollison (approved):
LGTM

ColinSullivan1 (approved):
LGTM.

PR #651, opened 2021-02-03
js: Remove PushDirect option

wallyqs:
Now that Ack can be used by non JetStream subscriptions, to subscriber directly it is only needed to use a regular subscription.


derekcollison (approved):
LGtM

PR #652, opened 2021-02-05
js: Add helpers and client reconnect tests for JS cluster mode

wallyqs:
Adds some tests for push/pull subscribers and queue subscribers when using JS clustering mode.

### DIFF
--- a/test/js_test.go
+++ b/test/js_test.go
@@ -16,12 +16,16 @@ package test
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
+	"net"
 	"os"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
 	"github.com/nats-io/nats-server/v2/server"
+	natsserver "github.com/nats-io/nats-server/v2/test"
 	"github.com/nats-io/nats.go"
 )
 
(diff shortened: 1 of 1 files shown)


test/js_test.go
```diff
@@ -2243,3 +2247,509 @@ func TestJetStream_UnsubscribeDeleteNoPermissions(t *testing.T) {
...
+		}
+		msg, err := sub.NextMsg(2 * time.Second)
+		if err != nil {
+			// t.Logf("Error fetching next message on %v: %v", sub.Subject, err)
+			continue Next
+		}
+
+		// Server will shutdown after a couple of messages which will result
+		// in empty messages with an status unavailable error.
+		if len(msg.Data) == 0 && msg.Header.Get("Status") == "503" {
```
wallyqs, line 2493: When a server restarts in cluster mode, these empty messages are very common.  I think maybe we need a new error type here to handle temporary unavailable state instead of checking from the header.

test/js_test.go
```diff
@@ -2243,3 +2247,509 @@ func TestJetStream_UnsubscribeDeleteNoPermissions(t *testing.T) {
...
+			// t.Logf("%v: Failed to get next message (n=%v): %+v", time.Now(), recvd, msg)
+			continue
+		}
+
+		got := string(msg.Data)
+		expected := fmt.Sprintf("i:%d", recvd)
```
wallyqs, line 2499: During the reconnect sometimes a pull based subscriber seems that can miss one message here sometimes. Opened PR as draft since right now this test is a flapper.

test/js_test.go
ColinSullivan1: Would this be a bug when there is still a quorum available?

test/js_test.go
wallyqs: Made this a warning instead for now.

test/js_test.go
```diff
@@ -2239,7 +2243,541 @@ func TestJetStream_UnsubscribeDeleteNoPermissions(t *testing.T) {
...
+				continue Retry
+			} else if err != nil {
+				t.Errorf("Unexpected error: %v", err)
+			}
+			published = true
+			break Retry
+		}
+
+		if !published {
+			failedPubs++
```
wallyqs, line 2657: Some publishes can fail since we can lose quorum when restarting some of the servers, we account for this instead and  log a warning so changed so that test would fail only when not receiving the messages that were successfully published.

test/js_test.go
```diff
@@ -2239,7 +2243,541 @@ func TestJetStream_UnsubscribeDeleteNoPermissions(t *testing.T) {
...
+		time.Sleep(100 * time.Millisecond)
+		return addr.String(), host, port
+	}
+
+	routes := []string{}
+	for i := 0; i < size; i++ {
+		o := natsserver.DefaultTestOptions
+		o.JetStream = true
+		o.ServerName = fmt.Sprintf("NODE_%d", i)
+		tdir, _ := ioutil.TempDir(os.TempDir(), fmt.Sprintf("%s_%s-", o.ServerName, clusterName))
```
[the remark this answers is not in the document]
wallyqs: Updated as part of this PR: https://github.com/nats-io/nats.go/pull/653/files

test/js_test.go
```diff
@@ -2239,7 +2243,541 @@ func TestJetStream_UnsubscribeDeleteNoPermissions(t *testing.T) {
...
+		recvd         int
+		srvA          = srvs[0]
+		totalMsgs     = 20
+		durable       = nats.Durable("d1")
+		reconnected   = make(chan struct{}, 2)
+		reconnectDone bool
+	)
+	nc, err := nats.Connect(srvA.ClientURL(),
+		nats.ReconnectHandler(func(nc *nats.Conn) {
+			// t.Logf("Reconnected!")
```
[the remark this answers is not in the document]
wallyqs: Thanks, ok removed while updating this one: https://github.com/nats-io/nats.go/pull/653/files

PR #653, opened 2021-02-05
js: Add Placement config to Stream

wallyqs:
Allows being able to define the cluster and tags on stream create.

```go
js.AddStream(&nats.StreamConfig{
	Name: "TEST",
	Placement: &nats.Placement{
		Cluster: cluster,
		Tags:    []string{"NODE_0"},
		},
	})
```


ripienaar:
Just a note atm server doesn’t do anything with tags. Will do before GA hopefully

jsm.go
```diff
@@ -64,20 +64,27 @@ type JetStreamManager interface {
...
+	MaxBytes     int64            `json:"max_bytes"`
+	Discard      DiscardPolicy    `json:"discard"`
+	MaxAge       time.Duration    `json:"max_age"`
+	MaxMsgSize   int32            `json:"max_msg_size,omitempty"`
+	Storage      StorageType      `json:"storage"`
+	Replicas     int              `json:"num_replicas"`
+	NoAck        bool             `json:"no_ack,omitempty"`
+	Template     string           `json:"template_owner,omitempty"`
+	Duplicates   time.Duration    `json:"duplicate_window,omitempty"`
+	Placement    *StreamPlacement `json:"placement,omitempty"`
```
derekcollison, line 81: Would just call it placement.

derekcollison (approved):
LGTM - I may just call it placement, but ok with StreamPlacement too.

jsm.go
wallyqs: ok will update to Placement

variadico (approved):

PR #654, opened 2021-02-06
js: Add GetMsg API to JetStreamManager

wallyqs:

### DIFF
--- a/jsm.go
+++ b/jsm.go
@@ -17,6 +17,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net/http"
 	"strings"
 	"time"
 )
@@ -40,6 +41,9 @@ type JetStreamManager interface {
 
 	// NewStreamLister is used to return pages of StreamInfo objects.
 	NewStreamLister() *StreamLister
+
+	// GetMsg retrieves a raw stream message stored in JetStream by sequence number.
+	GetMsg(name string, seq uint64) (*RawStreamMsg, error)
 
 	// DeleteMsg erases a message from a Stream.
 	DeleteMsg(name string, seq uint64) error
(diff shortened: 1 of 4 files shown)


jsm.go
```diff
@@ -462,6 +465,57 @@ func (js *js) DeleteStream(name string) error {
...
+
+	var resp apiMsgGetResponse
+	if err := json.Unmarshal(r.Data, &resp); err != nil {
+		return nil, err
+	}
+	if resp.Error != nil {
+		return nil, errors.New(resp.Error.Description)
+	}
+
+	msg := resp.Message
```
derekcollison, line 512: Probably should return storedMsg. Need access to timestamp etc. Also below you forget headers.

derekcollison: Maybe call it StreamMsg or RawStreamMsg.

wallyqs: Updated and renamed it `StreamMsg`.

jsm.go
```diff
@@ -462,6 +465,54 @@ func (js *js) DeleteStream(name string) error {
...
 
+type apiMsgGetRequest struct {
+	Seq uint64 `json:"seq"`
+}
+
+// StreamMsg is a raw message stored in JetStream.
+type StreamMsg struct {
+	Subject  string    `json:"subject"`
+	Sequence uint64    `json:"seq"`
+	Header   []byte    `json:"hdrs,omitempty"`
```
derekcollison, line 476: We should convert this if its there like a normal message no?

jsm.go
wallyqs: Synced with @ColinSullivan1 and I think now that `RawStreamMsg` might be better avoid confusion with regular `Msg` so updated again. Also it is now decoding the headers from the fetched message and  converting from `[]byte` to `http.Header`.

jsm.go
wallyqs: Updated to decode the header and use the http.Header type.

jsm.go
derekcollison: ok

derekcollison (approved):
LGTM

PR #657, opened 2021-02-10
js: Handle 503 status when JS API unavailable in pull mode

wallyqs:
The result of a `Poll()` publish request could be a 503 message from the API which when processed by the SubscribeSync iterator will result in an empty message.  This adds similar handling as in `Request` to make these special messages errors instead of being passed to the user.


derekcollison (approved):
LGTM

ColinSullivan1 (commented):
LGTM.

PR #658, opened 2021-02-10
js: Add QueueSubscribeSync to JS context

wallyqs:
Adds `QueueSubscribeSync` which can be used to create pull based queue groups.


derekcollison (approved):
LGTM

PR #659, opened 2021-02-11
jsm: Replace listers with receive only channels

wallyqs:
This adds the following two APIs to JSM which can be used instead of the listers.

```go
	// StreamsInfo can be used to retrieve a list of StreamInfo objects.
	StreamsInfo(ctx context.Context) <-chan *StreamInfo

	// ConsumersInfo is used to retrieve a list of ConsumerInfo objects.
	ConsumersInfo(ctx context.Context, stream string) <-chan *ConsumerInfo
}
```

### DIFF
--- a/jsm.go
+++ b/jsm.go
@@ -37,11 +37,11 @@ type JetStreamManager interface {
 	// StreamInfo retrieves information from a stream.
 	StreamInfo(stream string) (*StreamInfo, error)
 
-	// Purge stream messages.
+	// PurgeStream purges a stream messages.
 	PurgeStream(name string) error
 
-	// NewStreamLister is used to return pages of StreamInfo objects.
-	NewStreamLister() *StreamLister
+	// StreamsInfo can be used to retrieve a list of StreamInfo objects.
+	StreamsInfo(ctx context.Context) <-chan *StreamInfo
 
 	// StreamNames is used to retrieve a list of Stream names.
 	StreamNames(ctx context.Context) <-chan string
@@ -58,11 +58,11 @@ type JetStreamManager interface {
 	// DeleteConsumer deletes a consumer.
 	DeleteConsumer(stream, consumer string) error
 
-	// ConsumerInfo retrieves consumer information.
+	// ConsumerInfo retrieves information of a consumer from a stream.
 	ConsumerInfo(stream, name string) (*ConsumerInfo, error)
 
-	// NewConsumerLister is used to return pages of ConsumerInfo objects.
-	NewConsumerLister(stream string) *ConsumerLister
+	// ConsumersInfo is used to retrieve a list of ConsumerInfo objects.
+	ConsumersInfo(ctx context.Context, stream string) <-chan *ConsumerInfo
 
 	// ConsumerNames is used to retrieve a list of Consumer names.
 	ConsumerNames(ctx context.Context, stream string) <-chan string
@@ -273,9 +273,9 @@ func (js *js) ConsumerInfo(stream, consumer string) (*ConsumerInfo, error) {
 	return js.getConsumerInfo(stream, consumer)
 }
 
-// ConsumerLister fetches pages of ConsumerInfo objects. This object is not
+// consumerLister fetches pages of ConsumerInfo objects. This object is not
 // safe to use for multiple threads.
-type ConsumerLister struct {
+type consumerLister struct {
 	stream string
 	js     *js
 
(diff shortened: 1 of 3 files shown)


jsm.go
```diff
@@ -611,6 +627,23 @@ type StreamLister struct {
 	pageInfo *apiPaged
 }
 
+// Streams returns a receive only channel to iterate on the streams.
+func (js *js) StreamsInfo() <-chan *StreamInfo {
+	ach := make(chan *StreamInfo)
+	sl := &streamLister{js: js}
+	go func() {
```
derekcollison, line 630: Feels like it should have timeouts and a way to cancel it etc no?

jsm.go
```diff
@@ -600,9 +600,25 @@ func (js *js) PurgeStream(name string) error {
 	return nil
 }
 
-// StreamLister fetches pages of StreamInfo objects. This object is not safe
+// StreamLister fetches StreamInfo objects. This object is not safe
 // to use for multiple threads.
-type StreamLister struct {
+type StreamLister interface {
```
derekcollison: We expect this to be accessed by users directly still?

test/js_test.go
```diff
@@ -756,18 +756,18 @@ func TestJetStreamManagement(t *testing.T) {
 		t.Fatalf("ConsumerInfo is not correct %+v", si)
 	}
 
-	sl := js.NewStreamLister()
-	if !sl.Next() {
-		if err := sl.Err(); err != nil {
-			t.Errorf("Unexpected error: %v", err)
+	var i int
+	expected := "foo"
+	for stream := range js.StreamsInfo() {
```
derekcollison, line 752: I like the way this flows imo.

derekcollison (changes requested):

jsm.go
wallyqs: sounds good, we could make it take options like, `nats.MaxWait()` and `nats.Context()` as I did with `AckSync`

wallyqs: I tried adding options for both timeout and context (https://github.com/wallyqs/nats.go/commit/265ef17a0f0b00c5c7efce81cd34c0a95bfe2ff8) but now thinking that might be simpler if only support context.Context instead for this case, so change to something like:

```go
StreamsInfo(ctx context.Context) <-chan *StreamInfo
ConsumersInfo(ctx context.Context, stream string) <-chan *ConsumerInfo
```

jsm.go
```diff
@@ -24,41 +24,41 @@ import (
 
 // JetStreamManager is the public interface for managing JetStream streams & consumers.
 type JetStreamManager interface {
```
ColinSullivan1, line 26: fyi - Java also has an accessor to get stream names.

wallyqs: That is a good point, that API is not exposed in the `JetStreamManager` interface actually. Maybe we should add it too?

```go
GetStreamBySubject(subj string) (*StreamInfo, error)
```

https://github.com/nats-io/nats.go/blob/master/js.go#L655-L662

jsm.go
```diff
@@ -611,6 +627,26 @@ type StreamLister struct {
 	pageInfo *apiPaged
 }
 
+// StreamsInfo returns a receive only channel to iterate on the streams.
+func (js *js) StreamsInfo(ctx context.Context) <-chan *StreamInfo {
+	ach := make(chan *StreamInfo)
```
wallyqs, line 632: In case context is nil here, I think we should create one that defaults to the value of `js.wait`

jsm.go
```diff
@@ -611,6 +633,32 @@ type StreamLister struct {
 	pageInfo *apiPaged
 }
 
+// StreamsInfo returns a receive only channel to iterate on the streams.
+func (js *js) StreamsInfo(ctx context.Context) <-chan *StreamInfo {
+	if ctx == nil {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(context.Background(), js.wait)
+		defer cancel()
```
wallyqs, line 641: I think this `defer` should be moved to be within the goroutine below, otherwise would get canceled right away.

variadico: Whoops! Yeah.

wallyqs:
Need to have branch ready that replaces the lister APIs used by the server before merging.

variadico:
Ready to update the server side after this PR merges.
https://github.com/nats-io/nats-server/pull/1968

jsm.go
```diff
@@ -286,10 +286,25 @@ type consumerLister struct {
 }
 
 // ConsumersInfo returns a receive only channel to iterate on the consumers info.
-func (js *js) ConsumersInfo(ctx context.Context, stream string) <-chan *ConsumerInfo {
+func (js *js) ConsumersInfo(stream string, opts ...JSMOpt) <-chan *ConsumerInfo {
+	var o jsmOpts
+	if len(opts) > 0 {
+		for _, opt := range opts {
+			if err := opt.configureJSManager(&o); err != nil {
```
wallyqs, line 293: maybe add a helper/DRY for some of these?

wallyqs:
Closing in favor of this one https://github.com/nats-io/nats.go/pull/674

PR #662, opened 2021-02-22
js: Add nats.NoAutoNext() + msg.AckNext() option for Pull consumers 

wallyqs:
Adds `NoAutoNext` option which can be used to manually control the messages being delivered to a pull based consumer by disabling the initial Poll done when creating a pull consumer and the NXT when acking a message:

```go
js.SubscribeSync("tasks", nats.Pull(1), nats.Durable("pd1"), nats.NoAutoNext())
```

When AutoNext is disabled, to get new messages `sub.Poll` has to be called to get new messages or use `msg.AckNext()` to ACK+NXT a message a signal the server to have the next msg delivered.

### DIFF
--- a/js.go
+++ b/js.go
@@ -896,7 +896,7 @@ func (m *Msg) checkReply() (*js, bool, bool, error) {
 // ackReply handles all acks. Will do the right thing for pull and sync mode.
 // It ensures that an ack is only sent a single time, regardless of
 // how many times it is being called to avoid duplicated acks.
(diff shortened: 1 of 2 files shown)


ColinSullivan1:
From initial feedback it seems that users expect the default to be `NoAutoNext`, so wondering if we should flip that.

wallyqs:
@ColinSullivan1 closing in favor of this one which defaults to NoAutoNext: https://github.com/nats-io/nats.go/pull/664/files

PR #663, opened 2021-02-23
jsm: Add StreamNames and ConsumerNames

variadico:
This adds `ConsumerNames` and `StreamNames` to `JetStreamManager`.

```go
StreamNames(ctx context.Context) <-chan string
ConsumerNames(ctx context.Context, stream string) <-chan string
```

### DIFF
--- a/js.go
+++ b/js.go
@@ -52,6 +52,9 @@ const (
 	// apiConsumerListT is used to return all detailed consumer information
 	apiConsumerListT = "CONSUMER.LIST.%s"
 
+	// apiConsumerNamesT is used to return a list with all consumer names for the stream.
+	apiConsumerNamesT = "CONSUMER.NAMES.%s"
+
 	// apiStreams can lookup a stream by subject.
 	apiStreams = "STREAM.NAMES"
 
--- a/jsm.go
+++ b/jsm.go
@@ -14,6 +14,7 @@
 package nats
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -42,6 +43,9 @@ type JetStreamManager interface {
 	// NewStreamLister is used to return pages of StreamInfo objects.
 	NewStreamLister() *StreamLister
 
+	// StreamNames is used to retrieve a list of Stream names.
+	StreamNames(ctx context.Context) <-chan string
+
 	// GetMsg retrieves a raw stream message stored in JetStream by sequence number.
 	GetMsg(name string, seq uint64) (*RawStreamMsg, error)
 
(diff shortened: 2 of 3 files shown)


test/js_test.go
```diff
@@ -697,129 +697,164 @@ func TestJetStreamManagement(t *testing.T) {
...
-	// Delete a consumer using our client API.
-	if err := js.DeleteConsumer("", ""); err == nil {
-		t.Fatalf("Unexpected success")
-	}
-	if err := js.DeleteConsumer("foo", "dlc"); err != nil {
-		t.Fatalf("Unexpected error: %v", err)
-	}
+	t.Run("list consumer names", func(t *testing.T) {
+		var names []string
+		for name := range js.ConsumerNames(context.Background(), "foo") {
```
wallyqs, line 809: context.WithTimeout in case it could block?

test/js_test.go
```diff
@@ -697,129 +697,164 @@ func TestJetStreamManagement(t *testing.T) {
...
+		if si, err := js.StreamInfo("foo"); err != nil {
+			t.Fatalf("Unexpected error: %v", err)
+		} else if si.State.Msgs != 0 {
+			t.Fatalf("StreamInfo.Msgs is not correct")
+		}
+	})
+
+	t.Run("list stream names", func(t *testing.T) {
+		var names []string
+		for name := range js.StreamNames(context.Background()) {
```
wallyqs, line 839: add nil or timeout so that it cannot block

jsm.go
```diff
@@ -679,3 +778,90 @@ func (s *StreamLister) Err() error {
...
+func (l *streamNamesLister) Next() bool {
+	if l.err != nil {
+		return false
+	}
+	if l.pageInfo != nil && l.offset >= l.pageInfo.Total {
+		return false
+	}
+
+	subj := l.js.apiSubj(fmt.Sprintf(apiStreamNamesT))
+	r, err := l.js.nc.Request(subj, nil, l.js.wait)
```
wallyqs, line 808: Seems this is the same request as this one but with an empty payload? https://github.com/nats-io/nats.go/blob/master/js.go#L655-L676

js.go
```diff
@@ -73,6 +76,9 @@ const (
 	// apiStreamListT is the endpoint that will return all detailed stream information
 	apiStreamList = "STREAM.LIST"
 
+	// apiStreamNamesT is used to return a list with all stream names.
+	apiStreamNamesT = "STREAM.NAMES"
```
wallyqs, line 80: This is the same as `apiStreams = "STREAM.NAMES"` which above

jsm.go
variadico: Good catch. I deleted `apiStreamNamesT` and `streamNamesListResponse`.

wallyqs:
Let's merge this one after rebase, should be ok since this does not replace APIs used by the server.

variadico:
Rebased. 👍🏽

wallyqs:
ah we also need squash

PR #664, opened 2021-02-23
js: Disable Auto +NXT on Ack, add msg.AckNext()

wallyqs:
This disables sending `+NXT` by default on pull consumers when acking a message. To get the previous behavior now the client has to explicitly call `msg.AckNext()` to be delivered the following message.  Also `sub.Pull()` has to be called explicitly, it is no longer called the first time when `SubscribeSync` or `ChanSubscribe` are used.


wallyqs:
Will address this in a different PR

PR #665, opened 2021-02-25
jsm: Add fields for mirrors to StreamConfig

wallyqs:
Added extra fields to StreamConfig to use mirrors.

### DIFF
--- a/test/js_test.go
+++ b/test/js_test.go
@@ -2689,6 +2689,195 @@ func TestJetStream_ClusterPlacement(t *testing.T) {
 	})
 }
 
+func TestJetStreamStreamMirror(t *testing.T) {
+	withJSCluster(t, "MIRROR", 3, testJetStreamMirror_Source)
+}
+
+func testJetStreamMirror_Source(t *testing.T, nodes ...*jsServer) {
+	srvA := nodes[0]
+	nc, err := nats.Connect(srvA.ClientURL())
+	if err != nil {
+		t.Error(err)
+	}
+
+	js, err := nc.JetStream()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	_, err = js.AddStream(&nats.StreamConfig{
+		Name: "origin",
+		Placement: &nats.Placement{
+			Tags: []string{"NODE_0"},
+		},
+		Storage:  nats.MemoryStorage,
+		Replicas: 1,
+	})
+	if err != nil {
+		t.Fatalf("Unexpected error creating stream: %v", err)
+	}
+
+	totalMsgs := 10
+	for i := 0; i < totalMsgs; i++ {
+		payload := fmt.Sprintf("i:%d", i)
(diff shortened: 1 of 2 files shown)


test/js_test.go
```diff
@@ -2689,6 +2689,74 @@ func TestJetStream_ClusterPlacement(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Unexpected error creating stream: %v", err)
+	}
+	msgs := make([]*nats.RawStreamMsg, 0)
+
+GetNextMsg:
+	for i := 0; i < totalMsgs+1; i++ {
+		timeout := time.Now().Add(2 * time.Second)
+		for time.Now().Before(timeout) {
+			msg, err := js.GetMsg("m1", uint64(i))
```
wallyqs, line 2744: It seems that here we are getting the same sequence number twice: both `js.GetMsg("m1", 0)` and `js.GetMsg("m1", 1)` result in sequence number 0, then `js.GetMsg("m1", 2)` would get the sequence 1 from the original stream.

derekcollison: What version of the server do you have in go.mod?

wallyqs: It is using the version from this commit: https://github.com/nats-io/nats-server/commits/15fa5e21b858

derekcollison (approved):
LGTM - But let's sync tomorrow on double msgs.

test/js_test.go
wallyqs: fyi this seems to only occur when using FileStorage on the mirror:

```
    js_test.go:2746: -------------- i:0 (seq:0)
    js_test.go:2746: -------------- i:0 (seq:1)
    js_test.go:2746: -------------- i:1 ...
    js_test.go:2746: -------------- i:2
    js_test.go:2746: -------------- i:3
    js_test.go:2746: -------------- i:4
    js_test.go:2746: -------------- i:5
    js_test.go:2746: -------------- i:6
    js_test.go:2746: -------------- i:7
    js_test.go:2746: -------------- i:8
    js_test.go:2746: -------------- i:9
```

With MemoryStorage the stream sequence is as expected:

```
    js_test.go:2746: -------------- i:0
    js_test.go:2746: -------------- i:1
    js_test.go:2746: -------------- i:2
    js_test.go:2746: -------------- i:3
    js_test.go:2746: -------------- i:4
    js_test.go:2746: -------------- i:5
    js_test.go:2746: -------------- i:6
    js_test.go:2746: -------------- i:7
    js_test.go:2746: -------------- i:8
    js_test.go:2746: -------------- i:9
```

wallyqs: To update this comment, there aren't duplicated but a difference in how `GetMsg(0)` is handled when the stream is a mirror:

```
jsm.GetMsg(0) on source stream => No message found
jsm.GetMsg(0) on mirror file stream => First Message
jsm.GetMsg(0) on mirror mem stream => No message found
```

PR #666, opened 2021-02-25
js: Add more SubOpts: AckWait, MaxDeliver, MacAckPending

variadico:
This adds the following new options to Subscribe:

```go
js.SubscribeSync("foo",
	nats.Durable("test"),
	nats.AckWait(5*time.Millisecond),
	nats.MaxDeliver(5),
	nats.MaxAckPending(3),
)
```

This also updates so that `nats.AckWait` can be used to set the timeout to wait for the ack of a Publish or AckSync:

```go
js.Publish("baz", msg, nats.AckWait(5*time.Second))
msg.AckSync(nats.AckWait(2 * time.Second))
```

### DIFF
--- a/js.go
+++ b/js.go
@@ -332,13 +332,21 @@ func ExpectLastMsgId(id string) PubOpt {
 // MaxWait sets the maximum amount of time we will wait for a response.
 type MaxWait time.Duration
 
-func (ttl MaxWait) configurePublish(opts *pubOpts) error {
+func (ttl MaxWait) configureJSContext(js *js) error {
+	js.wait = time.Duration(ttl)
+	return nil
+}
+
+// AckWait sets the maximum amount of time we will wait for an ack.
+type AckWait time.Duration
+
+func (ttl AckWait) configurePublish(opts *pubOpts) error {
 	opts.ttl = time.Duration(ttl)
 	return nil
 }
 
-func (ttl MaxWait) configureJSContext(js *js) error {
-	js.wait = time.Duration(ttl)
+func (ttl AckWait) configureSubscribe(opts *subOpts) error {
+	opts.cfg.AckWait = time.Duration(ttl)
 	return nil
 }
 
@@ -794,13 +802,6 @@ func AckAll() SubOpt {
 func AckExplicit() SubOpt {
 	return subOptFn(func(opts *subOpts) error {
 		opts.cfg.AckPolicy = AckExplicitPolicy
-		return nil
-	})
-}
-
-func AckWait(d time.Duration) SubOpt {
-	return subOptFn(func(opts *subOpts) error {
-		opts.cfg.AckWait = d
 		return nil
 	})
 }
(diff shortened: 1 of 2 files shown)


js.go
```diff
@@ -798,14 +798,28 @@ func AckExplicit() SubOpt {
 	})
 }
 
-func MaxWaiting(n int) {
+func AckWait(d time.Duration) SubOpt {
```
wallyqs, line 801: Wonder if this could be confusing with `nats.MaxWait`, when publishing maybe someone could try to do:

```
js.Publish("foo", nats.AckWait(2*time.Second))
```

right now to get that behavior would need to do:

```
js.Publish("foo", nats.MaxWait(2*time.Second))
```

js.go
```diff
@@ -798,14 +798,28 @@ func AckExplicit() SubOpt {
 	})
```
wallyqs, line 798: we need to make sure that AckNone, AckAll is not attempted to be set for pull consumers too.

js.go
```diff
@@ -798,14 +798,28 @@ func AckExplicit() SubOpt {
...
+}
+
+func MaxDeliver(n int) SubOpt {
+	return subOptFn(func(opts *subOpts) error {
+		opts.cfg.MaxDeliver = n
+		return nil
+	})
+}
+
+func MaxWaiting(n int) SubOpt {
```
wallyqs, line 850: Maybe `PullMaxWaiting` to distinguish that it is a pull option?

js.go
```diff
@@ -798,6 +798,69 @@ func AckExplicit() SubOpt {
...
+}
+
+func MaxDeliver(n int) SubOpt {
+	return subOptFn(func(opts *subOpts) error {
+		opts.cfg.MaxDeliver = n
+		return nil
+	})
+}
+
+func FilterSubject(s string) SubOpt {
```
wallyqs, line 815: let's remove this one for now

js.go
variadico: Are you thinking something like this?
```go
func AckNone() SubOpt {
		if o.pull > 0 {
			return fmt.Errorf("cannot set AckNone for pull consumers")
		}
		opts.cfg.AckPolicy = AckNonePolicy
		return nil
	})
}
```

If so, then I think this would require users pass `Pull(n)` before they pass `AckNone()` so that `AckNone` can check `o.pull` that `Pull` set.

js.go
variadico: Ok. Removed `configurePublish` from `MaxWait` and made `AckWait` a `SubOpt` and a `PubOpt` aaaaand also updated the publish tests to use `AckWait` instead of `MaxWait`.

js.go
wallyqs: we could add the check to be done after the options have been processed on subscribe like: https://github.com/nats-io/nats.go/blob/master/js.go#L485-L488

derekcollison (approved):
LGTM - Nice work, I like the UX.

PR #669, opened 2021-03-02
js: Add sub.Fetch API for pull consumers

wallyqs:
This adds a new `Fetch` API to a subscription and changes the way that pull based consumers work so that the result of a request to be delivered new messages is processed synchronously.

```go
func (sub *Subscription) Fetch(batch int, opts ...PullOpt) ([]*Msg, error) {}
```

Example usage:

```go
sub, _ := js.SubscribeSync(subject, nats.Durable("durable"), nats.Pull())
defer sub.Unsubscribe()

// Blocking request to receive messages.
msgs, err := sub.Fetch(10, nats.MaxWait(2*time.Second))
for _, msg := range msgs {
	msg.AckSync()
}
```

Other changes in the branch that change how pull consumers work:

- Removed support for pull based consumers using `ChanSubscribe`
- Removed `sub.Pull()` to signal delivery of new messages which can be processed with `sub.NextMsg`
- Removed Auto +NXT on ack when using a pull consumer
- Changed `nats.Pull()` to mark a subscription as being in pull mode without setting the default batch size


js.go
```diff
@@ -408,9 +423,10 @@ type SequencePair struct {
 
 // nextRequest is for getting next messages for pull based consumers from JetStream.
 type nextRequest struct {
-	Expires *time.Time `json:"expires,omitempty"`
-	Batch   int        `json:"batch,omitempty"`
-	NoWait  bool       `json:"no_wait,omitempty"`
+	Expires   *time.Time `json:"expires,omitempty"`
+	Batch     int        `json:"batch,omitempty"`
+	NoWait    bool       `json:"no_wait,omitempty"`
+	DeliverTo string     `json:"deliver_to,omitempty"`
```
derekcollison: What is this for?

js.go
```diff
@@ -649,7 +665,7 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, opts []
 	// If we are pull based go ahead and fire off the first request to populate.
 	if isPullMode {
 		sub.jsi.pull = o.pull
-		sub.Poll()
+		// sub.Pull()
```
derekcollison: remove

js.go
```diff
@@ -939,7 +1048,7 @@ func (m *Msg) ackReply(ackType []byte, sync bool, opts ...PubOpt) error {
 
 	if isPullMode {
 		if bytes.Equal(ackType, AckAck) {
-			err = nc.PublishRequest(m.Reply, m.Sub.Subject, AckNext)
+			// err = nc.PublishRequest(m.Reply, m.Sub.Subject, AckNext)
```
derekcollison: remove

derekcollison (commented):
I would squash down, too many overwrites for review.

js.go
```diff
@@ -846,21 +839,195 @@ func (sub *Subscription) ConsumerInfo() (*ConsumerInfo, error) {
...
 
-	req, _ := json.Marshal(&nextRequest{Batch: batch})
+	// PullNoWait when enabled returns:
+	// - 404 right away if there are no more messages,
+	// - 409 in case it reached MaxAckPending.
+	//
+	// PullExpiresIn will give a duration to the server to expire our pull request
+	// to this inbox only and will serve the newest one.
+	//
+	// Default is to let the pull requests linger until there are 512 waiting requests,
```
aricart, line 928: so we are trying to do some sort of server protection for over pulling, which will eventually fail when 512 pull requests have been pending. This makes no sense as it is not related in any way to the pull arguments but to some unexpected consumer configuration that is not related to the number of live consumers. When the 512 hits, the subscription should be expected to fail? This is bizarre and complex. From a client perspective, it either gets messages and waits, or the request times out. It would be nice if the server responded to the requestor with a 404 after the expiration triggers (this still requires the client to keep some sort of timer).

To me, it seems that the issue of multiple consumers in a queue group should be handled by using a workqueue stream. In this case, the consumer can pull what it wants, and once handled it goes away. Each consumer is independent.

IMHO this is overly complex, and the pull subscriber should simply poll, get at most the messages requested, and repeated. If the behavior of the polling is that it shouldn't be too frequent it should be documented. The double configuration is really unexpected and overcomplicates the functionality.

wallyqs:
Closing, opening a new PR to address...

PR #670, opened 2021-03-03
js: Add js.PullSubscribe and sub.Fetch APIs for pull consumers

wallyqs:
This changes the way how Pull consumers work by adding the two following APIs:

```go
func (js *js) PullSubscribe(subj string, opts ...SubOpt) (*Subscription, error)
func (sub *Subscription) Fetch(batch int, opts ...PullOpt) ([]*Msg, error)
```

`PullSubscribe` replaces  using `SubscribeSync` or `QueueSubscribeSync` along with `nats.Pull` option to create pull based consumers.  When a subscription is called with `PullSubscribe`, it can use `sub.Fetch()` to get a synchronous response of a collection of messages.

Example usage:

```go
js, _ := nc.JetStream()

// Create pull based consumer with 128 max concurrent inflight pull requests.
sub, err := js.PullSubscribe("foo", nats.Durable("example"), nats.PullMaxWaiting(128))

// Simple loop to fetch one message is efficient:
// internally first makes a request to get a single message if
// there is one already, if none then makes longer request
// that waits up to 2 seconds
for {
        msgs, err := sub.Fetch(1, nats.MaxWait(2*time.Second))
        if err != nil {
                log.Println("Error during fetch:", err)
		continue
        }

	// If no errors, it is ensured that there would always be at
	// least a message at this point.
	msg := msgs[0]
	msg.Ack()
}

// When expecting for multiple messages, it will be awaited
// for the delivery of all the messages or yield the delivered
// results when the timeout expires.
for {
        msgs, err := sub.Fetch(10, nats.MaxWait(2*time.Second))
        if err != nil {
                log.Println("Error during fetch:", err)
		continue
        }
	for _, msg := range msgs {
		msg.Ack()
	}
}
```


derekcollison:
Let's squash down for review.

wallyqs:
Squashed and updated server version in go.mod

test/js_test.go
```diff
@@ -630,15 +635,22 @@ func TestJetStreamAckPending_Pull(t *testing.T) {
 
 	// 3 messages delivered 5 times.
 	expected := 15
```
wallyqs, line 637: Currently asking for 15 msgs with MaxkAckPending==3, means waiting for 5 redeliveries.  After including this change from the server this would work: https://github.com/nats-io/nats-server/pull/1958/files

```
	_, err = sub.Fetch(15)
	if err == nil {
		t.Fatal("Unexpected success")
	}

	// 3 messages re-delivered 5 times.
	expected := 15
	pending := 0
	msgs := make([]*nats.Msg, 0)
	timeout := time.Now().Add(2 * time.Second)
	for time.Now().Before(timeout) {
		ms, err := sub.Fetch(3)
		if err != nil || (ms != nil && len(ms) == 0) {
			continue
		}

		msgs = append(msgs, ms...)
		if len(msgs) >= expected {
			break
		}
		time.Sleep(10 * time.Millisecond)
	}
```

scottf:
What is this?
```
// Deletes durable consumer on unsubscribe
defer sub.Unsubscribe()
```

wallyqs:
@scottf that means that when Unsubscribe() is called then also makes the consumer go away:

https://github.com/nats-io/nats.go/blob/master/js.go#L427-L440
https://github.com/nats-io/nats.go/blob/master/nats.go#L3567-L3576

wallyqs:
Rebased and updated to test against latest version of the server.

js.go
```diff
@@ -846,21 +846,228 @@ func (sub *Subscription) ConsumerInfo() (*ConsumerInfo, error) {
...
+	// - 409 in case it reached MaxAckPending.
+	//
+	// PullExpiresIn will give a duration to the server to expire our pull request
+	// to this inbox only and will serve the newest one.
+	//
+	// Default is to let the pull requests linger until there are 512 waiting requests,
+	// this limit can be modified by using PullMaxWaiting when creating the consumer.
+	//
+	nr := &nextRequest{Batch: batch, NoWait: o.noWait, Expires: o.expires}
+	req, _ := json.Marshal(nr)
```
wallyqs, line 965: In case only Batch is set, could take advantage of the fast path and just the number as bytes.

js.go
```diff
@@ -846,21 +846,228 @@ func (sub *Subscription) ConsumerInfo() (*ConsumerInfo, error) {
...
+			if err == nil {
+				err = checkMsg(msg)
+			}
+		}
+		msgs = append(msgs, msg)
+	case <-ctx.Done():
+		if ctx.Err() == context.Canceled {
+			err = ctx.Err()
+		} else {
+			err = ErrTimeout
```
wallyqs, line 1023: @ColinSullivan1 If we don't get a single message, should we return a timeout error or an empty collection of messages? wdyt?

js.go
ColinSullivan1: Also results in a lighter protocol.

js.go
ColinSullivan1: I'd think a timeout would be consistent with the rest of the API, but am interested in other's opinions.  Polls typically wait for data and timeout when there's no response.

js.go
```diff
@@ -846,21 +846,228 @@ func (sub *Subscription) ConsumerInfo() (*ConsumerInfo, error) {
...
+		return nil, err
+	}
+
+	// Check for empty payload message and process synchronously
+	// any status messages.
+	checkMsg := func(msg *Msg) error {
+		if len(msg.Data) == 0 {
+			switch msg.Header.Get(statusHdr) {
+			case noResponders:
+				return ErrNoResponders
```
ColinSullivan1, line 998: Should there be another error type surfaced that might be more meaningful?

nats.go
```diff
@@ -3740,8 +3739,13 @@ func (s *Subscription) processNextMsgDelivered(msg *Msg) error {
 
 	// In case this is a JetStream message and in pull mode
 	// then check whether it is an JS API error.
-	if jsi != nil && jsi.pull > 0 && len(msg.Data) == 0 && msg.Header.Get(statusHdr) == noResponders {
-		return ErrNoResponders
+	if jsi != nil && jsi.pull && len(msg.Data) == 0 {
+		switch msg.Header.Get(statusHdr) {
+		case noResponders:
+			return ErrNoResponders
+		case "400", "404", "408", "409":
```
ColinSullivan1, line 3746: Convert this case for > 400 to future proof this?  (either convert to int or test the first char to '4' or '5')...  Alternatively, could add a default that results in an error.  wdyt?

test/js_test.go
```diff
@@ -1871,27 +1850,29 @@ func TestJetStreamPullBasedStall(t *testing.T) {
...
-			meta, _ := m.MetaData()
-			if meta.Consumer != uint64(received) {
-				t.Fatalf("Missed something, wanted %d but got %d", received, meta.Consumer)
-			}
-			m.Ack()
-		case <-time.After(time.Second):
-			t.Fatalf("Timeout waiting for messages, last received was %d", received)
-		}
-	}
+	// Chan based pull subscriber not possible.
```
ColinSullivan1, line 1853: Remove this?

ColinSullivan1 (approved):
LGTM; the suggested changes aren't critical to this PR.

wallyqs:
This is rebased and updated against latest version of the server.  The logic is now smarter using NoWait to be signaled in case there are no messages and fallback to a longer pull batch request in case there aren't any at the moment and wait for the delivery.  In case of batches of 1, also the way pull works was simplified to be a mix of new style request plus an old style request for the second one in case the initial `NoWait` request fails with no messages.

derekcollison (approved):
In general LGTM

js.go
wallyqs: In case both the no wait first request and longer request both fail, then it is an error. Otherwise the set of delivered messages is returned.

PR #672, opened 2021-03-05
js: Add EnableFlowControl option for push consumers

wallyqs:
This adds support to add opt-in flow control for the users when using push consumers (https://github.com/nats-io/nats-server/pull/1966).

```go
sub, err := js.SubscribeSync("foo", nats.AckWait(100*time.Millisecond), nats.EnableFlowControl())
```
```go
sub, err := js.Subscribe("foo", func(msg *nats.Msg) {}, nats.EnableFlowControl())
```

~~Opening PR as draft since a few questions~~

- In this PR, we are automatically handling the flow control for the user by detecting `100` status and responding automatically? ~~Should this be the default or an opt-in behavior like `nats.AutoFlowControl()` and let the user manually process the flow control messages?~~

✅  **A: Never surface to users and automatically handle when enabled**
 
- Currently the default so that flow control logic kicks is [4MB](https://github.com/nats-io/nats-server/blob/9813ae3f794a7833b7e1e0104ea0e9c148b5d316/server/consumer.go#L236) and FlowControl in the consumer request is a [bool](https://github.com/nats-io/nats-server/blob/9813ae3f794a7833b7e1e0104ea0e9c148b5d316/server/consumer.go#L65).  ~~In case it is later possible to add `flow_control_max_pending` to change the default?~~

 ✅ **A: Server will decide for now, no option for custom flow control needed**

### DIFF
--- a/js.go
+++ b/js.go
@@ -415,6 +415,7 @@ type ConsumerConfig struct {
 	SampleFrequency string        `json:"sample_freq,omitempty"`
 	MaxWaiting      int           `json:"max_waiting,omitempty"`
 	MaxAckPending   int           `json:"max_ack_pending,omitempty"`
+	FlowControl     bool          `json:"flow_control,omitempty"`
 }
 
 // ConsumerInfo is the info from a JetStream consumer.
@@ -868,6 +869,14 @@ func RateLimit(n uint64) SubOpt {
 func BindStream(name string) SubOpt {
 	return subOptFn(func(opts *subOpts) error {
 		opts.stream = name
+		return nil
+	})
+}
+
+// EnableFlowControl enables flow control for a push based consumer.
+func EnableFlowControl() SubOpt {
+	return subOptFn(func(opts *subOpts) error {
+		opts.cfg.FlowControl = true
 		return nil
 	})
 }
--- a/nats.go
+++ b/nats.go
@@ -2374,6 +2374,7 @@ func (nc *Conn) waitForMsgs(s *Subscription) {
 			s.delivered++
 			delivered = s.delivered
 		}
+		jsi := s.jsi
 		s.mu.Unlock()
 
 		if closed {
@@ -2382,7 +2383,14 @@ func (nc *Conn) waitForMsgs(s *Subscription) {
 
 		// Deliver the message.
 		if m != nil && (max == 0 || delivered <= max) {
-			mcb(m)
+			if jsi != nil && len(m.Data) == 0 && m.Header.Get(statusHdr) == controlMsg {
+				// Process and skip flow control messages automatically.
+				if m.Reply != "" {
+					m.Respond(nil)
+				}
+			} else {
+				mcb(m)
+			}
 		}
 		// If we have hit the max for delivered msgs, remove sub.
 		if max > 0 && delivered >= max {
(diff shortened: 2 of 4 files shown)


ColinSullivan1 (approved):
LGTM.

ColinSullivan1 (commented):
Code looks OK.  Once tests pass LGTM.

nats.go
```diff
@@ -2382,7 +2383,12 @@ func (nc *Conn) waitForMsgs(s *Subscription) {
 
 		// Deliver the message.
 		if m != nil && (max == 0 || delivered <= max) {
-			mcb(m)
+			if jsi != nil && len(m.Data) == 0 && m.Header.Get(statusHdr) == flowControl {
```
wallyqs, line 2386: Note: should ideally be done after Ack

js.go
```diff
@@ -828,6 +832,17 @@ func MaxAckPending(n int) SubOpt {
 	})
 }
 
+// FlowControl enables flow control for a push based consumer.
+func FlowControl(o ...uint64) SubOpt {
```
wallyqs, line 836: Remove and just make it bool

wallyqs: Renamed to `EnableFlowControl` and made it bool.

nats.go
```diff
@@ -3644,6 +3661,9 @@ func (s *Subscription) NextMsg(timeout time.Duration) (*Msg, error) {
 		if err := s.processNextMsgDelivered(msg); err != nil {
 			return nil, err
 		} else {
+			if jsi != nil {
+				return processControlFlow(msg)
```
derekcollison, line 3665: Should we loop back and see if real message is behind this one instead of returning here?

wallyqs: Added so that we peek at the channel and return the next queued message if the channel is ready to send.

nats.go
derekcollison: Change flowControl to controlMsg and check reply for flowControl vs Heartbeat.

nats.go
```diff
@@ -3613,6 +3620,35 @@ func (nc *Conn) unsubscribe(sub *Subscription, max int, drainMode bool) error {
...
+// processControlFlow checks whether it is a control flow message and
+// handles it automatically.
+func (s *Subscription) processControlFlow(msg *Msg) (*Msg, error) {
+	fc := len(msg.Data) == 0 && msg.Header.Get(statusHdr) == flowControl
+	if !fc {
+		return msg, nil
+	}
+
+	// Return in case of an i/o or connection error on publish.
+	err := msg.Respond(nil)
```
derekcollison, line 3632: This could be a heartbeat. Check for msg.Reply != ""

nats.go
```diff
@@ -3613,6 +3620,35 @@ func (nc *Conn) unsubscribe(sub *Subscription, max int, drainMode bool) error {
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Peek at the channel and return the next one already
+	// in case there is one waiting.
+	s.mu.Lock()
```
derekcollison, line 3639: RLock?

wallyqs: Updated to avoid taking the lock to get the sub.mch channel since already had a [snapshot copy of it from earlier](https://github.com/nats-io/nats.go/blob/5f6f07b870cf9652b7453f831ed83014eeaac411/nats.go#L3667-L3668) and passing it instead to the function as a receive only channel.

wallyqs:
Found a few issues with the approach in this PR while adding support for heartbeats, so closing in favor of this PR instead: https://github.com/nats-io/nats.go/pull/685/files

PR #673, opened 2021-03-06
js: Add ReplayOriginal policy and RateLimit

wallyqs:


test/js_test.go
```diff
@@ -2945,6 +2945,171 @@ func TestJetStream_UnsubscribeDeleteNoPermissions(t *testing.T) {
 	}
 }
 
+func TestJetStreamSubscribe_ReplayPolicy(t *testing.T) {
```
kozlovic, line 2948: This is general comment for all tests: we should make sure that the API is sending the proper requests, and not test that the server behaves properly. For instance, to check replay policy, it just making sure that the library sends the request with the proper fields, etc.. otherwise we are just duplicating tests that are done in the server here in the library...

PR #674, opened 2021-03-09
js: Replace JetStreamManager listers with iterators

variadico:
This adds the following two APIs to JSM which can be used instead of the listers.

```go
	// StreamsInfo can be used to retrieve a list of StreamInfo objects.
	StreamsInfo(opts ...JSOpt) <-chan *StreamInfo

	// ConsumersInfo is used to retrieve a list of ConsumerInfo objects.
	ConsumersInfo(stream string, opts ...JSOpt) <-chan *ConsumerInfo
}
```

### DIFF
--- a/js.go
+++ b/js.go
@@ -115,7 +115,8 @@ type JetStreamContext interface {
 
 // js is an internal struct from a JetStreamContext.
 type js struct {
-	nc *Conn
+	ctx context.Context
+	nc  *Conn
 	// For importing JetStream from other accounts.
 	pre string
 	// Amount of time to wait for API requests.
@@ -359,6 +360,11 @@ type ContextOpt struct {
 }
 
 func (ctx ContextOpt) configurePublish(opts *pubOpts) error {
+	opts.ctx = ctx
+	return nil
+}
+
+func (ctx ContextOpt) configureJSContext(opts *js) error {
 	opts.ctx = ctx
 	return nil
 }
--- a/jsm.go
+++ b/jsm.go
@@ -37,11 +37,11 @@ type JetStreamManager interface {
 	// StreamInfo retrieves information from a stream.
 	StreamInfo(stream string) (*StreamInfo, error)
 
-	// Purge stream messages.
+	// PurgeStream purges a stream messages.
 	PurgeStream(name string) error
 
-	// NewStreamLister is used to return pages of StreamInfo objects.
-	NewStreamLister() *StreamLister
+	// StreamsInfo can be used to retrieve a list of StreamInfo objects.
+	StreamsInfo(opts ...JSOpt) <-chan *StreamInfo
 
 	// StreamNames is used to retrieve a list of Stream names.
 	StreamNames(ctx context.Context) <-chan string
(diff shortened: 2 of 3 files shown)


jsm.go
```diff
@@ -340,18 +340,59 @@ func (c *ConsumerLister) Next() bool {
...
-	return &ConsumerLister{stream: stream, js: js}
+// ConsumersInfo is used to retrieve a list of ConsumerInfo objects.
+func (jsc *js) ConsumersInfo(stream string, opts ...JSOpt) <-chan *ConsumerInfo {
+	var o js
+	for _, opt := range opts {
+		if err := opt.configureJSContext(&o); err != nil {
+			return nil
+		}
+	}
+	// Check for option collisions. Right now just timeout and context.
```
wallyqs, line 360: Some of this code that repeats in `StreamsInfo` as well we I think should make into a function and reuse

js.go
```diff
@@ -115,7 +115,9 @@ type JetStreamContext interface {
 
 // js is an internal struct from a JetStreamContext.
 type js struct {
-	nc *Conn
+	ctx       context.Context
+	ctxCancel context.CancelFunc
```
wallyqs, line 119: Shorter variable?
```suggestion
	ctxc context.CancelFunc
```

jsm.go
```diff
@@ -340,18 +340,59 @@ func (c *ConsumerLister) Next() bool {
...
+	if o.ctx != nil && o.wait != 0 {
+		return nil
+	}
+	if o.wait == 0 && o.ctx == nil {
+		o.wait = jsc.wait
+	}
+	if o.ctx == nil && o.wait > 0 {
+		o.ctx, o.ctxCancel = context.WithTimeout(context.Background(), o.wait)
```
wallyqs, line 368: I think we do not need `o.ctxCancel` as part of the opts state since we can use the closure when creating the goroutine below.

jsm.go
```diff
@@ -340,18 +340,59 @@ func (c *ConsumerLister) Next() bool {
...
+	if o.pre == "" {
+		o.pre = jsc.pre
+	}
+	o.nc = jsc.nc
+
+	ch := make(chan *ConsumerInfo)
+	l := &consumerLister{js: &o, stream: stream}
+	go func() {
+		defer func() {
+			if o.ctxCancel != nil {
```
wallyqs, line 379: think we can just rely on the closure above to call `cancel`

variadico:
`ctxCancel` removed and common option logic moved into different function. 👍🏽

jsm.go
```diff
@@ -884,3 +934,30 @@ func (js *js) StreamNames(ctx context.Context) <-chan string {
 
 	return ch
 }
+
+func getJSContextOpts(defs *js, opts ...JSOpt) (*js, context.CancelFunc, error) {
+	var o js
```
wallyqs, line 939: feels like maybe this `var o js` should probably should have been a `jsOpts` similar to `pubOpts`/`subOpts`, and the interface should be taking those instead, wdyt?

```diff
- configureJSContext(opts *js)
+ configureJSContext(opts *jsOpts)
```

variadico: Yeah, makes sense to me. Coming right up.

variadico: Well, here it is:
https://github.com/nats-io/nats.go/pull/674/commits/105e1aaf2862d60150883c3917e74baec8f797ed

Not sure if we want this here or in another PR.

wallyqs: I mean more to split the `type js` which has the behaviors from the options `type jsOpts`.

wallyqs (approved):
LGTM

jsm.go
variadico: jsOpts commit reverted. Will follow up in new PR. 👍🏽

PR #675, opened 2021-03-10
Create jsOpts

variadico:
This creates `jsOpts` which contains all of the options users can pass to
methods that receive `opts ...JSOpt`. It's similar to the existing `pubOpts`
and `subOpts`.

Follow up to https://github.com/nats-io/nats.go/pull/674

### DIFF
--- a/jsm.go
+++ b/jsm.go
@@ -172,7 +172,7 @@ type accountInfoResponse struct {
 
 // AccountInfo retrieves info about the JetStream usage from the current account.
 func (js *js) AccountInfo() (*AccountInfo, error) {
-	resp, err := js.nc.Request(js.apiSubj(apiAccountInfo), nil, js.wait)
+	resp, err := js.nc.Request(js.apiSubj(apiAccountInfo), nil, js.opts.wait)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


jsm.go
```diff
@@ -935,8 +935,8 @@ func (js *js) StreamNames(ctx context.Context) <-chan string {
 	return ch
 }
 
-func getJSContextOpts(defs *js, opts ...JSOpt) (*js, context.CancelFunc, error) {
-	var o js
+func getJSContextOpts(defs *js, opts ...JSOpt) (*jsOpts, context.CancelFunc, error) {
```
wallyqs, line 938: could just pass the options here?

```suggestion
func getJSContextOpts(defs *jsOpts, opts ...JSOpt) (*jsOpts, context.CancelFunc, error) {
```

jsm.go
```diff
@@ -410,7 +410,7 @@ func (c *consumerNamesLister) Next() bool {
 	}
 
 	clSubj := c.js.apiSubj(fmt.Sprintf(apiConsumerNamesT, c.stream))
-	r, err := c.js.nc.Request(clSubj, nil, c.js.wait)
+	r, err := c.js.nc.Request(clSubj, nil, c.js.opts.wait)
```
wallyqs, line 413: If a context was passed then here would `RequestWithContext`

wallyqs (approved):
LGTM

PR #676, opened 2021-03-12
Add optional JSOpts to management API

variadico:
This changes the management APIs to receive options.

```diff
-	AddStream(cfg *StreamConfig) (*StreamInfo, error)
+	AddStream(cfg *StreamConfig, opts ...JSOpt) (*StreamInfo, error)
```

This also updates calls that accepted a timeout to accept a context instead. The context is created from the old timeout value.

```diff
-	r, err := js.nc.Request(usSubj, req, js.opts.wait)
+	r, err := js.nc.RequestWithContext(o.ctx, usSubj, req)
```

### DIFF
--- a/js.go
+++ b/js.go
@@ -899,8 +899,14 @@ func (sub *Subscription) Poll() error {
 }
 
 func (js *js) getConsumerInfo(stream, consumer string) (*ConsumerInfo, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), js.opts.wait)
+	defer cancel()
+	return js.getConsumerInfoContext(ctx, stream, consumer)
+}
+
+func (js *js) getConsumerInfoContext(ctx context.Context, stream, consumer string) (*ConsumerInfo, error) {
 	ccInfoSubj := fmt.Sprintf(apiConsumerInfoT, stream, consumer)
-	resp, err := js.nc.Request(js.apiSubj(ccInfoSubj), nil, js.opts.wait)
+	resp, err := js.nc.RequestWithContext(ctx, js.apiSubj(ccInfoSubj), nil)
 	if err != nil {
 		if err == ErrNoResponders {
 			err = ErrJetStreamNotEnabled
--- a/jsm.go
+++ b/jsm.go
@@ -26,49 +26,49 @@ import (
 // JetStreamManager is the public interface for managing JetStream streams & consumers.
 type JetStreamManager interface {
 	// AddStream creates a stream.
-	AddStream(cfg *StreamConfig) (*StreamInfo, error)
+	AddStream(cfg *StreamConfig, opts ...JSOpt) (*StreamInfo, error)
 
 	// UpdateStream updates a stream.
-	UpdateStream(cfg *StreamConfig) (*StreamInfo, error)
+	UpdateStream(cfg *StreamConfig, opts ...JSOpt) (*StreamInfo, error)
 
 	// DeleteStream deletes a stream.
-	DeleteStream(name string) error
+	DeleteStream(name string, opts ...JSOpt) error
 
 	// StreamInfo retrieves information from a stream.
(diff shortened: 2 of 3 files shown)


wallyqs (approved):

jsm.go
```diff
@@ -26,49 +26,49 @@ import (
 // JetStreamManager is the public interface for managing JetStream streams & consumers.
 type JetStreamManager interface {
 	// AddStream creates a stream.
-	AddStream(cfg *StreamConfig) (*StreamInfo, error)
+	AddStream(cfg *StreamConfig, opts ...JSOpt) (*StreamInfo, error)
```
wallyqs, line 29: AddStream is getting a panic here: https://github.com/nats-io/nats.go/blob/7ea5697e139f81d56e991d486b07cec9e8199858/test/js_test.go#L106-L116

wallyqs: maybe the error is not being returned some how after the attempts?

variadico: Ooops, that's my bad. I purposely added some bad code to make it fail. I forgot to take that out! 😄

jsm.go
```diff
@@ -987,5 +1112,22 @@ func getJSContextOpts(defs *jsOpts, opts ...JSOpt) (*jsOpts, context.CancelFunc,
 		o.pre = defs.pre
 	}
 
+	// 1 normal try plus the number of retries.
+	o.maxTries++
+	if o.retryBackoff == time.Duration(0) {
+		o.retryBackoff = 10 * time.Millisecond
```
wallyqs, line 1118: 10ms maybe too short for a delay, how about 100ms?

```suggestion
		o.retryBackoff = 100 * time.Millisecond
```

js.go
```diff
@@ -192,6 +195,20 @@ func DirectOnly() JSOpt {
...
+type MaxRetries uint
+
+func (v MaxRetries) configureJSContext(opts *jsOpts) error {
+	opts.maxTries = int(v)
+	return nil
+}
+
+type RetryBackoff time.Duration
+
+func (v RetryBackoff) configureJSContext(opts *jsOpts) error {
```
wallyqs, line 207: Maybe apply these to some of the tests to simplify? Like:

https://github.com/nats-io/nats.go/blob/master/test/js_test.go#L3205-L3218

and

https://github.com/nats-io/nats.go/blob/master/test/js_test.go#L3843-L3855

test/js_test.go
```diff
@@ -3841,30 +3832,24 @@ func testJetStream_ClusterReconnectDurableQueueSubscriber(t *testing.T, subject
...
-		stream, err = js.StreamInfo(subject)
-		if err == nats.ErrTimeout {
-			time.Sleep(100 * time.Millisecond)
-			continue
-		} else if err != nil {
-			t.Fatalf("Unexpected error: %v", err)
-		}
-		break
+	stream, err := js.StreamInfo(subject,
+		nats.MaxRetries(10), nats.RetryBackoff(100*time.Millisecond))
```
wallyqs, line 3836: I think to replace this test it would be something like:

```go
ctx, err := context.WithTimeout(context.Background(), 5 * time.Second)
stream, err := js.StreamInfo(subject, nats.Context(ctx), nats.MaxRetries(-1), nats.RetryBackoff(100*time.Millisecond))
```

Would that work?

variadico: Ah, do we want to allow infinite retries? With `nats.MaxRetries(-1)`?

wallyqs: I think -1 would be fine to support, since the context would ensure that it gets canceled, also iirc a default context based on the default js.wait 5seconds default is getting created so would not block.

variadico:
I removed the retry code and only left the change that allows the API to accept variadic JSOpts to keep things simpler.
I'll create a follow up PR for the retries.

PR #679, opened 2021-03-17
js: Add option to consume from Streams that are mirrors or sourced

wallyqs:
This adds a `nats.Stream` option that can be used to set stream from where messages can be consumed. 

```go
js.SubscribeSync("origin", nats.BindStream("mirror"))
js.SubscribeSync("origin", nats.BindStream("source"))
```

Signed-off-by: Jaime Piña <jaime@synadia.com>

### DIFF
--- a/js.go
+++ b/js.go
@@ -561,10 +561,14 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, opts []
 			deliver = NewInbox()
 		}
 	} else {
-		// Find the stream mapped to the subject.
-		stream, err = js.lookupStreamBySubject(subj)
-		if err != nil {
-			return nil, err
+		// Find the stream mapped to the subject if not bound to a stream already.
+		if o.stream == _EMPTY_ {
+			stream, err = js.lookupStreamBySubject(subj)
+			if err != nil {
+				return nil, err
+			}
+		} else {
+			stream = o.stream
 		}
 
 		// With an explicit durable name, then can lookup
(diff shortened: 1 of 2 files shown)


js.go
```diff
@@ -541,10 +541,14 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, opts []
...
-		if err != nil {
-			return nil, err
+		// Find the stream mapped to the subject if not bound to a stream already.
+		if o.stream == _EMPTY_ {
+			stream, err = js.lookupStreamBySubject(subj)
+			if err != nil {
+				return nil, err
+			}
+		} else {
+			stream = o.stream
```
derekcollison, line 551: If we ask to bind to a stream this should be a hard failure if not in the list returned from the server, no?
Do I need to do any work on server side for matching? Right now I would imagine you will not see those streams in normal names lookup yes?

js.go
```diff
@@ -860,6 +864,14 @@ func RateLimit(n uint64) SubOpt {
 	})
 }
 
+// Stream binds a consumer to a stream explicitly based on a name.
+func Stream(name string) SubOpt {
```
derekcollison, line 868: BindStream() to be more explicit about what we want the system to do?

js.go
wallyqs: Right now the [stream names API](https://github.com/nats-io/nats-server/blob/ee92cc9a5ba23a07f7678109ed5af3b36d38b393/server/jetstream_api.go#L1286) only returns the origin stream (does not include mirrors or stream with sources), so we avoid the lookup and just use what is presented in BindStream.
If the server is changed to also include all the mirrors/stream sources that match that subject, then in that case we would have to change the check here.

js.go
wallyqs: Updated to `BindStream()`

js.go
derekcollison: ok so we need some changes on server side to return those streams as well when matching based on subject.

derekcollison: Until then though I think this works, but let's revisit when names will return all streams etc..

derekcollison (approved):
LGTM

PR #680, opened 2021-03-17
Update server dep and tests

variadico:
This updates the server dependency to latest master. In addition, it updates some tests that don't pass with the new server updates.


wallyqs (approved):
LGTM

PR #684, opened 2021-03-24
Fix auth revoke err not being invoked in async error handler

wallyqs:


nats.go
```diff
@@ -2757,6 +2758,9 @@ func checkAuthError(e string) error {
 	if strings.HasPrefix(e, AUTHENTICATION_EXPIRED_ERR) {
 		return ErrAuthExpired
 	}
+	if strings.HasPrefix(e, `User Authentication Revoked`) {
```
kozlovic, line 2761: Make this a constant and all lower case. We convert the content of error string to lower (see line 2773)

kozlovic (changes requested):

nats.go
kozlovic: (add after `AUTHENTICATION_EXPIRED_ERR` const)

matthiashanel:
@kozlovic just for completeness. this will change how that error is surfaced. (for the better, but it's a change)
without this we'd invoke DisconnectErrHandler without error. The revoked error then could be found in nc.LastError()
```go
nats.DisconnectErrHandler(func(nc *nats.Conn, err error) {
	log.Printf("Disconnected due to: %s, will attempt reconnect", err, nc.LastError())
}),
```

With this change the error handler is invoked separately
```
19:46:00 Unexpected NATS error: nats: authentication revoked <--- ErrorHandler
19:46:00 Disconnected due to: EOF, will attempt reconnect <---- DisconnectErrHandler
19:46:02 Unexpected NATS error: nats: authorization violation <--- ErrorHandler
19:46:04 Unexpected NATS error: nats: authorization violation <--- ErrorHandler
```

matthiashanel (commented):
I concur with Ivan.

kozlovic (approved):
LGTM, and good to have added a test!

PR #685, opened 2021-03-25
js: Add support for FlowControl and Heartbeats

wallyqs:
This adds the following options to push based consumers:

```go
sub, err := js.Subscribe("foo", func(msg *nats.Msg) {}, nats.EnableFlowControl(), nats.IdleHeartbeat(2*time.Second))

sub, err := js.SubscribeSync("foo", nats.EnableFlowControl(), nats.IdleHeartbeat(2*time.Second))
```

When they are enabled, the client will automatically handle/skip the control messages sent by the server.  In case of push based consumers with Heartbeats, an asynchronous typed error will be dispatched so that the subscription could deleted and restarted from the given sequence:

```go
errHandler := nats.ErrorHandler(func(c *nats.Conn, sub *nats.Subscription, err error) {
	if info, ok := err.(*nats.ErrConsumerSequenceMismatch); ok {
		t.Logf("Subscription should be restarted from stream sequence at %+v", info.StreamResumeSequence)
	}
})
```

### DIFF
--- a/context.go
+++ b/context.go
@@ -133,6 +133,7 @@ func (s *Subscription) NextMsgWithContext(ctx context.Context) (*Msg, error) {
 
 	// snapshot
 	mch := s.mch
+	jsi := s.jsi
 	s.mu.Unlock()
 
 	var ok bool
@@ -147,9 +148,36 @@ func (s *Subscription) NextMsgWithContext(ctx context.Context) (*Msg, error) {
 		if err := s.processNextMsgDelivered(msg); err != nil {
 			return nil, err
 		} else {
+			// JetStream Push consumers may get extra status messages
+			// that the client will process automatically.
+			if jsi != nil {
+				if isControlMessage(msg) {
+					if err := jsi.handleControlMessage(s, msg); err != nil {
+						return nil, err
+					}
+					// Skip and wait for next message.
+					break
+				} else if jsi.hbs {
+					jsi.trackSequences(msg)
+				}
+				return msg, nil
+			}
 			return msg, nil
 		}
 	default:
+	}
+
+	if jsi != nil {
+		// Skip any control messages that may have been delivered
+		// until there is a valid message or a timeout error.
+		msg, err = s.processControlFlow(mch, jsi, nil, ctx.Done())
+		if err != nil {
+			if err == ErrTimeout {
+				err = ctx.Err()
+			}
+			return nil, err
+		}
+		return msg, nil
 	}
 
 	select {
--- a/js.go
+++ b/js.go
@@ -415,6 +415,8 @@ type ConsumerConfig struct {
 	SampleFrequency string        `json:"sample_freq,omitempty"`
 	MaxWaiting      int           `json:"max_waiting,omitempty"`
 	MaxAckPending   int           `json:"max_ack_pending,omitempty"`
+	FlowControl     bool          `json:"flow_control,omitempty"`
+	Heartbeat       time.Duration `json:"idle_heartbeat,omitempty"`
 }
 
 // ConsumerInfo is the info from a JetStream consumer.
@@ -454,6 +456,19 @@ type jsSub struct {
 	pull     bool
 	durable  bool
 	attached bool
+
+	// Heartbeats handling from push consumers.
+	hbs bool
+
+	// cmeta is holds metadata from a push consumer
+	// for when heartbeats are enabled.
+	cmeta atomic.Value
+}
+
+// controlMetadata is metadata used to be able to detect sequence mismatch
+// errors in push based consumers that have heartbeats enabled.
+type controlMetadata struct {
+	meta string
 }
 
 func (jsi *jsSub) unsubscribe(drainMode bool) error {
@@ -535,6 +550,7 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, opts []
 
 	isPullMode := ch == nil && cb == nil
 	badPullAck := o.cfg.AckPolicy == AckNonePolicy || o.cfg.AckPolicy == AckAllPolicy
+	hasHeartbeats := o.cfg.Heartbeat > 0
 	if isPullMode && badPullAck {
 		return nil, fmt.Errorf("invalid ack mode for pull consumers: %s", o.cfg.AckPolicy)
 	}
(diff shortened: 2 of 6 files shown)


js.go
```diff
@@ -456,6 +456,20 @@ type jsSub struct {
 	pull     bool
 	durable  bool
 	attached bool
+
+	// Heartbeats handling from push consumers.
+	hbs bool
+
+	// cmeta is holds metadata from a push consumer
+	// for when heartbeats are enabled.
+	cmeta atomic.Value
```
derekcollison, line 465: atomic.Value new?

nats.go
```diff
@@ -3623,9 +3625,45 @@ func (nc *Conn) unsubscribe(sub *Subscription, max int, drainMode bool) error {
 	return nil
 }
 
-func (s *Subscription) handleControlMessage(msg *Msg) (bool, error) {
+// ErrConsumerSequenceMismatch represents an error where the consumer sequence.
```
derekcollison, line 3628: Finish sentence ;)

nats.go
```diff
@@ -3623,9 +3625,45 @@ func (nc *Conn) unsubscribe(sub *Subscription, max int, drainMode bool) error {
...
+type ErrConsumerSequenceMismatch struct {
+	// StreamResumeSequence is the stream sequence from where the consumer
+	// should resume consuming from the stream.
+	StreamResumeSequence uint64
+}
+
+func (ecs *ErrConsumerSequenceMismatch) Error() string {
+	return fmt.Sprintf("nats: Sequence mismatch error, should restart consumer from stream sequence %d", ecs.StreamResumeSequence)
```
derekcollison, line 3636: I think has to be all lower case.

nats.go
```diff
@@ -3623,9 +3625,45 @@ func (nc *Conn) unsubscribe(sub *Subscription, max int, drainMode bool) error {
 	return nil
 }
 
-func (s *Subscription) handleControlMessage(msg *Msg) (bool, error) {
+// ErrConsumerSequenceMismatch represents an error where the consumer sequence.
+type ErrConsumerSequenceMismatch struct {
+	// StreamResumeSequence is the stream sequence from where the consumer
+	// should resume consuming from the stream.
+	StreamResumeSequence uint64
```
derekcollison, line 3632: Should we put in the consumer sequence too in case they want access to it for some reason?

nats.go
```diff
@@ -3635,13 +3673,38 @@ func (s *Subscription) handleControlMessage(msg *Msg) (bool, error) {
...
+	} else {
+		// Process heartbeat received, get latest control metadata if present.
+		var ctrl *controlMetadata
+		cmeta := jsi.cmeta.Load()
+		if cmeta == nil {
+			return isControl, nil
+		}
+
+		ctrl = cmeta.(*controlMetadata)
+		tokens, err := getMetadataFields(ctrl.meta)
```
derekcollison, line 3685: Is this from last message we received successfully?

nats.go
```diff
@@ -3635,13 +3673,38 @@ func (s *Subscription) handleControlMessage(msg *Msg) (bool, error) {
...
+		dseq := tokens[6]
+		ldseq := msg.Header.Get(lastConsumerSeqHdr)
+
+		// Detect consumer sequence mismatch and whether
+		// should restart the consumer.
+		if ldseq != dseq {
+			// Dispatch async error including the sequence
+			// from where the consumer could be restarted.
+			sseq := parseNum(tokens[5])
+			handleConsumerSequenceMismatch(s, uint64(sseq+1))
```
derekcollison, line 3699: If not consumer sequence maybe the gap size or something?

js.go
wallyqs: Available since [Go 1.4](https://golang.org/doc/go1.4)

nats.go
wallyqs: Yes these would be from the raw metadata of a previous message.

nats.go
```diff
@@ -3613,6 +3623,51 @@ func (nc *Conn) unsubscribe(sub *Subscription, max int, drainMode bool) error {
...
+// processControlFlow checks whether it is a control message and
+// handles it automatically.
+func (s *Subscription) processControlFlow(t *time.Timer, mch <-chan *Msg) (*Msg, error) {
+	// We will peek at the channel and return the next message
+	// that is not a control message or a timeout error.
+	var msg *Msg
+	var ok bool
+	for {
+		select {
+		case msg, ok = <-mch:
```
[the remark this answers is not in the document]
wallyqs: There is only one of these now since passing the channels from the timer and ctx, in case of either being `nil` then it would just block allowing the other one to receive.

nats.go
wallyqs: Added the consumer sequence and gap size as part of the error.

context.go
```diff
@@ -147,11 +148,39 @@ func (s *Subscription) NextMsgWithContext(ctx context.Context) (*Msg, error) {
 		if err := s.processNextMsgDelivered(msg); err != nil {
 			return nil, err
 		} else {
+			// JetStream Push consumers may get extra status messages
+			// that the client will process automatically.
+			if jsi != nil {
+				if isControlMessage(msg) {
+					err := jsi.handleControlMessage(s, msg)
```
derekcollison, line 155: Make one line.

context.go
```diff
@@ -147,11 +148,39 @@ func (s *Subscription) NextMsgWithContext(ctx context.Context) (*Msg, error) {
...
 	default:
 	}
 
+	if jsi != nil {
+		// Skip any control messages that may have been delivered
+		// until there is a valid message or a timeout error.
+		msg, err = s.processControlFlow(mch, jsi, nil, ctx.Done())
+		if err != nil {
+			if err == ErrTimeout {
+				return nil, ctx.Err()
```
derekcollison, line 177: err = ctx.Err()

nats.go
```diff
@@ -3651,6 +3799,20 @@ func (s *Subscription) NextMsg(timeout time.Duration) (*Msg, error) {
 		if err := s.processNextMsgDelivered(msg); err != nil {
 			return nil, err
 		} else {
+			// JetStream Push consumers may get extra status messages
+			// that the client will process automatically.
+			if jsi != nil {
+				if isControlMessage(msg) {
+					err := jsi.handleControlMessage(s, msg)
```
derekcollison, line 3806: One line

wallyqs:
One remaining issue of this PR is that it does not support `ChanSubscribe`, a fix for that might make the handling simpler: https://github.com/nats-io/nats.go/commit/4ee7ddfff1cba16b7e95b89e43198e8e8d3d0076

wallyqs:
Third time's a charm maybe, several issues related to chan subscribers plus FC / HBs. Opening new PR for FlowControl + Heartbeats here: https://github.com/nats-io/nats.go/pull/692

PR #686, opened 2021-03-25
js: Use old request style for sub.Fetch(1)

wallyqs:
In case of using multiple PullSubscribers concurrently under the same connection, some of the pull messages could have been missed when calling `sub.Fetch(1)`.  The reason was due to the special case added for when batch == 1 to use the global request handler, but due to the [subject rewriting of JetStream](https://github.com/nats-io/nats.go/blob/master/nats.go#L3032-L3034) messages happening then this meant that some of the replies were not being returned.  As a workaround, we just switch to an old style request for the case of batch == 1 as well.


derekcollison (approved):
LGTM

PR #688, opened 2021-03-27
Async publish for JetStream

derekcollison:
This PR introduces async publishing for JetStream.

We wanted to support 3 basic modes of operation.

1. Get a PubAckFuture for each call and be able to query it and select on status.
2. Register interest for errors coming back from the results so a user can just worry about the errors.
3. Be able to wait after sending messages for all to complete.

I also implemented a max outstanding window for benchmarks and normal operations etc.

Basics

```go
	paf, err := js.PublishAsync("foo", []byte("Hello JS"))
	if err != nil {
	}

	select {
	case pa := <-paf.Ok():
		fmt.Printf("Got the PubAck: %+v\n", pa)
	case err := <-paf.Err():
		fmt.Printf("Got an err: %v\n", err)
		// Should be able to get the message back to resend, etc.
		js.PublishMsg(paf.Msg())
	case <-time.After(time.Second):
		t.Fatalf("Did not receive an error in time")
	}
```

Attaching an error handler.

```go
	// Make sure we can register an async err handler for these.
	errHandler := func(js nats.JetStream, originalMsg *nats.Msg, err error) {
		fmt.Printf("Got error %v while sending msg: %+v\n", err, originalMsg)
	}

	js, err = nc.JetStream(nats.PublishAsyncErrHandler(errHandler))
	if err != nil {
	}
```

Setting a window for max outstanding and waiting for all outstanding in a JetStream context ot be complete.

```go
	// Now test that we can wait on all prior outstanding if we want.
	js, err = nc.JetStream(nats.PublishAsyncMaxPending(256))
	if err != nil {
	}

	for i := 0; i < 500; i++ {
		if _, err = js.PublishAsync("bar", []byte("Hello JS ASYNC PUB")); err != nil {
		}
	}

	select {
	case <-js.PublishAsyncComplete():
	case <-time.After(5 * time.Second):
		t.Fatalf("Did not receive completion signal")
	}
```

### DIFF
--- a/js.go
+++ b/js.go
@@ -183,10 +183,6 @@ func (nc *Conn) JetStream(opts ...JSOpt) (JetStreamContext, error) {
 		}
 	}
 
-	if js.opts.direct {
-		return js, nil
-	}
-
 	// If we have check recently we can avoid another account lookup here.
 	// We want these to be lighweight and created at will.
 	nc.mu.Lock()
@@ -228,14 +224,6 @@ func APIPrefix(pre string) JSOpt {
 		if !strings.HasSuffix(js.pre, ".") {
 			js.pre = js.pre + "."
 		}
-		return nil
-	})
-}
-
-// DirectOnly makes a JetStream context avoid using the JetStream API altogether.
-func DirectOnly() JSOpt {
-	return jsOptFn(func(js *jsOpts) error {
-		js.direct = true
 		return nil
 	})
 }
@@ -805,13 +793,7 @@ func (jsi *jsSub) unsubscribe(drainMode bool) error {
 		// consumers when using drain mode.
 		return nil
 	}
-
-	// Skip if in direct mode as well.
 	js := jsi.js
-	if js.opts.direct {
-		return nil
-	}
-
 	return js.DeleteConsumer(jsi.stream, jsi.consumer)
 }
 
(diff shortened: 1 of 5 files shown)


kozlovic (approved):
LGTM (but would have to rebase probably to resolve conflicts)

derekcollison:
Will fix up the bad merge in a second.

js.go
```diff
@@ -300,6 +347,304 @@ func (js *js) Publish(subj string, data []byte, opts ...PubOpt) (*PubAck, error)
...
+}
+
+func (paf *PubAckFuture) Err() <-chan error {
+	paf.js.mu.Lock()
+	defer paf.js.mu.Unlock()
+
+	if paf.errCh == nil {
+		paf.errCh = make(chan error, 1)
+	}
+	if paf.err != nil {
```
wallyqs, line 379: Think here we need to check if the channel is empty before sending, otherwise we will block when sending until previous consumer is done.
```suggestion
	if paf.err != nil && len(paf.errCh) == 0 {
```

derekcollison: We create buffered with size of 1 and will at most only ever send 1 error to this channel.

kozlovic (approved):
LGTM

js.go
wallyqs: yes but thinking of the case when `Err()` is called twice without but consuming, for example if doing this right now the second call will block. 

```
		first := paf.Err()
		second := paf.Err()
```

With the suggested change then both would get the same buffered channel and the first to consume from it would receive the error, the other one would miss it.

derekcollison: ok I understand what you mean now.. I think we should only place the error on there if we create the channel..

derekcollison: ok I think I got it now, see CL below.

wallyqs (approved):
LGTM

PR #689, opened 2021-03-28
js: Fix cluster reconnect test

wallyqs:
Fixes`TestJetStream_ClusterReconnect/qsub_durable` which was flapping due to a couple of server restarts, also makes these tests wait for the server shutdown properly.


derekcollison (approved):
LGTM

PR #690, opened 2021-03-28
js: Remove DirectOnly JS Option

wallyqs:
Removes the `nats.DirectOnly` option which was used to lock down the client not making API lookups.

### DIFF
--- a/test/js_test.go
+++ b/test/js_test.go
@@ -1568,6 +1568,8 @@ func TestJetStreamImportDirectOnly(t *testing.T) {
 				jetstream: enabled
 				users: [ {user: dlc, password: foo} ]
 				exports [
+					# For now have to expose the API to enable JS context across account.
+					{ service: "$JS.API.INFO" }
 					# For the stream publish.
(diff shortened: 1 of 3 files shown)


test/js_test.go
```diff
@@ -1582,6 +1584,7 @@ func TestJetStreamImportDirectOnly(t *testing.T) {
 			U: {
 				users: [ {user: rip, password: bar} ]
 				imports [
+					{ service: { subject: "$JS.API.INFO", account: JS } }
```
wallyqs, line 1587: This is needed so that it is possible to call `nc.JetStream()` and create a context to be able to publish.

derekcollison (commented):
This still allows consumers to be created on mirrors or sourced streams yes? I have not looked at the code yet.

wallyqs:
@derekcollison yes that should still be possible when JS is enabled in the local account as well, covered here: https://github.com/nats-io/nats.go/blob/master/test/js_test.go#L1817-L1851

derekcollison (approved):
LGTM

PR #691, opened 2021-03-29
Make Durable mandatory for PullSubscribe

variadico:
nats.Durable is a SubOpt that may be passed to PullSubscribe or other
methods. If a durable name isn't passed to PullSubscribe, then we return
a runtime error.

Instead of returning a runtime error for this case, this change updates
the PullSubscribe method to require a durable name parameter.

### DIFF
--- a/js.go
+++ b/js.go
@@ -125,7 +125,7 @@ type JetStream interface {
 	QueueSubscribeSync(subj, queue string, opts ...SubOpt) (*Subscription, error)
 
 	// PullSubscribe creates a Subscription that can fetch messages.
-	PullSubscribe(subj string, opts ...SubOpt) (*Subscription, error)
+	PullSubscribe(subj, durable string, opts ...SubOpt) (*Subscription, error)
 }
 
 // JetStreamContext allows JetStream messaging and stream management.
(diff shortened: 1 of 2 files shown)


wallyqs (approved):
LGTM

js.go
```diff
@@ -843,8 +843,8 @@ func (js *js) ChanSubscribe(subj string, ch chan *Msg, opts ...SubOpt) (*Subscri
 }
 
 // PullSubscribe creates a pull subscriber.
-func (js *js) PullSubscribe(subj string, opts ...SubOpt) (*Subscription, error) {
-	return js.subscribe(subj, _EMPTY_, nil, nil, opts)
+func (js *js) PullSubscribe(subj, durable string, opts ...SubOpt) (*Subscription, error) {
+	return js.subscribe(subj, _EMPTY_, nil, nil, append(opts, Durable(durable)))
```
wallyqs, line 847: should we prevent here that nats.Durable is passes so that it is not ambiguous?

variadico: Ooh, yeah. Good idea.

variadico: I updated `nats.Durable` to check if it has already been set. If so, it returns an error. 

```go
func Durable(name string) SubOpt {
		if opts.cfg.Durable != "" {
			return fmt.Errorf("nats: option Durable set more than once")
		}
		if strings.Contains(name, ".") {
			return ErrInvalidDurableName
		}

		opts.cfg.Durable = name
		return nil
	})
}
```

I also added this new test code.

```go
sub, err = js.PullSubscribe("bar", "foo", nats.Durable("bar"))
if err == nil {
	t.Fatalf("Unexpected success")
}
if err != nil && err.Error() != `nats: option Durable set more than once` {
	t.Errorf("Expected consumer in pull mode error, got %v", err)
}
```

derekcollison (commented):
LGTM

PR #692, opened 2021-03-30
js: FlowControl + Heartbeats for Push Consumers

wallyqs:
This adds the following options to push based consumers:

```go
sub, err := js.Subscribe("foo", func(msg *nats.Msg) {}, nats.EnableFlowControl(), nats.IdleHeartbeat(2*time.Second))

sub, err := js.SubscribeSync("foo", nats.EnableFlowControl(), nats.IdleHeartbeat(2*time.Second))

mch := make(chan *Msg, 512)
sub, err := js.ChanSubscribe("foo", mch, nats.EnableFlowControl(), nats.IdleHeartbeat(2*time.Second))
```

When they are enabled, the client will automatically handle/skip the control messages sent by the server.  In case of push based consumers with Heartbeats, an asynchronous typed error will be dispatched so that the subscription could deleted and restarted from the given sequence:

```go
errHandler := nats.ErrorHandler(func(c *nats.Conn, sub *nats.Subscription, err error) {
	if info, ok := err.(*nats.ErrConsumerSequenceMismatch); ok {
		t.Logf("Subscription should be restarted from stream sequence at %+v", info.StreamResumeSequence)
	}
})
```

### DIFF
--- a/js.go
+++ b/js.go
@@ -765,6 +765,8 @@ type ConsumerConfig struct {
 	SampleFrequency string        `json:"sample_freq,omitempty"`
 	MaxWaiting      int           `json:"max_waiting,omitempty"`
 	MaxAckPending   int           `json:"max_ack_pending,omitempty"`
+	FlowControl     bool          `json:"flow_control,omitempty"`
+	Heartbeat       time.Duration `json:"idle_heartbeat,omitempty"`
 }
 
 // ConsumerInfo is the info from a JetStream consumer.
@@ -804,6 +806,19 @@ type jsSub struct {
 	pull     bool
 	durable  bool
 	attached bool
+
+	// Heartbeats and Flow Control handling from push consumers.
+	hbs bool
+	fc  bool
+
+	// cmeta is holds metadata from a push consumer when HBs are enabled.
+	cmeta atomic.Value
+}
+
+// controlMetadata is metadata used to be able to detect sequence mismatch
+// errors in push based consumers that have heartbeats enabled.
+type controlMetadata struct {
+	meta string
 }
 
 func (jsi *jsSub) unsubscribe(drainMode bool) error {
@@ -879,6 +894,8 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync 
 
 	isPullMode := ch == nil && cb == nil
 	badPullAck := o.cfg.AckPolicy == AckNonePolicy || o.cfg.AckPolicy == AckAllPolicy
+	hasHeartbeats := o.cfg.Heartbeat > 0
+	hasFC := o.cfg.FlowControl
 	if isPullMode && badPullAck {
 		return nil, fmt.Errorf("nats: invalid ack mode for pull consumers: %s", o.cfg.AckPolicy)
 	}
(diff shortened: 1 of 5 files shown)


js.go
```diff
@@ -814,40 +829,40 @@ func (js *js) Subscribe(subj string, cb MsgHandler, opts ...SubOpt) (*Subscripti
...
 }
 
 // PullSubscribe creates a pull subscriber.
 func (js *js) PullSubscribe(subj, durable string, opts ...SubOpt) (*Subscription, error) {
-	return js.subscribe(subj, _EMPTY_, nil, nil, append(opts, Durable(durable)))
+	return js.subscribe(subj, _EMPTY_, nil, nil, false, append(opts, Durable(durable)))
 }
 
-func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, opts []SubOpt) (*Subscription, error) {
+func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync bool, opts []SubOpt) (*Subscription, error) {
```
wallyqs, line 865: ChanSubscribers were accidentally becoming SyncSubscribers which triggered a few unexpected issues related to pending limits.  I will try to abstract this change into a different PR.

nats.go
```diff
@@ -2494,30 +2501,41 @@ func (nc *Conn) processMsg(data []byte) {
...
+				sub.pTail.next = m
+				sub.pTail = m
+			}
+		}
+	} else if jsi != nil && jsi.fch != nil {
+		// Process flow control and heartbeat messages
+		// automatically for JetStream Push consumers.
 		select {
-		case sub.mch <- m:
+		case jsi.fch <- m:
```
wallyqs, line 2528: In order to support ChanSubscribe, SyncSubscribe and Async Subscribers, a goroutine that expects signals via a channel is started to process the status messages outside of the message delivery.

js.go
```diff
@@ -1005,11 +1022,61 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync
 	// If we are pull based go ahead and fire off the first request to populate.
 	if isPullMode {
 		sub.jsi.pull = o.pull
+	} else if hasFC || hasHeartbeats {
+		// Start goroutine to handle flow control messages for a push consumer.
+		sub.jsi.fch = make(chan *Msg, 64)
+		go sub.processControlFlow()
```
derekcollison, line 1028: Why do we need this?

wallyqs: I moved the control messages handling to be done before delivering the message at `processMsg` and added an extra goroutine that would be handling the flow control messages from a subscription instead: https://github.com/nats-io/nats.go/pull/692/files#diff-1d2fb9af677137c0182213770500dd5625a51ba7286301cb11d0197b21e0f948R2533

derekcollison: Anyway to avoid having another Go routine?

wallyqs: I removed the goroutine and now responding a flow control message happens as soon as we receive the message, we have to do the processing of control messages there to be able to support ChanSubscribe:

- https://github.com/nats-io/nats.go/pull/692/files#diff-1d2fb9af677137c0182213770500dd5625a51ba7286301cb11d0197b21e0f948R2543
- https://github.com/nats-io/nats.go/pull/692/files#diff-176a2342c55218b70f165bed6a7731dc83d901f9513382f3c98bbd3d08ce0678R1047

derekcollison (approved):
LGTM

PR #693, opened 2021-03-30
js: Fix ChanSubscribe acting as SubscribeSync

wallyqs:
ChanSubscribe was incorrectly becoming a SubscriptionSync type, so there were some unexpected pending limits related slow consumer issues.


kozlovic (approved):
LGTM

PR #694, opened 2021-03-30
js: Add AckOpts for msg.Ack()

wallyqs:
`msg.AckSync()` was reusing some of the `PubOpts` to be able to customize the timeout, instead we replace it and add a new AckOpt type which can be the parameter for all types of acks. This also means that it is possible now to optionally await for the response of other types of Ack, and leaves the door option to potentially add more options to the acks:

```go
wait := nats.AckWait(2*time.Second)
msg.Ack(wait) // same as AckSync(wait), but AckSync() without option is always synchronous.
msg.Term(wait)
msg.InProgress(wait)
msg.Nak(wait)
```

```go
ctx, _ := context.WithTimeout(2*time.Second)
wait := nats.Context(ctx)
msg.Ack(wait)
msg.Term(wait)
msg.InProgress(wait)
msg.Nak(wait)
```

### DIFF
--- a/js.go
+++ b/js.go
@@ -677,6 +677,15 @@ func ExpectLastMsgId(id string) PubOpt {
 	})
 }
 
+type ackOpts struct {
+	ttl time.Duration
+	ctx context.Context
+}
+
+type AckOpt interface {
+	configureAck(opts *ackOpts) error
+}
+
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

PR #695, opened 2021-03-30
Update handling of account authentication expired error

wallyqs:
Applies similar handling `-ERR 'Account Authentication Expired` error as with others auth expired errors.

### DIFF
--- a/nats_test.go
+++ b/nats_test.go
@@ -1436,8 +1436,8 @@ func TestUserCredentialsChainedFile(t *testing.T) {
 	nc.Close()
 }
 
-func TestExpiredUserCredentials(t *testing.T) {
-	// The goal of this test was to check how a client with an expiring JWT
+func TestExpiredAuthentication(t *testing.T) {
+	// The goal of these tests was to check how a client with an expiring JWT
 	// behaves. It should receive an async -ERR indicating that the auth
 	// has expired, which will trigger reconnects. There, the lib should
 	// received -ERR for auth violation in response to the CONNECT (instead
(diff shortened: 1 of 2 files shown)


wallyqs:
After this merge, this test should be updated in the server: https://github.com/nats-io/nats-server/compare/master...wallyqs:auth-acc-expired-err

nats_test.go
```diff
@@ -1651,6 +1651,107 @@ func TestRevokedUserCredentials(t *testing.T) {
 	wg.Wait()
 }
 
+func TestExpiredAccountAuthentication(t *testing.T) {
```
wallyqs, line 1654: Think some of these tests should be refactored into table driven tests since too similar, I'll update the PR

kozlovic:
@wallyqs I don't understand.. you already merged a PR that did that: https://github.com/nats-io/nats.go/pull/684

wallyqs:
@kozlovic there were three types of expired/revoked errors that had to be handled in the same way, I changed into table driven tests so that it is easier to add/remove these special cases /cc @matthiashanel 
```go
		{"expired users credentials", AUTHENTICATION_EXPIRED_ERR, ErrAuthExpired},
		{"revoked users credentials", AUTHENTICATION_REVOKED_ERR, ErrAuthRevoked},
		{"expired account", ACCOUNT_AUTHENTICATION_EXPIRED_ERR, ErrAccountAuthExpired},
```

kozlovic (approved):
LGTM (sorry I got confused with 684 and thought that you added the same auth error that was there, but it is different).

matthiashanel (approved):
LGTM

PR #696, opened 2021-03-31
js: Renaming of msg metadata fields

wallyqs:
This renames the `msg.MetaData` into `msg.Metadata` since a closed form compound word ([more commonly used as well](https://github.com/search?l=Golang&o=desc&q=%22Metadata%22&s=&type=Code)) plus other changes to the metadata fields from a message:

- `meta.Delivered` to `meta.NumDelivered` similar to `NumRedelivered` field in `sub.ConsumerInfo`: https://github.com/nats-io/nats.go/blob/1b16e34671b1b5add561c0b6ede3453009c837d0/js.go#L781
- `meta.Pending` to `meta.NumPending` as in `sub.ConsumerInfo`: https://github.com/nats-io/nats.go/blob/1b16e34671b1b5add561c0b6ede3453009c837d0/js.go#L783
- `meta.Stream` and `meta.Consumer` to `meta.StreamSeq` and `meta.ConsumerSeq` respectively
- `meta.StreamName` changes to `meta.Stream`

### DIFF
--- a/js.go
+++ b/js.go
@@ -1715,13 +1715,12 @@ func (m *Msg) InProgress(opts ...AckOpt) error {
 }
 
 // MsgMetadata is the JetStream metadata associated with received messages.
-type MsgMetaData struct {
-	Consumer   uint64
-	Stream     uint64
-	Delivered  uint64
-	Pending    uint64
-	Timestamp  time.Time
-	StreamName string
+type MsgMetadata struct {
+	Sequence     SequencePair
+	NumDelivered uint64
+	NumPending   uint64
+	Timestamp    time.Time
+	Stream       string
 }
 
 func getMetadataFields(subject string) ([]string, error) {
@@ -1743,9 +1742,9 @@ func getMetadataFields(subject string) ([]string, error) {
 	return tokens, nil
 }
 
-// MetaData retrieves the metadata from a JetStream message. This method will
+// Metadata retrieves the metadata from a JetStream message. This method will
 // return an error for non-JetStream Msgs.
-func (m *Msg) MetaData() (*MsgMetaData, error) {
+func (m *Msg) Metadata() (*MsgMetadata, error) {
 	if _, _, err := m.checkReply(); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


derekcollison (commented):
Not sure I agree with all of these changes. Best to discuss.

variadico (approved):
Seems good to me. It wasn't immediately obvious to me what `MsgMetadata.Stream` was before, I had to look up the documentation for that. `StreamSeq` makes sense to me. And we've gotten feedback that `Delivered` sounds like a boolean, so this makes it clear that it's the number of messages delivered, instead of whatever the message was simply delivered.

js.go
```diff
@@ -1715,13 +1715,13 @@ func (m *Msg) InProgress(opts ...AckOpt) error {
...
 // MsgMetadata is the JetStream metadata associated with received messages.
-type MsgMetaData struct {
-	Consumer   uint64
-	Stream     uint64
-	Delivered  uint64
-	Pending    uint64
-	Timestamp  time.Time
-	StreamName string
+type MsgMetadata struct {
+	StreamSeq    uint64
```
derekcollison, line 1719: ok, trying to look at this in a different way as you both described. I am ok with NumDelivered and NumPending, was wondering if the sequences should there own struct?

```go
type Sequence struct {
  Consumer uint64
  Stream   uint64
}
```

derekcollison (approved):
LGTM - Minor suggestion.

js.go
wallyqs: Adding the SequencePair sounds good, so would look like this?

```go
type MsgMetadata struct {
	Sequence     SequencePair
	NumDelivered uint64
	NumPending   uint64
	Timestamp    time.Time
	Stream       string
}
```

where `SequencePair` is the same type that is common in other types:

```
type SequencePair struct {
	Consumer uint64 `json:"consumer_seq"`
	Stream   uint64 `json:"stream_seq"`
}
```

https://github.com/wallyqs/nats.go/commit/01f3157fba1687b2890cb6c54599c09321c3d01f

wallyqs: Updated to:

```go
type MsgMetadata struct {
	Sequence     SequencePair
	NumDelivered uint64
	NumPending   uint64
	Timestamp    time.Time
	Stream       string
}

meta, _ := m.Metadata()
// Stream sequence
meta.Sequence.Stream
// Consumer sequence
meta.Sequence.Consumer
```

derekcollison (approved):
LGTM

PR #705, opened 2021-04-05
Add go_test.mod and minimal go.mod

wallyqs:
This is so that users that depend on `nats.go` do not depend explicitly on the test dependencies of the package.  In case users want to fetch the test dependencies as we run in CI, they can either run: `go test -modfile=go_test.modfile test ./...` from this repo, or run `go mod tidy` to fetch all testing deps.

The minimal go mod can be generated by building one of the examples:

```
go build -mod=mod examples/nats-sub/main.go 
```


wallyqs:
@kozlovic I think we should include this, the v1.10.0 go.mod for example is a minimal go.mod and has been working well for us: https://github.com/nats-io/nats.go/blob/v1.10.0/go.mod
In case there is an error with test dependencies during development `go mod tidy` would fix it automatically anyway.

kozlovic (approved):
LGTM

PR #707, opened 2021-04-08
Make Msg.Header fields case-preserving when accessed directly

### LINKED ISSUE #706: NATS Jetstream message header case becomes mangled
nats.go uses http.Header for transmitting message headers.

http.Header is used to represent a set of key value pairs, where the value is a slice.

http.Header has additional side effects resulting in canonicalization of the case according to the HTTP spec.

Recommend moving to a different structure based on a normal go map[string].

wallyqs:
There is an [issue](https://github.com/golang/go/issues/37834) with Go and the HTTP libraries in that responses handled by its HTTP clients become represented in 'canonical form' (`X-Foo-Bar`) sometimes unexpectedly causing integration issues with proxies, APIs, legacy apps and likely compatibility with http/2 headers to some degree which require headers in lowercase, (unlike http v1 ones that are case insensitive). As a workaround, the Go docs suggest to access the [Header map directly](https://golang.org/pkg/net/http/#Header.Set) to use the fields not in canonical form.

In this PR, it is avoided during the decoding of the headers calling `CanonicalMIMEHeaderKey` in order to be able to preserve the raw version of the headers when traversing over multiple hops without causing unexpected side effects or changing the format over the wire.

If users add headers via `Header.Add()` or `Header.Set()` for example, those methods already call `CanonicalMIMEHeaderKey` when creating the header so they will be presented in canonical form without changes.

Fixes #706

### DIFF
--- a/test/js_test.go
+++ b/test/js_test.go
@@ -169,6 +169,16 @@ func TestJetStreamPublish(t *testing.T) {
 	// Messages should have been rejected.
 	expect(0, 1)
 
+	// Using PublishMsg API and accessing directly the Header map.
+	msg2 := nats.NewMsg("foo")
+	msg2.Header[nats.ExpectedLastSeqHdr] = []string{"10"}
+	pa, err = js.PublishMsg(msg2)
+	if err == nil || !strings.Contains(err.Error(), "wrong last sequence") {
+		t.Fatalf("Expected an error, got %v", err)
+	}
+	// Messages should have been rejected.
+	expect(0, 1)
+
 	// Send in a stream with a msgId
 	pa, err = js.Publish("foo", msg, nats.MsgId("ZZZ"))
 	if err != nil {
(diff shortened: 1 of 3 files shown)


ripienaar:
Hmm, did we test if this will also work for messages retrieved from JetStream where I fear we might have lost the original casing once saved using go mangling - untested so curious how this works there

wallyqs:
@ripienaar added a test and seems fine, this would not affect JetStream headers either https://github.com/nats-io/nats.go/pull/707/files#diff-9e6bdee4051a6e36d6d507dcef63b5f9a39b281e421c08f02ec74a0fb6639e3aR1833

test/js_test.go
```diff
@@ -208,6 +218,16 @@ func TestJetStreamPublish(t *testing.T) {
 	}
 	expect(3, 3)
 
+	// JetStream Headers are case-sensitive right now,
```
wallyqs, line 221: Filed separate issue for this, issue unrelated to this PR but might be good to have to avoid issues when a header changes case for some reason in one of the libraries: https://github.com/nats-io/nats-server/pull/2094

kozlovic (approved):
LGTM

PR #710, opened 2021-04-12
Add godoc example for JS context functions

wallyqs:

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -796,7 +796,8 @@ func ExampleContext() {
 	}, nctx)
 
 	// Custom context with timeout
-	tctx, _ := context.WithTimeout(nctx, 2*time.Second)
+	tctx, tcancel := context.WithTimeout(nctx, 2*time.Second)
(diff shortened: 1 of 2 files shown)


example_test.go
```diff
@@ -772,3 +772,39 @@ func ExamplePullOpt() {
...
+	nctx := nats.Context(ctx)
+
+	// JetStreamManager functions all can use context.
+	js.AddStream(&nats.StreamConfig{
+		Name:     "FOO",
+		Subjects: []string{"foo"},
+	}, nctx)
+
+	// Custom context with timeout
+	tctx, _ := context.WithTimeout(nctx, 2*time.Second)
```
variadico, line 799: I think we need to add this, but not sure if we can edit wally's repo. 😅 
```suggestion
	tctx, tcancel := context.WithTimeout(nctx, 2*time.Second)
	defer tcancel()
```

kozlovic (approved):
LGTM

PR #711, opened 2021-04-14
js: Improvements for push consumers with flow control

wallyqs:
- Move responding to flow control messages to be done after latest Ack

- Set larger pending limits for async subscribers with flow control

### DIFF
--- a/js.go
+++ b/js.go
@@ -146,6 +146,7 @@ type js struct {
 	pafs map[string]*pubAckFuture
 	stc  chan struct{}
 	dch  chan struct{}
+	rr   *rand.Rand
 }
 
 type jsOpts struct {
@@ -443,10 +444,11 @@ func (js *js) newAsyncReply() string {
 			return _EMPTY_
 		}
 		js.rsub = sub
+		js.rr = rand.New(rand.NewSource(time.Now().UnixNano()))
 	}
 	var sb strings.Builder
 	sb.WriteString(js.rpre)
-	rn := js.nc.respRand.Int63()
+	rn := js.rr.Int63()
 	var b [aReplyTokensize]byte
 	for i, l := 0, rn; i < len(b); i++ {
 		b[i] = rdigits[l%base]
@@ -850,11 +852,10 @@ type jsSub struct {
 	attached bool
 
 	// Heartbeats and Flow Control handling from push consumers.
-	hbs bool
-	fc  bool
-
-	// cmeta is holds metadata from a push consumer when HBs are enabled.
-	cmeta atomic.Value
+	hbs   bool
+	fc    bool
+	cmeta string
+	fcs   map[uint64]string
 }
 
 // newFetchReply generates a unique inbox used for a fetch request.
@@ -929,12 +930,6 @@ func (jsi *jsSub) fetchNoWait(ctx context.Context, subj string, payload []byte) 
 	}
 
 	return msg, nil
-}
-
-// controlMetadata is metadata used to be able to detect sequence mismatch
-// errors in push based consumers that have heartbeats enabled.
-type controlMetadata struct {
-	meta string
 }
 
 func (jsi *jsSub) unsubscribe(drainMode bool) error {
@@ -1096,6 +1091,12 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync 
 		return nil, err
 	}
 
+	// With flow control enabled async subscriptions we will disable msgs
+	// limits, and set a larger pending bytes limit by default.
+	if !isPullMode && cb != nil && hasFC {
+		sub.SetPendingLimits(DefaultSubPendingMsgsLimit*16, DefaultSubPendingBytesLimit)
+	}
+
 	// If we are creating or updating let's process that request.
 	if shouldCreate {
 		// If not set default to ack explicit.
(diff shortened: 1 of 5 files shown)


js.go
```diff
@@ -1053,40 +1055,99 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync
...
-			// from where the consumer could be restarted.
-			sseq := parseNum(tokens[5])
-			ecs := &ErrConsumerSequenceMismatch{
-				StreamResumeSequence: uint64(sseq),
-				ConsumerSequence:     parseNum(dseq),
-				LastConsumerSequence: parseNum(ldseq),
-			}
-			nc.handleConsumerSequenceMismatch(s, ecs)
+	// ConsumerSequence is the sequence of the consumer that is behind.
+	ConsumerSequence int64
```
derekcollison, line 1067: Should be uint64?

js.go
```diff
@@ -1053,40 +1055,99 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync
...
+
+	// LastConsumerSequence is the sequence of the consumer when the heartbeat
+	// was received.
+	LastConsumerSequence int64
```
derekcollison, line 1071: Same

js.go
```diff
@@ -139,12 +140,13 @@ type js struct {
...
-	pafs map[string]*pubAckFuture
-	stc  chan struct{}
-	dch  chan struct{}
+	mu       sync.RWMutex
+	rpre     string
+	rsub     *Subscription
+	pafs     map[string]*pubAckFuture
+	stc      chan struct{}
+	dch      chan struct{}
+	respRand *rand.Rand
```
wallyqs, line 149: This also fixes a data race that was happening when PublishAsync was being called in a goroutine.

js.go
```diff
@@ -984,6 +981,12 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync
 		if err != nil {
 			return nil, err
 		}
+
+		// With flow control enabled async subscriptions we will disable msgs limits,
+		// and set a larger pending bytes limit by default.
+		if cb != nil && hasFC {
+			sub.SetPendingLimits(-1, DefaultSubPendingBytesLimit*4)
```
wallyqs, line 988: This would mean 256M max pending bytes for flow control based push consumers. Disabled the limit of number of messages but maybe could have done something similar and change this to *4: https://github.com/nats-io/nats.go/blob/aa4ab64c8ba0289099145ca795b6e51c90e8ca99/nats.go#L4119

test/js_test.go
```diff
@@ -1421,6 +1421,106 @@ Loop:
...
+		t.Logf("WARN: %s : %+v", err, sub)
+		errCh <- err
+	})
+	nc, err := nats.Connect(s.ClientURL(), errHandler)
+	if err != nil {
+		t.Fatalf("Unexpected error: %v", err)
+	}
+	defer nc.Close()
+
+	const totalMsgs = 10000
```
wallyqs, line 1443: To hit the limit locally, I need to make this a higher value like more than 30000 msgs here, but that will fail in Travis since environment is slower.

wallyqs:
Rebased to latest and build is green now

js.go
```diff
@@ -984,6 +981,12 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync
 		if err != nil {
 			return nil, err
 		}
+
+		// With flow control enabled async subscriptions we will disable msgs limits,
```
derekcollison, line 985: Not sure we should disable. We should make sure FC on its on will not trip internal pending limits under normal operations.

js.go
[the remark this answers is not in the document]
wallyqs: Updated

js.go
wallyqs: Updated comment to mention that not disabling but instead increasing the default limits, I did some changes to the way that flow control is done so that it happens after `Ack()` but I can see it still overrun the default pending bytes sometimes.

derekcollison: Pending Bytes or Messages?

js.go
```diff
@@ -984,6 +981,11 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync
 		if err != nil {
 			return nil, err
 		}
+
+		// With flow control enabled async subscriptions we will set a larger pending bytes and msgs limit than the default.
+		if cb != nil && hasFC {
+			sub.SetPendingLimits(DefaultSubPendingMsgsLimit*4, DefaultSubPendingBytesLimit*4)
```
derekcollison, line 987: We can match bytes with the FC internals plus a bit of a fudge factor.

I think it is 1M can be outstanding without an FC ack.

js.go
wallyqs: The pending bytes if I use 4K payloads.

derekcollison: FC will allow a max of 1M outstanding. So if FC acks are after processing then pending bytes default should be easily sufficient IIRC.

js.go
```diff
@@ -1096,6 +1091,12 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync
 		return nil, err
 	}
 
+	// With flow control enabled async subscriptions we will disable msgs
+	// limits, and set a larger pending bytes limit by default.
+	if !isPullMode && cb != nil && hasFC {
+		sub.SetPendingLimits(DefaultSubPendingMsgsLimit*4, DefaultSubPendingBytesLimit*4)
```
derekcollison, line 1097: Would leave Bytes alone, but upping messages ok by me.

js.go
```diff
@@ -1194,40 +1195,124 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync
...
+	if len(jsi.fcs) == 0 {
+		return
+	}
+
+	if reply := jsi.fcs[delivered]; reply != _EMPTY_ {
+		delete(jsi.fcs, delivered)
+		nc.Publish(reply, nil)
+	}
+
+	// FIXME(dlc) - Debugging only, should not go above 1M
```
derekcollison, line 1254: Remove this, this was for testing.

nats.go
```diff
@@ -2683,8 +2687,19 @@ func (nc *Conn) processMsg(data []byte) {
 				sub.pTail = m
 			}
 		}
-		if hasFC {
-			jsi.trackSequences(m)
+		if jsi != nil && (hasHBs || hasFC) {
```
derekcollison, line 2690: I thought I needed these tracked before if hasFC but the way the impl worked we do not need to track if hasFC per se.

derekcollison (changes requested):

js.go
variadico: The current limit is:
```
DefaultSubPendingMsgsLimit*4  ->   262,144 pending messages
```

Maybe we could pick one of these instead?

```
DefaultSubPendingMsgsLimit*8  ->   524,288 pending messages
DefaultSubPendingMsgsLimit*16 -> 1,048,576 pending messages
```

js.go
variadico: Cool, removed the whole block.
```diff
-	// FIXME(dlc) - Debugging only, should not go above 1M
-	if sub.pBytes > 1024*1024 {
-		fmt.Println(delivered, sub.pMsgs, sub.pBytes)
-	}
```

nats.go
variadico: Should this block get moved to another location?

js.go
derekcollison: Well its the most amount of properly FC'd messages within 1MB that we could receive properly. *16 is good I think..

nats.go
derekcollison: I think this is only applicable for HBs now.

js.go
variadico: Updated message pending limit. 👍🏽 
```diff
-sub.SetPendingLimits(DefaultSubPendingMsgsLimit*4, DefaultSubPendingBytesLimit*4)
+sub.SetPendingLimits(DefaultSubPendingMsgsLimit*16, DefaultSubPendingBytesLimit*4)
```

nats.go
variadico: Removed the `hasFC` check.
```diff
-if jsi != nil && (hasHBs || hasFC) {
+if jsi != nil && hasHBs {
```

variadico:
Hm. Seeing this error on CI and locally. Not exactly sure why that's happening. Looking into it.
```
$ go test -count=5 -run TestJetStreamPushFlowControl_SubscribeAsyncAndChannel
--- FAIL: TestJetStreamPushFlowControl_SubscribeAsyncAndChannel (13.98s)
    js_test.go:1513: Expected 33072 messages, got: 14
FAIL
exit status 1
FAIL    github.com/nats-io/nats.go/test 28.207s
```

variadico:
Woo, green! Squashed and rebased commits. Thanks for the help, @derekcollison!

derekcollison (approved):
LGTM

kozlovic (approved):

PR #713, opened 2021-04-17
js: Fix issue with concurrent queue subscribers creating consumer

wallyqs:
When multiple concurrent subscribers start without an existing consumer, one of them would create the consumer, some of them would attach and others fail with a `consumer already exists` error.  In this change, we make the subscriptions that would have failed with `consumer already exists` error to do an extra lookup of the latest consumer info to attach instead.


derekcollison (approved):
LGTM

PR #716, opened 2021-04-20
js: Fixes to parallel creation of durable consumers

wallyqs:


js.go
```diff
@@ -1019,52 +1019,80 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync
 		}
 
 		resp, err := js.nc.Request(js.apiSubj(ccSubj), j, js.opts.wait)
-		if err != nil {
+		if err == ErrTimeout {
```
derekcollison, line 1022: If we fix the behavior in the server will this change?

wallyqs: Yes could be improved if the server sends a response such as `consumer already exists` at this point.

js.go
```diff
@@ -1020,31 +1020,30 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync
...
+			sub.Drain()
 
 			// Multiple subscribers could compete in creating the first consumer
 			// that will be shared using the same durable name. If this happens, then
 			// do a lookup of the consumer info and resubscribe using the latest info.
-			if consumer != _EMPTY_ && strings.Contains(cinfo.Error.Description, `consumer already exists`) {
+			if consumer != _EMPTY_ && (strings.Contains(cinfo.Error.Description, `consumer already exists`) || strings.Contains(cinfo.Error.Description, `consumer name already in use`)) {
 				info, err = js.ConsumerInfo(stream, consumer)
-				if err != nil && err.Error() != "nats: consumer not found" {
+				if err != nil {
```
wallyqs, line 1046: @derekcollison the scope of this PR changed since server fixed most of the issues plus some added tests, now it only is a bugfix to error when consumer lookup fails and handle `consumer name already in use` errors.

js.go
wallyqs: Fixed by the server, so removed this code path.

derekcollison (approved):
LGTM

PR #717, opened 2021-04-21
jsm: Check that stream name is valid on AddStream // StreamInfo

wallyqs:

### DIFF
--- a/jsm.go
+++ b/jsm.go
@@ -532,6 +532,10 @@ func (js *js) AddStream(cfg *StreamConfig, opts ...JSOpt) (*StreamInfo, error) {
 		return nil, ErrStreamNameRequired
 	}
 
+	if strings.Contains(cfg.Name, ".") {
+		return nil, ErrInvalidStreamName
+	}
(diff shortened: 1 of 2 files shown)


kozlovic (approved):
LGTM

jsm.go
```diff
@@ -532,6 +532,10 @@ func (js *js) AddStream(cfg *StreamConfig, opts ...JSOpt) (*StreamInfo, error) {
 		return nil, ErrStreamNameRequired
 	}
 
+	if strings.Contains(cfg.Name, ".") {
```
derekcollison, line 535: There is a function in the server.

```go
func isValidName(name string) bool {
	if name == "" {
		return false
	}
	return !strings.ContainsAny(name, ".*>")
}
```

derekcollison (commented):
Would switch to this function to match server.

jsm.go
wallyqs: ok will open new PR to switch to that one

PR #718, opened 2021-04-23
Fixed travis script

kozlovic:
In PR #705, we introduced use of `-modfile` with a separate go_test.mod
file. As part of this update, we used to have a `go test -i -race ./...`
and it was mentioned that `-i` was deprecated. But the whole line should
have been removed, not use `-i`. The way it is since that PR is that
all tests are run with `-race` and potentially twice.
Also, a `-modfile` was missing when running the tests under to go version
conditional.


derekcollison (approved):
LGTM

wallyqs (approved):
LGTM.

PR #720, opened 2021-04-28
Add nats.Header type based on http.Header

wallyqs:
This adds a `nats.Header` type that can be used interchangeably with the common `http.Header` from the Go std lib, but allowing us to control the case of the keys and avoid any type of normalization happening in the API. It implements a subset of the interface of the `http.Header` with exception of [WriteSubset](https://golang.org/pkg/net/http/#Header.WriteSubset) and [Write](https://golang.org/pkg/net/http/#Header.Write) which do not apply to NATS.  

Internally,  the client still uses `net/http` when writing the header, and to get http v1.1 like canonical names in a header, it is also possible to just cast into `http.Header`:

```go
	hdr := nats.Header{}
	http.Header(hdr).Set("accept-encoding", "json")
	http.Header(hdr).Add("AUTHORIZATION", "s3cr3t")
```


nats.go
```diff
@@ -3001,11 +3001,49 @@ func (nc *Conn) Publish(subj string, data []byte) error {
 	return nc.publish(subj, _EMPTY_, nil, data)
 }
 
+// Header represents the optional Header for a NATS message,
+// based on the implementation of http.Header.
+type Header map[string][]string
+
+// Add adds the key, value pair to the header.
+// It appends to any existing values associated with key.
+func (h Header) Add(key, value string) {
```
kozlovic, line 3010: I would be explicit and have a one liner for each APIs that says that the key is case-sensitive. I understand that we say that in the PR, but I want user to be able to see it when using this API.

kozlovic (changes requested):
Make it explicit in doc of Set/Get/etc.. APIs that the key is case sensitive.

nats.go
wallyqs: Good idea, ok I updated the godoc. Thanks.

kozlovic (approved):
LGTM

ColinSullivan1 (approved):
LGTM.

PR #721, opened 2021-04-29
js: Add async sub for PullSubscribe to make fetch requests

wallyqs:
This makes the pull based subscription have their own requests handler to make fetch requests using a single wildcard subscription similar to how `nc.Request` works. Unlike `nc.Request` though, it does not use a token to identify response to requestor since the inbox gets rewritten when the JS message is delivered, and just sends the latest response to the first available requestor.

### DIFF
--- a/js.go
+++ b/js.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"math/rand"
 	"strconv"
 	"strings"
 	"sync"
@@ -398,6 +399,27 @@ func (paf *pubAckFuture) Msg() *Msg {
 	paf.js.mu.RLock()
 	defer paf.js.mu.RUnlock()
 	return paf.msg
+}
+
+// pullSubscribe creates the wildcard subscription used per pull subscriber
+// to make fetch requests.
+func (js *js) pullSubscribe(subj string) (*Subscription, error) {
+	jsi := &jsSub{js: js, pull: true}
+
+	// Similar to async request handler we create a wildcard subscription for making requests,
+	// though we do not use the token based approach since we cannot match the response to
+	// the requestor due to JS subject being remapped on delivery.  Instead, we just use an array
+	// of channels similar to how ping/pong interval is handled and send the message to the first
+	// available requestor via a channel.
+	jsi.rr = rand.New(rand.NewSource(time.Now().UnixNano()))
+	jsi.rpre = fmt.Sprintf("%s.", NewInbox())
+	sub, err := js.nc.Subscribe(fmt.Sprintf("%s*", jsi.rpre), jsi.handleFetch)
+	if err != nil {
+		return nil, err
+	}
+	jsi.psub = sub
+
+	return &Subscription{Subject: subj, conn: js.nc, typ: PullSubscription, jsi: jsi}, nil
 }
 
 // For quick token lookup etc.
@@ -810,7 +832,15 @@ type nextRequest struct {
 
 // jsSub includes JetStream subscription info.
 type jsSub struct {
-	js       *js
+	js *js
+
+	// To setup request mux handler for pull subscribers.
+	mu    sync.RWMutex
+	psub  *Subscription
+	rpre  string
+	rr    *rand.Rand
+	freqs []chan *Msg
+
 	consumer string
 	stream   string
 	deliver  string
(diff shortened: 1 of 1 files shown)


js.go
```diff
@@ -400,6 +401,34 @@ func (paf *pubAckFuture) Msg() *Msg {
...
+func (js *js) pullSubscribe(subj string) (*Subscription, error) {
+	jsi := &jsSub{js: js, pull: true}
+
+	// Similar to async request handler we create a wildcard subscription for making requests,
+	// though we do not use the token based approach since we cannot match the response to
+	// the requestor due to JS subject being remapped on delivery.  Instead, we just use an array
+	// of channels similar to how ping/pong interval is handled and send the message to the first
+	// available requestor via a channel.
+	sha := sha256.New()
+	sha.Write([]byte(nuid.Next()))
```
derekcollison, line 415: If using sha can just pull crypto.Rand() as well. Not sure if this was copied.

js.go
```diff
@@ -400,6 +401,34 @@ func (paf *pubAckFuture) Msg() *Msg {
 	return paf.msg
 }
 
+// pullSubscribe creates the wildcard subscription used per pull subscriber
+// to make fetch requests.
+func (js *js) pullSubscribe(subj string) (*Subscription, error) {
+	jsi := &jsSub{js: js, pull: true}
+
+	// Similar to async request handler we create a wildcard subscription for making requests,
```
derekcollison, line 409: Is this for all pull subscribers? There should be one of these for each pull subscriber instance I believe.

js.go
```diff
@@ -810,7 +839,15 @@ type nextRequest struct {
 
 // jsSub includes JetStream subscription info.
 type jsSub struct {
-	js       *js
+	js *js
+
+	// To setup request mux handler for pull subscribers.
+	mu    sync.RWMutex
```
derekcollison, line 845: I don't think we want a mux here. We want one per pull subscriber. Otherwise we will have issues.

js.go
```diff
@@ -826,6 +863,67 @@ type jsSub struct {
...
+	sb.Write(b[:])
+	return sb.String()
+}
+
+// handleFetch is delivered a message requested by pull subscribers
+// when calling Fetch.
+func (jsi *jsSub) handleFetch(m *Msg) {
+	jsi.mu.Lock()
+	if len(jsi.freqs) > 0 {
+		mch := jsi.freqs[0]
```
derekcollison, line 888: This pattern will bite us ;)

derekcollison (changes requested):

js.go
wallyqs: yes it is one per pull subscribers, because they have their own queue of inflight requests

js.go
wallyqs: I wonder if this would work better as a channel of channels...

derekcollison: Let's do the append trick we discussed.

derekcollison: And add to ping/pong as needed.

js.go
```diff
@@ -400,6 +401,28 @@ func (paf *pubAckFuture) Msg() *Msg {
...
+func (js *js) pullSubscribe(subj string) (*Subscription, error) {
+	jsi := &jsSub{js: js, pull: true}
+
+	// Similar to async request handler we create a wildcard subscription for making requests,
+	// though we do not use the token based approach since we cannot match the response to
+	// the requestor due to JS subject being remapped on delivery.  Instead, we just use an array
+	// of channels similar to how ping/pong interval is handled and send the message to the first
+	// available requestor via a channel.
+	jsi.rr = rand.New(rand.NewSource(time.Now().UnixNano()))
+	jsi.rpre = fmt.Sprintf("%s.", NewInbox())
```
wallyqs, line 415: Changed to just use `nats.NewInbox` instead which should work for this case.

js.go
wallyqs: Sending different PR for the ping/pong change

wallyqs: https://github.com/nats-io/nats.go/pull/723

js.go
```diff
@@ -826,6 +857,67 @@ type jsSub struct {
...
+	sb.Write(b[:])
+	return sb.String()
+}
+
+// handleFetch is delivered a message requested by pull subscribers
+// when calling Fetch.
+func (jsi *jsSub) handleFetch(m *Msg) {
+	jsi.mu.Lock()
+	if len(jsi.freqs) == 0 {
+		jsi.mu.Unlock()
```
derekcollison, line 882: This something we should alert user about?

js.go
```diff
@@ -826,6 +857,67 @@ type jsSub struct {
...
+		return
+	}
+	mch := jsi.freqs[0]
+	jsi.freqs = append(jsi.freqs[:0], jsi.freqs[1:]...)
```
derekcollison, line 886: Check if we have more than 1, if not optimize that case with a reslice to [:0]

derekcollison (changes requested):

js.go
wallyqs: In theory this should not happen (unless there is a race in the code), but maybe we should send an async error just in case it occurs

wallyqs: Added dispatching the error to the async error handler just in case it might happen

js.go
wallyqs: Updated this one too

js.go
```diff
@@ -400,6 +401,28 @@ func (paf *pubAckFuture) Msg() *Msg {
...
+	jsi := &jsSub{js: js, pull: true}
+
+	// Similar to async request handler we create a wildcard subscription for making requests,
+	// though we do not use the token based approach since we cannot match the response to
+	// the requestor due to JS subject being remapped on delivery.  Instead, we just use an array
+	// of channels similar to how ping/pong interval is handled and send the message to the first
+	// available requestor via a channel.
+	jsi.rr = rand.New(rand.NewSource(time.Now().UnixNano()))
+	jsi.rpre = fmt.Sprintf("%s.", NewInbox())
+	jsi.freqs = make([]chan *Msg, 0)
```
derekcollison, line 416: Would remove.

derekcollison (approved):
LGTM - one minor suggestion.

derekcollison (approved):
LGTM

PR #723, opened 2021-04-29
Fix to ping/pong processing

wallyqs:

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -2810,7 +2810,7 @@ func (nc *Conn) processPong() {
 	nc.mu.Lock()
 	if len(nc.pongs) > 0 {
 		ch = nc.pongs[0]
-		nc.pongs = nc.pongs[1:]
+		nc.pongs = append(nc.pongs[:0], nc.pongs[1:]...)
 	}
 	nc.pout = 0
 	nc.mu.Unlock()


nats.go
```diff
@@ -2810,7 +2810,7 @@ func (nc *Conn) processPong() {
 	nc.mu.Lock()
 	if len(nc.pongs) > 0 {
 		ch = nc.pongs[0]
-		nc.pongs = nc.pongs[1:]
+		nc.pongs = append(nc.pongs[:0], nc.pongs[1:]...)
```
kozlovic, line 2813: Any issue specifically that this is addressing? Is that to prevent reallocating/expanding the slice?

derekcollison: Yes this pattern which I have used I realized with JS work is unbounded growth..

derekcollison (approved):
LGTM

nats.go
wallyqs: before this change noticed it would be reallocating and expanding the capacity each time, so len + cap on the array would have look like this:

```
len cap
245 280
244 279
243 278
242 277
241 276
240 275
239 274
```

and with this change the capacity grows like this instead:

```
len cap
10 16
14 16
16 16
...
29 32
32 32
31 32
...
43 64
42 64
41 64
...
```

kozlovic: I tried to search in our code base if we do that in other places (which I thought we were doing) but could not find. Anyway, let's keep an eye on that pattern and fix if we stumble on it.

kozlovic (approved):
LGTM

PR #724, opened 2021-04-29
Fixed SubscribeSync that could possibly share channel across 2 subs

kozlovic:
Also added missing subscription's lock when setting some fields
at the end of initialization.

Fixed some flappers and missing defers.

### DIFF
--- a/test/js_test.go
+++ b/test/js_test.go
@@ -3465,6 +3465,7 @@ func TestJetStream_UnsubscribeCloseDrain(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Unexpected error: %v", err)
 	}
+	defer mc.Close()
 	jsm, err := mc.JetStream()
 	if err != nil {
 		t.Errorf("Unexpected error: %v", err)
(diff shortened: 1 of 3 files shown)


test/js_test.go
```diff
@@ -3844,7 +3845,9 @@ func TestJetStreamSubscribe_RateLimit(t *testing.T) {
 	}
 
 	// Change rate limit.
-	recvd := make(chan *nats.Msg)
+	// Make the receive channel able to possibly hold ALL messages, but
+	// we expect it to hold less due to rate limiting.
+	recvd := make(chan *nats.Msg, totalMsgs)
```
kozlovic, line 3850: Not sure exactly about the intent of this test, but since recvd channel was not buffered, it would block in the message callback trying to send message to it (since no one is reading from this channel).
Maybe you wanted to make sure that we don't put in the channel more than the rate limit? Anyway, feel free to modify this, but clearly the recvd channel cannot be unbuffered if no routine receives from it.

derekcollison (approved):
LGTM - Still concerned on sub details post creating the subscription itself.

wallyqs (approved):

variadico (approved):

test/js_test.go
wallyqs: Thanks, it should have been trying to buffer all messages. I used a channel there to avoid races when checking length with an array of messages.

PR #726, opened 2021-05-03
[FIXED] Possible data race on auto-unsubscribe

kozlovic:
There was a possible data race between AutoUnsubscribe() and the go routine dispatching messages.


derekcollison (approved):
LGTM

PR #730, opened 2021-05-14
Add ability to check certificates for OCSP and verify connection

variadico:
This adds the option `nats.EnforceOCSP()` which checks if a server certificate has `status_request` set. If so, then it checks for a valid OCSP response. If the response is invalid, then the TLS connection is aborted.

TODO:
- [ ] Add tests

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -19,6 +19,7 @@ import (
 	"bytes"
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/asn1"
 	"encoding/base64"
 	"encoding/json"
 	"errors"
@@ -42,6 +43,7 @@ import (
 	"github.com/nats-io/nats.go/util"
 	"github.com/nats-io/nkeys"
 	"github.com/nats-io/nuid"
+	"golang.org/x/crypto/ocsp"
 )
 
 // Default Constants
@@ -430,6 +432,11 @@ type Options struct {
 	// For websocket connections, indicates to the server that the connection
 	// supports compression. If the server does too, then data will be compressed.
 	Compression bool
+
+	// EnforceOCSP enables verifying OCSP responses in server certificates. The
+	// certificate needs to have OCSP enabled as well. If verification fails,
+	// then the TLS connection will be aborted.
+	EnforceOCSP bool
 }
 
 const (
@@ -734,6 +741,16 @@ func Secure(tls ...*tls.Config) Option {
 		if len(tls) == 1 {
 			o.TLSConfig = tls[0]
 		}
+		return nil
+	}
+}
+
+// EnforceOCSP enables verifying OCSP responses in server certificates. The
+// certificate needs to have OCSP enabled as well. If verification fails, then
+// the TLS connection will be aborted.
+func EnforceOCSP() Option {
+	return func(o *Options) error {
+		o.EnforceOCSP = true
 		return nil
 	}
 }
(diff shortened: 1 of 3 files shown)


nats.go
```diff
@@ -1683,6 +1700,72 @@ func (nc *Conn) createConn() (err error) {
...
+}
+
+// verifyOCSP verifies that if a server certificate has "status_request" set,
+// then the certificate has a valid OCSP response.
+func (nc *Conn) verifyOCSP(s tls.ConnectionState) error {
+	if !nc.Opts.EnforceOCSP {
+		return nil
+	}
+
+	statusStr := map[int]string{
```
ColinSullivan1, line 1734: This could be moved inside the if below, or since there is only two values you care about might be able to be removed entirely with an if.

variadico: Great suggestion!

wallyqs:
Note that this option is mostly a helper since it is possible to enable OCSP response verification with the following code already:

```go
func getOCSPStatus(s tls.ConnectionState) (*ocsp.Response, error) {
	if len(s.VerifiedChains) == 0 {
		return nil, fmt.Errorf("missing TLS verified chains")
	}
	chain := s.VerifiedChains[0]

	if got, want := len(chain), 2; got < want {
		return nil, fmt.Errorf("incomplete cert chain, got %d, want at least %d", got, want)
	}
	leaf, issuer := chain[0], chain[1]

	resp, err := ocsp.ParseResponseForCert(s.OCSPResponse, leaf, issuer)
	if err != nil {
		return nil, fmt.Errorf("failed to parse OCSP response: %w", err)
	}
	if err := resp.CheckSignatureFrom(issuer); err != nil {
		return resp, err
	}
	return resp, nil
}
	nc, err := nats.Connect(fmt.Sprintf("tls://localhost:%d", opts.Port),
		nats.Secure(&tls.Config{
			VerifyConnection: func(s tls.ConnectionState) error {
				resp, err := getOCSPStatus(s)
				if err != nil {
					return err
				}
				if resp.Status != ocsp.Good {
					return fmt.Errorf("invalid staple")
				}
				return nil
			},
		}),
		nats.RootCAs(caCert),
		nats.ErrorHandler(noOpErrHandler),
	)
```

nats.go
```diff
@@ -1683,6 +1700,70 @@ func (nc *Conn) createConn() (err error) {
...
+	// https://tools.ietf.org/html/rfc6066
+
+	for _, ext := range cert.Extensions {
+		if !ext.Id.Equal(tlsFeatures) {
+			continue
+		}
+		if len(ext.Value) != tlsFeaturesLen {
+			continue
+		}
+		return ext.Value[tlsFeaturesLen-1] == statusRequestExt
```
philpennock, line 1721: This has the same problem as the similar logic in the nats-server PR, and needs the same fix of using `asn1.Unmarshal()`

nats.go
```diff
@@ -1683,6 +1700,70 @@ func (nc *Conn) createConn() (err error) {
...
+	}
+
+	return false
+}
+
+// verifyOCSP verifies that if a server certificate has "status_request" set,
+// then the certificate has a valid OCSP response.
+func (nc *Conn) verifyOCSP(s tls.ConnectionState) error {
+	if !nc.Opts.EnforceOCSP {
```
philpennock, line 1730: This doesn't match the function description.
AIUI the desired logic is:
 **IF _either_ the certificate has status_request set _or_ the client was explicitly told to require OCSP, THEN verify OCSP**

nats.go
```diff
@@ -1683,6 +1700,70 @@ func (nc *Conn) createConn() (err error) {
...
+	chain := s.VerifiedChains[0]
+
+	if got, want := len(chain), 2; got < want {
+		return fmt.Errorf("incomplete cert chain, got %d, want at least %d", got, want)
+	}
+	leaf, issuer := chain[0], chain[1]
+
+	if !hasOCSPStatusRequest(leaf) {
+		// "status_request" not set, no need to do anything.
+		return nil
```
philpennock, line 1746: Similarly here, it's "the cert says OCSP is required" OR "the client app says OCSP is required".

philpennock (changes requested):

variadico:
Closing for now.

PR #733, opened 2021-05-24
Update nats-server test dep

variadico:
This updates the nats-server dependency used in tests to latest master.

I ran these commands.
```
go get -modfile=go_test.mod github.com/nats-io/nats-server/v2@master
go mod tidy -modfile=go_test.mod
```

I also had to update `TestJetStreamCrossAccountMirrorsAndSources` because the server now returns an error if replicas is greater than 1, but the server isn't clustered.


kozlovic:
@variadico Do we need that right away? We may release 2.2.6 today, so maybe we can wait for the server release?

variadico:
@kozlovic Ah, sure thing. This can definitely wait for the server release.

wallyqs (approved):
LGTM

kozlovic (approved):
LGTM

PR #734, opened 2021-05-24
[added] option to not create durable on subscribe

matthiashanel:
Signed-off-by: Matthias Hanel <mh@synadia.com>

When exporting/importing say the next subject, the consumer exists already and does not need to be created.
Thus the client should also not try to look it up.

We went down the road of a don't create durable option (1st commit) but altered it to have a dedicated JS create call.
We believe that using `BindJetStream` might be easier to explain

### DIFF
--- a/js.go
+++ b/js.go
@@ -183,21 +183,24 @@ func (nc *Conn) JetStream(opts ...JSOpt) (JetStreamContext, error) {
 	}
 
 	// If we have check recently we can avoid another account lookup here.
-	// We want these to be lighweight and created at will.
-	nc.mu.Lock()
-	now := time.Now()
-	checkAccount := now.Sub(nc.jsLastCheck) > defaultAccountCheck
-	if checkAccount {
-		nc.jsLastCheck = now
-	}
-	nc.mu.Unlock()
+	// We want these to be lightweight and created at will.
+	// The account check is permanently disabled when the api prefix is set to anything but default
+	if js.opts.pre == defaultAPIPrefix {
+		nc.mu.Lock()
+		now := time.Now()
+		checkAccount := now.Sub(nc.jsLastCheck) > defaultAccountCheck
+		if checkAccount {
+			nc.jsLastCheck = now
+		}
+		nc.mu.Unlock()
 
-	if checkAccount {
-		if _, err := js.AccountInfo(); err != nil {
-			if err == ErrNoResponders {
-				err = ErrJetStreamNotEnabled
+		if checkAccount {
+			if _, err := js.AccountInfo(); err != nil {
+				if err == ErrNoResponders {
+					err = ErrJetStreamNotEnabled
+				}
+				return nil, err
 			}
-			return nil, err
 		}
 	}
 
@@ -1041,7 +1044,9 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync 
 	// With an explicit durable name, then can lookup
 	// the consumer to which it should be attaching to.
 	consumer = o.cfg.Durable
-	if consumer != _EMPTY_ {
+	if consumer == _EMPTY_ && o.noDurCreate {
+		return nil, fmt.Errorf("nats: NoCreate requires a durable name")
+	} else if consumer != _EMPTY_ && !o.noDurCreate {
 		// Only create in case there is no consumer already.
 		info, err = js.ConsumerInfo(stream, consumer)
 		if err != nil && err.Error() != "nats: consumer not found" {
@@ -1049,7 +1054,10 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync 
 		}
 	}
 
-	if info != nil {
+	if o.noDurCreate {
+		attached = true
+		deliver = subj
+	} else if info != nil {
 		// Attach using the found consumer config.
 		ccfg = &info.Config
 		attached = true
@@ -1351,6 +1359,8 @@ type subOpts struct {
 	mack bool
 	// For creating or updating.
 	cfg *ConsumerConfig
+	// Assert consumer exists and do not create it
+	noDurCreate bool
 }
 
 // ManualAck disables auto ack functionality for async subscriptions.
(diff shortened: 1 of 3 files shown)


js.go
```diff
@@ -166,8 +169,9 @@ const (
 	defaultAccountCheck = 20 * time.Second
 )
 
-// JetStream returns a JetStreamContext for messaging and stream management.
-func (nc *Conn) JetStream(opts ...JSOpt) (JetStreamContext, error) {
+// BindJetStream returns a JetStreamContext for messaging and stream management that will NOT create
+// underlying objects like stream or durable. It assumes they exist already.
```
ColinSullivan1, line 173: I don't think this is necessary here.

derekcollison (commented):
Help me understand when this option would be used?

matthiashanel:
@derekcollison  This new call would be used whenever you don't want the client to create durables/streams automagically.
This will help with export/import or when you have restrictive permissions

js.go
ColinSullivan1: Reading the diff, this looks like it's on an internal API.  No problem keeping it if you feel it's important, it's just identical to the exported API below on line 191.

derekcollison:
I guess that is what I was looking for clarity on since now we bind to existing durables if they exist and don't always create them, so wanted to make sure I understood the difference in this case.

matthiashanel:
if you call `JetStream` then consumer etc will be created. 
if you call `BindJetStream` we assume they exist. 

I can add an exception if no durable name was provided an an ephemeral durable needs creating.
comments?  @wallyqs

derekcollison:
When is their a case where you can to have a durable option like we do now but not create?

This because in those situations you do not want the client lib doing its probing to see which direction it should go with a durable etc?

matthiashanel:
Exactly, when I only export the next subject or have restrictive permissions I want to avoid probing around.

js.go
```diff
@@ -181,9 +185,28 @@ func (nc *Conn) JetStream(opts ...JSOpt) (JetStreamContext, error) {
 			return nil, err
 		}
 	}
+	return js, nil
+}
+
+// BindJetStream returns a JetStreamContext for messaging and stream management that will NOT create
+// underlying objects like stream or durable. It assumes they exist already.
+func (nc *Conn) BindJetStream(opts ...JSOpt) (JetStreamContext, error) {
```
wallyqs, line 193: I think just JetStream interface might work better since JetStreamManagement requires being able to use the API.
```suggestion
func (nc *Conn) BindJetStream(opts ...JSOpt) (JetStream, error) {
```

matthiashanel: I would be ok if you can do that as these calls to create streams/durables are explicit.

js.go
matthiashanel: sorry my bad. removed

js.go
wallyqs: ok sounds good

nats.go
```diff
@@ -135,6 +135,8 @@ var (
 	ErrNoContextOrTimeout           = errors.New("nats: no context or timeout given")
 	ErrPullModeNotAllowed           = errors.New("nats: pull based not supported")
 	ErrJetStreamNotEnabled          = errors.New("nats: jetstream not enabled")
+	ErrBoundJetStreamStream         = errors.New("nats: a bound JS requires a stream name")
+	ErrBoundJetStreamDurable        = errors.New("nats: a bound JS requires a stream name")
```
wallyqs, line 139:
```suggestion
	ErrBoundJetStreamDurable        = errors.New("nats: a bound JS requires a durable name")
```

test/js_test.go
```diff
@@ -2344,14 +2344,14 @@ func TestJetStreamImportDirectOnly(t *testing.T) {
 	}
 
 	// Cannot subscribe with JS context from another account right now.
```
wallyqs, line 2346: We can remove this line now.
```suggestion
```

wallyqs:
As it is, this is replacing the `nats.Direct` option that used to do something similar to disable using the JetStream API, but instead of being an option it adds a new constructor called `js, err := nc.BindJetStream()`, that similar to the `nats.BindStream()` option assumes that there are JetStream streams/consumers are already bound to the current account due to an import for example, so it does not lookup or attempts to create them which can't be done in environments with limited permissions.

ColinSullivan1:
IIRC, with direct mode there was a potential problem of overlapping acks with duplicate consumers.   Does that issue present itself here?

matthiashanel:
> IIRC, with direct mode there was a potential problem of overlapping acks with duplicate consumers. Does that issue present itself here?

@ColinSullivan1 can you explain how this issue showed itself?

wallyqs:
I think that @ColinSullivan1 means that if you run Subscribe multiple times and they are not a queue subscriber with the same durable name, they can all receive the same message that can be acked.

test/js_test.go
```diff
@@ -2343,15 +2351,21 @@ func TestJetStreamImportDirectOnly(t *testing.T) {
...
 	}
-	if _, err = js.PullSubscribe("ORDERS", "d1", nats.BindStream("ORDERS")); err != nats.ErrJetStreamNotEnabled {
-		t.Fatalf("Expected an error of '%v', got '%v'", nats.ErrJetStreamNotEnabled, err)
+	if sub, err = js.PullSubscribe("ORDERS", "d1", nats.BindStream("ORDERS")); err != nil {
+		t.Fatalf("Expected no error, got '%v'", err)
+	} else if m, err := sub.Fetch(1); err != nil {
+		t.Fatalf("expected no error, got %v", err)
+	} else if len(m) != 1 {
+		t.Fatalf("expected one message, got %d", len(m))
+	} else if err = m[0].Ack(); err != nil {
```
wallyqs, line 2367: Can you change to `AckSync` I think it would work as well

```suggestion
	} else if err = m[0].AckSync(); err != nil {
```

js.go
```diff
@@ -181,9 +183,29 @@ func (nc *Conn) JetStream(opts ...JSOpt) (JetStreamContext, error) {
 			return nil, err
 		}
 	}
+	return js, nil
+}
+
+// BindJetStream returns a JetStreamContext for messaging and stream management that will NOT create
+// underlying objects like stream or durable. It assumes they exist already.
```
wallyqs, line 190: Maybe expand a bit on the godoc that is to bind to a JetStream from another account?
```suggestion
// underlying objects like stream or durable on Subscribe. It assumes they exist already and that
// the JetStream context is bound to another account via an import with limited permissions for example.
```

wallyqs (approved):
I think whenever JetStream is imported from another account we could refer to it as being bound to the account as an import, LGTM.

matthiashanel:
@ColinSullivan1 @derekcollison is this change in it's current form ok with you?

derekcollison:
Let's setup a quick meeting for tomorrow to discuss. You and Colin and I and Wally..

wallyqs:
Fixed via #740

PR #737, opened 2021-05-27
Remove unused code from readMIMEHeader

rutgerbrf:
The `strs` slice is not used - it seems to have stuck around from [`net/textproto.ReadMIMEHeader`](https://golang.org/pkg/net/textproto/#Reader.ReadMIMEHeader).


variadico (approved):
Ah, ok. `strs` never gets initialized or modified so its length is always zero. And because it's always zero, we don't need that if statement.

Looks like in the original code there was a possibility of `strs` being initialized, but our code is a bit different and doesn't include that case.

LGTM! Thanks!

PR #739, opened 2021-06-01
[changed] JetStream function to not lookup account info & moved it

matthiashanel:
The new function is JSAccountInfo

Signed-off-by: Matthias Hanel <mh@synadia.com>

This separates concerns into give me a jetstream object and check jetstream.
Moving that logic to the nats connection to avoid dependencies with the JetStream call.
In our meeting the group generally favored opt in over opt out. 

I'm clearly expecting ppl to either copy a sample anyway and there we add the call of course.
Or if they run into this, to just ask & find this new call.
it's also opt in because the main thing that the lookup does is provide a different error. 

I noticed that the unit test for the js object appears to be broken right now.  (at least for me, even without my change)
Deleted the checks so I could show you this sooner.

If that approach is not desirec, I'll revert and add an option to opt out of the lookup.

### DIFF
--- a/js.go
+++ b/js.go
@@ -167,6 +167,7 @@ const (
 )
 
 // JetStream returns a JetStreamContext for messaging and stream management.
+// Errors are only returned if inconsistent options are provided.
 func (nc *Conn) JetStream(opts ...JSOpt) (JetStreamContext, error) {
 	js := &js{
 		nc: nc,
@@ -181,26 +182,6 @@ func (nc *Conn) JetStream(opts ...JSOpt) (JetStreamContext, error) {
 			return nil, err
 		}
 	}
-
-	// If we have check recently we can avoid another account lookup here.
-	// We want these to be lighweight and created at will.
-	nc.mu.Lock()
-	now := time.Now()
-	checkAccount := now.Sub(nc.jsLastCheck) > defaultAccountCheck
-	if checkAccount {
-		nc.jsLastCheck = now
-	}
-	nc.mu.Unlock()
-
-	if checkAccount {
-		if _, err := js.AccountInfo(); err != nil {
-			if err == ErrNoResponders {
-				err = ErrJetStreamNotEnabled
-			}
-			return nil, err
-		}
-	}
-
 	return js, nil
 }
 
--- a/jsm.go
+++ b/jsm.go
@@ -170,6 +170,8 @@ type accountInfoResponse struct {
 }
 
 // AccountInfo retrieves info about the JetStream usage from the current account.
+// If JetStream is not enabled, this will return ErrJetStreamNotEnabled
+// Other errors can happen but are generally considered retryable
 func (js *js) AccountInfo(opts ...JSOpt) (*AccountInfo, error) {
 	o, cancel, err := getJSContextOpts(js.opts, opts...)
 	if err != nil {
@@ -181,6 +183,10 @@ func (js *js) AccountInfo(opts ...JSOpt) (*AccountInfo, error) {
 
 	resp, err := js.nc.RequestWithContext(o.ctx, js.apiSubj(apiAccountInfo), nil)
 	if err != nil {
+		// todo maybe nats server should never have no responder on this subject and always respond if they know there is no js to be had
+		if err == ErrNoResponders {
+			err = ErrJetStreamNotEnabled
+		}
 		return nil, err
 	}
 	var info accountInfoResponse
--- a/nats.go
+++ b/nats.go
@@ -496,9 +496,6 @@ type Conn struct {
 	respMux   *Subscription        // A single response subscription
 	respMap   map[string]chan *Msg // Request map for the response msg channels
 	respRand  *rand.Rand           // Used for generating suffix
-
-	// JetStream Contexts last account check.
-	jsLastCheck time.Time
 }
 
 type natsReader struct {
--- a/test/js_test.go
+++ b/test/js_test.go
@@ -42,7 +42,11 @@ func TestJetStreamNotEnabled(t *testing.T) {
 	}
 	defer nc.Close()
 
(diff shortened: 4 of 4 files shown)


js.go
```diff
@@ -192,15 +192,6 @@ func (nc *Conn) JetStream(opts ...JSOpt) (JetStreamContext, error) {
 	}
 	nc.mu.Unlock()
 
-	if checkAccount {
```
kozlovic, line 195: If that disappears, then above `jsLastCheck` should disappear too and be removed from the `Conn` object.

nats.go
```diff
@@ -2424,6 +2424,47 @@ func (ac *asyncCallbacksHandler) asyncCBDispatcher() {
...
+func (nc *Conn) jsAccountInfo(jOpts *jsOpts, opts ...JSOpt) (*AccountInfo, error) {
+	o, cancel, err := getJSContextOpts(jOpts, opts...)
+	if err != nil {
+		return nil, err
+	}
+	if cancel != nil {
+		defer cancel()
+	}
+
+	resp, err := nc.RequestWithContext(o.ctx, jOpts.apiSubj(apiAccountInfo), nil)
```
kozlovic, line 2439: If user provides a different prefix than the one registered when initially getting the context, then should not this prefix be used instead? In other words, should it be `o.apiSubj()` here? (if I understand correctly `o` is the mix of the default's js options and the user provided one.

kozlovic (commented):
Not commenting on the validity of the move, just noticed inconsistencies.

This is concerning since I just started working on the C client (was waiting for the Go client to be done) and the first function I implemented is the the equivalent of JetStream() (get the context, which would call js.AccountInfo()) and that is already changing :-)

nats.go
matthiashanel: fixed. instead of moving code we now use js.AcccountInfo (that existed before)

js.go
matthiashanel: fixed. JetStream() is now a constructor only. because it can't reallly fail, we can use js.AccountInfo()

jsm.go
```diff
@@ -170,6 +170,8 @@ type accountInfoResponse struct {
 }
 
 // AccountInfo retrieves info about the JetStream usage from the current account.
+// If JetStream is not enabled, this will return ErrJetStreamNotEnabled
```
derekcollison, line 173: This now has to be the case for every API call to the JS context, but I think that is ok.

nats.go
```diff
@@ -496,9 +496,6 @@ type Conn struct {
 	respMux   *Subscription        // A single response subscription
 	respMap   map[string]chan *Msg // Request map for the response msg channels
 	respRand  *rand.Rand           // Used for generating suffix
-
-	// JetStream Contexts last account check.
-	jsLastCheck time.Time
```
derekcollison, line 501: I think we all agreed that JS contexts should be lightweight and easy to create. In many tests I create multiple ones with different characteristics.

Question though, do we expect code to look like this all the time or folks will skip the AccountInfo and just check errors on the other API calls?

```go
nc, _ := nc.Connect("demo.nats.io")
js, _ := nc.JetStream()
if _, err := js.AccountInfo(); err != nil {}
```

kozlovic: Help me understand: can the enabled/disabled nature of JetStream for this connection change overtime? Meaning, after calling JetStream() (and lets say we still do an account check inside, and it returned ok), is there a possibility that JS may no longer be enabled for this account? What if the connection reconnects to a server where this account does not have JS enabled, which likely would be misconfiguration?

If it can't change, then instead of a time check, it should be simply a "already checked and here is the result" type of things (say 2 booleans).

If it can change overtime and since all APIs will return "not enabled" if JS is or no longer is enabled, not sure what we gain by doing it inside JetStream(). Say you already have the context, but then JS is disabled, then the context is "valid" and yet all other APIs call will fail with "not enabled". Same if you recently got a context and we use time-check, then if we are within the window, JetStream() would return a context (which won't work when calling APIs), but if you happen to ask after the time-check has passed, this call would return "not enabled".

I am not against doing the account check inside JetStream() (with the time-check bug fixed), since it can give a "fail fast" behavior where if you get "not enabled", there is no need to proceed further, but it also can be non deterministic (as described above), and all APIs should be checked for success/failure anyway.

jsm.go
matthiashanel: @wallyqs we need to probably go through our public calls and make sure that is so

nats.go
matthiashanel: due to reload or jwt update Js can be legitimately enabled disabled on the fly.
We have code that turns a no responder into the same error, so flapping because of that can happen as well.
If you want to fail fast, perhaps for debugging, I'd suggest call js.AccountInfo. 
I'd also suggest calling that function on retries. you fails for some reason now you can better determine if retrying makes sense or not.

jsm.go
wallyqs: yes could happen if JS becomes disabled

nats.go
wallyqs: I think for users that want to ensure that JS is ready and available, a bit more code would be needed which might be better to have outside of the creation of the `JetStream()` context:

```go
nc, _ := nc.Connect("demo.nats.io")
js, _ := nc.JetStream()

ctx, done := context.WithTimeout(context.Background(), 5*time.Second)
defer done()

// IsJSAvailable()
for {
        select {
        case <-ctx.Done():
                return
        default:
        }

	if _, err := js.AccountInfo(); err != nil {
		// Handle temporary errors:
		//
		// - Quorum not ready
		// - JS API not yet ready
		// - No responders (temporary while JS booting up)

		// Then: timeout when too many failures or no response
	}
}
```

There would be some retries when this happens for example and backoff until is ready.

derekcollison (approved):
LGTM

PR #740, opened 2021-06-01
js: Add nats.Bind option to disable creating consumers

### LINKED ISSUE #735: Option to turn off auto consumer create with PullSub
## Feature Request
Have an option on PullSub to turn off the auto creation of consumer when it doesn't exist.

#### Use Case:
Developers are not creating the consumer with the needed configuration, because the auto create is a hidden behavior in the Pull Subscription. This is leading to subjects and the consumer not matching and messages being lost, test failing, and developers having to deep dive Pull Subscription behavior.

#### Proposed Change:
Add an option to Pull Sub that turns off the creation of the consumer when not found and returns an error.

Everyone 

#### Alternative Approaches
Having to layer on top of the Pull Sub a wrapper that checks if the consumer exists and returning an error before calling the Pull Sub.

wallyqs:
This allows creating subscriptions that are bound to a consumer that matches the durable name, and does not attempt to create the consumer in case it is not present.

Scenarios where this option could be used is when the consumer is bound to a JetStream instance from another account and there are limited permissions, or when a consumer is created/deleted out of band and the subscription should not attempt creation to avoid misconfigurations/configuration drift.

Fixes #735

### DIFF
--- a/js.go
+++ b/js.go
@@ -1008,7 +1008,7 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync 
 		}
 	}
 
-	isPullMode := ch == nil && cb == nil
+	isPullMode := ch == nil && cb == nil && !isSync
 	badPullAck := o.cfg.AckPolicy == AckNonePolicy || o.cfg.AckPolicy == AckAllPolicy
 	hasHeartbeats := o.cfg.Heartbeat > 0
 	hasFC := o.cfg.FlowControl
@@ -1017,16 +1017,25 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync 
 	}
 
 	var (
-		err          error
-		shouldCreate bool
-		ccfg         *ConsumerConfig
-		info         *ConsumerInfo
-		deliver      string
-		attached     bool
-		stream       = o.stream
-		consumer     = o.consumer
-		isDurable    = o.cfg.Durable != _EMPTY_
+		err           error
+		shouldCreate  bool
+		ccfg          *ConsumerConfig
+		info          *ConsumerInfo
+		deliver       string
+		attached      bool
+		stream        = o.stream
+		consumer      = o.consumer
+		isDurable     = o.cfg.Durable != _EMPTY_
+		consumerBound = o.bound
+		notFoundErr   bool
+		lookupErr     bool
 	)
+
+	// In case a consumer has not been set explicitly, then the
+	// durable name will be used as the consumer name.
+	if consumer == _EMPTY_ {
+		consumer = o.cfg.Durable
+	}
 
 	// Find the stream mapped to the subject if not bound to a stream already.
 	if o.stream == _EMPTY_ {
@@ -1038,18 +1047,16 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync 
 		stream = o.stream
 	}
 
-	// With an explicit durable name, then can lookup
-	// the consumer to which it should be attaching to.
-	consumer = o.cfg.Durable
+	// With an explicit durable name, then can lookup the consumer first
+	// to which it should be attaching to.
 	if consumer != _EMPTY_ {
-		// Only create in case there is no consumer already.
 		info, err = js.ConsumerInfo(stream, consumer)
-		if err != nil && err.Error() != "nats: consumer not found" {
-			return nil, err
-		}
+		notFoundErr = err != nil && strings.Contains(err.Error(), "consumer not found")
+		lookupErr = err == ErrJetStreamNotEnabled || err == ErrTimeout || err == context.DeadlineExceeded
 	}
 
-	if info != nil {
+	switch {
+	case info != nil:
 		// Attach using the found consumer config.
 		ccfg = &info.Config
 		attached = true
@@ -1059,12 +1066,25 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync 
 			return nil, ErrSubjectMismatch
 		}
 
+		// Prevent binding a subscription against incompatible consumer types.
+		if isPullMode && ccfg.DeliverSubject != _EMPTY_ {
+			return nil, ErrPullSubscribeToPushConsumer
+		} else if !isPullMode && ccfg.DeliverSubject == _EMPTY_ {
+			return nil, ErrPullSubscribeRequired
+		}
+
 		if ccfg.DeliverSubject != _EMPTY_ {
 			deliver = ccfg.DeliverSubject
-		} else {
+		} else if !isPullMode {
 			deliver = NewInbox()
 		}
-	} else {
+	case (err != nil && !notFoundErr) || (notFoundErr && consumerBound):
+		// If the consumer is being bound got an error on pull subscribe then allow the error.
+		if !(isPullMode && lookupErr && consumerBound) {
+			return nil, err
+		}
+	default:
+		// Attempt to create consumer if not found nor using Bind.
 		shouldCreate = true
 		deliver = NewInbox()
 		if !isPullMode {
@@ -1351,6 +1371,8 @@ type subOpts struct {
 	mack bool
 	// For creating or updating.
 	cfg *ConsumerConfig
+	// For binding a subscription to a consumer without creating it.
+	bound bool
 }
 
 // ManualAck disables auto ack functionality for async subscriptions.
@@ -1362,16 +1384,19 @@ func ManualAck() SubOpt {
 }
 
 // Durable defines the consumer name for JetStream durable subscribers.
-func Durable(name string) SubOpt {
+func Durable(consumer string) SubOpt {
 	return subOptFn(func(opts *subOpts) error {
-		if opts.cfg.Durable != "" {
+		if opts.cfg.Durable != _EMPTY_ {
 			return fmt.Errorf("nats: option Durable set more than once")
 		}
-		if strings.Contains(name, ".") {
+		if opts.consumer != _EMPTY_ && opts.consumer != consumer {
+			return fmt.Errorf("nats: duplicate consumer names (%s and %s)", opts.consumer, consumer)
+		}
+		if strings.Contains(consumer, ".") {
 			return ErrInvalidDurableName
 		}
 
-		opts.cfg.Durable = name
+		opts.cfg.Durable = consumer
 		return nil
 	})
 }
@@ -1482,9 +1507,39 @@ func RateLimit(n uint64) SubOpt {
 }
 
 // BindStream binds a consumer to a stream explicitly based on a name.
-func BindStream(name string) SubOpt {
+func BindStream(stream string) SubOpt {
 	return subOptFn(func(opts *subOpts) error {
-		opts.stream = name
+		if opts.stream != _EMPTY_ && opts.stream != stream {
+			return fmt.Errorf("nats: duplicate stream name (%s and %s)", opts.stream, stream)
+		}
+
+		opts.stream = stream
+		return nil
+	})
+}
+
+// Bind binds a subscription to an existing consumer from a stream without attempting to create.
+// The first argument is the stream name and the second argument will be the consumer name.
+func Bind(stream, consumer string) SubOpt {
+	return subOptFn(func(opts *subOpts) error {
+		if stream == _EMPTY_ {
+			return ErrStreamNameRequired
+		}
+		if consumer == _EMPTY_ {
+			return ErrConsumerNameRequired
+		}
+
+		// In case of pull subscribers, the durable name is a required parameter
+		// so check that they are not different.
+		if opts.cfg.Durable != _EMPTY_ && opts.cfg.Durable != consumer {
+			return fmt.Errorf("nats: duplicate consumer names (%s and %s)", opts.cfg.Durable, consumer)
+		}
+		if opts.stream != _EMPTY_ && opts.stream != stream {
+			return fmt.Errorf("nats: duplicate stream name (%s and %s)", opts.stream, stream)
+		}
+		opts.stream = stream
+		opts.consumer = consumer
+		opts.bound = true
 		return nil
 	})
 }
--- a/nats.go
+++ b/nats.go
@@ -146,9 +146,12 @@ var (
 	ErrInvalidJSAck                 = errors.New("nats: invalid jetstream publish response")
 	ErrMultiStreamUnsupported       = errors.New("nats: multiple streams are not supported")
 	ErrStreamNameRequired           = errors.New("nats: stream name is required")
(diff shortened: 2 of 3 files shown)


matthiashanel (approved):
LGTM

js.go
```diff
@@ -1044,9 +1045,17 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync
 	if consumer != _EMPTY_ {
 		// Only create in case there is no consumer already.
 		info, err = js.ConsumerInfo(stream, consumer)
-		if err != nil && err.Error() != "nats: consumer not found" {
-			return nil, err
+		if err != nil {
+			if strings.Contains(err.Error(), "consumer not found") {
```
derekcollison, line 1049: This feels like could be simplified and possibly documented just a bit better as to what is going on here.

js.go
```diff
@@ -1489,6 +1500,14 @@ func BindStream(name string) SubOpt {
 	})
 }
 
+// BindConsumer binds a subscription to a consumer based on a durable name.
+func BindConsumer() SubOpt {
```
derekcollison, line 1504: By default this happens sometimes on its own if the consumer exists. So possibly the naming might be a bit off here.

BindOnly or NoCreate. Probably not these but hopefully this gets to the point I am trying to make a bit.

nats.go
```diff
@@ -149,6 +149,7 @@ var (
 	ErrConsumerConfigRequired       = errors.New("nats: consumer configuration is required")
 	ErrStreamSnapshotConfigRequired = errors.New("nats: stream snapshot configuration is required")
 	ErrDeliverSubjectRequired       = errors.New("nats: deliver subject is required")
+	ErrDurableNameRequired          = errors.New("nats: durable name of consumer is required")
```
derekcollison, line 152: If we know the name of an ephemeral can we bind to that? I am not sure how that would act today or not but we may not want to restrict to durables here per se.

test/js_test.go
```diff
@@ -5899,3 +5932,72 @@ func TestJetStreamPublishAsyncPerf(t *testing.T) {
...
+	if si == nil || si.Config.Name != "foo" {
+		t.Fatalf("StreamInfo is not correct %+v", si)
+	}
+
+	for i := 0; i < 25; i++ {
+		js.Publish("foo", []byte("hi"))
+	}
+
+	// Durable name required to bind a consumer.
+	_, err = js.SubscribeSync("foo", nats.BindConsumer())
```
derekcollison, line 5971: So looking at these examples, maybe BindOnly?

```go
js.SubscribeSync("foo", nats.BindOnly())
```

derekcollison (commented):
Some comments, mostly around naming. I agree this is useful.

test/js_test.go
wallyqs: I think `nats.BindOnly` sounds good, will update

js.go
wallyqs: Tried to refactor a bit this part

js.go
wallyqs: Updated to use `BindOnly` for now

nats.go
wallyqs: We cannot right now due to some filter subject checks, but it might look something like this.  First need to get the name and delivery subject of the ephemeral then use bind stream and bind only to attach to that consumer without looking up the stream.

```go
	sub, err := js.SubscribeSync("foo")
	if err != nil {
	}
	cinfo, err := sub.ConsumerInfo()
	if err != nil {
		t.Fatal(err)
	}
	sub, err = js.SubscribeSync(cinfo.Config.DeliverSubject, nats.Durable(cinfo.Name), nats.BindStream("foo"), nats.BindOnly())
```

derekcollison: ConsumerInfo should work for both durables and ephemerals.

derekcollison: Which checks?

test/js_test.go
```diff
@@ -2343,15 +2360,31 @@ func TestJetStreamImportDirectOnly(t *testing.T) {
 		}
 	}
 
-	// Cannot subscribe with JS context from another account right now.
-	if _, err := js.SubscribeSync("ORDERS"); err != nats.ErrJetStreamNotEnabled {
-		t.Fatalf("Expected an error of '%v', got '%v'", nats.ErrJetStreamNotEnabled, err)
+	// Cannot create ephemeral subscriber with JS context.
+	if _, err := js.SubscribeSync("ORDERS", nats.BindOnly()); err != nats.ErrDurableNameRequired {
```
derekcollison, line 2364: I think we should be able to bind to any pre-existing consumer.

nats.go
wallyqs: I meant these [filter subject checks](https://github.com/nats-io/nats.go/blob/master/js.go#L1039-L1041) but found that it would work using the `nats.Durable` name for the consumer name even though it is ephemeral since that helps making the lookup of the consumer. Added examples of using BindOnly with ephemeral consumers, all these should work atm:

```go
	// Create ephemeral consumer
	sub1, err := js.SubscribeSync("foo")
	if err != nil {
	}
	cinfo, err := sub1.ConsumerInfo()
	if err != nil {
		t.Fatal(err)
	}

	// Bind to ephemeral consumer by setting the name using nats.Durable
	sub2, err := js.SubscribeSync("foo", nats.Durable(cinfo.Name), nats.BindStream("foo"), nats.BindOnly())
	if err != nil {
	}
	sub3, err := nc.SubscribeSync(cinfo.Config.DeliverSubject)
	if err != nil {
	}
```

test/js_test.go
wallyqs: It is possible to bind to ephemerals as well but have to use `nats.Durable` to set the consumer name

js.go
```diff
@@ -1059,12 +1063,23 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync
 			return nil, ErrSubjectMismatch
 		}
 
+		// Prevent binding a subscription against incompatible consumer types.
+		if isPullMode && ccfg.DeliverSubject != _EMPTY_ {
+			return nil, fmt.Errorf("nats: cannot pull subscribe to push based consumer")
```
derekcollison, line 1068: Let's make these errors constants since we don't fill anything in.

js.go
```diff
@@ -1059,12 +1063,23 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync
...
+		}
+
 		if ccfg.DeliverSubject != _EMPTY_ {
 			deliver = ccfg.DeliverSubject
-		} else {
+		} else if !isPullMode {
 			deliver = NewInbox()
 		}
-	} else {
+	case err != nil && !notFoundErr || notFoundErr && consumerBound:
```
derekcollison, line 1078: Let's put parens around these to clarify intent.

```go
case (err != nil && !notFoundErr) || (notFoundErr && consumerBound)
```

js.go
```diff
@@ -1346,11 +1361,13 @@ func (js *js) lookupStreamBySubject(subj string) (string, error) {
...
 type subOpts struct {
 	// For attaching.
-	stream, consumer string
+	stream string
 	// For manual ack
 	mack bool
 	// For creating or updating.
 	cfg *ConsumerConfig
+	// For binding a subscription to a consumer without creating it.
+	cbound bool
```
derekcollison, line 1370: I think just bound since subOpts.

nats.go
derekcollison: Feels off though to use Durable when its an ephemeral, at least to me.

In the call above with BindOnly, will we always need to provide the stream and consumer name? If so why not just make them args to BindOnly?

test/js_test.go
derekcollison: See above.

derekcollison (changes requested):

js.go
```diff
@@ -1482,9 +1503,49 @@ func RateLimit(n uint64) SubOpt {
...
+		if stream == _EMPTY_ {
+			return ErrStreamNameRequired
+		}
+
+		// In case of pull subscribers, the durable name is a required parameter
+		// so only needed to use BindOnly with stream name.
+		if opts.cfg.Durable != _EMPTY_ {
+			consumer = opts.cfg.Durable
+		}
+		if len(params) >= 2 {
```
matthiashanel, line 1534: How about a check for > 2 resulting in an too many params error?

matthiashanel: possibly also return that error when using a pull subscriber and providing two arguments

wallyqs: You already would get an error if using PullSubscribe and setting both durable and different consumer name.  About too many args, I was not sure yet on adding another constant for a too many args error, could revisit later.

js.go
```diff
@@ -1059,12 +1066,25 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync
...
 			deliver = NewInbox()
 		}
-	} else {
+	case (err != nil && !notFoundErr) || (notFoundErr && consumerBound):
+		// If the consumer is being bound got an error on pull subscribe then allow the error.
+		if !(isPullMode && lookupErr && consumerBound) {
+			return nil, err
+		}
+	default:
+		// Attempt to create consumer if not found or using BindOnly.
```
derekcollison, line 1087: I think BindOnly means not create yes?

js.go
```diff
@@ -1482,9 +1507,51 @@ func RateLimit(n uint64) SubOpt {
 }
 
 // BindStream binds a consumer to a stream explicitly based on a name.
-func BindStream(name string) SubOpt {
+func BindStream(stream string) SubOpt {
 	return subOptFn(func(opts *subOpts) error {
-		opts.stream = name
+		if opts.stream != "" {
```
derekcollison, line 1512: "" -> _EMPTY_

js.go
```diff
@@ -1482,9 +1507,51 @@ func RateLimit(n uint64) SubOpt {
...
+
+		opts.stream = stream
+		return nil
+	})
+}
+
+// BindOnly binds a subscription to a consumer from a stream without attempting to create.
+// The first argument is the stream name and the second argument will be the consumer name,
+// any extra arguments will be ignored.
+func BindOnly(params ...string) SubOpt {
```
derekcollison, line 1524: I think we make these explicit.

```go
Bind(stream, consumer string)
```

derekcollison (changes requested):

js.go
matthiashanel: when we do that, we will have to specify the durable name on pull consumer again.
With it being optional, we don't have to

matthiashanel (approved):
LGTM

js.go
```diff
@@ -1482,9 +1507,39 @@ func RateLimit(n uint64) SubOpt {
 }
 
 // BindStream binds a consumer to a stream explicitly based on a name.
-func BindStream(name string) SubOpt {
+func BindStream(stream string) SubOpt {
 	return subOptFn(func(opts *subOpts) error {
-		opts.stream = name
+		if opts.stream != _EMPTY_ {
```
derekcollison, line 1512: If they are same do we care?

js.go
```diff
@@ -1482,9 +1507,39 @@ func RateLimit(n uint64) SubOpt {
...
+			return fmt.Errorf("nats: duplicate stream name (%s and %s)", opts.stream, stream)
+		}
+
+		opts.stream = stream
+		return nil
+	})
+}
+
+// Bind binds a subscription to a consumer from a stream without attempting to create.
```
derekcollison, line 1521: to an **existing** consumer from a...

js.go
```diff
@@ -1482,9 +1507,39 @@ func RateLimit(n uint64) SubOpt {
...
+		if consumer == _EMPTY_ {
+			return ErrConsumerNameRequired
+		}
+
+		// In case of pull subscribers, the durable name is a required parameter
+		// so check that they are not different.
+		if opts.cfg.Durable != _EMPTY_ && opts.cfg.Durable != consumer {
+			return fmt.Errorf("nats: duplicate consumer names (%s and %s)", opts.cfg.Durable, consumer)
+		}
+		if opts.stream != "" {
```
derekcollison, line 1537: _EMPTY_

test/js_test.go
```diff
@@ -2343,15 +2378,58 @@ func TestJetStreamImportDirectOnly(t *testing.T) {
...
+
+	// Bind has the same effect as above since it would not attempt to create and lookup will work.
+	sub, err = js.SubscribeSync("ORDERS", nats.Bind("ORDERS", "d4"))
+	if err != nil {
+		t.Fatalf("Unexpected error: %v", err)
+	}
+
+	// Even if there are no permissions or import to check that a consumer exists,
+	// it is still possible to bind subscription to it.
+	sub, err = js.PullSubscribe("ORDERS", "d1", nats.Bind("ORDERS", "d1"))
```
derekcollison, line 2396: This does look a bit funny, but might be ok.

I think it might be better to say that nats.Bind only applicable to  normal Subscribe() calls?

test/js_test.go
```diff
@@ -2343,15 +2378,58 @@ func TestJetStreamImportDirectOnly(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Unexpected error: %v", err)
+	}
+	defer nc.Close()
+
+	js, err = nc.JetStream()
+	if err != nil {
+		t.Fatal(err)
+	}
+	sub, err = js.PullSubscribe("ORDERS", "d1", nats.Bind("ORDERS", "d1"))
```
derekcollison, line 2421: Not to continue the march to LGTM but should we just have..

js.BindConsumer(stream, consumer string, optCallback)?

test/js_test.go
wallyqs: But if we don't use `nats.Bind()` here then `PullSubscribe` it would attempt to create the consumer

derekcollison: I meant you just do a 

```go
js.Subscribe("ORDERS", nats.Bind("MYSTREAM", "d1"))
```

The bound consumer is a pull subscriber already, but maybe that is not the best way to approach.

test/js_test.go
wallyqs: I like that idea but thinking on how it could work: think it would be a lookup to find a consumer, then based on the consumer info dynamically become a push or pull subscriber. 
So for example could call `sub, err := js.BindConsumer(stream, consumer)` and a consumer is found then it would result `sub` becoming a `SubscribeSync` like subscription or `PullSubscriber` based on the already present ConsumerInfo.  If a callback is passed `js.BindConsumer(stream, consumer, optCallback)`, then would have to be a 
push based consumer.  In case the lookup fails, then maybe default into becoming a pull subscriber.

derekcollison: Good point..

test/js_test.go
wallyqs: Ah ok I see now, then maybe based on the consumer info lookup would make the client either a push or pull subscriber.

matthiashanel (approved):
LGTM

derekcollison (approved):
LGTM

PR #755, opened 2021-06-21
[FIXED] Websocket decompression of continuation frames

kozlovic:
When compression is used, the full message payload need to be
assembled (in case there are continuation frames) before being
decompressed.


derekcollison (approved):
LGTM

PR #757, opened 2021-06-23
when run nats_test.go will report our go.mod not include all lib

dtest11:
we need add
google.golang.org/protobuf
github.com/nats-io/nats-server/v2

these lib


wallyqs:
To run the tests you need to use the module for for tests like this:

```
go -modfile=go_test.mod test ./...
```

Or you could also run `go mod tidy` first which is designed to fetch test dependencies not used by the core library:

```
go mod tidy
go test ./... -v
```

Issue #751, opened 2021-06-17
JS: DeleteConsumer not found errors should be sentinel errors


sloveridge:
Some of the errors reported from DeleteConsumer look to be just passed on from Nats Server and should probably be sentinel errors to be consistent with the rest of nats.go

Example:

```
package main

import (
	"fmt"
	"github.com/nats-io/nats.go"
)

func main() {
	nc, _ := nats.Connect(nats.DefaultURL)
	js, _ := nc.JetStream()
	if err := js.DeleteConsumer("TEST", "test"); err != nil {
		fmt.Println(err)
	}
	js.AddStream(&nats.StreamConfig{Name: "TEST"})
	if err := js.DeleteConsumer("TEST", "test"); err != nil {
		fmt.Println(err)
	}
	js.DeleteStream("TEST")
	nc.Close()
}
```

I would expect something like:

```
var (
    ErrStreamNotFound    = errors.New("nats: stream not found")
    ErrConsumerNotFound   = errors.New("nats: consumer not found")
)
```

wallyqs:
Fixed via #760

PR #760, opened 2021-06-23
Add two new sentinel errors. ErrNoStream and ErrNoConsumer

actatum:
These sentinel errors will allow for better handling of managing streams and consumers.
Examples:
```go
// Creating a stream when a stream does not exist.
si, err := js.StreamInfo(cfg.streamName)
if err != nil {
	if !errors.Is(err, nats.ErrNoStream) {
		log.Fatalf("js.StreamInfo(): %v", err)
	}
	_, err = js.AddStream(&nats.StreamConfig{
		Name:     "foo",
		Subjects: []string{"foo.>"},
	})
	if err != nil {
		log.Fatalf("js.AddStream(): %v", err)
	}
}

// Creating a consumer when one does not exist.
ci, err := js.ConsumerInfo(cfg.streamName, cfg.consumerName)
if err != nil {
	if !errors.Is(err, nats.ErrNoConsumer) {
		log.Fatalf("js.ConsumerInfo(): %v", err)
	}
	_, err = js.AddConsumer(cfg.streamName, &nats.ConsumerConfig{
		AckPolicy:     nats.AckExplicitPolicy,
		DeliverPolicy: nats.DeliverNewPolicy,
		Durable:       cfg.consumerName,
	})
	if err != nil {
		log.Fatalf("js.AddConsumer(): %v", err)
	}
}
```

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -146,6 +146,8 @@ var (
 	ErrInvalidJSAck                 = errors.New("nats: invalid jetstream publish response")
 	ErrMultiStreamUnsupported       = errors.New("nats: multiple streams are not supported")
 	ErrStreamNameRequired           = errors.New("nats: stream name is required")
+	ErrStreamNotFound               = errors.New("stream not found")
+	ErrConsumerNotFound             = errors.New("consumer not found")
 	ErrConsumerNameRequired         = errors.New("nats: consumer name is required")
 	ErrConsumerConfigRequired       = errors.New("nats: consumer configuration is required")
 	ErrStreamSnapshotConfigRequired = errors.New("nats: stream snapshot configuration is required")
--- a/nats.go
+++ b/nats.go
@@ -147,7 +147,7 @@ var (
 	ErrMultiStreamUnsupported       = errors.New("nats: multiple streams are not supported")
 	ErrStreamNameRequired           = errors.New("nats: stream name is required")
 	ErrStreamNotFound               = errors.New("stream not found")
-	ErrConsumerNotFound             = errors.New("consumer not found")
+	ErrConsumerNotFound             = errors.New("nats: consumer not found")
 	ErrConsumerNameRequired         = errors.New("nats: consumer name is required")
 	ErrConsumerConfigRequired       = errors.New("nats: consumer configuration is required")
 	ErrStreamSnapshotConfigRequired = errors.New("nats: stream snapshot configuration is required")
--- a/nats.go
+++ b/nats.go
@@ -146,7 +146,7 @@ var (
 	ErrInvalidJSAck                 = errors.New("nats: invalid jetstream publish response")
 	ErrMultiStreamUnsupported       = errors.New("nats: multiple streams are not supported")
 	ErrStreamNameRequired           = errors.New("nats: stream name is required")
-	ErrStreamNotFound               = errors.New("stream not found")
+	ErrStreamNotFound               = errors.New("nats: stream not found")
 	ErrConsumerNotFound             = errors.New("nats: consumer not found")
 	ErrConsumerNameRequired         = errors.New("nats: consumer name is required")
 	ErrConsumerConfigRequired       = errors.New("nats: consumer configuration is required")
--- a/test/js_test.go
+++ b/test/js_test.go
@@ -16,6 +16,7 @@ package test
 import (
 	"context"
 	"crypto/rand"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"net"
@@ -1742,6 +1743,16 @@ func TestJetStreamManagement(t *testing.T) {
 		js.Publish("foo", []byte("hi"))
 	}
 
+	t.Run("stream not found", func(t *testing.T) {
+		si, err = js.StreamInfo("bar")
+		if !errors.Is(err, nats.ErrStreamNotFound) {
+			t.Fatalf("Expected error: %v, got: %v", nats.ErrStreamNotFound, err)
+		}
+		if si != nil {
+			t.Fatalf("StreamInfo should be nil %+v", si)
+		}
+	})
+
 	t.Run("stream info", func(t *testing.T) {
 		si, err = js.StreamInfo("foo")
 		if err != nil {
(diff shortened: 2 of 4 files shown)


nats.go
```diff
@@ -146,6 +146,8 @@ var (
 	ErrInvalidJSAck                 = errors.New("nats: invalid jetstream publish response")
 	ErrMultiStreamUnsupported       = errors.New("nats: multiple streams are not supported")
 	ErrStreamNameRequired           = errors.New("nats: stream name is required")
+	ErrNoStream                     = errors.New("nats: no stream found")
+	ErrNoConsumer                   = errors.New("nats: no consumer found")
```
wallyqs, line 150: Can we change to this so that it is backwards compatible in case matching against the string?

```suggestion
	ErrNoConsumer                   = errors.New("nats: consumer not found")
```

Thanks!

nats.go
```diff
@@ -146,6 +146,8 @@ var (
 	ErrInvalidJSAck                 = errors.New("nats: invalid jetstream publish response")
 	ErrMultiStreamUnsupported       = errors.New("nats: multiple streams are not supported")
 	ErrStreamNameRequired           = errors.New("nats: stream name is required")
+	ErrNoStream                     = errors.New("nats: no stream found")
```
wallyqs, line 149: Also maybe change to:
```suggestion
	ErrNoStream                     = errors.New("nats: stream not found")
```

So that it is similar to the error that was returned from server: https://github.com/nats-io/nats-server/blob/77d54dabeb97c4f7e2d1cbe98e4cd33108e71560/server/errors.json#L351-L354

nats.go
actatum: Will make this change now!

variadico: I think there are tests that depend on the error being `consumer not found`, not `nats: consumer not found`.
Here's one that's failing.
```
js_test.go:3906: Expected consumer not found error, got: nats: consumer not found
```

I think we have to either drop the prefix or update the test.

Maybe based on https://github.com/nats-io/nats.go/pull/760#discussion_r657423111, we drop the prefix?

actatum: I noticed this and was going in to change from a string compare to use errors.Is() with the new error types in the tests. I added the prefix to keep consistency with the rest of the sentinel errors. For backwards compatibility sake I could remove the prefix

actatum: Making the change to remove the prefix for backwards compatibility. All the effected tests are passing locally. This should close issue #751 

This is my first real contribution to anything open source, so I appreciate how fast you all have responded. I'm a huge fan of the NATS project and hope I can contribute more in the future as well.

variadico: Awesome! That's great to hear! Thank you very much for the contribution! 🎉

test/js_test.go
```diff
@@ -6099,13 +6120,13 @@ func TestJetStreamBindConsumer(t *testing.T) {
 		t.Fatalf("Unexpected error: %v", err)
 	}
 	_, err = js.SubscribeSync("foo", nats.Bind("foo", "push"))
-	if err == nil || err != nil && !strings.Contains(err.Error(), "consumer not found") {
+	if err == nil || err != nil && !errors.Is(err, nats.ErrNoConsumer) {
```
variadico, line 6123: I figure you probably wanted to make a minimal change here, but we can simplify this a little further.
```suggestion
	if err == nil || !errors.Is(err, nats.ErrNoConsumer) {
```

test/js_test.go
```diff
@@ -6099,13 +6120,13 @@ func TestJetStreamBindConsumer(t *testing.T) {
...
 		t.Fatalf("Unexpected error: %v", err)
 	}
 
 	// Pull consumer
 	_, err = js.PullSubscribe("foo", "pull", nats.Bind("foo", "pull"))
-	if err == nil || err != nil && !strings.Contains(err.Error(), "consumer not found") {
+	if err == nil || err != nil && !errors.Is(err, nats.ErrNoConsumer) {
```
variadico, line 6129: Same here.
```suggestion
	if err == nil || !errors.Is(err, nats.ErrNoConsumer) {
```

variadico: I think this one got missed. 😄

actatum: 🤦 fixed.

wallyqs:
Thanks @actatum for making the changes! Could you squash the changes into a single commit? 
Also I think I'd like to change to `ErrStreamNotFound` and `ErrConsumerNotFound` so that they are similar to the errors from the server:

https://github.com/nats-io/nats-server/blob/77d54dabeb97c4f7e2d1cbe98e4cd33108e71560/server/errors.json#L351
https://github.com/nats-io/nats-server/blob/77d54dabeb97c4f7e2d1cbe98e4cd33108e71560/server/errors.json#L95

actatum:
> Thanks @actatum for making the changes! Could you squash the changes into a single commit?
> Also I think would be good to change to `ErrStreamNotFound` and `ErrConsumerNotFound` so that they are similar to the errors from the server:
> 
> https://github.com/nats-io/nats-server/blob/77d54dabeb97c4f7e2d1cbe98e4cd33108e71560/server/errors.json#L351
> https://github.com/nats-io/nats-server/blob/77d54dabeb97c4f7e2d1cbe98e4cd33108e71560/server/errors.json#L95

No problem, had a lot of fun doing this. Everything should now be squashed into a single commit!

nats.go
```diff
@@ -146,6 +146,8 @@ var (
 	ErrInvalidJSAck                 = errors.New("nats: invalid jetstream publish response")
 	ErrMultiStreamUnsupported       = errors.New("nats: multiple streams are not supported")
 	ErrStreamNameRequired           = errors.New("nats: stream name is required")
+	ErrStreamNotFound               = errors.New("stream not found")
```
wallyqs, line 149: Thanks again for the contribution, looks like it would be ok to include with the `nats: ` prefix to be consistent here:

```suggestion
	ErrStreamNotFound               = errors.New("nats: stream not found")
```

nats.go
```diff
@@ -146,6 +146,8 @@ var (
...
+	ErrConsumerNotFound             = errors.New("consumer not found")
```
wallyqs, line 150:
```suggestion
	ErrConsumerNotFound             = errors.New("nats: consumer not found")
```

wallyqs (approved):
LGTM :)

PR #767, opened 2021-06-29
support custom inbox prefixes

ripienaar:

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -1109,7 +1109,7 @@ func Compression(enabled bool) Option {
 // CustomInboxPrefix configures the request + reply inbox prefix
 func CustomInboxPrefix(p string) Option {
 	return func(o *Options) error {
-		if p == "" || strings.HasSuffix(p, ">") || strings.HasSuffix(p, "*") || strings.HasSuffix(p, ".") || strings.HasPrefix(p, ">") || strings.HasPrefix(p, "*") {
+		if p == "" || strings.Contains(p, ">") || strings.Contains(p, "*") || strings.HasSuffix(p, ".") {
 			return fmt.Errorf("nats: invald custom prefix")
 		}
 		o.InboxPrefix = p
@@ -3359,7 +3359,7 @@ func (nc *Conn) createNewRequestAndSend(subj string, hdr, data []byte) (chan *Ms
 	// Create new literal Inbox and map to a chan msg.
 	mch := make(chan *Msg, RequestChanLen)
 	respInbox := nc.newRespInbox()
-	token := respInbox[nc.respSubLen+1:]
+	token := respInbox[nc.respSubLen:]
 
 	nc.respMap[token] = mch
 	if nc.respMux == nil {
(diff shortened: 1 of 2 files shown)


ripienaar:
This has been discussed a few times, Derek mentioned he has some other ideas but that I should open this to start a discussion.

Let me know your thoughts, I think supporting this is pretty important for peoples ability to create private comms channels as it lets them make predictable inbox patterns

/cc @derekcollison @wallyqs @kozlovic

nats.go
```diff
@@ -1101,6 +1106,17 @@ func Compression(enabled bool) Option {
 	}
 }
 
+// CustomInboxPrefix configures the request + reply inbox prefix
+func CustomInboxPrefix(p string) Option {
+	return func(o *Options) error {
+		if p == "" || strings.HasSuffix(p, ">") || strings.HasSuffix(p, "*") || strings.HasSuffix(p, ".") || strings.HasPrefix(p, ">") || strings.HasPrefix(p, "*") {
```
kozlovic, line 1112: This would not catch wildcards that would be in a middle, say `my.*.prefix`.

derekcollison:
So at a very high level we want to present a way for folks to make sure only they can see any given response and not have to abandon use of nc.Request() and the like correct?

ripienaar:
Well, that might be one possible outcome? I think there are many reasons to avoid a one-size-fits-all inbox pattern.

Even if not private in many of my use cases I prefix all related traffic by some string - I want the inboxes there too.  That's not about privacy as such but about just managing ACLs 

now of course I think we need private reply subjects that just work and don't require special ACLs and such, but thats probably more than I wish to tackle - I think we'll have inboxes like this for a long time

nats.go
```diff
@@ -3343,7 +3359,8 @@ func (nc *Conn) createNewRequestAndSend(subj string, hdr, data []byte) (chan *Ms
 	// Create new literal Inbox and map to a chan msg.
 	mch := make(chan *Msg, RequestChanLen)
 	respInbox := nc.newRespInbox()
-	token := respInbox[respInboxPrefixLen:]
+	token := respInbox[nc.respSubLen+1:]
```
kozlovic, line 3362: Seems that respSubLen is used only here (and set in initNewResp), and looks more like the token offset. So instead of doing a "+1" here and "-1" when setting it, if it was renamed to respSubToken (or something similar), we could simply use it here without arithmetic, no?

nats.go
```diff
@@ -3490,10 +3507,23 @@ func NewInbox() string {
...
+	sb.WriteString(nuid.Next())
+	return sb.String()
+}
+
 // Function to init new response structures.
 func (nc *Conn) initNewResp() {
-	// _INBOX wildcard
-	nc.respSub = fmt.Sprintf("%s.*", NewInbox())
+	nc.respSubPrefix = fmt.Sprintf("%s.", nc.newInbox())
+	nc.respSubLen = len(nc.respSubPrefix) - 1
```
kozlovic, line 3525: See my comment about being more of an "offset" than needing the len of the prefix (minus last '.')

nats.go
```diff
@@ -3505,15 +3535,17 @@ func (nc *Conn) newRespInbox() string {
 	if nc.respMap == nil {
 		nc.initNewResp()
 	}
-	var b [respInboxPrefixLen + replySuffixLen]byte
-	pres := b[:respInboxPrefixLen]
-	copy(pres, nc.respSub)
+
```
kozlovic, line 3538: Looks like respInboxPrefixLen (and replySuffixLen) may no longer be used? If so, will have to remove.

nats.go
ripienaar: oh yeah thats a good catch, sorted

kozlovic (approved):
Code LGTM, but will let Derek decide if we go with that or other approach.

ripienaar:
@derekcollison what are your thoughts here?

derekcollison:
I will try to find some time this week to think harder about it.

ripienaar:
@derekcollison this PR seems ok to me as is as discussed

Here's the matching feature in nats.java https://github.com/nats-io/nats.java/commit/03c22e41d738bed704511b6d748dda5ebcb41664 and .Net https://github.com/nats-io/nats.net/pull/322

derekcollison (commented):
If I supply a custom prefix of say "derek", does my inbox look like derek.NUID.NUID or does it look like derek._INBOX.NUID.NUID?

ripienaar:
_INBOX is replaced with prefix.

derekcollison (commented):
LGTM - Have not tested it but if it replaces _INBOX and tests pass looks good. I could use my own prefix of "_INBOX.derek" if I wanted as well for the use case we discussed.

derekcollison (approved):
LGTM

PR #770, opened 2021-07-08
Fixes to RespondMsg usage when crossing accounts

wallyqs:
When receiving a request, if the msg is responded using the `RespondMsg` API, the published response currently accidentally includes the inbox of the original request which will not be used, and in case of the request/response happening across accounts it also causes a permissions error due to the responder trying to publish a request. 

This could happen when doing the following for example:

```go
_, err := nc.Subscribe("foo", func(m *nats.Msg) {
	m.Data = []byte("response")
	m.RespondMsg(m)
// ...
```

That would result in a permissions error right now due to the reply inbox not being cleared out.

```
nats: Permissions Violation for Publish with Reply of "_R_.T3CyfT.TnlvbL"
```

In order to avoid the error which might be common,  `Reply` is always cleared out when needed if using `RespondMsg` and also a copy of the message is used instead for the publish to avoid changing the state of the original request.


nats.go
```diff
@@ -4190,12 +4190,26 @@ func (m *Msg) RespondMsg(msg *Msg) error {
 	if m.Reply == "" {
 		return ErrMsgNoReply
 	}
-	msg.Subject = m.Reply
```
wallyqs, line 4193: fyi @scottf @ColinSullivan1 this could happen as well in Java/.NET APIs with RespondMsg

nats.go
```diff
@@ -4190,12 +4190,26 @@ func (m *Msg) RespondMsg(msg *Msg) error {
...
+
+	resp := &Msg{
+		Subject: m.Reply,
+		Header:  msg.Header,
+		Data:    msg.Data,
+	}
+	// Discard the reply inbox unless it is set to be a different one,
+	// for example when using it like this:
+	//
+	// m.Data = []byte("response")
```
kozlovic, line 4205: You meant `m.Reply = "response"`?

kozlovic (commented):
I am not clear on "copy of the message is used instead for the publish to avoid changing the state of the original request". What state would be changed? (except for the cleared reply subject obviously).

wallyqs:
@kozlovic I meant that cannot clear out the Reply from the original Msg struct but it is just that, because in case of using a stream response need to call multiple times `RespondMsg` and the reply becomes the subject, so first time it would do `msg.Subject = m.Reply` but then the following would fail since the Reply was cleared.

kozlovic:
@wallyqs  I understand why we clear the Reply subject. I was just asking what other state is changed that justifies using a "copy" of the message. I am not saying that we should not use a different message object, just curious.

derekcollison:
Reusing the message as above in the example feels like a strange pattern. And we do not want to randomly clear the reply subject IMO since this could be a workflow situation.

nats.go
kozlovic: I see what you meant now.

test/auth_test.go
```diff
@@ -359,3 +360,281 @@ func TestPermViolation(t *testing.T) {
...
+			t.Fatal(err)
+		}
+		_, err = ncA.Subscribe("baz", func(m *nats.Msg) {
+			// Publish new message with different reply from to sub in Account A.
+			msg := &nats.Msg{
+				Header: m.Header,
+				Data:   []byte("baz"),
+				Reply:  inbox,
+			}
+			err = m.RespondMsg(msg)
```
kozlovic, line 605: You get a RACE when running the tests and I think it is because of `err` here. I would use a local variable here.
```
if err := m.RespondMsg(msg); err != nil {
    errCh <- err
}
```

kozlovic (changes requested):
Check for this pattern in rest of the test.

wallyqs:
Main thing that we want to avoid is `m.RespondMsg(m)` causing `PUB _R_.same_inbox _R_.same_inbox` when responding to a request, maybe we should disallow `m.RespondMsg(m)` so that correct usage is always like this with a new Msg to avoid reusing the original?

```go
		ncA.Subscribe("foo", func(m *nats.Msg) {
			msg := nats.NewMsg(m.Reply)
			msg.Data = []byte("ok")
			msg.Header = m.Header // take original headers if needed
			m.RespondMsg(msg)
		})
```

Since RespondMsg replaces the reply with the subject, cannot avoid the permissions error by clearing `m.Reply` manually since that would mean that `m.RespondMsg(m)` cannot be called multiple times either for workflow scenarios.

test/auth_test.go
wallyqs: ah right, thanks Ivan

test/auth_test.go
```diff
@@ -359,3 +360,281 @@ func TestPermViolation(t *testing.T) {
...
+		if !ok {
+			t.Error("Missing header in response")
+		}
+	})
+
+	t.Run("multiple RespondMsg stream", func(t *testing.T) {
+		_, err := ncA.Subscribe("quux", func(m *nats.Msg) {
+			m.Data = []byte("start")
+			m.Header["Task-Progress"] = []string{"0%"}
+			err = m.RespondMsg(m)
```
kozlovic, line 498: Again here, to be safe, would use a local variable.

kozlovic (approved):
LGTM

wallyqs:
Closing for now

PR #771, opened 2021-07-09
expose server version to clients

ripienaar:

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -644,6 +644,7 @@ type srv struct {
 type serverInfo struct {
 	ID           string   `json:"server_id"`
 	Name         string   `json:"server_name"`
+	Version      string   `json:"version"`
 	Proto        int      `json:"proto"`
 	Host         string   `json:"host"`
 	Port         int      `json:"port"`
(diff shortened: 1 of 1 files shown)


nats.go
```diff
@@ -1783,6 +1784,21 @@ func (nc *Conn) ConnectedServerName() string {
 	return nc.info.Name
 }
 
+// ConnectedServerVersion reports the connected server's version
+func (nc *Conn) ConnectedServerVersion() string {
```
derekcollison, line 1788: Feels like we need a better name.

ripienaar: Goes with `ConnectedServerId` and `ConnectedServerName`

derekcollison: Oh geez we have those? Must have missed that :/

derekcollison: We should move to just Server() and return and interface or struct that has ID, Name, Node, Version etc..

ripienaar:
So Export the current `serverInfo` and then `func Server() ServerInfo`?

derekcollison:
I think that is cleaner. Keep old ones but place deprecated label, IMO.

PR #774, opened 2021-07-13
In-process connections

neilalexander:
This PR is an extension of the work from nats-io/nats-server#2360 by allowing a client to use an in-process connection to the NATS server, rather than a TCP connection. Much of the justification is in the other PR, but this lets us do something like:

```
natsServer, err = natsserver.NewServer(&server.Options{
	ServerName: "monolith",
	DontListen: true,
	JetStream: true,
})
if err != nil {
	panic(err)
}

go natsServer.Start()

if !natsServer.ReadyForConnections(time.Second * 10) {
	panic("not ready in time")
}

nc, err := natsclient.Connect("", natsclient.InProcessServer(natsServer))
if err != nil {
	panic(err)
}
```

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -283,6 +283,10 @@ type CustomDialer interface {
 	Dial(network, address string) (net.Conn, error)
 }
 
+type InProcessConnProvider interface {
+	InProcessConn() (net.Conn, error)
+}
+
 // Options can be used to create a customized connection.
 type Options struct {
 
@@ -290,6 +294,11 @@ type Options struct {
 	// will be connecting. If the Servers option is also set, it
 	// then becomes the first server in the Servers array.
 	Url string
+
+	// InProcessServer represents a NATS server running within the
+	// same process. If this is set then we will attempt to connect
+	// to the server directly rather than using external TCP conns.
+	InProcessServer InProcessConnProvider
 
 	// Servers is a configured set of servers which this client
 	// will use when attempting to connect.
@@ -779,6 +788,15 @@ func Connect(url string, options ...Option) (*Conn, error) {
 func Name(name string) Option {
 	return func(o *Options) error {
 		o.Name = name
+		return nil
+	}
+}
+
+// InProcessServer is an Option that will try to establish a direction to a NATS server
+// running within the process instead of dialing via TCP.
+func InProcessServer(server InProcessConnProvider) Option {
+	return func(o *Options) error {
+		o.InProcessServer = server
 		return nil
 	}
 }
@@ -1747,6 +1765,18 @@ func (nc *Conn) createConn() (err error) {
 	}
 	if _, cur := nc.currentServer(); cur == nil {
 		return ErrNoServers
+	}
+
+	// If we have a reference to an in-process server then establish a
+	// connection using that.
+	if nc.Opts.InProcessServer != nil {
+		conn, err := nc.Opts.InProcessServer.InProcessConn()
+		if err != nil {
+			return fmt.Errorf("failed to get in-process connection: %w", err)
+		}
+		nc.conn = conn
+		nc.bindToNewConn()
+		return nil
 	}
 
 	// We will auto-expand host names if they resolve to multiple IPs
--- a/nats_test.go
+++ b/nats_test.go
@@ -2883,3 +2883,25 @@ func TestRespInbox(t *testing.T) {
 		t.Fatalf("Error: %s", resp.Data)
 	}
 }
+
+func TestInProcessConn(t *testing.T) {
+	s := RunServerOnPort(-1)
+	defer s.Shutdown()
+
+	nc, err := Connect("", InProcessServer(s))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	defer nc.Close()
+
+	// Status should be connected.
+	if nc.Status() != CONNECTED {
+		t.Fatal("should be status CONNECTED")
+	}
+
+	// The server should respond to a request.
+	if _, err := nc.RTT(); err != nil {
+		t.Fatal(err)
+	}
+}


nats.go
```diff
@@ -1634,6 +1649,13 @@ func (nc *Conn) createConn() (err error) {
 		return ErrNoServers
 	}
 
+	// If a custom conn has been provided then use that.
```
kozlovic, line 1652: Curious why not before selecting current server above? Is it because rest of code would fail if currentServer() was never invoked? If so, would not that be a problem if user does not specify any URL? (although I think that we add the localhost:4222 internally by default if none is specified).

kozlovic (changes requested):
See comments on the server side PR. I am worried about the behavior of the net.Pipe when connection on one side is closed. Until we figure this out, I would not want to merge the PRs.

nats.go
neilalexander: In a bit of local testing it doesn't appear to matter, so I'll move it up to before `currentServer()`.

kozlovic: And i think that `doReconnect()` would not work since we try to get a server from the pool and if we don't have any more we exit the reconnect loop.

neilalexander: In which case is it better just to handle this case in `connect()` instead, and only loop through servers in the usual way if `nc.Opts.Conn` isn't set?

kozlovic: Actually reconnect (as it stands today) won't work at all, because you would need to reassign a new s.InProcessConn(). By definition, if there is a disconnect it means that one side (or both) of the connection is closed, so it needs to be reassigned a new one.

kozlovic: So the issue is not the server pool, since it has 1 URL (always), so it would find one, but then we call createConn() that uses the option one but this connection is no longer valid.

neilalexander: I suppose this is unavoidable unless we give a reference to the NATS Server directly (via something like an `interface { InProcessConn() (net.Conn, error)` or similar) so that we can attempt to request a new in-process conn, rather than just expecting to be given a `net.Conn`?

kozlovic: I guess that would be better. Plus, it would even simply the steps when creating the connection since you would not have to directly call s.InProcessConn() and then pass it to nats.Conn().

neilalexander: Have added this so that you can `nc, err := natsclient.Connect("", natsclient.InProcessServer(natsserver))` now.

nats.go
```diff
@@ -1634,6 +1652,18 @@ func (nc *Conn) createConn() (err error) {
 		return ErrNoServers
 	}
 
+	// If we have a reference to an in-process server then establish a
+	// connection using that.
+	if nc.Opts.InProcessServer != nil {
+		conn, err := nc.Opts.InProcessServer.InProcessConn()
```
kozlovic, line 1658: I am bad at naming so I am not going to complain, but you will agree that there is a lot of stuttering here ;-) (InProcessServer.InProcessConn()).

kozlovic (approved):
Changes here look good to me. Let's wait for the PR on the server to be discussed internally to see what we do. I am going to move this PR to Draft mode so that it does not get merged by mistake until we make a decision.
Thanks again for the contribution!

nats.go
neilalexander: Indeed, if anyone has any better naming ideas I’m open to them!

variadico: Maybe?
```
type InProcessConnProvider interface {
	Conn() (net.Conn, error)
}
conn, err := nc.Opts.InProcessServer.Conn()
```

neilalexander:
Rebased against `main`.

nats.go
```diff
@@ -283,6 +283,10 @@ type CustomDialer interface {
 	Dial(network, address string) (net.Conn, error)
 }
 
+type InProcessConnProvider interface {
+	InProcessConn() (net.Conn, error)
```
kozlovic, line 287: Following Jaime's recommendation, I think this could be simply `Conn()` here..

nats.go
kozlovic: Agree with Jaime (@variadico). Change the interface and call here.

nats.go
neilalexander: Done.

nats.go
```diff
@@ -283,6 +283,10 @@ type CustomDialer interface {
 	Dial(network, address string) (net.Conn, error)
 }
 
+type InProcessConnProvider interface {
+	Conn() (net.Conn, error)
```
kozlovic, line 287: Stupid me... unless we change the server code (from your PR: https://github.com/nats-io/nats-server/pull/2360), it has to be InProcessConn(). By the way, since your server PR has been merged, I would recommend that you update this PR to add a test that checks the new feature. You would need to update the go_test.mod file to point to the main branch of the server and then do a `go mod tidy -modfile=go_test.mod`.

kozlovic (changes requested):

neilalexander:
@kozlovic Yes, good point :-) I've added a test to create an in-process connection and to perform an RTT test.

neilalexander:
@kozlovic Seems to be looking better now!

kozlovic (approved):
LGTM

ripienaar:
Be great to get this merged before next release.

kozlovic:
Oops, that fell through the cracks. Thanks @ripienaar for the reminder. @neilalexander Apologies, could you resolve conflicts and I promise I will merge this right after :-)

kozlovic:
@neilalexander while at it, would not hurt to rebase from main and squash your commits if possible.

neilalexander:
@kozlovic Done!

kozlovic (approved):
LGTM

PR #782, opened 2021-07-28
js: Add note on ephemeral consumers and Subscribe

wallyqs:
Add some notes on ephemeral consumers being the default.

### DIFF
--- a/js.go
+++ b/js.go
@@ -952,7 +952,7 @@ func (opt subOptFn) configureSubscribe(opts *subOpts) error {
 }
 
 // Subscribe will create a subscription to the appropriate stream and consumer.
-// By default, it will automatically create an JetStream consumer in the server in case
+// By default, it will automatically create a JetStream consumer in the server in case
 // one cannot be found. Unless the `Durable` option is passed, the consumer will be
 // ephemeral and automatically cleaned up by the server if the client that created
 // the consumer goes away.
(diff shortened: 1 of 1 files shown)


js.go
```diff
@@ -952,6 +952,10 @@ func (opt subOptFn) configureSubscribe(opts *subOpts) error {
 }
 
 // Subscribe will create a subscription to the appropriate stream and consumer.
+// By default, it will automatically create an JetStream consumer in the server in case
```
kozlovic, line 955: `an`->`a` in `an JetStream consumer`.

js.go
```diff
@@ -952,6 +952,10 @@ func (opt subOptFn) configureSubscribe(opts *subOpts) error {
...
+// one cannot be found. Unless the `Durable` option is passed, the consumer will be
+// ephemeral and automatically cleaned up by the server if the client that created
+// the consumer goes away.
```
kozlovic, line 958: Correct, but then the library is still trying to delete on Unsusbcribe(), so we will have to address that.

kozlovic (approved):
LGTM except for minor typo.

js.go
wallyqs, line 955:
```suggestion
// By default, it will automatically create a JetStream consumer in the server in case
```

kozlovic:
Closing in favor of PR #794

Issue #787, opened 2021-08-04
may avoid unnecessary data copy in processMsg for large message payloads


Endeavourken:
In function `func (nc *Conn) processMsg(data []byte)` there's a snippet of code
```go
	// FIXME(dlc): Need to copy, should/can do COW?
	msgPayload := make([]byte, len(data))
	copy(msgPayload, data)
```
It appears that every time we see a new message coming in, the payload gets copied into a new buffer. This is reasonable for small payloads since we are reusing the scratch buffer `scratch     [MAX_CONTROL_LINE_SIZE]byte`. But for large payloads a new buffer area sized as that of the actual message payload has already been created by invoking `make([]byte)`
```go
		// If we will overflow the scratch buffer, just create a
		// new buffer to hold the split message.
		if nc.ps.ma.size > cap(nc.ps.scratch)-len(nc.ps.argBuf) {
			lrem := len(buf[nc.ps.as:])
			nc.ps.msgBuf = make([]byte, lrem, nc.ps.ma.size)
			copy(nc.ps.msgBuf, buf[nc.ps.as:])
		} else {
			nc.ps.msgBuf = nc.ps.scratch[len(nc.ps.argBuf):len(nc.ps.argBuf)]
			nc.ps.msgBuf = append(nc.ps.msgBuf, (buf[nc.ps.as:])...)
		}
```
Here I think we may avoid this unnecessary copy for large message payloads within `processMsg(data []byte)` by introducting a bool flag, say `largeMsgBuf`, indicating that we don't need to copy the message payload a second time
```go
type parseState struct {
    ...
	msgBuf      []byte
	largeMsgBuf bool
	scratch     [MAX_CONTROL_LINE_SIZE]byte
}
```
```go
func (nc *Conn) processMsg(data []byte)
    ...
	// FIXME(dlc): Need to copy, should/can do COW?
	var msgPayload = data
	if !nc.ps.largeMsgBuf {
		msgPayload = make([]byte, len(data))
		copy(msgPayload, data)
	}
```
I've done some preliminary benchmark tests locally. The result showed that this fix could gain approximately 5%~6% performance boost for large message mayloads sized at 5K.
How does this small fix look here?

kozlovic:
[...]

PR #788, opened 2021-08-05
[CHANGED] avoid unnecessary data copy in processMsg

Endeavourken:
This PR intends to address the github issue
> may avoid unnecessary data copy in processMsg for large message payloads #787 

I've also  tweaked the benchmark test in file *test/bench_test.go* for local testing, but not included it in the PR because I am not sure of its necessity.
```go
func BenchmarkPubSubSpeed(b *testing.B)
...
    //msg := []byte("Hello World")
    msg := []byte(strings.Repeat("Hello", 1000))

```

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -191,7 +191,7 @@ func (nc *Conn) parse(buf []byte) error {
 				}
 			} else if i-nc.ps.as >= nc.ps.ma.size {
 				nc.processMsg(buf[nc.ps.as:i])
-				nc.ps.argBuf, nc.ps.msgBuf, nc.ps.state = nil, nil, MSG_END
+				nc.ps.argBuf, nc.ps.msgBuf, nc.ps.msgCopied, nc.ps.state = nil, nil, false, MSG_END
 			}
 		case MSG_END:
 			switch b {
--- a/parser.go
+++ b/parser.go
@@ -28,14 +28,15 @@ type msgArg struct {
 const MAX_CONTROL_LINE_SIZE = 4096
 
 type parseState struct {
-	state   int
-	as      int
-	drop    int
-	hdr     int
-	ma      msgArg
-	argBuf  []byte
-	msgBuf  []byte
-	scratch [MAX_CONTROL_LINE_SIZE]byte
+	state     int
+	as        int
+	drop      int
+	hdr       int
+	ma        msgArg
+	argBuf    []byte
+	msgBuf    []byte
+	msgCopied bool
+	scratch   [MAX_CONTROL_LINE_SIZE]byte
 }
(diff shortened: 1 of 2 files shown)


parser.go
```diff
@@ -167,7 +168,7 @@ func (nc *Conn) parse(buf []byte) error {
 			if nc.ps.msgBuf != nil {
 				if len(nc.ps.msgBuf) >= nc.ps.ma.size {
 					nc.processMsg(nc.ps.msgBuf)
-					nc.ps.argBuf, nc.ps.msgBuf, nc.ps.state = nil, nil, MSG_END
+					nc.ps.argBuf, nc.ps.msgBuf, nc.ps.msgCopied, nc.ps.state = nil, nil, false, MSG_END
```
kozlovic, line 171: For consistency, I wonder if we should also reset this boolean here: https://github.com/nats-io/nats.go/pull/788/files#diff-83eb8e32639d01cf443d6d8bde24c1c8be78766090d8c5f8586c36250cfedca6R194
I agree that at this location it means that the message fit in the current read buffer, so it is clearly not copied, so this boolean should not be `true`..

kozlovic (approved):
LGTM but small comment if we should reset this boolean after the other call to processMsg() (although I believe it is not needed, but same for nc.ps.msg.Buf which is clearly already nil, yet we reset).

parser.go
Endeavourken: At first I thought it was not needed, but I think I agree with you we should reset the flag as well for code consistency. 
PR updated :-)

Endeavourken:
There's a test case failure in Travis CI.
图片
I tried to duplicate it locally on go version 1.15.2 mimicing the command in the Travis test failure log
`go1.15.2 test -modfile=go_test.mod -race -v -p=1 ./... --failfast`
Apart from a few warning-like output the test suites PASS
> === RUN   TestJetStreamImportDirectOnly
> nats: Permissions Violation for Publish to "$JS.API.CONSUMER.INFO.ORDERS.d1" on connection [21]
> ...
>=== RUN   TestJetStream_ClusterMultipleSubscribe/qsub_n=1_r=1
> js_test.go:5055: 0 WARN: Timeout waiting for next message: nats: timeout
> js_test.go:5055: 0 WARN: Timeout waiting for next message: nats: timeout
> js_test.go:5055: 0 WARN: Timeout waiting for next message: nats: timeout
> js_test.go:5055: 0 WARN: Timeout waiting for next message: nats: timeout

Can you help me with what goes wrong here?

kozlovic (approved):
LGTM. Thank you for the contribution!

Issue #775, opened 2021-07-13
JS: Possible lock inversion


kozlovic:
When the library receives a JS flow control message and is ready to reply to it (pending messages is 0), it sends an empty message to the reply subject: https://github.com/nats-io/nats.go/blob/c7fc3c7fdf53ba0fef1a28400253d7f8d2c1a037/nats.go#L2708

The problem is that this is done under the subscription lock, which can cause a lock inversion since proper order is: connection -> subscription, not the other way around.

PR #789, opened 2021-08-06
Ordered Consumers

derekcollison:
This introduces ordered consumers. They are a convenience over ephemeral, no ack, no redelivery, only deliver things in strict order setups. We have the swap out when we detect gaps and change out the underlying sub and JetStream consumer, and we process the heartbeats as well, detecting gaps at the end or if a stream or consumer is pulled out from underneath of us.


js.go
```diff
@@ -1204,13 +1247,55 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync
...
 	}
-	sub.mu.Lock()
-	sub.jsi.stream = stream
-	sub.jsi.consumer = consumer
-	sub.jsi.durable = isDurable
-	sub.jsi.attached = attached
-	sub.jsi.deliver = deliver
-	sub.mu.Unlock()
+
+	// Recent servers are more tolerant for ephemerals not immediately being present so we can do sub here.
```
kozlovic, line 1251: Actually we can't delay the creation of the subscription to that point. The reason is that if user provides a durable name and say create 10 go routines that try to create the durable at the same time, all AddConsumer will return OK but the server will update the deliver subject with whatever latest call was. Out of the 10 durable subs, only 1 is going to receive this.
I have a test in the C client to test for that (originally I had also delayed creation of the sub here, but realized in code coverage analysis that the "already exists" condition was never reached and figured out what was happening, as described above).

derekcollison: Will write that test and see if I can move things around where sub is prior to consumer create but still have the jsi context that is needed.

js.go
```diff
@@ -1250,6 +1335,127 @@ func (jsi *jsSub) trackSequences(reply string) {
...
+	return false
+}
+
+// We are here if we have detected a gap with an ordered consumer.
+// We will create a new consumer and rewire the low level subscription.
+// Lock should be held.
+func (sub *Subscription) resetOrderedConsumer(sseq uint64) {
+	nc := sub.conn
+	closed := sub.closed
+	if sub.jsi == nil || nc == nil || closed || nc.IsClosed() {
```
kozlovic, line 1373: Don't check nc.IsClosed() since this creates a lock inversion (allowed is CONN->SUB, not the other way around). Note that when connection is closed, the connection will mark all its sub.closed to true, so checking sub.closed should be enough here.

js.go
```diff
@@ -1250,6 +1335,127 @@ func (jsi *jsSub) trackSequences(reply string) {
...
+	nc.subs[sub.sid] = sub
+	nc.subsMu.Unlock()
+
+	// Grab new inbox.
+	newDeliver := NewInbox()
+	sub.Subject = newDeliver
+
+	// Unsubscribe and subscribe with new inbox and sid.
+	// Remap a new low level sub into this sub since its client accessible.
+	nc.mu.Lock()
```
kozlovic, line 1393: Ouch. If we are under the subscription lock, then this is not good for reasons described above.

kozlovic (changes requested):
I am afraid that we are introducing lots of lock inversion with work on JetStream. See also https://github.com/nats-io/nats.go/issues/775 that I could address later with some code cleanup.

js.go
derekcollison: ok changed.

js.go
derekcollison: Good point, will remove.

js.go
derekcollison: Will move to under Go routine. Good catch, thanks.

js.go
```diff
@@ -1250,6 +1369,128 @@ func (jsi *jsSub) trackSequences(reply string) {
...
+	if err != nil {
+		return false
+	}
+	sseq, dseq := uint64(parseNum(tokens[5])), uint64(parseNum(tokens[6]))
+
+	jsi := sub.jsi
+	jsi.mu.Lock()
+	if dseq != jsi.dseq {
+		jsi.mu.Unlock()
+		sub.resetOrderedConsumer(jsi.sseq + 1)
```
kozlovic, line 1392: Should you compute `jsi.sseq+1` while under the jsi lock?

js.go
```diff
@@ -1250,6 +1369,128 @@ func (jsi *jsSub) trackSequences(reply string) {
...
+// Lock should be held.
+func (sub *Subscription) resetOrderedConsumer(sseq uint64) {
+	nc := sub.conn
+	closed := sub.closed
+	if sub.jsi == nil || nc == nil || closed {
+		return
+	}
+
+	// Quick unsubscribe. Since we know this is a simple push subscriber we do in place.
+	nc.subsMu.Lock()
```
kozlovic, line 1412: Another one: subsMu->sub.mu, not the opposite. Maybe make resetOrderedConsumer outside of the sub lock (and grab sub.mu lock above only to gather closed/etc..)

js.go
```diff
@@ -1180,37 +1292,44 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync
 					return nil, ErrSubjectMismatch
 				}
 
+				// Update attached status.
```
kozlovic, line 1295: What if you add a LABEL before processing of `switch{}` where we have `case info != nil`, then line 1288 (after calling js.ConsumerInfo() when return is OK, we could here simply make sure that the subscription is destroyed, 
set `deliver` to info.Config.DeliverSubject (although likely set in the processing of the info in switch statement), set `shouldCreate` to false, so we don't attempt again to add the consumer, then a simple goto <LABEL>. That should take care of validation and recreating the sub/jsi from scratch.

note1: recreate also the `ch` for sync subscriptions because it has been assigned to another subscription, so you can't reused. This is an issue I found in the past.

note2: you would also have to move the declaration of `ccreq *createConsumerRequest` and some other at the top of the function).

That's how I do in the C client and find it a bit more clean. That being said, this is with my pull subscribe implementation that uses a single sync subscribe, so basically I have only a single js.nc.subscribe() call, it's just that the deliver subject is an inbox in case of pull subscribe (and a sync flag).

kozlovic (changes requested):
Still some lock inversion and a comment on how the sub could be recreated when getting "already exists"

js.go
derekcollison: We can grab the value yes. Will update.

js.go
derekcollison: I need to make sure no more messages are received after this one in the stack. I could release the sub lock here though.

kozlovic: From my reading of the code, this is invoked from the readLoop isn't it? So messages would not be added to the sub, even if we release the lock.

derekcollison: Will clean up.

js.go
derekcollison: Code is working, so rather not redo completely again.

js.go
derekcollison: Yes just can't do it another go routine is what I meant.

kozlovic (approved):
LGTM

PR #791, opened 2021-08-10
Reworked PullSubscribe implementation

kozlovic:
Use a SyncSubscription instead. The only visible change from the
user is that calling Fetch() after Unsubscribe() returns ErrBadSubscription
instead of timeout or context deadline exceeded, which makes more
sense to me. This is the only test that I had to tweak.

### DIFF
--- a/js.go
+++ b/js.go
@@ -400,26 +400,6 @@ func (paf *pubAckFuture) Msg() *Msg {
 	return paf.msg
 }
 
-// pullSubscribe creates the wildcard subscription used per pull subscriber
-// to make fetch requests.
-func (js *js) pullSubscribe(subj string, jsi *jsSub) (*Subscription, error) {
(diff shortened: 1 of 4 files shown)


js.go
```diff
@@ -1433,7 +1337,7 @@ func (sub *Subscription) resetOrderedConsumer(sseq uint64) {
 	osid := sub.applyNewSID()
 
 	// Grab new inbox.
-	newDeliver := NewInbox()
+	newDeliver := nc.newInbox()
```
kozlovic, line 1340: @derekcollison Unrelated to this PR, but while I was changing references to NewInbox() in subscribe, I found this one and changed it too. If it should not be changed, let me know.

derekcollison (approved):
LGTM

wallyqs (approved):
LGTM

PR #792, opened 2021-08-10
[FIXED] Possible lock inversion and incorrect delete of JS consumer

### LINKED ISSUE #775: JS: Possible lock inversion

### LINKED ISSUE #776: JS: Library possibly removes a consumer when it should not
I think it is dangerous to delete (js.DeleteConsumer) a consumer when user calls sub.Unsubscribe(). Users may be used to call this function as a good practice when being done with a subscription.

The issue here is that if a user subscribes to a consumer that it does not "own" (meaning that the library did not create it), then Unsubscribe() will delete it: https://github.com/nats-io/nats.go/blob/c7fc3c7fdf53ba0fef1a28400253d7f8d2c1a037/js.go#L928

I believe that the library should ONLY delete JS consumer if it has created it for an ephemeral consumer (called AddConsumer inside js.subscribe()) and that there is a guarantee that no other subscription is using it (so not queue sub, or when AddConsumer returned an error meaning that someone else called AddConsumer()).

kozlovic:
- Some refactoring to prevent lock inversion (no connection publish
should be done under the subscription lock).
- Remove used of jsSub mutex since so far everything can be done
under the protection of the subscription's lock.
- Attempt to delete JS consumer on Unsubscribe *only* if the library
called AddConsumer and got a success.

Resolves #775
Resolves #776

### DIFF
--- a/js.go
+++ b/js.go
@@ -822,7 +822,7 @@ type jsSub struct {
 	deliver  string
 	pull     bool
 	durable  bool
-	delCons  bool
+	dc       bool // Delete JS consumer
 
 	// Ordered consumers
 	ordered bool
@@ -841,30 +841,20 @@ type jsSub struct {
 	fcd    uint64
 }
 
-func (sub *Subscription) jsiUnsubscribe(jsi *jsSub, drainMode bool) error {
+// Deletes the JS Consumer.
+// No connection nor subscription lock must be held on entry.
+func (sub *Subscription) deleteConsumer() error {
 	sub.mu.Lock()
-	durable, delCons := jsi.durable, jsi.delCons
+	jsi := sub.jsi
+	if jsi == nil {
+		sub.mu.Unlock()
+		return nil
+	}
 	stream, consumer := jsi.stream, jsi.consumer
 	js := jsi.js
-	if jsi.hbc != nil {
-		jsi.hbc.Stop()
-		jsi.hbc = nil
-	}
 	sub.mu.Unlock()
 
-	// Delete the JS consumer only if the library created the JS consumer,
-	// in which case delCons==true, but even then, in drain mode if this
-	// is a durable, do not delete.
-	if !delCons || (drainMode && durable) {
-		return nil
-	}
-	// We don't want to possibly fail a drain because we were not able to
-	// delete the consumer.
-	err := js.DeleteConsumer(stream, consumer)
-	if drainMode {
-		return nil
-	}
-	return err
+	return js.DeleteConsumer(stream, consumer)
 }
 
 // SubOpt configures options for subscribing to JetStream consumers.
@@ -1117,6 +1107,7 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync,
 		dseq:     1,
 		pull:     isPullMode,
 		nms:      nms,
+		dc:       o.dc,
 	}
 
 	sub, err = nc.subscribe(deliver, queue, cb, ch, isSync, jsi)
@@ -1222,15 +1213,10 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync,
 				}
 				return nil, fmt.Errorf("nats: %s", cinfo.Error.Description)
 			}
-		} else {
-			// Since we created the JS consumer internally, mark that we should
-			// delete it on Unsubscribe().
-			sub.mu.Lock()
-			sub.jsi.delCons = true
+		} else if consumer == _EMPTY_ {
 			// Update our consumer name here which is filled in when we create the consumer.
-			if consumer == _EMPTY_ {
-				sub.jsi.consumer = info.Name
-			}
+			sub.mu.Lock()
+			sub.jsi.consumer = info.Name
 			sub.mu.Unlock()
 		}
 	}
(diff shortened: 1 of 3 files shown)


js.go
```diff
@@ -823,7 +822,7 @@ type jsSub struct {
 	deliver  string
 	pull     bool
 	durable  bool
-	attached bool
```
derekcollison, line 826: I preferred attached here, was more meaningful then delCons.

js.go
```diff
@@ -838,27 +837,34 @@ type jsSub struct {
...
 		jsi.hbc.Stop()
 		jsi.hbc = nil
 	}
-	jsi.mu.Unlock()
+	sub.mu.Unlock()
 
-	if drainMode && (durable || attached) {
-		// Skip deleting consumer for durables/attached
-		// consumers when using drain mode.
+	// Delete the JS consumer only if the library created the JS consumer,
```
derekcollison, line 855: We have been going back and forth here and I am still not convinced we know the answer.

For instance if an app creates a durable and unsubscribes at the end (meaning it wants to do cleanup), but the app gets interrupted and a new version of the app is created, it will not cleanup the durable at the end here.

js.go
```diff
@@ -1222,10 +1222,15 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync,
 				}
 				return nil, fmt.Errorf("nats: %s", cinfo.Error.Description)
 			}
-		} else if consumer == _EMPTY_ {
-			// Update our consumer name here which is filled in when we create the consumer.
+		} else {
+			// Since we created the JS consumer internally, mark that we should
+			// delete it on Unsubscribe().
```
derekcollison, line 1227: I think the intention to delete an underlying consumer may need to be more explicit.

js.go
```diff
@@ -1403,86 +1405,71 @@ func (sub *Subscription) resetOrderedConsumer(sseq uint64) {
...
-		if !closed {
-			nc.mu.Lock()
-			errCB := nc.Opts.AsyncErrorCB
-			if errCB != nil {
-				nc.ach.push(func() { errCB(nc, sub, ErrConsumerNotActive) })
-			}
-			nc.mu.Unlock()
+	if !active && !closed {
+		nc.mu.Lock()
+		errCB := nc.Opts.AsyncErrorCB
```
derekcollison, line 1454: Can be one line since errCB  does not escape this block.

derekcollison (commented):
In general looks good. Not a fan of delCons vs attached and I think we need to think harder on when a sub.Unsubscribe() is supposed to delete an underlying JetStream consumer.

js.go
kozlovic: I have strong opinion that the library should try to delete (and ok if it does not since if really it is ephemeral the server will delete it anyway) *only* if it called AddConsumer() and got OK in the subscribe() call.

derekcollison: Again think of app that wants to use durable just to survive restart but once its done it wants the consumer deleted too. If the app crashes, the second version will never delete the durable.

kozlovic: Then app should call AddConsumer() -> subscribe -> process all -> DeleteConsumer. Again, we are trying to do too much in the library and assume we know what the user's end goal is.

kozlovic: (but as you suggested on Slack, if we make it an option, then I am fine with that)

derekcollison: Let's do that, we never delete the underlying durable by default, you can DeleteConsumer or we can add a SubOpt to js.Subscribe() that instructs the system to do so.

js.go
```diff
@@ -1775,6 +1762,15 @@ func IdleHeartbeat(duration time.Duration) SubOpt {
 	})
 }
 
+// DeleteConsumer instructs the library to delete the JetStream consumer
+// when calling Subscription.Unsubscribe().
```
derekcollison, line 1766: And Drain()?

test/js_test.go
```diff
@@ -3709,19 +3710,19 @@ func TestJetStream_Unsubscribe(t *testing.T) {
 		deleteAllConsumers(t)
 	})
 
-	t.Run("durable consumers not deleted on drain", func(t *testing.T) {
-		subB, err := js.SubscribeSync("foo.B", nats.Durable("B"))
+	t.Run("durable consumers deleted on drain", func(t *testing.T) {
+		subB, err := js.Subscribe("foo.B", func(_ *nats.Msg) {}, nats.Durable("B"), nats.DeleteConsumer())
```
derekcollison, line 3714: All these changes till needed?

kozlovic: Meaning? These tests are about checking that consumers are deleted, but the semantic have changed a bit. I could refactor them completely to match the current expectations as opposed to try "fixing" them. Will do that tomorrow.

js.go
kozlovic: will fix.

test/js_test.go
derekcollison: Sounds good. Looking at the examples, would you add the opt to Unsubscribe() vs Subscribe()?

kozlovic: Although it would make more sense, I would not change Unsubscribe()/Drain() (even adding var option may cause backward compatibility, and may not be possible in other lang).

kozlovic:
Closing this PR and will open a new one that is more comprehensive.

PR #801, opened 2021-08-18
Removed new option SubjectIsDelivery and some other changes

kozlovic:
This is to align with newer version of the ADR describing the
subscription workflow:
- Removed option SubjectIsDelivery that was introduced in main branch
(but not released yet).
- Take into consideration the new ACK layout
- Make sure that OrderedConsumer rejects user configuring DeliverSubject


wallyqs (approved):
LGTM

PR #802, opened 2021-08-23
[CHANGED] ConsumerInfo Delivered and AckFloor are now SequenceInfo

kozlovic:
A new object `SequenceInfo` is added for `Delivered` and `AckFloor`'s
`ConsumerInfo` fields. This object contains `Last` which represents
the last activity time (in UTC).
Having this field under `SequencePair` was wrong since we used
`SequencePair` in other object, namely `MsgMetadata`, and in
that context, `Last` did not make sense.


derekcollison (approved):
LGTM

PR #803, opened 2021-08-23
[IMPROVED] Subscribe returns error if consumer config does not match

### LINKED ISSUE #796: Return error if actual consumer configuration is different
## Feature Request

When creating subscriptions I can pass extra options, e.g.:
```go
sub, err := js.PullSubscribe("TEST.*", "testing", nats.MaxDeliver(3))
```
The consumer is created in NATS if it didn't exist yet:
```
$ nats con info TEST testing
Information for Consumer TEST > testing created 2021-08-17T11:37:10+02:00

Configuration:

        Durable Name: testing
           Pull Mode: true
      Filter Subject: TEST.*
         Deliver All: true
          Ack Policy: Explicit
            Ack Wait: 30s
       Replay Policy: Instant
  Maximum Deliveries: 3
     Max Ack Pending: 20,000
   Max Waiting Pulls: 512

State:
...
```

If I then change the options, e.g.:
```go
sub, err := js.PullSubscribe("TEST.*", "testing", nats.MaxDeliver(5))
```
Then the new value is silently ignored and the consumer still has `Maximum Deliveries: 3`. The options seems to be used only when creating a new consumer, but ignored otherwise (some of them at least).

While it might not be feasible to transparently update the consumer, I would expect to get an error telling me that the requested options do not match the actual consumer configuration.

#### Use Case:

I can very easily imagine a situation where a developer changes the options (adds a new one, removes one, changes a value) thinking that it will work. For someone unfamiliar with NATS JS it's not obvious that some consumer object exists in NATS itself and that these options are not just client-side connection options. It's easy for the code to diverge from the actual configuration.

#### Proposed Change:

Compare the actual consumer configuration with the requested one an return an error if there's a mismatch.

It seems that the existing configuration is already read (the `info` variable), so this shouldn't affect performance.

Unwitting developers trying to change the consumer configuration.

#### Alternative Approaches

Update the consumer configuration to match the requested one.

kozlovic:
This is done to alert the user that consumer config is not changed
as one would expect.

This change is more tricky than expected. We can't simply compare
user's consumer config with what is returned from the server.
For once, there are configurations that may not be set by the
user that were set when the consumer was created, that the lib
should not fail. Let's say the consumer is created with a description,
why fail the user if they don't pass the Description() option at all?

The Description() option was actually missing, but this is a good
example. But same could be say with for instance "optional start
sequence". This may be something that is set when the consumer
is first created (possibly outside of the app with CLI), but when
user calls js.Subscribe(), that option should not have to be
set to the value that was used when creating the consumer.

I had to add "not set" values for replay and deliver policy, similar
to ack policy.

I added an extensive test that checks proper error when trying
to make configuration changes, but also that if the options are
not set, then the lib does not fail the subscribe call.

Resolves #796

### DIFF
--- a/test/js_test.go
+++ b/test/js_test.go
@@ -18,7 +18,6 @@ import (
 	"crypto/rand"
 	"errors"
 	"fmt"
-	"github.com/nats-io/nuid"
 	"io/ioutil"
 	"net"
 	"os"
@@ -31,6 +30,7 @@ import (
 
 	"github.com/nats-io/nats-server/v2/server"
 	"github.com/nats-io/nats.go"
+	"github.com/nats-io/nuid"
 
 	natsserver "github.com/nats-io/nats-server/v2/test"
 )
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

test/js_test.go
```diff
@@ -3596,6 +3597,134 @@ func TestJetStreamSubscribe_RateLimit(t *testing.T) {
...
+				t.Fatalf("Error on subscribe: %v", err)
+			}
+			// If not explicitly asked by the user, we are ok
+			_, err = js.PullSubscribe("foo", durName)
+			if err != nil {
+				t.Fatalf("Unexpected error: %v", err)
+			}
+			sub.Unsubscribe()
+		})
+	}
```
wallyqs, line 3714: I was curious on the transition from unset with defaults to set but with same defaults as server like:

```
# Errors as expected since server default is 512.
sub, err := js.PullSubscribe("foo", "dur2")
sub, err := js.PullSubscribe("foo", "dur2", nats.PullMaxWaiting(513))

# No error as expected.
sub, err := js.PullSubscribe("foo", "dur")
sub, err := js.PullSubscribe("foo", "dur", nats.PullMaxWaiting(512))
```

It seems it is fine which is great, added some these extra tests in this commit:
https://github.com/wallyqs/nats.go/commit/e2c6dffbb7b273eb67dc51527b246d7aae9453d0

wallyqs (approved):
LGTM

test/js_test.go
wallyqs: ok just updated

PR #804, opened 2021-08-25
Fixed check for JS.ACK tokens count

kozlovic:
Do not reject if it has only 11 tokens.


matthiashanel (approved):
LGTM

PR #808, opened 2021-09-01
[FIXED] Inboxes suffix would contain many zeros

kozlovic:
This was introduced by PR https://github.com/nats-io/nats.go/pull/767
when the nuidSize was used instead of replySuffixLen to just build
the reply suffix. This meant that the suffix part with 22 characters
long and had lots of zeros at the end.

Added a test to make sure we don't break that in the future.


derekcollison (approved):
LGTM

ripienaar (approved):
LGTM

PR #812, opened 2021-09-02
[ADDED] Stringer for connection status

JosephWoodward:
This change updates the connection status by implementing the Stringer interface, enabling users to log out the connection status in a human readable format.

I've updated an existing tests to verify the string output. It's not exhaustive, but covers the same cases the existing test exercises. If you'd rather I add a separate test that covers each permutation then let me know and I can update the pull request.

### DIFF
--- a/test/conn_test.go
+++ b/test/conn_test.go
@@ -49,7 +49,7 @@ func TestConnectionStatus(t *testing.T) {
 	nc := NewDefaultConnection(t)
 	defer nc.Close()
 
-	if nc.Status() != nats.CONNECTED || fmt.Sprintf("%s", nc.Status()) != "CONNECTED"{
+	if nc.Status() != nats.CONNECTED || fmt.Sprintf("%s", nc.Status()) != "CONNECTED" {
 		t.Fatal("Should have status set to CONNECTED")
 	}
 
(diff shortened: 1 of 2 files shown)


wallyqs (approved):
LGTM

test/conn_test.go
```diff
@@ -49,14 +49,15 @@ func TestConnectionStatus(t *testing.T) {
 	nc := NewDefaultConnection(t)
 	defer nc.Close()
 
-	if nc.Status() != nats.CONNECTED {
+	if nc.Status() != nats.CONNECTED || fmt.Sprintf("%s", nc.Status()) != "CONNECTED"{
```
wallyqs, line 52: To fix `go fmt`:
```suggestion
	if nc.Status() != nats.CONNECTED || fmt.Sprintf("%s", nc.Status()) != "CONNECTED" {
```

JosephWoodward: Oops, thanks for that!

JosephWoodward:
Updated the test to use `String()` in instead of `fmt.Sprintf(...)` to satisfy the linter.

kozlovic (approved):
LGTM. Thank you for your contribution!

PR #821, opened 2021-09-13
[FIXED] Websocket: possible panic when decoding CLOSE frame

kozlovic:

### DIFF
--- a/ws.go
+++ b/ws.go
@@ -372,7 +372,6 @@ func (r *websocketReader) handleControlFrame(frameType wsOpCode, buf []byte, pos
 	var payload []byte
 	var err error
 
-	statusPos := pos
 	if rem > 0 {
 		payload, pos, err = wsGet(r.r, buf, pos, rem)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM - One small comment on readability.

ws.go
```diff
@@ -382,17 +381,19 @@ func (r *websocketReader) handleControlFrame(frameType wsOpCode, buf []byte, pos
...
 		// that represent the status code and then optional payload.
 		if len(payload) >= 2 {
-			status = int(binary.BigEndian.Uint16(buf[statusPos : statusPos+2]))
-			body = string(buf[statusPos+2 : statusPos+len(payload)])
-			if body != "" && !utf8.ValidString(body) {
-				// https://tools.ietf.org/html/rfc6455#section-5.5.1
-				// If body is present, it must be a valid utf8
-				status = wsCloseStatusInvalidPayloadData
-				body = "invalid utf8 body in close frame"
+			status = int(binary.BigEndian.Uint16(payload[:2]))
```
kozlovic, line 388: @derekcollison If you could suggest how I could make it more readable I would appreciate, although I have a suggestion below.

The spec says that if there is a body, the first 2 bytes MUST be the status, then it can be followed by optional "reason". (see below for spec text)

What I could do is this:
```
// If there is a payload, it MUST be a 2-byte unsigned integer (in network byte order) representing a status code
if len(payload) >= 2 {
    status = int(binary.BigEndian.Uint16(payload[:2]))

    // Now check for optional UTF-8-encoded data for the close reason.
    if len(payload) > 2 {
        ...
```
The reason I have the check of length of payload > 2 inside is because to me it makes sense that if > 2, it is necessarily >=2. So it looks illogical to have it at the "same level" than the check for the status.

I could rename things, create intermediate slices (status, reason, etc..), but let me know what exactly you found difficult to read.

--
The Close frame MAY contain a body (the "Application data" portion of
   the frame) that indicates a reason for closing, such as an endpoint
   shutting down, an endpoint having received a frame too large, or an
   endpoint having received a frame that does not conform to the format
   expected by the endpoint.  If there is a body, the first two bytes of
   the body MUST be a 2-byte unsigned integer (in network byte order)
   representing a status code with value /code/ defined in Section 7.4.
   Following the 2-byte integer, the body MAY contain UTF-8-encoded data
   with value /reason/, the interpretation of which is not defined by
   this specification.

derekcollison:
```go
lp, hasBody := len(payload), lp > 2
if lp >= min {
  ...
  if hasBody {
  }
}
```

Maybe something like that? At first it looked to me like you were repeating same if test.

kozlovic: Great, I will do that. Thanks!

kozlovic: except that you can't use lp in the right hand, but I can either duplicate `len(payload)` or do it in different line. Will see if that reads better.

ws.go
```diff
@@ -382,17 +381,24 @@ func (r *websocketReader) handleControlFrame(frameType wsOpCode, buf []byte, pos
...
-		// that represent the status code and then optional payload.
-		if len(payload) >= 2 {
-			status = int(binary.BigEndian.Uint16(buf[statusPos : statusPos+2]))
-			body = string(buf[statusPos+2 : statusPos+len(payload)])
-			if body != "" && !utf8.ValidString(body) {
-				// https://tools.ietf.org/html/rfc6455#section-5.5.1
-				// If body is present, it must be a valid utf8
-				status = wsCloseStatusInvalidPayloadData
-				body = "invalid utf8 body in close frame"
+		body := _EMPTY_
```
derekcollison, line 384: var body string

kozlovic:
@derekcollison Did update and "push force" so that you can check against original code to see if it reads better. Happy to change again if not readable enough (once ok, I will apply the same changes to the server PR).

ws.go
```diff
@@ -382,17 +381,24 @@ func (r *websocketReader) handleControlFrame(frameType wsOpCode, buf []byte, pos
...
+		lp := len(payload)
+		// If there is a payload, the status is represented as a 2-byte
+		// unsigned integer (in network byte order). Then, there may be an
+		// optional body.
+		hasStatus, hasBody := lp >= 2, lp > 2
```
derekcollison, line 389: Make 2 a const?

ws.go
kozlovic: Changed

ws.go
kozlovic: Done, please review commit: https://github.com/nats-io/nats.go/pull/821/commits/949abffd71d12ae021dfc2dbd0fddd3d8acb9924

derekcollison (approved):
LGTM

PR #823, opened 2021-09-14
js: unblock batch requests on 408 with at least a message

wallyqs:
This makes it so that fetch requests that receive a 408 message signal to be canceled become unblocked in case they have received a single message at least.

### DIFF
--- a/js.go
+++ b/js.go
@@ -2105,9 +2105,9 @@ func checkMsg(msg *Msg, checkSts bool) (usrMsg bool, err error) {
 	case reqTimeoutSts:
 		// Older servers may send a 408 when a request in the server was expired
 		// and interest is still found, which will be the case for our
-		// implementation. Regardless, ignore 408 errors, the caller will
-		// go back to wait for the next message.
-		err = nil
+		// implementation. Regardless, ignore 408 errors until receiving at least
+		// one message.
+		err = ErrTimeout
 	default:
 		err = fmt.Errorf("nats: %s", msg.Header.Get(descrHdr))
 	}
@@ -2252,6 +2252,10 @@ func (sub *Subscription) Fetch(batch int, opts ...PullOpt) ([]*Msg, error) {
 					// wait this time.
 					noWait = false
 					err = sendReq()
+				} else if err == ErrTimeout && len(msgs) == 0 {
+					// If we get a 408, we will bail if we already collected some
+					// messages, otherwise ignore and go back calling nextMsg.
+					err = nil
 				}
 			}
 		}
--- a/test/js_test.go
+++ b/test/js_test.go
@@ -4498,6 +4498,287 @@ func testJetStreamMirror_Source(t *testing.T, nodes ...*jsServer) {
 	})
 }
 
+func TestJetStream_PullSubscribeMaxWaiting(t *testing.T) {
+	nodes := []int{1, 3}
+	replicas := []int{1}
+
+	for _, n := range nodes {
+		for _, r := range replicas {
+			if r > 1 && n == 1 {
+				continue
+			}
+			t.Run(fmt.Sprintf("psub n=%d r=%d", n, r), func(t *testing.T) {
+				name := fmt.Sprintf("PSUBMAX%d%d", n, r)
+				stream := &nats.StreamConfig{
+					Name:     name,
+					Replicas: n,
+				}
+				withJSClusterAndStream(t, name, n, stream, testJetStream_PullSubscribeMaxWaiting)
+			})
+		}
+	}
+}
+
+func testJetStream_PullSubscribeMaxWaiting(t *testing.T, subject string, srvs ...*jsServer) {
+	srv := srvs[0]
+	nc, err := nats.Connect(srv.ClientURL())
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer nc.Close()
+
+	js, err := nc.JetStream()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Create pull subscriber with a lower max waiting limit.
+	sub, err := js.PullSubscribe(subject, "durable", nats.PullMaxWaiting(5))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Delay for a bit the first message being received.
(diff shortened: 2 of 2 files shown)


test/js_test.go
```diff
@@ -4498,6 +4498,283 @@ func testJetStreamMirror_Source(t *testing.T, nodes ...*jsServer) {
...
+	}
+
+	// Create pull subscriber with a lower max waiting limit.
+	sub, err := js.PullSubscribe(subject, "durable", nats.PullMaxWaiting(5))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Delay for a bit the first message being received.
+	go func() {
```
kozlovic, line 4542: You don't need go func() here since time.AfterFunc() is already going to execute the closure function asynchronously after 200ms.

test/js_test.go
```diff
@@ -4498,6 +4498,283 @@ func testJetStreamMirror_Source(t *testing.T, nodes ...*jsServer) {
...
+		t.Fatalf("Unexpected error: %v", err)
+	}
+	if len(msgs) != 1 {
+		t.Fatalf("Expected one message to be delivered, got: %v", len(msgs))
+	}
+	msg := msgs[0]
+	expected := "hello"
+	got := string(msg.Data)
+	if got != expected {
+		t.Errorf("Expected: %v, got: %v", expected, got)
```
kozlovic, line 4559: Why do you have t.Errorf()? Any error that we have is fatal, there is no point continuing the test I think.
If you agree, then this comment applies to all other references of t.Errorf() you have in this test.

test/js_test.go
```diff
@@ -4498,6 +4498,283 @@ func testJetStreamMirror_Source(t *testing.T, nodes ...*jsServer) {
...
+	expected = "bar"
+	got = string(msg.Data)
+	if got != expected {
+		t.Errorf("Expected: %v, got: %v", expected, got)
+	}
+	if len(msgs) != 1 {
+		t.Fatalf("Expected one message to be delivered, got: %v", len(msgs))
+	}
+
+	// There should be no waiting pull requests since they got expired after fetch succeeded.
```
kozlovic, line 4635: Why though? Is the server not "force expiring" just 1 at a time? I mean all other requests have expiration set so they will expire, but this is because of their MaxWait value that has now passed, no? You may be right and that is a behavior of the server that I don't know about though...

test/js_test.go
```diff
@@ -4498,6 +4498,283 @@ func testJetStreamMirror_Source(t *testing.T, nodes ...*jsServer) {
...
+		)
+		for i := 0; i < 5; i++ {
+			expectedMap[fmt.Sprintf("quux:%d", i)] = false
+			go func() {
+				// Can only have at most 5 inflight fetch requests so
+				// these will block until they receive at least a message
+				// and a 408 response status.
+				msgs, err := sub.Fetch(100, nats.MaxWait(30*time.Second))
+				if err != nil {
+					errCh <- err
```
kozlovic, line 4661: Should you return after that? The contract is that if Fetch() returns any message, then it will return `nil` anyway. So if there is an error, then there should be no message.

js.go
```diff
@@ -2105,9 +2105,13 @@ func checkMsg(msg *Msg, checkSts bool) (usrMsg bool, err error) {
 	case reqTimeoutSts:
 		// Older servers may send a 408 when a request in the server was expired
 		// and interest is still found, which will be the case for our
-		// implementation. Regardless, ignore 408 errors, the caller will
-		// go back to wait for the next message.
-		err = nil
+		// implementation. Regardless, ignore 408 errors until receiving at least
```
kozlovic, line 2108: Alternatively, set err to ErrTimeout here, (without the need of cancelWhen408), but just treat the possible ErrTimeout from the caller perspective, see below.

js.go
```diff
@@ -2243,7 +2247,12 @@ func (sub *Subscription) Fetch(batch int, opts ...PullOpt) ([]*Msg, error) {
 			if err == nil {
 				var usrMsg bool
 
-				usrMsg, err = checkMsg(msg, true)
+				// If we already got at least a message but we are prompted
+				// with a 408 status, then unblock the request and return
+				// the messages that have arrived already.
+				cancelWhen408 := len(msgs) >= 1
+
+				usrMsg, err = checkMsg(msg, true, cancelWhen408)
```
kozlovic, line 2255: If you do without cancelWhen408, just check for err == ErrTimeout at the end if this if/else by adding:
```
..
} else if err == ErrTimeout && len(msgs) == 0 {
    err = nil
}
```
Keep in mind that even if we get an error in this function, at the end we will return no error if `len(msgs)` is > 0 (see at the end).

kozlovic (changes requested):
Proposing some changes...

test/js_test.go
wallyqs: Added the return

test/js_test.go
wallyqs: Once a waiting queue for a consumer is full, the server starts expiring all the ones that were expired here:
https://github.com/nats-io/nats-server/blob/main/server/consumer.go#L2090-L2095

js.go
wallyqs: I didn't get this change completely but sounds interesting, could you share me the whole snippet of where to apply the change?

kozlovic: Remove the cancelWhen408 variable definition above and restore checkMsg() (remove boolean), but set `err = ErrTimeout` when processing 408 header.
Then here the code would be:
```
usrMsg, err = checkMsg(msg, true)
if err == nil && usrMsg {
    msgs = append(msgs, msg)
} else if noWait && (err == errNoMessages) && len(msgs) == 0 {
    // If we have a 404 for our "no_wait" request and have
    // not collected any message, then resend request to
    // wait this time.
    noWait = false
    err = sendReq()
} else if err == ErrTimeout && len(msgs) == 0 {
    // If we get a 408, we will bail if we already collected some
    // messages, otherwise ignore and go back calling nextMsg
    err = nil
}
```

wallyqs: Thanks! I updated now.

kozlovic (approved):
LGTM

PR #824, opened 2021-09-15
Subject interning

moredure:
@derekcollison Please review. The price of this comparison near 2ns per operation but in case of plain subscriptions it reduces single allocation per operation and results in speedup. Looking forward for your comments. If somehow to know beforehand that subscription using wildcard then it's possible to make it no cost. So I suggest to merge this pull request first, than make some changes to subscription struct to contain isWildcard bool field and then switch to using it. so no need to compare strings just single if isWildcard and nothings else.

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -577,6 +577,9 @@ type Subscription struct {
 	// Type of Subscription
 	typ SubscriptionType
 
+	// Whether subject in messages and Subject in subscription may differ
+	smd bool
+
 	// Async linked list
 	pHead *Msg
 	pTail *Msg
@@ -2692,7 +2695,10 @@ func (nc *Conn) processMsg(data []byte) {
 	}
 
 	// Copy them into string
-	subj := string(nc.ps.ma.subject)
+	subj := sub.Subject
+	if sub.smd {
+		subj = string(nc.ps.ma.subject)
+	}
 	reply := string(nc.ps.ma.reply)
 
 	// Doing message create outside of the sub's lock to reduce contention.
@@ -3746,6 +3752,11 @@ func badQueue(qname string) bool {
 	return strings.ContainsAny(qname, " \t\r\n")
 }
 
+// wildcard will check a subject name for wildcardness.
+func wildcard(subj string) bool {
+	return strings.ContainsAny(subj, "*>")
+}
+
 // subscribe is the internal subscribe function that indicates interest in a subject.
 func (nc *Conn) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync bool, js *jsSub) (*Subscription, error) {
 	if nc == nil {
@@ -3779,7 +3790,14 @@ func (nc *Conn) subscribeLocked(subj, queue string, cb MsgHandler, ch chan *Msg,
 		return nil, ErrBadSubscription
 	}
 
-	sub := &Subscription{Subject: subj, Queue: queue, mcb: cb, conn: nc, jsi: js}
+	sub := &Subscription{
+		Subject: subj,
+		Queue:   queue,
+		mcb:     cb,
+		conn:    nc,
+		jsi:     js,
+		smd:     wildcard(subj) || js != nil,
+	}
 	// Set pending limits.
 	if ch != nil {
 		sub.pMsgsLimit = cap(ch)


moredure:
will be good if subscription with wildcard subject will have some flag then we can simplify the flow a littlebit

moredure:
If it's possible to add some bool flag `wildcard` to subscription it will be awesome, any ideas?

derekcollison:
Thanks, will have a few others take a look.

nats.go
```diff
@@ -2692,7 +2692,10 @@ func (nc *Conn) processMsg(data []byte) {
 	}
 
 	// Copy them into string
-	subj := string(nc.ps.ma.subject)
+	subj := sub.Subject
+	if subj != string(nc.ps.ma.subject) {
```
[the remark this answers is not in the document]
derekcollison: When you do a subscription those subjects can contain wildcards, and we have utilities to test for that. When a NATS application receives a message they are always on literal subjects (no wildcards).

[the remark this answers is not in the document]
moredure: Can you point me where can I find them? Somewhere in the nats-server?

derekcollison:
I believe if you only store and retrieve from the map when a message is received that would solve the problem.

moredure:
in case of high amount of different subjects we can overflow map

derekcollison:
I was assuming you would use an LRU of some sort etc..

nats.go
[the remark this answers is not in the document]
kozlovic: But for wildcard, are 2 copies made here? since we call string(nc.ps.ma.subject) twice? So yes, maybe evaluating if the subscription is wildcarded when created (so only once) and use that here would be better.

derekcollison:
Each message you receive has a non-wildcard subject. I think a low contention LRU would be fine here when receiving a message.

Note that with JetStream the subject will match the origin subject not the low level one, so even if you did a non-wildcard subscription you still need to check the subject.

kozlovic:
> Note that with JetStream the subject will match the origin subject not the low level one, so even if you did a non-wildcard subscription you still need to check the subject.

Ah right, with JetSteam the incoming message subject will likely NEVER match the subscription's subject, even if non-wildcard'ed. At this point (adding a map, etc..) I am really not sure if that will yield any benefit. We may trade one mem copy with CPU time that will make things worse. But if it can be evaluated that it does improve (for wildcard, non wildcard and JetStream), why not.

moredure:
I am not known with the way JetStream works yet, I'll check it.

moredure:
@kozlovic it's not much subject passed to js.Subject*() but it may match internal created Subscription.Subject or not match depending on wildcard it is or not.

kozlovic:
>  it's not much subject passed to js.Subject*() but it may match internal created Subscription.Subject or not match depending on wildcard it is or not.

No. The subject passed to js.Subscribe() is not the one going to be used by the internal subscription created in js.Subscribe() call. This will likely be a non-wc, but the point is that it could be a subscription on say "_INBOX.1", but the server will send to this subscription a message on subject "foo". This is because the server knows the subscription ID, but rewrites the subject. So you can't rely on the fact that the incoming message's subject will be the subscription's subject (even with no WC), at least for JetStream.

moredure:
Looks like I had a typo.

nats.go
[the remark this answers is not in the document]
moredure: @kozlovic no because of golang compiler optimisation

moredure:
Maybe it's time to start doing smth like this:
```go
sub := &Subscription{
  Subject: subj, 
  Queue: queue, 
  mcb: cb, 
  conn: nc, 
  jsi: js, 
  wildcard: wildcard(subj),
}
```

kozlovic:
But again, for JetStream, the subscription may be done on "_INBOX.1" but message may arrive on "foo". This has nothing to do with wildcards. At the very least, you should do this change only for `sub.jsi == nil` and leave current code for JS.

moredure:
I'll write the test case for that ` "_INBOX.1" but message may arrive on "foo".`. why is that happening?

kozlovic:
> Why is that happening, is JetStream some sort of broker that's why messages could be messed up?

JetStream is the persistence layer added to core NATS. It rewrites subjects. So you can't really write a test for this outside of using JetStream, unless you write your own "hand crafted" server, which we have in some tests...

moredure:
@kozlovic sounds reasonable, added check for jsi != nil || wildcard

moredure:
Even with two check it should be faster than comparing strings.

nats.go
```diff
@@ -2692,7 +2695,11 @@ func (nc *Conn) processMsg(data []byte) {
 	}
 
 	// Copy them into string
-	subj := string(nc.ps.ma.subject)
+	subj := sub.Subject
+	jsi := sub.jsi
```
kozlovic, line 2699: Yeah, but this is not entirely immutable (there is a case where sub.jsi is set to nil when a js subscription fails at some point), so you may or may not get race reports here...

kozlovic: Forgot to mention that where we had captured it down below, it was under the subscription lock... but now you move it outside of it.

kozlovic: Actually even sub.Subject can mutate (for a JS subscription in resetOrderedConsumer). Hum... this is getting a bit out of hand, not sure if this is all worth the effort. Have you actually measured how much we gain? (with a go bench report)?

moredure: I guess in this case 
sub.mu.Lock()

// Check if closed.
if sub.closed {
  sub.mu.Unlock()
  return
}
shows that subscription closed?

moredure: I have an another idea

kozlovic: But that's not the issue, the issue is that you access `sub.Subject` and `sub.jsi` outside of the subscription's lock while in some other go routines, those values may be changed at the same time (under the sub's lock). So you may get race reports.

moredure: Check latest changes

nats.go
```diff
@@ -2692,7 +2695,10 @@ func (nc *Conn) processMsg(data []byte) {
 	}
 
 	// Copy them into string
-	subj := string(nc.ps.ma.subject)
+	subj := sub.Subject
+	if sub.subjectMayDiffer {
```
kozlovic, line 2699: But again, not enough for the above `subj := sub.Subject`.. I am sorry to be a pain, but I am just stating that the way it is right now, you may still get DATA RACE report one day because of that.
So maybe something like:
```
var subj string
if sub.subjectMayDiffer {
    subj = string(nc.pas.ma.subject)
} else {
   subj = sub.Subject
}
```

nats.go
```diff
@@ -577,6 +577,9 @@ type Subscription struct {
 	// Type of Subscription
 	typ SubscriptionType
 
+	// Whether subject in messages and Subject in subscription may differ
+	subjectMayDiffer bool
```
wallyqs, line 581: would prefer short variable style for the optimization along with the comment, maybe:
```suggestion
	smd bool
```

nats.go
moredure: hmm, it's just a way to get rid of race condition checker message, because of I don't see any usage or changing Subject of subscription?

nats.go
moredure: Why not

nats.go
kozlovic: Yes we do, look for resetOrderedConsumer()...

moredure: Yes, but previous string already allocated and attached to subj anyway

kozlovic: We are ok because indeed the resetOrderSubject() is invoked only from this same go routine and after that line. Ok, we are good.

kozlovic (approved):
LGTM

nats.go
moredure:
```go
var g Subscription
g.Subject = "fun"
k := g.Subject
g.Subject = "not fun"
println(k) // fun
```

kozlovic: Oh, this is not what I was pointing out, what I was afraid is that resetOrderConsumer() could be invoked from a different go routine, which apparently and as of now is not. But if it was, you would basically have a situation where processMsg() from thread 1 would be doing `subj := sub.Subject` while thread 2 would be doing `sub.Subject = newSubject`. But again, seems like resetOrderConsumer() is invoked from the processMsg() go routine only, and after you capture the subject.

wallyqs (approved):
LGTM, let's squash into single commit before merging

moredure:
utilities @derekcollison mentioned inside nats-server
```go
// Determine if a subject has any wildcard tokens.
func subjectHasWildcard(subject string) bool {
	return !subjectIsLiteral(subject)
}

// Determine if the subject has any wildcards. Fast version, does not check for
// valid subject. Used in caching layer.
func subjectIsLiteral(subject string) bool {
	for i, c := range subject {
		if c == pwc || c == fwc {
			if (i == 0 || subject[i-1] == btsep) &&
				(i+1 == len(subject) || subject[i+1] == btsep) {
				return false
			}
		}
	}
	return true
}
```
But they are unexported.

kozlovic:
We could port, or tokenize the subject, which makes the code simpler (the one in server is written for performance, not readability). But even your current approach is ok to me since the worse case is that a string does not really have real wildcard, just `*` or `>` as part of a token, and in that case your change would make a copy of nc.ps.ma.subject (which is the original, safe, code).

kozlovic:
@moredure For your information: we had to release version v1.12.3 that reverts changes from your PR. There was unfortunately issues with non JS subscriptions used to receive messages from a JetStream consumer. In that case, message's subject was rewritten but we would use the subscription's subject (say the inbox, instead of "foo"). I added a test that will now make sure that any change to try to optimize things still don't break this use case.
So we still like the idea to try reduce memory copies, but this would need more tweaking.

kozlovic:
> as I understand plain subscription without wildcard can receive message with absolutely different subject name?

Yes, at least in the context of JetStream (for now).

moredure:
@kozlovic I don't know, to make another pull request with such a behaviour or is it better postpone until better times?)

kozlovic:
Would have to profile because it seems to me that doing what you suggest may still result in a copy (simply for the comparison, or maybe compiler optimize this, or we use bytes.Equal(), etc..). If you *want* and can provide some bench with various subscription types (wildcard, non wildcard, and various subjects length) and see if performance actually increases, etc.. this would help. But I understand that all this takes time and you may not have the time since your modification works well for your use case. Thanks again for the contribution, regardless.

moredure:
Exactly.

PR #827, opened 2021-09-21
[FIXED] Received JetStream message may have wrong subject

kozlovic:
The recent PR #824 tried to optimize reducing memory copy
by using the non JS, non wildcard subscription's subject
as the message subject.

However, in case where a NATS subscription is used for
the delivery subject of a AddConsumer call, then this
breaks and the message would be received with the subscription
subject instead of the subject that we get from MSG parsing.

This PR basically reverts changes from #824 and add a subject
to make sure that we catch such issue in the future.


derekcollison (approved):
LGTM - We can revisit subject interning, like the idea, just needs a bit tweaking.

PR #832, opened 2021-09-27
First Pass at KeyValue and ObjectStore

derekcollison:
This PR introduces two new materialized views to NATS/JetStream.

The KeyValue store. This is meant to also serve as a [kine](https://github.com/k3s-io/kine/blob/master/pkg/client/client.go#L23) implementation by design.

```go
type KeyValue interface {
	// Get returns the latest value for the key.
	Get(key string) (value []byte, revision uint64, err error)
	// Put will place the new value for the key into the store.
	Put(key string, value []byte) (revision uint64, err error)
	// Create will add the key/value pair iff it does not exist.
	Create(key string, value []byte) (revision uint64, err error)
	// Update will update the value iff the latest revision matches.
	Update(key string, value []byte, last uint64) (revision uint64, err error)
	// Delete the key and all revisions.
	Delete(key string) error
	// WatchAll will invoke the callback for all updates.
	WatchAll(cb KeyValueUpdate) (*Subscription, error)
	// Watch will invoke the callback for any keys that match keyPattern when they update.
	Watch(keys string, cb KeyValueUpdate) (*Subscription, error)
	// List will return all values for the key.
	List(key string) ([]*KeyValueEntry, error)
	// Bucket returns the current bucket name (JetStream stream).
	Bucket() string
}
```

And the ObjectStore.

```go
type ObjectStore interface {
	// PutObject will place the contents from the reader into a new stream.
	PutObject(cfg *ObjectConfig, reader io.Reader) error
	// GetObject will pull the object from the stream and write to the writer.
	GetObject(name string) (*ObjectResult, error)
	// PutFile is convenience function to put a file into an object store.
	PutFile(cfg *ObjectConfig, filename string) error
	// GetFile is a convenience function to pull and object and place in a file.
	GetFile(name, outfile string) error
}
```

I will setup a meeting to discuss some items with the extended team.


kv.go
```diff
@@ -0,0 +1,387 @@
...
+	}
+	return kv, nil
+}
+
+// AddKeyValue will create a KeyValue store with the following configuration.
+func (nc *Conn) AddKeyValue(cfg *KeyValueConfig, opts ...JSOpt) (KeyValue, error) {
+	if cfg == nil || cfg.Bucket == _EMPTY_ {
+		return nil, ErrBucketNameRequired
+	}
+	if strings.Contains(cfg.Bucket, ".") {
```
ripienaar, line 129: as above about name checks

kv.go
```diff
@@ -0,0 +1,387 @@
...
+
+	// Build by hand since simple and avoids stdlib JSON.
+	var b strings.Builder
+	b.WriteString("{\"filter\":\"")
+	b.WriteString(kv.pre)
+	b.WriteString(key)
+	b.WriteString("\"}")
+
+	// Send request.
+	subj := kv.js.apiSubj(fmt.Sprintf(apiStreamPurgeT, kv.stream))
```
ripienaar, line 275: delete works by writing an entry with a delete operation, not by purging

kv.go
```diff
@@ -0,0 +1,387 @@
...
+type kvs struct {
+	name   string
+	stream string
+	pre    string
+	nc     *Conn
+	js     *js
+}
+
+// Get returns the latest value for the key.
+func (kv *kvs) Get(key string) (value []byte, revision uint64, err error) {
```
ripienaar, line 187: Should return KeyValueEntry, languages can add language specific typed gets like this would be GetValue or GetBytes or something rather.  But important that we give access to the KeyValueEntry else you have no idea about historical state, delete operations etc

kv.go
```diff
@@ -0,0 +1,387 @@
...
+func (kv *kvs) Delete(key string) error {
+	o, cancel, err := getJSContextOpts(kv.js.opts)
+	if err != nil {
+		return err
+	}
+	if cancel != nil {
+		defer cancel()
+	}
+
+	// Build by hand since simple and avoids stdlib JSON.
```
ripienaar, line 267: simple yes, but with no checking of validity of key also trivial to break based on value of `key`

kv.go
```diff
@@ -0,0 +1,387 @@
...
+		return err
+	}
+	if resp.Error != nil {
+		return errors.New(resp.Error.Description)
+	}
+	return nil
+}
+
+// List will return all values for the key.
+func (kv *kvs) List(key string) ([]*KeyValueEntry, error) {
```
ripienaar, line 291: I'd have thought you'd make this an iterator or channel since you made the same for things like StreamNames? I find that completely unusable - but it seems it would be more consistent with the design of this library?

kv.go
```diff
@@ -0,0 +1,387 @@
...
+	}
+}
+
+// WatchAll watches all keys.
+func (kv *kvs) WatchAll(cb KeyValueUpdate) (*Subscription, error) {
+	return kv.Watch(AllKeys, cb)
+}
+
+// Watch will fire the callback when a key that matches the keys pattern is updated.
+// keys needs to be a valid NATS subject.
```
ripienaar, line 358: but in this implementation - due to not using operation on delete - would be unaware of deletes meaning one could not build a cache or ever actually accurate watched view of a key

ripienaar (commented):
looks good in general, a few comments of things the ADR covered and deviations on ADR design

kv.go
```diff
@@ -0,0 +1,387 @@
...
+	kvSubjectsPreTmpl = "$KV.%s."
+	kvNoPending       = "0"
+)
+
+// KeyValue will lookup and bind to an existing KeyValue store.
+func (nc *Conn) KeyValue(bucket string, opts ...JSOpt) (KeyValue, error) {
+	if bucket == _EMPTY_ {
+		return nil, ErrBucketNameRequired
+	}
+	if strings.Contains(bucket, ".") {
```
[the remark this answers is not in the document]
derekcollison: Beyond being a token? This was copied from elsewhere and felt like a top level "catch" for convenience.

object.go
```diff
@@ -0,0 +1,284 @@
...
+	"github.com/nats-io/nuid"
+)
+
+type ObjectStore interface {
+	// PutObject will place the contents from the reader into a new stream.
+	PutObject(cfg *ObjectConfig, reader io.Reader) error
+	// GetObject will pull the object from the stream and write to the writer.
+	GetObject(name string) (*ObjectResult, error)
+	// PutFile is convenience function to put a file into an object store.
+	PutFile(cfg *ObjectConfig, filename string) error
```
aricart, line 38: Put/GetFile are simply conveniences on a name that is a path - I don't think they are needed.

kv.go
derekcollison: I am not sure, be good to discuss.

kv.go
derekcollison: Another good discussion, these seem more like "one shot" vs paging and I was taking a lead from how Fetch() was done which returns []*nats.Msg.

kv.go
derekcollison: How would you be aware if I did do deletes?

kv.go
ripienaar: https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-8.md#deleting-values

you put a nil byte message with `KV-Operation: DEL` header.

object.go
```diff
@@ -0,0 +1,284 @@
...
+	"net"
+	"os"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/nats-io/nuid"
+)
+
+type ObjectStore interface {
```
aricart, line 32: It would seem that all KVs are a simplification of the ObjectStore, the blob storage is simply saying that the payload is larger than it fits in some message and some checksum is applied to validate against corruption. Wondering if merging the concepts has any value.

kv.go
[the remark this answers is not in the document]
ripienaar: https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-8.md#overview

```Valid keys are \A[-/_=\.a-zA-Z0-9]+\z, additionally they may not start or end in ., after encoding```

```Valid buckets are \A[a-zA-Z0-9_-]+\z```

Also covered here https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-6.md

File: kv.go (hunk 3 above)
Comment (ripienaar) [reply kv.go:187]: We did discuss these things at length during the ADR desgin process.  You gave a lot of feedback on the various data types of the `Item` property etc.

File: kv.go (hunk 6 above)
Comment (ripienaar) [reply kv.go:358]: as per ADR - a nil body with the correct header set. I would not know if you did a Purge for example though.

Comment (derekcollison) [reply kv.go:358]: Seems this is part of ADR I should go back and take a look at to see if that would actually work.

Comment (ripienaar) [reply kv.go:358]:
```
$ nats kv put TEST x.y z
z
$ nats kv del TEST x.y -f
$ nats kv get TEST x.y
nats: error: unknown key
$ nats kv history TEST x.y
╭──────────────────────────────────────────────────╮
│              History for TEST > x.y              │
├─────┬─────┬─────────────────────┬────────┬───────┤
│ Seq │ Op  │ Created             │ Length │ Value │
├─────┼─────┼─────────────────────┼────────┼───────┤
│ 1   │ PUT │ 27 Sep 21 14:15 UTC │ 1      │ z     │
│ 2   │ DEL │ 27 Sep 21 14:15 UTC │ 0      │       │
╰─────┴─────┴─────────────────────┴────────┴───────╯
$ ats s get KV_TEST --last-for '$KV.TEST.x.y' --json
{
  "subject": "$KV.TEST.x.y",
  "seq": 2,
  "hdrs": "TkFUUy8xLjANCktWLU9wZXJhdGlvbjogREVMDQoNCg==",
  "time": "2021-09-27T14:15:03.844010066Z"
}

aricart (commented):
Part of the reason why KvEntry is OK is that it is effectively a StoredMessage, with additional metadata about the bucket and the operation. With that said, I would much rather have a presentation where such values are in the headers. But there we get into additional issues: headers from the system may collide with user headers, and if the user sets a header that collides what is the meaning of that.

For replication and watches, the understanding of `DELETE` as an operation is important. `Empty` is really open to interpretation. The store could a have value by simply indicating that the key exists, thus the need for the operation to be telegraphed.

Outside of applications that deal with replication or that need to do something special on delete, the absence of the value is simpler. For NGS, my KV wrappers actually hide all the unpacking and simply code DELETEs as a null return value (ie I don't differentiate). But In some cases for example where keys should have a TTL, it is important to know the value doesn't exist even if it did.

kv.go
```diff
@@ -0,0 +1,387 @@
...
+	"errors"
+	"fmt"
+	"strconv"
+	"strings"
+	"time"
+)
+
+type KeyValue interface {
+	// Get returns the latest value for the key.
+	Get(key string) (value []byte, revision uint64, err error)
```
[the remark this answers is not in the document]
scottf: In java I had get for the value and getEntry to get the full entry, so this works for me. Java Get actually calls GetEntry and extracts the value for return

kv.go
```diff
@@ -0,0 +1,387 @@
...
+	// Put will place the new value for the key into the store.
+	Put(key string, value []byte) (revision uint64, err error)
```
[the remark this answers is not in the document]
scottf: In the ADR, there is no create, just put, and the behavior is create or replace

kv.go
```diff
@@ -0,0 +1,387 @@
...
+	// Update will update the value iff the latest revision matches.
+	Update(key string, value []byte, last uint64) (revision uint64, err error)
+	// Delete the key and all revisions.
+	Delete(key string) error
+	// WatchAll will invoke the callback for all updates.
+	WatchAll(cb KeyValueUpdate) (*Subscription, error)
+	// Watch will invoke the callback for any keys that match keyPattern when they update.
+	Watch(keys string, cb KeyValueUpdate) (*Subscription, error)
+	// List will return all values for the key.
+	List(key string) ([]*KeyValueEntry, error)
```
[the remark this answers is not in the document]
scottf: List is inefficient right now, but I thought there is plan to address this by being able to get the subjects for a stream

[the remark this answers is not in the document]
scottf: As a reference, here is the java interface: https://github.com/nats-io/nats.java/blob/main/src/main/java/io/nats/client/KeyValue.java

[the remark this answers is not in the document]
scottf: The java client has not implemented watches. Should watch / watch all be separate, or should watch rely on wildcard subject. ADR needs to be updated.

kv.go
scottf: Yeah, I found little utility for getting the entry except for the ability to know if a key was deleted. But it makes sense to expose the `getEntry` call because you get the entry anyway, it is the entire message. I in turn made `get`  just return the value (or null when key is missing or deleted) because to me that's the most common use. Probably should update the ADR

kv.go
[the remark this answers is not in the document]
ripienaar: ADR does cover watches extensively.

kv.go
scottf: If we want to fully delete the history we can directly purge it's subject

kv.go
scottf: yeah, right now I read the entire stream and put all subjects (keys) into a set, so at the end I have a set of the keys. Clearly not efficient.

kv.go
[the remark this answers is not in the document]
aricart: would `Init` be more indicative than `Create`?

object.go
```diff
@@ -0,0 +1,284 @@
...
+	// PutObject will place the contents from the reader into a new stream.
+	PutObject(cfg *ObjectConfig, reader io.Reader) error
+	// GetObject will pull the object from the stream and write to the writer.
+	GetObject(name string) (*ObjectResult, error)
```
aricart, line 36: Clients will possibly need a `MaxWindowSize` (max size of message payload on a message), and also `Range` - The start and end byte for the operation. This would allow a client to resume the download if it got disconnected.

ColinSullivan1:
With respect to the object store we should allow the design to support features (external or internal) such as:

Lifecycle Management (delete)
Event Notifications (add/delete/lock)
Locking
Archiving (tiered storage)
Searching/Indexing (tagging)
Versioning

With the proposed design it seems that all of this could be possible through an external system (primarily via object nomenclature) while keeping functionality as simple as possible here.  Just want to avoid making it difficult to build a future feature due to a design decision today.

derekcollison:
**Lifecycle Management (delete, TTL)**
 - We have TTL, but not explicit delete, can add though the delete.
 
**Event Notifications (add/delete/lock)**
  - We have these in general for streams (sans lock/seal). @ripienaar can confirm
  
**Locking**
 - I did a soft seal so to speak as we discussed. Would need for formalize to avoid deletes of messages, purges etc.
 - That can all be done now with perms but I think seal would be good eventually.
 
**Archiving (tiered storage)**
 - That is way bigger then just object stuff, that is a whole core JetStream thing.
 
**Searching/Indexing (tagging)**
 - Tagging maybe if we promote to stream itself.
 - Indexing feels external as we discussed.
 
**Versioning**
 - What do you mean here?
 -

kv.go
```diff
@@ -0,0 +1,436 @@
...
+
+// Create will add the key/value pair iff it does not exist.
+func (kv *kvs) Create(key string, value []byte) (revision uint64, err error) {
+	v, err := kv.Update(key, value, 0)
+	if err == nil {
+		return v, nil
+	}
+	// TODO(dlc) - Since we have tombstones for DEL ops for watchers, this could be from that
+	// so we need to double check.
+	if e, err := kv.Get(key); err == ErrMSgNotFound {
```
ripienaar, line 276: will this code ever actually be reached? since the error handling above wold bail if the first update failed, we'll never get a chance to correct that to be a update with sequence

kv.go
```diff
@@ -0,0 +1,436 @@
...
+		return kv.Update(key, value, e.Revision())
+	}
+	return 0, err
+}
+
+// Update will update the value iff the latest revision matches.
+func (kv *kvs) Update(key string, value []byte, revision uint64) (uint64, error) {
```
ripienaar, line 283: The godoc should clearly indicate these deviations from the spec if you wish to keep them - and keep adding more functions over time - vs the options to put approach.  And the godoc should provide a indication that in other implementation if you did that now you do this etc

kv.go
```diff
@@ -0,0 +1,436 @@
...
+	var b strings.Builder
+	b.WriteString(kv.pre)
+	b.WriteString(key)
```
ripienaar, line 286: needs validation

kv.go
```diff
@@ -0,0 +1,436 @@
...
+}
+
+// Delete the key and all revisions.
+func (kv *kvs) Delete(key string) error {
+	var b strings.Builder
+	b.WriteString(kv.pre)
+	b.WriteString(key)
+
+	// DEL op marker. For watch functionality.
+	m := Msg{Subject: b.String(), Header: Header{}}
```
ripienaar, line 305: use NewMsg() instead

kv.go
```diff
@@ -0,0 +1,436 @@
...
+	m.Header.Set(kvop, kvdel)
+	paf, err := kv.js.PublishMsgAsync(&m)
+	if err != nil {
+		return err
+	}
+	err = kv.js.purgeStream(kv.stream, &streamPurgeRequest{Subject: b.String(), Keep: 1})
```
ripienaar, line 311: wouldnt using a future here and an unconditional purge end up in cases where the delete op message failed AND the stream got purged? that seems lke a bad outcome. The delete op is critical for watchers functioning and we can then follow that with a purge.  And we should purge up the sequence of the delete op, not to keep 1, since a write could have come in between the delete op and the purge.  Purging to the sequence of the delete op means no data will be lost in that case

kv.go
```diff
@@ -0,0 +1,436 @@
...
+
+	pa, err := kv.js.PublishMsg(&m)
+	if err != nil {
+		return 0, err
+	}
+	return pa.Sequence, err
+}
+
+// Delete the key and all revisions.
+func (kv *kvs) Delete(key string) error {
```
ripienaar, line 299: Since recent discussions this is `Purge(key string)`, delete should only put the delete op message so history is preserved

kv.go
```diff
@@ -0,0 +1,436 @@
...
+	select {
+	case <-paf.Ok():
+		return nil
+	case err := <-paf.Err():
+		return err
+	}
+}
+
+// List will return all values for the key.
+func (kv *kvs) List(key string) ([]KeyValueEntry, error) {
```
ripienaar, line 326: Should be `History()`

kv.go
```diff
@@ -0,0 +1,436 @@
...
+func (kv *kvs) WatchAll(cb KeyValueUpdate) (*Subscription, error) {
+	return kv.Watch(AllKeys, cb)
+}
+
+// Watch will fire the callback when a key that matches the keys pattern is updated.
+// keys needs to be a valid NATS subject.
+func (kv *kvs) Watch(keys string, cb KeyValueUpdate) (*Subscription, error) {
+	var b strings.Builder
+	b.WriteString(kv.pre)
+	b.WriteString(keys)
```
ripienaar, line 401: needs validation

kv.go
```diff
@@ -0,0 +1,436 @@
...
+			})
+			if pending == kvNoPending {
+				done <- nil
+			}
+		}
+	}
+
+	var b strings.Builder
+	b.WriteString(kv.pre)
+	b.WriteString(key)
```
ripienaar, line 370: needs validation

ripienaar (changes requested):

kv.go
```diff
@@ -0,0 +1,436 @@
...
+// Retrieved entry for Get or List or Watch.
+type KeyValueEntry interface {
+	// Bucket is the bucket the data was loaded from.
+	Bucket() string
+	// Key is the key that was retrieved.
+	Key() string
+	// Value is the retrieved value.
+	Value() []byte
+	// Revision is a unique sequence for this value.
+	Revision() uint64
```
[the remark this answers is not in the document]
derekcollison: https://github.com/k3s-io/kine/blob/master/pkg/client/client.go#L23

kv.go
```diff
@@ -0,0 +1,436 @@
...
+)
+
+type KeyValueOp uint8
+
+const (
+	KeyValuePut KeyValueOp = iota
+	KeyValueDelete
+)
+
+// Retrieved entry for Get or List or Watch.
```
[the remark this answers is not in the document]
derekcollison: Will take a look. You mean like linter stuff yes?

kv.go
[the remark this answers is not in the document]
derekcollison: Sequence notation does not feel right to me since will most likely have gaps.

kv.go
```diff
@@ -0,0 +1,436 @@
...
+
+const (
+	kvBucketNameTmpl  = "KV_%s"
+	kvSubjectsTmpl    = "$KV.%s.>"
+	kvSubjectsPreTmpl = "$KV.%s."
+	kvNoPending       = "0"
+)
+
+// KeyValue will lookup and bind to an existing KeyValue store.
+func (js *js) KeyValue(bucket string) (KeyValue, error) {
```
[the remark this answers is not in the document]
derekcollison: I think the point was to be consistent though no? I did add in a sanitizeName helper for ObjectStore which we could also use here.

kv.go
```diff
@@ -0,0 +1,436 @@
...
+func (e *kve) Revision() uint64      { return e.revision }
+func (e *kve) Created() time.Time    { return e.created }
+func (e *kve) Delta() uint64         { return e.delta }
+func (e *kve) Operation() KeyValueOp { return e.op }
+
+// Get returns the latest value for the key.
+func (kv *kvs) Get(key string) (KeyValueEntry, error) {
+	var b strings.Builder
+	b.WriteString(kv.pre)
+	b.WriteString(key)
```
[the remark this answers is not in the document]
derekcollison: I could use sanitize here as well, do we have any mentions in ADR of what makes a proper key?

kv.go
```diff
@@ -0,0 +1,436 @@
...
+		key:      key,
+		value:    m.Data,
+		revision: m.Sequence,
+		created:  m.Time,
+	}
+
+	// Double check here that this is not a DEL Operation marker.
+	if len(m.Header) > 0 && m.Header.Get(kvop) == kvdel {
+		entry.op = KeyValueDelete
+		return entry, ErrMSgNotFound
```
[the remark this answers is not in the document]
derekcollison: Should we create specific errors?

The whole errors thing needs some attention I noticed.

kv.go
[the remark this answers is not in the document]
ripienaar: Other implementations also use Sequence, so right or not, consistent is better than right.

kv.go
derekcollison: Will take a look.

kv.go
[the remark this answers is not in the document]
ripienaar: yeah

kv.go
derekcollison: Which spec are you referring to, the ADR?

Help me understand what you are looking for here.

kv.go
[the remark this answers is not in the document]
ripienaar: indeed, but you should be able to bind to any properly prepared stream (that was your request), and I think for replicas being specifically read only (being mirrors, typically) we would want to bind to them by specific name rather than have someone accidently access them as if they are a full on bucket.

kv.go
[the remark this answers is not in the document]
ripienaar: yes near the top, its quite strict, dont want to entirely close the door to future other backends

kv.go
derekcollison: ok will update.

kv.go
ripienaar: yeah, ADR would have `Put(k, v, opts...)` to implement Update and Creates and whatever other things we wish to add in future. Though I agree Creates() reads nicer, so feel free to keep them instead, just point out in the comments that this is equivalent to `Put(k, v, OnlyIfLastValueSequence(0))` though that's not nice think we're still thinking about actual correct names.

kv.go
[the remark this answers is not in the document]
derekcollison: Well some of the largest I know of use Revision, never seen one call it sequence.

kv.go
ripienaar: yes, just use a rollup

kv.go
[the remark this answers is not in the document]
derekcollison: I agree originally I thought you should be able to free-form KV and ObjectStores and we would *allow* them, but do not think that now. Could be a support nightmare imo.

Read-Replicas and all of that need a bunch more thought around it so not touching that here. Also is applicable to more than just KV.

kv.go
derekcollison: Gotcha.

kv.go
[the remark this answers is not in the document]
ripienaar: ok, anyway we can add later if its a vararg it should at least not break too much :)

kv.go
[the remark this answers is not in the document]
ripienaar: Index in consul - its largely irrelevant, best to be consistent

kv.go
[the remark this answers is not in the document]
derekcollison: Yes agree, wanted to try to avoid duplicating the JSOpt so went with binding to JS which solves that, now we can add in KVOpt or ObjOpts if needed but prefer to see how far we can get without them.

kv.go
```diff
@@ -0,0 +1,387 @@
...
+}
+
+// Used to watch all keys.
+const AllKeys = ">"
+
+// Value is for listing all values with revisions.
+type KeyValueEntry struct {
+	Key      string
+	Data     []byte
+	Revision uint64
```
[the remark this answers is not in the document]
derekcollison: If I update an existing key, does created move to that time? Should it be Modified vs Created?

kv.go
derekcollison: Does Purge leave a marker or really make everything go away?

kv.go
derekcollison: I don't see people doing crazy big histories do we?

kv.go
```diff
@@ -0,0 +1,387 @@
...
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strconv"
+	"strings"
+	"time"
+)
+
+type KeyValue interface {
```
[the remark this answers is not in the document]
derekcollison: My fault.. But, I do think we need to have a discussion around the ADR process internally.

kv.go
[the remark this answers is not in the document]
derekcollison: Create only succeeds if not present, Update only if it is present and has not changed, and Put just works all the time sans other issues.

kv.go
ripienaar: Purge is a roll up with delete.

kv.go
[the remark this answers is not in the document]
derekcollison: List, now History is very efficient actually, but question is do we think there will be very very large history per key? Should we limit to say 128/256 max as a soft limit that we can revisit if needed?

kv.go
ripienaar: I considered capping it to some number below 100

object.go
derekcollison: I have a few toy apps that combine them already I was playing with over the weekend.

object.go
derekcollison: We have ChunkSize, agree about range eventually but we can add it in.

kv.go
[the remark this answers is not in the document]
scottf: I just meant that this impl has `Watch` and `WatchAll` but the adr only shows `Watch` which supports wildcards or empty, so is `WatchAll` necessary?

kv.go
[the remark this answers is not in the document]
derekcollison: Have ErrKeyNotFound and ErrKeyDeleted now.

kv.go
derekcollison: Fixed server and updated.

object.go
scottf: I had start byte in the chunk meta data and chunk size in the object meta data which would support figuring out a sequence number to start with.

kv.go
derekcollison: How do we delete the delete markers? Do they always just accumulate? I think we place the marker for connected watchers but purge whole stream after.

ripienaar: Yeah I don’t think purge will be that common that accumulating tbem will be a problem. Also why I didn’t think purge just deleting everything is a problem - it’s a unusual situation indicating we are done with this key. Shouldn’t have any watchers. 

One use case for purge is if someone just put like 1MB values in filling history and someone just wants to fix that. Then a delete marker is no biggie

derekcollison: What I have now is Purge places delete marker through sync js.Publish() then Purges the key in the stream.

kv.go
derekcollison: I am ok with that.. 64 to start?

kv.go
ripienaar: Roll up didn’t work? We added that feature for this

kv.go
ripienaar: Yes

kv.go
derekcollison: Rollup works like designed. But when you Delete, we place marker, or when you do a Purge we do as well now. But, who deletes the marker?

For now in Purge() I do same as Delete with sync js.PublishMsg() for the empty delete marker to kick connected watchers then I purge complete key including the marker.

kv.go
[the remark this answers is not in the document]
derekcollison: Just a helper for Watch(ALL_KEYS)

kv.go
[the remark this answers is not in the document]
derekcollison: I believe they should all have this functionality.

kv.go
ripienaar: Delete ads a entry keeping history. Purge is for removing history entirely leaving a tiny marker. 

I had it with no marker. You said that’s completely useless since it breaks the entire concept of a watch. 

So now purge deleted history and leaves a tiny marker. That’s why you added the roll up feature to deliver that in a single shot. 

What you are doing only works if everything is working. It a watch is just disconnected for a short while it gets stuck and unaware leaving us back to square one with broken watchers which you said is entirely unacceptable

ripienaar: Having history showing when the purge was done is valuable and I think in line with the general design of hjstoric values as a concept. So I really think the right behaviour is a roll up delete with marker

derekcollison: Then we need another call to clear those markers. I think the way I implemented it makes sense, will trigger connected watchers, but purge whole key on Purge including the marker.

On a new Watch if you have local state and you will need to compare anyway for things that have been deleted from your local cache. Most of the time you will have a marker I know.

Even Delete() feels like it should maybe work like Purge. We will get a bunch of folks wondering why they run out of memory doing lots of Puts and Deletes IMO.

derekcollison:
ok I pushed updates to KV. Will do small update to Object in a bit for versions and Purge as well, etc.

kv.go
jnmoyne: > Even Delete() feels like it should maybe work like Purge. We will get a bunch of folks wondering why they run out of memory doing lots of Puts and Deletes IMO.

Remember that (history filling up with put/deletes, rathe than just puts) would only happen if the developers specifically set the history depth to something else than the default (1)? You probably wouldn't set history to more than one if you know you're going to be doing puts and deletes (rather than just puts to update a values).

Once you do set history to > 1 then you would understand (would have to be documented too) that deletes are part of the history and therefore that if you do lots of puts and deletes that could fill up your history.
If you set history > 1 then the function of a purge is one of 'delete the value and compact the history' keeping a marker of the purge is then equivalent to 'compacting the history to 1' vs 'compacting the history to 0' (personally given that this purge marker would anyway be very small I'm ok with it being there and compacting history to 1.

By default with history = 1 there functionally no difference between a delete and a purge, history remains at 1 the only difference is you can tell after the fact that a delete vs a purge was used.

kozlovic:
@derekcollison Following the PR with the FC fix, this PR has now conflict that will need to be addressed. We need to be careful in reverting your FC changes that we are not putting back the way it was prior to the FC fix PR :-)

kv.go
derekcollison: They are needed to some degree for watchers but also if you want to have an audit trail.

I did ad in PurgeDeletes() if that is needed.

kv.go
```diff
@@ -0,0 +1,488 @@
...
+			key:      subj,
+			value:    m.Data,
+			revision: uint64(parseNum(tokens[ackStreamSeqTokenPos])),
+			created:  time.Unix(0, parseNum(tokens[ackTimestampSeqTokenPos])),
+			delta:    uint64(parseNum(tokens[ackNumPendingTokenPos])),
+			op:       op,
+		})
+	}
+	// Used ordered consumer to deliver results.
+	return kv.js.Subscribe(b.String(), update, OrderedConsumer(), DeliverLastPerSubject())
```
ripienaar, line 532: There's a missing behavior here mentioned in the ADR:

```
When a stream is empty at create time - `NumPending+Delivered.Consumer==0` - a nil is sent across the channel to signal to the caller that they should not expect any values.
```

The purpose of this is in creating in-memory caches, you want the cache to be ready and usable as soon as possible, generally the cache can figure this out by looking at delta in the received entries (delta==0 is last message).

But when there are no messages at all in the stream the watch will appear to just wait, the calling code will never be notified that there is no waiting messages and that its in effect ready even when no entries came out of the watch.

Without this a cache would never know if its ready - or perhaps just disconnected or have a broken path between it and the stream

ripienaar: not at the moment required and it would be an opt in for using the KV not on by default. We're focussing on getting core behaviors right now.

But we need to make it possible to build such a cache and I have a implementation that I use to test that its possible and makes sense.

scottf: @derekcollison Is ordered consumer necessary here? If someone wants to audit and not miss an entry, then fine, but if I'm building a cache and miss an entry but get the latest, what's the harm? Maybe ordered should be an option to the watch.

ripienaar: Building a cache and missing an entry is pretty serious, when that entry is a delete.

Your cache will not be consistent with reality.

scottf: Also about a local cache. Is it the expectation that when a KV context is created and a local cache is started, that we warm the entire cache by reading everything in the kv? It seems that it's sufficient to watch and use the results of get, and even expensive to try to warm a cache, when a user wants a cache, again there certainly could be reasons for a client to not want a cache.

ripienaar: Cache design not final, the implementation I have do warm itself and acts as a pass through when not ready, but a simple on demand read cache is also of course possible.

A cache built with a watcher must be able to delete items when they are deleted / purged elsewhere though, hence pointing out the missing behavior here.

derekcollison: For cache it should not be considered to be 1:1 match for system of record, so what I mean is cache is ready even before any callbacks or populations happen since if you miss (no cache record) you have to ask system of record anyway.

So not sure we need the empty marker, WDYT?

ripienaar: No, I 100% think we need the marker. You can make a simple read cache but one populated by a watcher is slightly different and the marker will help in various scenarios.

You can trust me, I've written several of these caches and the feature was added to the ADR since its needed

derekcollison: I have written quite a few as well so maybe not understanding. What does the boolean tell you besides that you have received all first (last) values?

What I mean is that the delete markers will populate, say you don't wait (which I am suggesting we should not) when you hook the watcher. A Get that does not see any information in the cache, no data or delete marker, has to be passed to system of record.

So if you wait you might save the first Get that would miss, maybe.. But nothing else imo but happy to understand more here.

ripienaar: when you start the watcher on an empty stream, its not going to get anything, it will never call the callback until the first write to the bucket happens.

From the perspective of the code that started the watch, what does it mean when that nothing is happening?

 * Is there no data?
 * Is the ordered consumer disconnected and trying to connect again?
 * Did the bucket vanish and it's repeatedly trying to recreate the consumers?

With the marker saying "there is no data" the calling code can know it can go ahead and do whatever it felt like doing with the data it gathers from the watch.

This marker only happens on EMPTY buckets.

ripienaar: And remember, it's not always that easy in watch we can do watch `x.>`, you cannot do a Get() trick to find out the empty stream scenario. 

The only way to do this is with the method described, users would have no way to build something as basic as "dump all values under x.>" without having signal indicating that there's no data matching the pattern.

derekcollison: After speaking directly with @ripienaar I understand better what he was aiming at, will add it in with Keys() as well.

object.go
```diff
@@ -0,0 +1,706 @@
...
+)
+
+type obs struct {
+	name   string
+	stream string
+	js     *js
+}
+
+// CreateObjectStore will create an object store.
+func (js *js) CreateObjectStore(cfg *ObjectStoreConfig) (ObjectStore, error) {
```
wallyqs, line 167: I think maybe we should update the interface to have the partial opts like:
```suggestion
func (js *js) CreateObjectStore(cfg *ObjectStoreConfig, opts ...JSOpt) (ObjectStore, error) {
```
so that later on can pass the arguments below for example like: `js.AddStream(scfg, opts)`
That would make it possible to change the timeout or context for example.

PR #837, opened 2021-10-05
[FIXED] JetStream flow control may stall in some conditions

kozlovic:
Keep track of the "sequence" of incoming messages that are then compared with the number of delivered message.
However, ChanSubscription were broken: the FC would *always* be responded to when received, because pMsgs (and pBytes) would always be 0 for these type of subscriptions.

We now check for flow control on all incoming messages for ChanSubscription (and evaluate the "delivered" count based on the incoming sequence minus length of the user's channel). We also have a long-lived go routine to cover situations where the evaluation during the incoming message did not result in the target to send the FC, but then the user consumes enough messages to bring the "delivered" at or past the target. Without this go routine, the client would stall and have to wait for an heartbeat with the "fc stalled" header in order to be unblocked.

Alternatively, we could fail creation of ChanSubscription subscriptions if flow control is in use... 

Finally, this PR fixes some issues with the ordered consumer test:
- We send a "EOF" empty message which should not be counted as the number of chunks used to reconstitute the asset
- Some "message filters" that are removed as part of the execution of the filter's callback would not be put back for the following "sync" test (we test async then sync).

### DIFF
--- a/js.go
+++ b/js.go
@@ -906,6 +906,7 @@ type jsSub struct {
 	fcr    string
 	fcd    uint64
 	fciseq uint64
+	csfct  *time.Timer
 }
 
 // Deletes the JS Consumer.
@@ -1467,7 +1468,7 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync,
 	// start a go routine that evaluates the number of delivered messages
 	// and process flow control.
 	if sub.Type() == ChanSubscription && hasFC {
-		go sub.chanSubcheckForFlowControlResponse()
+		sub.chanSubcheckForFlowControlResponse()
 	}
 
 	return sub, nil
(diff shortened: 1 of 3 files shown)


js.go
```diff
@@ -1581,7 +1636,7 @@ func (sub *Subscription) resetOrderedConsumer(sseq uint64) {
 		// Reset some items in jsi.
 		jsi.dseq = 1
 		jsi.cmeta = _EMPTY_
-		jsi.fcr, jsi.fcd = _EMPTY_, 0
+		jsi.fcr, jsi.fcd, jsi.fciseq = _EMPTY_, 0, 0
```
kozlovic, line 1639: I just made that change (with push force) after submitting the PR but I think it is wrong, since the subscription is maintained, the sub.delivered count is not reset either, so this value should not be reset. Will revert this change and push force again.

js.go
```diff
@@ -1442,10 +1463,41 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync,
 	if hasHeartbeats {
 		sub.scheduleHeartbeatCheck()
 	}
+	// For ChanSubscriptions, if we know that there is flow control, we will
+	// start a go routine that evaluates the number of delivered messages
+	// and process flow control.
+	if sub.Type() == ChanSubscription && hasFC {
+		go sub.chanSubcheckForFlowControlResponse()
```
derekcollison, line 1470: Why does it have to be in Go routine?

js.go
```diff
@@ -1442,10 +1463,41 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync,
...
+// This long-lived routine is used per ChanSubscription to check
+// on the number of delivered messages and check for flow control response.
+func (sub *Subscription) chanSubcheckForFlowControlResponse() {
+	sub.mu.Lock()
+	nc := sub.conn
+	sub.mu.Unlock()
+	if nc == nil {
+		return
+	}
+	t := time.NewTicker(chanSubFCCheckInterval)
```
derekcollison, line 1485: Should probably have a chan here that is closed when sub is unsubscribed or connection is closed.

js.go
```diff
@@ -1442,10 +1463,41 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync,
...
+	for range t.C {
+		sub.mu.Lock()
+		if sub.closed {
+			sub.mu.Unlock()
+			t.Stop()
```
derekcollison, line 1490: Just do top level defer for when you leave the function.

js.go
```diff
@@ -1442,10 +1463,41 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync,
...
+			return
+		}
+		fcReply := sub.checkForFlowControlResponse()
+		sub.mu.Unlock()
+		if fcReply != _EMPTY_ {
```
derekcollison, line 1495: If we don't check for errors just send blindly? No need to check fcReply?

js.go
kozlovic: Doesn't have to be... could add a timer.

js.go
kozlovic: Could add a chan too (in addition to the timer). I thought that it was ok like this, meaning that this go routine may linger for 250ms and then we would detect that the subscription is gone.

js.go
kozlovic: Have to check for each tick if the subscription is closed, so will need to get the lock here anyway.

js.go
kozlovic: Well, sure, Conn.Publish() checks for empty subject and returns an error, but I thought that it makes more readable that we know that fcReply will be non empty only when we need to reply back to FC.

js.go
kozlovic: Changed to use a timer instead.

js.go
kozlovic: Refactored that as a timer's func

js.go
kozlovic: Since I use timer now, it is stopped when the subscription is stopped/removed.

derekcollison (approved):
LGTM

Issue #833, opened 2021-09-28
JetStream Work Queue - PullSubscribe & Fetch


T-J-L:
Hello,

I’m working on a POC with nats JetStream, to implement a durable work queue. The jobs will be low throughput, long running jobs, which may run for as long as 20 minutes. There will be multiple services publishing and receiving from a single subject.

I’m trying to accept the message, mark it as in progress using an upper bound for the processing time, perform the work, then ack once complete.

The issue I’m having is with js.PullSubscribe and sub.Fetch. If I use nats.Context(ctx) with Fetch, I seem to receive zero or one messages, the subscription then stops receiving any further messages. 

I'm new to nats so I expect this is user error, any help would be appreciated. See an example below which reflects the scenario. I am expecting this to continuously publish and receive messages.

Versions:
- go version go1.17.1 darwin/amd64
- github.com/nats-io/nats.go v1.12.3
- Docker image nats:2.6.1 (command `-js -m 8222 -n nats_js_1`)

```go
package main

import (
	"context"
	"errors"
	"fmt"
	"log"
	"time"

	"github.com/nats-io/nats.go"
)

func main() {

	// Connect to NATS
	nc, err := nats.Connect(nats.DefaultURL)
	if err != nil {
		log.Fatalln(err)
	}
	defer nc.Close()

	// Create JetStream Context
	js, err := nc.JetStream(nats.PublishAsyncMaxPending(256))
	if err != nil {
		log.Fatalln(err)
	}

	// Create WorkQueue stream
	_, err = js.AddStream(&nats.StreamConfig{
		Name:       "WORK",
		Subjects:   []string{"WORK.*"},
		Storage:    nats.FileStorage,
		Retention:  nats.WorkQueuePolicy,
		Discard:    nats.DiscardNew,
		Duplicates: 5 * time.Minute,
		MaxMsgs:    200,
		MaxBytes:   -1,
	})
	if err != nil {
		log.Fatalln(err)
	}

	// Create a pull subscription.
	sub, err := js.PullSubscribe("WORK.request", "WORK")
	if err != nil {
		log.Fatalln(err)
	}

	// Monitor sub
	go func() {
		ticker := time.NewTicker(10 * time.Second)
		for range ticker.C {
			con, err := sub.ConsumerInfo()
			if err != nil {
				log.Fatalln(err)
			}
			log.Printf("NumPending=%v NumWaiting=%v\n", con.NumPending, con.NumWaiting)
		}
	}()

[...]

PR #838, opened 2021-10-05
js: Fix context usage with sub.Fetch

### LINKED ISSUE #834: Fetch ignores context timeout.
nats.go version: **1.13.3**
nats version: **2.6.1**

I have examined at the Fetch code in `js.go` and I think there may be some issues with the logic:

When a context is passed in that has a context timeout, and no explicit `PullOpts` timeout, the code sets the internal `ttl` to the default timeout value.  It uses this default value `ttl` to time out the request returning `context.DeadlineExceeded` thereby ignoring the context timeout value.

When a timeout value is provided through `PullOpts` without a context, the timeout value is respected.

The check at the start of the function prevents supplying both, so therefore it is therefore impossible to provide a cancellable context to Fetch, and and also a timeout.

Providing no timeout value is causing Fetch to stall for us, but I also need to be able to cancel the context during a graceful shutdown of the server.

wallyqs:
The deadline of a context is now used to calculate the time used for `expires` instead of the default `ttl` of the JetStream context which was `5s` by default.  This was preventing library users from passing a context with a custom timeout.

This also disallows the usage of `context.Background` to make it explicit that `sub.Fetch` has to be used with a context that has a timeout since each fetch request has to include an expire time anyway.

In case `context.WithCancel` is used, then a child context with the same duration as the JetStream context default timeout will be created. 

It is also updated to prevent using both timeout and context options with `msg.InProgress()`, `msg.AckSync()`.

Fixes #834 #833 

### DIFF
--- a/js.go
+++ b/js.go
@@ -2148,6 +2148,8 @@ func (sub *Subscription) Fetch(batch int, opts ...PullOpt) ([]*Msg, error) {
 	js := sub.jsi.js
 	pmc := len(sub.mch) > 0
 
+	// All fetch requests have an expiration, in case of no explicit expiration
+	// then the default timeout of the JetStream context is used.
 	ttl := o.ttl
 	if ttl == 0 {
 		ttl = js.opts.wait
@@ -2161,8 +2163,19 @@ func (sub *Subscription) Fetch(batch int, opts ...PullOpt) ([]*Msg, error) {
 		err    error
 		cancel context.CancelFunc
 	)
-	if o.ctx == nil {
+	if ctx == nil {
 		ctx, cancel = context.WithTimeout(context.Background(), ttl)
+		defer cancel()
+	} else if _, hasDeadline := ctx.Deadline(); !hasDeadline {
+		// Prevent from passing the background context which will just block
+		// and cannot be canceled either.
+		if octx, ok := ctx.(ContextOpt); ok && octx.Context == context.Background() {
+			return nil, ErrNoDeadlineContext
+		}
+
+		// If the context did not have a deadline, then create a new child context
+		// that will use the default timeout from the JS context.
+		ctx, cancel = context.WithTimeout(ctx, ttl)
 		defer cancel()
 	}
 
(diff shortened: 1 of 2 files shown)


js.go
```diff
@@ -2166,6 +2168,19 @@ func (sub *Subscription) Fetch(batch int, opts ...PullOpt) ([]*Msg, error) {
 		defer cancel()
 	}
 
+	deadline, hasDeadline := ctx.Deadline()
```
kozlovic, line 2171: Few lines above, we set `ctx = o.ctx`, which means we use what is set in the option. Then we say `if o.ctx == nil` we create a context with timeout. So I feel like we should, to make it clear, check `if ctx == nil` (instead of o.ctx) and then have a `else` where we check for "hasDeadline" in the user provided context.


The `deadline` can then be gotten down below, before the `ttl = time.Until(deadline)`

js.go
```diff
@@ -2166,6 +2168,19 @@ func (sub *Subscription) Fetch(batch int, opts ...PullOpt) ([]*Msg, error) {
...
+	if !hasDeadline {
+		// Prevent from passing the background context which will just block
+		// and cannot be canceled either.
+		if octx, ok := ctx.(ContextOpt); ok && octx.Context == context.Background() {
+			return nil, ErrNoDeadlineContext
+		}
+		// If the context did not have a deadline, then create a new child context
+		// that will use the default timeout from the JS context.
+		ctx, cancel = context.WithTimeout(ctx, ttl)
```
kozlovic, line 2180: Do we need to create a context with timeout from the user-given context here? If not, then code could be even more refactored to check for `ctx = o.ctx` being not nil first, just to check for "has deadline" and if so set ctx to nil so there is then the check for `nil` and creation with background() context. If we need to use the user-given context, then ok with this code.

kozlovic (changes requested):
Would recommend a little bit of refactor and more depending on answer to question regarding user-given context without deadline.

js.go
wallyqs: Yes we need to chain the context from the parent one so that if it is canceled then the child one is canceled as well.

kozlovic (approved):
LGTM

PR #845, opened 2021-10-11
add kv and object status functionality

ripienaar:


derekcollison:
I am not sure that buys us anything when asking about the status (optimization), and feel that there should be some consistency between StreamInfo, KVStatus and ObjectStoreStatus.

derekcollison:
Thanks that is what we did, just made ObjectStore the same.

PR #846, opened 2021-10-13
reimplement bucket status to avoid unspecified maps

ripienaar:


ripienaar:
Updated as per call, obj also hide the implementation specific function now

PR #856, opened 2021-11-04
Proposal: Conditional delete and purge for KV

steveh:
I've been testing out the experimental KV functionality, and conditional delete and purge would be very useful to us.

It appears to work out of the box on the server too in my limited testing. If you're happy with the naming and implementation I will flesh out the tests more.

### DIFF
--- a/kv.go
+++ b/kv.go
@@ -54,9 +54,9 @@ type KeyValue interface {
 	// Update will update the value iff the latest revision matches.
 	Update(key string, value []byte, last uint64) (revision uint64, err error)
 	// Delete will place a delete marker and leave all revisions.
-	Delete(key string) error
+	Delete(key string, opts ...DeleteOpt) error
 	// Purge will place a delete marker and remove all previous revisions.
-	Purge(key string) error
+	Purge(key string, opts ...DeleteOpt) error
 	// Watch for any updates to keys that match the keys argument which could include wildcards.
 	// Watch will send a nil entry when it has received all initial values.
 	Watch(keys string, opts ...WatchOpt) (KeyWatcher, error)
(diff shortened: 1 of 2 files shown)


ripienaar:
I like the idea but we need better function names I think, any suggestions @derekcollison ?

kv.go
```diff
@@ -45,14 +45,18 @@ type KeyValue interface {
 	Put(key string, value []byte) (revision uint64, err error)
 	// PutString will place the string for the key into the store.
 	PutString(key string, value string) (revision uint64, err error)
-	// Create will add the key/value pair iff it does not exist.
+	// Create will add the key/value pair if it does not exist.
```
derekcollison, line 48: iff is not a mis-spelling, it is used commonly to mean 'if and only if'..

derekcollison:
I would add an option to those that takes a revision for the conditional delete vs adding new methods.

steveh:
Updated to use options instead of additional functions.

kv.go
```diff
@@ -495,16 +519,35 @@ func (kv *kvs) delete(key string, purge bool) error {
...
+		m.Header.Set(ExpectedLastSubjSeqHdr, strconv.FormatUint(o.revision, 10))
+	}
+
 	_, err := kv.js.PublishMsg(m)
 	return err
 }
 
+// Purge will remove the key and all revisions.
+func (kv *kvs) Purge(key string, opts ...DeleteOpt) error {
+	return kv.Delete(key, append(opts, Purge())...)
```
ripienaar, line 548: the `Purge()` option is a bit weird and might be confusing to end users that `Purge(Purge())` is valid.

Really `Purge()` is internal behavior only so maybe we do not export the `Purge()` option and use it internally only?

ripienaar (approved):
LGTM

PR #872, opened 2021-12-14
js: make js.Subscribe context aware

wallyqs:
Can now attach a context to a subscription so that it is unsubscribed and/or consumer deleted via propagation of cancellation via parent context.  This helps specially with `kv.Watch` functions that have an ordered consumer subscription.

### DIFF
--- a/js.go
+++ b/js.go
@@ -859,6 +859,11 @@ func (ctx ContextOpt) configurePublish(opts *pubOpts) error {
 	return nil
 }
 
+func (ctx ContextOpt) configureSubscribe(opts *subOpts) error {
+	opts.ctx = ctx
+	return nil
+}
+
 func (ctx ContextOpt) configurePull(opts *pullOpts) error {
 	opts.ctx = ctx
 	return nil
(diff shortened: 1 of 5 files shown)


ripienaar:
Nice, I am not that concerned to make this work for kv atm though your change is small enough there.

For kv we must not accept WatchOpts on various places where its accepted now - we need to add a new kind of opt like CancelOpt that only accepts `nats.Context()` since that the only valid option for `Keys()` for example.  I can take care of that once the basic cancelable subscribe and order consumer is in place

js.go
```diff
@@ -1413,6 +1430,14 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync,
 		return nil, err
 	}
 
+	// Wait for context to get canceled if there is one.
```
kozlovic, line 1433: You should move this block at the end of the function, on success. You will note that at around line 1530 (or 1505 of original code), the subscription may be recreated.

kozlovic (changes requested):

kozlovic (approved):
LGTM

tpihl:
Will this work as a cancel more messages or will it delete a durable subscriber created in the subscribe call with WithDurable?

wallyqs:
If the app auto-created the durable (that is, without using `AddConsumer`) then the durable subscriber will be deleted when the context is canceled as well.

PR #873, opened 2021-12-15
Fix jetstream removeSub logic

boxboatmatt:
Signed-off-by: Matthew DeVenny <matt@boxboat.com>

When `removeSub` calls `jsi.hbc.Stop()` it is possible that the timer has already expired
https://github.com/nats-io/nats.go/blob/main/nats.go#L3919

from `timer.Stop()`
```
For a timer created with AfterFunc(d, f), if t.Stop returns false, then the timer has already expired 
and the function f has been started in its own goroutine; Stop does not wait for f to complete before 
returning. If the caller needs to know whether f is completed, it must coordinate with f explicitly.
```

This PR checks to see if `sub.closed == true` and returns

### DIFF
--- a/js.go
+++ b/js.go
@@ -1838,7 +1838,7 @@ func (sub *Subscription) scheduleFlowControlResponse(reply string) {
 func (sub *Subscription) activityCheck() {
 	sub.mu.Lock()
 	jsi := sub.jsi
-	if jsi == nil {
+	if jsi == nil || sub.closed {
 		sub.mu.Unlock()
 		return
 	}
@@ -1847,10 +1847,9 @@ func (sub *Subscription) activityCheck() {
 	jsi.hbc.Reset(jsi.hbi)
 	jsi.active = false
 	nc := sub.conn
-	closed := sub.closed
 	sub.mu.Unlock()
 
-	if !active && !closed {
+	if !active {
 		nc.mu.Lock()
 		if errCB := nc.Opts.AsyncErrorCB; errCB != nil {
 			nc.ach.push(func() { errCB(nc, sub, ErrConsumerNotActive) })


js.go
```diff
@@ -1844,7 +1844,14 @@ func (sub *Subscription) activityCheck() {
 	}
 
 	active := jsi.active
-	jsi.hbc.Reset(jsi.hbi)
+	// the timer could still have gone off even though it was stopped in removeSub,
+	// in this case it will be nil because removeSub sets it to nil
+	if jsi.hbc != nil {
+		jsi.hbc.Reset(jsi.hbi)
+	} else {
+		active = false
```
boxboatmatt, line 1852: Not 100% sure whether we should be setting `active` or `closed` to false here in this dangling execution

js.go
```diff
@@ -1844,7 +1844,14 @@ func (sub *Subscription) activityCheck() {
 	}
 
 	active := jsi.active
-	jsi.hbc.Reset(jsi.hbi)
+	// the timer could still have gone off even though it was stopped in removeSub,
```
kozlovic, line 1847: Would revert the changes here and instead few lines above change the condition to:
```
if jsi == nil || sub.closed {
    sub.mu.Unlock()
    return
}
```
(or maybe even `if sub.closed` would be enough)

kozlovic (changes requested):

js.go
boxboatmatt: Yup.. that would work

kozlovic: Like I said, this change can be reverted. Since jsi.hbc is set to nil at the same time (and under sub.mu.Lock) we set sub.closed = true in removeSub, we are ok with original code.

kozlovic (changes requested):

js.go
```diff
@@ -1838,7 +1838,7 @@ func (sub *Subscription) scheduleFlowControlResponse(reply string) {
 func (sub *Subscription) activityCheck() {
 	sub.mu.Lock()
 	jsi := sub.jsi
-	if jsi == nil {
+	if jsi == nil || sub.closed {
```
kozlovic, line 1841: Sorry about that.. So now, since here we exit if the subscription is closed, the capture down there where we do:
```
closed := sub.closed
sub.mu.Unlock()

if !active && !closed {
..
```
is silly... we can simply by removing `closed := sub.closed` and simply have `if !active {..`

kozlovic (changes requested):

kozlovic (approved):
LGTM. Thank you for your contribution!

PR #875, opened 2021-12-27
Prepare for KV1.0 release

ripienaar:
This closes the loop on supporting ADR-8:

 * We fix the bugs identified by @scottf where WatchOpts can confused Keys, History etc by now only accepting a context. 
 * We support arbitrary subject prefixes per bucket using BucketSubjectPrefix()

### DIFF
--- a/kv.go
+++ b/kv.go
@@ -23,9 +23,7 @@ import (
 	"time"
 )
 
-// Notice: Experimental Preview
-//
-// This functionality is EXPERIMENTAL and may be changed in later releases.
+// KeyValueManager creates, deletes or binds to Key-Value stores
 type KeyValueManager interface {
 	// KeyValue will lookup and bind to an existing KeyValue store.
 	KeyValue(bucket string) (KeyValue, error)
@@ -35,9 +33,7 @@ type KeyValueManager interface {
 	DeleteKeyValue(bucket string) error
 }
 
-// Notice: Experimental Preview
-//
-// This functionality is EXPERIMENTAL and may be changed in later releases.
+// KeyValue manages data in a Key-Value store
 type KeyValue interface {
 	// Get returns the latest value for the key.
 	Get(key string) (entry KeyValueEntry, err error)
@@ -54,18 +50,17 @@ type KeyValue interface {
 	// Purge will place a delete marker and remove all previous revisions.
 	Purge(key string) error
 	// Watch for any updates to keys that match the keys argument which could include wildcards.
-	// Watch will send a nil entry when it has received all initial values.
 	Watch(keys string, opts ...WatchOpt) (KeyWatcher, error)
 	// WatchAll will invoke the callback for all updates.
 	WatchAll(opts ...WatchOpt) (KeyWatcher, error)
 	// Keys will return all keys.
-	Keys(opts ...WatchOpt) ([]string, error)
+	Keys(opts ...CancelOpt) ([]string, error)
 	// History will return all historical values for the key.
-	History(key string, opts ...WatchOpt) ([]KeyValueEntry, error)
+	History(key string, opts ...CancelOpt) ([]KeyValueEntry, error)
 	// Bucket returns the current bucket name.
 	Bucket() string
 	// PurgeDeletes will remove all current delete markers.
-	PurgeDeletes(opts ...WatchOpt) error
+	PurgeDeletes(opts ...CancelOpt) error
 	// Status retrieves the status and configuration of a bucket
 	Status() (KeyValueStatus, error)
 }
@@ -91,9 +86,17 @@ type KeyValueStatus interface {
 // KeyWatcher is what is returned when doing a watch.
 type KeyWatcher interface {
 	// Updates returns a channel to read any updates to entries.
+	//
+	// A nil value is sent over the channel indicating that the end of initial data was reached,
+	// initial data may be the last available message or an indication that no messages are available
+	// at the time the watch was created.
 	Updates() <-chan KeyValueEntry
 	// Stop will stop this watcher.
 	Stop() error
+}
+
+type CancelOpt interface {
+	configureCancel(opts *cancelOpts) error
 }
 
 type WatchOpt interface {
@@ -104,6 +107,16 @@ type WatchOpt interface {
 func (ctx ContextOpt) configureWatcher(opts *watchOpts) error {
 	opts.ctx = ctx
 	return nil
+}
+
+// For nats.Context() support.
+func (ctx ContextOpt) configureCancel(opts *cancelOpts) error {
+	opts.ctx = ctx
+	return nil
+}
+
+type cancelOpts struct {
+	ctx context.Context
 }
 
 type watchOpts struct {
@@ -507,8 +520,22 @@ func (kv *kvs) delete(key string, purge bool) error {
 
 // PurgeDeletes will remove all current delete markers.
 // This is a maintenance option if there is a larger buildup of delete markers.
-func (kv *kvs) PurgeDeletes(opts ...WatchOpt) error {
-	watcher, err := kv.WatchAll(opts...)
+func (kv *kvs) PurgeDeletes(opts ...CancelOpt) error {
+	var o cancelOpts
+	for _, opt := range opts {
+		if opt != nil {
+			if err := opt.configureCancel(&o); err != nil {
+				return err
+			}
+		}
+	}
+
+	wopts := []WatchOpt{MetaOnly()}
+	if o.ctx != nil {
+		wopts = append(wopts, Context(o.ctx))
+	}
+
+	watcher, err := kv.WatchAll(wopts...)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


kv.go
```diff
@@ -542,18 +569,31 @@ func (kv *kvs) PurgeDeletes(opts ...WatchOpt) error {
...
 }
 
-// Keys() will return all keys.
-func (kv *kvs) Keys(opts ...WatchOpt) ([]string, error) {
-	opts = append(opts, IgnoreDeletes(), MetaOnly())
-	watcher, err := kv.WatchAll(opts...)
+// Keys will return all keys.
+func (kv *kvs) Keys(opts ...CancelOpt) ([]string, error) {
+	var o cancelOpts
+	for _, opt := range opts {
```
derekcollison, line 637: Maybe make this a small utility function.

kv.go
```diff
@@ -95,21 +95,42 @@ type KeyWatcher interface {
...
-type CancelOpt interface {
-	configureCancel(opts *cancelOpts) error
+// KeyValueOption configures the Key-Value store instance
+type KeyValueOption func(*kvs) error
+
+// BucketSubjectPrefix sets a custom prefix to use for write operations like Put() to the bucket.
+// Usually keys are accessed as $KV.BUCKET.KEY, this would replace the $KV.BUCKET. part of
+// that subject.
+//
+// When a APIPrefix() was used to create the JetStream context the KV Stream API and Subjects
```
derekcollison, line 105: Still a bit concerned about monkeying with data subjects outside of the server itself through imports.

derekcollison: I do believe there should be an easy way to match service import mappings into the client.

ripienaar: This is essentially a stop gap and to support people with weird stuff.  Matthias wants to add server-side helpers that would mean things would have fewer moving parts and more strict convention.

Right now though, KV simply can't be used cross account at all - so until above is finalized we need something. 

Can you expand on what your concerns are and what you mean with matching imports? We cannot overlap subjects and someone might want to import 2 x FOO buckets from different accounts.

derekcollison: To date we have not monkeyed with the data subjects, only control via APIPrefix etc. 

This is a departure from that, so I just want us to be very sure about this path.

test/kv_test.go
```diff
@@ -106,6 +106,51 @@ func TestKeyValueBasics(t *testing.T) {
...
+		v, err := kv.Get("FOO")
+		expectOk(t, err)
+		if string(v.Value()) != "BAR" {
+			t.Fatalf("received wrong value: %q", v.Value())
+		}
+	}
+
+	testAccount(t, "i", "i", "fromA", "")
+	testAccount(t, "o", "o", "fromA", "X.$KV.CROSS")
+	testAccount(t, "o", "o", "fromA", "X.$KV.CROSS.")
```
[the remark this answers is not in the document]
ripienaar: For the purpose of an arbitrary bucket prefix - thats exactly what it is, an arbitrary bucket prefix.  Its for accessing a single bucket.

As you can see from the example configuration, users can put in whatever they like we have no conventions, this supports that while we build out such conventions.

kv.go
ripienaar: Agree, personally do not really like the proposed structure in ADR-19, would rather just keep to arbitrary prefix for the moment while we work on that.  However it gets super tedious to require both ApiPrefix() and BucketSubjectPrefix() so figured I might default to the adr behavior.

I am do not like at all what the adr propose where it has `fromA.STREAM.INFO.FOO` and `fromA.$KV.B.K` the fact that it mounts these 2 subjects speaces to overlap each other is quite funky.

So I had it without the ApiPrefix() handling at first, but would also rather remove it while we work out ADR-19

ripienaar: Removed the ApiPrefix() handling, we should remove ADR-19 while we go back to the drawing board on that.

test/kv_test.go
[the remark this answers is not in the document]
ripienaar: Imagine you have 100 buckets, you export 1 of those buckes to me.  I have to import that 1 bucket, thats it.  I can import them on any arbitrary subject I choose - because we have no conventions.  So it has to import a single subject, I figured its better to support being explicit in that case.  It's the most supportable thing we can do while we figure out conventions for cross account access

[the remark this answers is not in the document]
scottf: maybe it's just confusing seeing a kv like subject as the example anything prefix. Is there another example config that would make the prefix something like `bucketsFromA` or 'kvsFromA`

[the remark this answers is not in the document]
ripienaar: well, its arbitrary you can put whatever you like there.  I updated the example a bit to make it clearer that there is no relation to names etc

test/configs/kv_accounts.conf
```diff
@@ -0,0 +1,23 @@
...
+
+jetstream: enabled
+
+accounts: {
+    A: {
+        users: [ {user: a, password: a} ]
+        jetstream: enabled
+        exports: [
+            {service: '$JS.API.>' }
+            {service: '$KV.CROSS.>'}
```
derekcollison, line 12: Does this support watchers also?

derekcollison (approved):
LGTM

test/configs/kv_accounts.conf
ripienaar: This specific example probably not - but a watcher doesnt need any fiddling with subjects other than normal API related fiddling which of course is just up to if JS work or not. Ie. watch is up to ApiPrefix() passed to the JetStream context, nothing to do with the actual KV code.

Right? I think so anyway.

derekcollison: Not sure, be good to flesh out a full example eventually, not needed here IMO. Thanks.

ripienaar: Yeah need some stuff to make ordered consumer work cross account, not sure whats up - but yeah not for this PR its unrelated.

test/kv_test.go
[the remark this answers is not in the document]
scottf: perfect. LGTM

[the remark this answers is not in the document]
scottf: That's clearer, thanks.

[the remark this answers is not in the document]
scottf: Does this replace the use of the inbox prefix of `forB.` which is what was used with this config? https://github.com/nats-io/nats.deno/pull/239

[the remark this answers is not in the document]
ripienaar: No that’s needed - and this is probably why I couldn’t get watch to work. 

But that’s not really about this PR, clearly this is all too complex I don’t think any normal user can get this going. So I am ok to leave as is for now 

later when we work on adr 19 we will see about macros to simplify this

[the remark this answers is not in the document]
scottf: ok yeah, I'm all for just putting something out there and revisiting it later when this is all nailed down.

kv.go
```diff
@@ -91,21 +86,76 @@ type KeyValueStatus interface {
...
 }
 
-type WatchOpt interface {
-	configureWatcher(opts *watchOpts) error
+// KeyValueOption configures the Key-Value store instance
+type KeyValueOption func(*kvs) error
+
+// BucketSubjectPrefix sets a custom prefix to use for write operations like Put() to the bucket.
+// Usually keys are accessed as $KV.BUCKET.KEY, this would replace the $KV.BUCKET. part of
+// that subject.
```
scottf, line 103: Just read this again. This comment says replace `$KV.<BUCKET>.` but @matthiashanel told me something different, that is to put the js prefix in front of the key, so for publishes, we publish to `<prefix>.$KV.<BUCKET>.<KEY>` and the server strips the prefix. So maybe we need to wait until he's back to settle on this.

scottf (dismissed):

kv.go
ripienaar: No, this is what we are dong. We are supporting ARBITRARY prefixes, not the conventions in ADR-19. Arbitrary means any random prefix users might choose.

See comments from Mathias here https://github.com/nats-io/nats-architecture-and-design/pull/82

ADR-19 should not have been merged tbh, its way too early and the design there is not going to work. So this is what we are doing so KV can work cross account.

scottf: Matthias and I had a slack conversation on the afternoon of the 23rd and conversation with Alberto on the 24th. I have unit test that works cross account against server v2.6.6 based on those conversations. Alberto has also coded the same as me, he is the one who provided the inbox prefix config. So, I removed my review request. I'm just going leave the java pr un-merged for now and work on something else.

ripienaar: Before those called Mathias and I agreed on arbitrary prefixes, if he agreed with you on non arbitrary prefixes I am afraid that's not something I can talk about. I know what was agreed as to finalise KV 1.0 before we all went on holiday and it was not `$KV.`

ripienaar: in the java code I believe `toKeySubjectConsiderPrefix` implements only the jetstream context prefix option, not an arbitrary prefix - we discussed that, you have no support for arbitrary prefixes.

PR #883, opened 2022-01-14
object store add link changes and some refactoring

scottf:
- remove the ability to link to a link
- prevent a link  from overwriting an object [by name]. an object is still allowed to replace an object and a link is still allowed to replace a link.
- prevent adding links to deleted objects
- misc error checks.


object.go
```diff
@@ -281,22 +281,60 @@ func (js *js) DeleteObjectStore(bucket string) error {
...
-func sanitizeName(name string) string {
-	stream := strings.ReplaceAll(name, ".", "_")
-	return strings.ReplaceAll(stream, " ", "_")
+func encodeName(name string) string {
+	return base64.URLEncoding.EncodeToString([]byte(name));
+}
+
+// toDefaultMetaSubject and toPublishMetaSubject will diverge once there is account support
+func toDefaultMetaSubject(obs *obs, name string) string {
+	return fmt.Sprintf(objMetaPreTmpl, obs.name, encodeName(name))
```
derekcollison, line 290: This looks same as func below. Also if 1 liners why not inline?

object.go
```diff
@@ -281,22 +281,60 @@ func (js *js) DeleteObjectStore(bucket string) error {
...
 
-func sanitizeName(name string) string {
-	stream := strings.ReplaceAll(name, ".", "_")
-	return strings.ReplaceAll(stream, " ", "_")
+func encodeName(name string) string {
+	return base64.URLEncoding.EncodeToString([]byte(name));
+}
+
+// toDefaultMetaSubject and toPublishMetaSubject will diverge once there is account support
+func toDefaultMetaSubject(obs *obs, name string) string {
```
derekcollison, line 289: Should this be a func on (*obs)?

object.go
```diff
@@ -578,77 +613,82 @@ func (obs *obs) Delete(name string) error {
 		return ErrBadObjectMeta
 	}
 
-	// Place a rollup delete marker.
+	// Place a rollup delete marker / publish the meta
```
derekcollison, line 616: / -> 'and'

object.go
```diff
@@ -578,77 +613,82 @@ func (obs *obs) Delete(name string) error {
...
 		return err
 	}
-	mm.Header.Set(MsgRollup, MsgRollupSubject)
-	_, err = obs.js.PublishMsg(mm)
+
+	// Purge chunks for the object
+	return purgeChunks(obs, info.NUID)
+}
+
+func doesntExistOrIsDeleted(obs *obs, name string) error {
```
derekcollison, line 627: Should this be a func on (*obs)?

object.go
```diff
@@ -578,77 +613,82 @@ func (obs *obs) Delete(name string) error {
...
+	info, err := obs.GetInfo(name)
 	if err != nil {
+		if err == ErrObjectNotFound {
```
derekcollison, line 630: Combine into one if I think.

object.go
```diff
@@ -578,77 +613,82 @@ func (obs *obs) Delete(name string) error {
...
-		if err := obs.UpdateMeta(obj.Name, &info.ObjectMeta); err != nil {
-			return nil, err
-		}
-		return obs.GetInfo(name)
-	}
+	// If the link name already exists in this store, you cannot add as a link
+	// If the link name exists as a deleted object, that's fine
+	// @derek Maybe we need an addOrUpdateLink or just an updateLink function,
+	// or add a flag/option to the method?
+	// either way, we should not update a non link to a link
```
derekcollison, line 652: What happens if you put something in under same name today? Meaning behavior of non-link put when something already there?

object.go
```diff
@@ -754,19 +793,28 @@ func (obs *obs) UpdateMeta(name string, meta *ObjectMeta) error {
...
+	info.Name = meta.Name
+	info.Description = meta.Description
+	info.Headers = meta.Headers
+
+	if err = publishMeta(obs, meta.Name, info); err != nil { return err }
+
+	// did the name of this object change? We just stored the meta under the new name
+	// so delete the meta from the old name
+	if err == nil && name != meta.Name {
+		metaSubj := fmt.Sprintf(objMetaPreTmpl, obs.name, name)
```
derekcollison, line 814: I think you have a helper for this no?

test/object_test.go
```diff
@@ -345,27 +337,45 @@ func TestObjectMetadata(t *testing.T) {
 	_, err = obs.PutString("A", "AAA")
 	expectOk(t, err)
 
-	meta := &nats.ObjectMeta{Name: "B"}
+	meta := &nats.ObjectMeta{Name: "A"}
```
derekcollison, line 340: Using "B" was on purpose.

object.go
scottf: right now it is the same as the func below. To handle cross account we need to handle subjects differently in certain spots. These are just placeholders for when that work gets done.

object.go
scottf: That's where I don't know go well enough

object.go
scottf: again, lack of go knowledge

object.go
scottf: I could use an example

object.go
scottf: the previous add link would do different things if the link was bucket to same bucket or bucket to different bucket.

bucket to same bucket: this code had errors. 

so I wasn't sure what original line 617 did. I thought that info and obj now point to the exact same object versus info being a deep copy.
In original line 618 you set that info.Name to the (sanitized) name. I assumed that this also set obj.Name

if info and obj is the same line 619 calls UpdateMeta with some sanitized object name that isn't lookup-able.
if info and obj diverge 619 would overwrite the object you are trying top link to.

bucket to diff bucket. This code also used the sanitized name when it should not have, but otherwise is the foundation for the re-write I did.

scottf: I actually wanted to change this api to 
```
func (obs *obs) AddLink(name string, targetBucket string, targetName string) (*ObjectInfo, error)
```

Or at least verify the obj that is being sent in, but that would require a getInfo possibly to another bucket understanding that the target bucket may be different.

But that was a lot of work so I landed where I did.

object.go
scottf: yeah this should be the toDefaultMetaSubject since the subject will be used as data in a JS.API call versus publishing

test/object_test.go
scottf: B is further down. I added meta data with update since the PutString doesn't take meta data directly.

object.go
derekcollison: If you always have a first arg that looks like self, that is a good test.

object.go
derekcollison:
```go
info, err := obs.GetInfo(name)
if err == ErrObjectNotFound || (info != nil && info.Deleted) {
  return nil
} else if err != nil {
  return err
}
return errors.New("nats: name already used for an existing object")
```

derekcollison: Not sure that's any better ;)

object.go
derekcollison: I meant non-link normal put when object already exists.

object.go
scottf: yeah, that's about what I had. It's six of one

object.go
scottf: So I just looked and realized the put is a problem with the same name. It always makes a new [nu]id. It will clean up (purge) the chunks from the original name's nuid, so if put was called from addlink, it does what you would think and completely replaces the object with a link.
I think the behavior or add link should be like create, also possibly allow overwrite of a name that is a link, but not allow overwrite of a name this is an object.

scottf: Much like KV, I wonder if we need put/add link with options where the user tells us create/update/allow overwrite, etc. and then what the default behavior is

scottf: are you thinking that if they want to put something over something else, that's their prerogative?

object.go
scottf: removed / re-inlined with comments

object.go
scottf: So I'm just simplifying and not doing any checking, just making sure that put does the right think and the add link code is correct. We can always add more granularity later

object.go
```diff
@@ -430,15 +448,15 @@ func (obs *obs) Put(meta *ObjectMeta, r io.Reader, opts ...ObjectOpt) (*ObjectIn
...
 	if einfo != nil && !einfo.Deleted {
-		chunkSubj := fmt.Sprintf(objChunksPreTmpl, obs.name, einfo.NUID)
-		obs.js.purgeStream(obs.stream, &streamPurgeRequest{Subject: chunkSubj})
+		purgeChunks(obs, einfo.NUID)
 	}
 
-	return info, nil
+	// Yes this is an extra call, but GetInfo sets the modified time correctly
+	// to the server message time, instead of just setting the time based on the client
+	return obs.GetInfo(meta.Name)
```
scottf, line 459: Derek - the original code just set the modified time on the info after it was publish, which isn't really the actual message timestamp. If we don't mind the extra server round trip here, then leave it, otherwise I would prefer leaving the modified time nil instead of giving it time from the client.

object.go
```diff
@@ -578,55 +596,34 @@ func (obs *obs) Delete(name string) error {
...
-		info := *obj
-		info.Name = name
-		if err := obs.UpdateMeta(obj.Name, &info.ObjectMeta); err != nil {
-			return nil, err
-		}
-		return obs.GetInfo(name)
-	}
-
-	link := &ObjectLink{Bucket: obj.Bucket, Name: obj.Name}
+	// create the meta object
```
scottf, line 619: I don't know why the original code was trying to do something with same vs different bucket, but I'm pretty sure there is no need to.  I'm not even sure of the intention of the original 616 - 623 block the way it's coded, what was the intent here?

object.go
```diff
@@ -759,14 +756,24 @@ func (obs *obs) UpdateMeta(name string, meta *ObjectMeta) error {
 	if err != nil {
 		return err
 	}
-	// Copy new meta
-	info.ObjectMeta = *meta
-	mm := NewMsg(fmt.Sprintf(objMetaPreTmpl, obs.name, sanitizeName(meta.Name)))
-	mm.Data, err = json.Marshal(info)
-	if err != nil {
+
```
scottf, line 759: Derek - is it an error to update meta for a deleted object?

object.go
scottf: Also, if we don't care if they put an object or link over an existing object, do we care if they link to a deleted object?

object.go
derekcollison: I think so.

object.go
derekcollison: Will need to spend some time here but right now have other priorities.

object.go
derekcollison: I was trying to use modified time as a signal, but would need to dig in further.

scottf:
@kozlovic This fell off my radar b/c of other priorities. There were some questions that we had not gotten around to being resolved and there was never an approval and that's why it never got merged.

jnmoyne (approved):
LGTM (see my added commit to remove skip on windows from one of the tests)

scottf:
Open Questions
1. Should we allow links to links?
2. Should we return ErrObjectNotFound for Get/GetInfo when an object is deleted? This would match what KV does on get where the entry is deleted. This would also affect links to things that were linked then deleted.

object.go
```diff
@@ -434,18 +432,21 @@ func (obs *obs) Put(meta *ObjectMeta, r io.Reader, opts ...ObjectOpt) (*ObjectIn
...
+			}
 			return nil, err
 		}
 	case <-time.After(obs.js.opts.wait):
 		return nil, ErrTimeout
 	}
-	info.ModTime = time.Now().UTC()
 
-	// Delete any original one.
+	info.ModTime = time.Now().UTC() // TODO get the time from the PublishAck
```
scottf, line 444: can someone help me with this? The PublishAck has the message time, I just don't know how to get it.

test/object_test.go
```diff
@@ -486,27 +520,86 @@ func TestObjectLinks(t *testing.T) {
...
+
+	// Check Error Link to a Link
+	_, err = root.AddBucketLink("", root)
+	expectErr(t, err)
+
+	_, err = root.AddBucketLink("Nil Bucket", nil)
+	expectErr(t, err)
+}
+
+func expectLinkIsCorrect(t *testing.T, originalObject *nats.ObjectInfo, linkObject *nats.ObjectInfo) {
```
piotrpio, line 583: Usually in go you check for error conditions first and if no error, continue execution. Also, we can easily avoid nesting if blocks here. So how about:
```
func expectLinkIsCorrect(t *testing.T, originalObject *nats.ObjectInfo, linkObject *nats.ObjectInfo) {
	if linkObject.Opts.Link == nil || !expectLinkPartsAreCorrect(t, linkObject, originalObject.Bucket, originalObject.Name) {
		t.Fatalf("Link info not what was expected:\nActual: %+v\nTarget: %+v", linkObject, originalObject)
	}
}
```

test/object_test.go
```diff
@@ -471,12 +497,20 @@ func TestObjectLinks(t *testing.T) {
...
-	_, err = root.AddLink("LA", info)
+	// Link to individual object.
+	infoLA, err := root.AddLink("LA", infoA)
 	expectOk(t, err)
+	expectLinkIsCorrect(t, infoA, infoLA)
+
+	// link to a link
+	_, err = root.AddLink("LALA", infoLA)
+	expectErr(t, err)
+	//infoLALA, err := root.AddLink("LALA", infoLA)
```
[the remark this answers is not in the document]
scottf: These are testing a link to a link. Originally it was allowed to do this, but I have asked for a design review to disallow link to link. When that is decided one way or the other I will fix the tests.

test/object_test.go
```diff
@@ -353,27 +352,51 @@ func TestObjectMetadata(t *testing.T) {
...
+	meta.Headers.Set("color", "blue")
 
+	// simple update that does not change the name, just adds data
 	err = obs.UpdateMeta("A", meta)
 	expectOk(t, err)
 
-	info, err := obs.GetInfo("B")
-	expectOk(t, err)
-	if info.Name != "B" {
+	info, _ := obs.GetInfo("A")
```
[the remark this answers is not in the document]
scottf: fixing

test/object_test.go
```diff
@@ -486,27 +520,86 @@ func TestObjectLinks(t *testing.T) {
...
 	expectOk(t, err)
 
-	if dbl != "DIR-BBB" {
-		t.Fatalf("Expected %q but got %q", "DIR-BBB", dbl)
+	if getLA != "AAA" {
+		t.Fatalf("Expected %q but got %q", "AAA", getLA)
+	}
+
+	// Test the link to a link get
+	//getLALA, err := root.GetString("LALA")
```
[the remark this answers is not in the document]
scottf: same as above, re link to a link

test/object_test.go
```diff
@@ -486,27 +520,86 @@ func TestObjectLinks(t *testing.T) {
...
+func expectLinkIsCorrect(t *testing.T, originalObject *nats.ObjectInfo, linkObject *nats.ObjectInfo) {
+	if linkObject.Opts.Link != nil {
+		if expectLinkPartsAreCorrect(t, linkObject, originalObject.Bucket, originalObject.Name) {
+			return
+		}
 	}
+	t.Fatalf("Link info not what was expected:\nActual: %+v\nTarget: %+v", linkObject, originalObject)
+}
+
+func expectLinkPartsAreCorrect(t *testing.T, linkObject *nats.ObjectInfo, bucket, name string) bool {
```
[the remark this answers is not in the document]
scottf: This is not my code. It's just in a different place. But I can change it

PR #886, opened 2022-01-20
Extended comment for Subscribe method

ipromax:
Currently, the comment for `Subscribe` method contains information about the wrong separator for subjects.

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -3738,8 +3738,13 @@ func (nc *Conn) respToken(respInbox string) string {
 }
 
 // Subscribe will express interest in the given subject. The subject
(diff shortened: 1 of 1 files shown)


kozlovic (changes requested):
I think the meaning was to say: "partial: this is the `*` wildcard" and "full: this is the '>' wildcard".

kozlovic:
@ipromax I was not suggesting that you close the PR, we could make it more obvious of what the meaning was. If you got it wrong, some other users may get it wrong too.

kozlovic (approved):
LGTM

PR #894, opened 2022-02-01
[ADDED] JetStream: Nak delay and BackOff lists

kozlovic:

### DIFF
--- a/test/js_test.go
+++ b/test/js_test.go
@@ -2752,7 +2752,7 @@ func TestJetStreamSubscribe_AckPolicy(t *testing.T) {
 		if err != nil {
 			t.Fatalf("Error on NextMsg: %v", err)
 		}
-		if err := msg.Nak(nats.NakDelay(500 * time.Millisecond)); err != nil {
+		if err := msg.NakWithDelay(500 * time.Millisecond); err != nil {
 			t.Fatalf("Error on Nak: %v", err)
 		}
 		// We should not get redelivery before 500ms+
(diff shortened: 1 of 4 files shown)


test/js_test.go
```diff
@@ -2741,6 +2741,100 @@ func TestJetStreamSubscribe_AckPolicy(t *testing.T) {
...
+		js.Publish("bar", []byte("msg"))
+		sub, err := js.SubscribeSync("bar", nats.Durable("nak_dur"))
+		if err != nil {
+			t.Fatalf("Error on subscribe: %v", err)
+		}
+		msg, err := sub.NextMsg(time.Second)
+		if err != nil {
+			t.Fatalf("Error on NextMsg: %v", err)
+		}
+		if err := msg.Nak(nats.NakDelay(500 * time.Millisecond)); err != nil {
```
derekcollison, line 2755: Hmm.. Looking at this suggest maybe NakWithDelay(d time.Duration) might be better? WDYT?

kozlovic: Sure. But if we were to add more options to Nak() (which is unlikely), then it may be weird too. Will change for now.

kozlovic:
@derekcollison Updated in this commit: https://github.com/nats-io/nats.go/pull/894/commits/f1d50b84afd1c6cb3c7465f51343a66b6b0d407d

derekcollison (commented):
In general LGTM, wondering if we can place nak delay into opts to avoid passing around explicit arg that is usually 0.

kozlovic:
@derekcollison I think I was able to do as you asked: https://github.com/nats-io/nats.go/pull/894/commits/d34c12d8a5817d32d5c6ad043fe302bf8ddc1730

derekcollison (approved):
LGTM

PR #895, opened 2022-02-02
Fixed tests to run on Windows + some cleanup

kozlovic:
- JS Servers need a special sequence on shutdown. Most of the time
the store was attempted to be removed before the server was shutdown,
which on Windows would result in an error and the next test may
recover the previous state.
- Made use of the jsClient() helper to remove typical connect+jetstream
code.
- Added some nats.Timeout() during connect so that Windows connect
to a server that is not running is faster than the default 2sec, which
may cause tests to run very long or simply fail.

### DIFF
--- a/js_test.go
+++ b/js_test.go
@@ -81,6 +81,7 @@ func createConfFile(t *testing.T, content []byte) string {
 }
 
 func shutdownJSServerAndRemoveStorage(t *testing.T, s *server.Server) {
+	t.Helper()
 	var sd string
 	if config := s.JetStreamConfig(); config != nil {
 		sd = config.StoreDir
--- a/js_test.go
+++ b/js_test.go
@@ -1,4 +1,4 @@
-// Copyright 2012-2021 The NATS Authors
+// Copyright 2012-2022 The NATS Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
@@ -35,6 +35,25 @@ import (
 	natsserver "github.com/nats-io/nats-server/v2/test"
 )
 
+func client(t *testing.T, s *server.Server, opts ...Option) *Conn {
+	t.Helper()
+	nc, err := Connect(s.ClientURL(), opts...)
+	if err != nil {
+		t.Fatalf("Unexpected error: %v", err)
+	}
+	return nc
+}
+
+func jsClient(t *testing.T, s *server.Server, opts ...Option) (*Conn, JetStreamContext) {
+	t.Helper()
+	nc := client(t, s, opts...)
+	js, err := nc.JetStream(MaxWait(10 * time.Second))
+	if err != nil {
+		t.Fatalf("Unexpected error getting JetStream context: %v", err)
+	}
+	return nc, js
+}
+
 func RunBasicJetStreamServer() *server.Server {
 	opts := natsserver.DefaultTestOptions
 	opts.Port = -1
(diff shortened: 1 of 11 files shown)


nats_test.go
```diff
@@ -137,7 +137,10 @@ func TestExpandPath(t *testing.T) {
 			{path: "/Foo/Bar", userProfile: `C:\Foo\Bar`, wantPath: "/Foo/Bar"},
 			{path: "Foo/Bar", userProfile: `C:\Foo\Bar`, wantPath: "Foo/Bar"},
 			{path: "~/Fizz", userProfile: `C:\Foo\Bar`, wantPath: `C:\Foo\Bar\Fizz`},
-			{path: `${HOMEDRIVE}${HOMEPATH}\Fizz`, userProfile: `C:\Foo\Bar`, wantPath: `C:\Foo\Bar\Fizz`},
+			// That one would fail because expandPath(), if not finding `~` returns
+			// the given path, which since ${HOMEDRIVE}${HOMEPATH} is not set,
+			// would return `\Fizz` but test expects `C:\Foo\Bar\Fizz`?
+			// {path: `${HOMEDRIVE}${HOMEPATH}\Fizz`, userProfile: `C:\Foo\Bar`, wantPath: `C:\Foo\Bar\Fizz`},
```
kozlovic, line 143: @variadico I had to comment out this line specific test:
```
{path: `${HOMEDRIVE}${HOMEPATH}\Fizz`, userProfile: `C:\Foo\Bar`, wantPath: `C:\Foo\Bar\Fizz`},
```
My reading of the code, since there is no `~`, expandPath() would simply returned what is given, in this case `\Fizz` since the 2 env variables are not set, and so the test would fail.
Could you have a look at what was the intent or if this need tweaking?

test/object_test.go
```diff
@@ -141,8 +142,15 @@ func TestDefaultObjectStatus(t *testing.T) {
 }
 
 func TestObjectFileBasics(t *testing.T) {
+	// The filename is likely to be something like C:\xxx and we don't allow
```
kozlovic, line 145: @derekcollison @scottf Want to bring this to your attention regarding ObjectStore (low priority, I know, but understand that this is not likely to work on Windows as long as user uses files like `C:\foo\bar`).

derekcollison (approved):
LGTM

test/object_test.go
scottf: I think this is fine now. The key encoding has changed from simple replacement to base64 url encoding of the filename for use as in a subject (key). You can see the change here: https://github.com/nats-io/nats.go/pull/883/files#diff-12229ea844f42fd705168380fb7870bcd7d66f0c94b53ef0efe8b9ca50f05ec7R284

After this PR is merged and incorporated into my branch I will update the test.

nats_test.go
variadico: I think perhaps I meant
```
wantPath: `\Fizz`
```
But, it should probably be ok to remove this case.

If `${HOMEDRIVE}${HOMEPATH}` are set, then the correct path will be returned. Otherwise, the user will get an error and they'll know to give us a different path. Probably a path with `${USERPROFILE}` so we can expand that.

variadico (approved):

nats_test.go
kozlovic: > If ${HOMEDRIVE}${HOMEPATH} are set, then the correct path will be returned.

But that's the thing, in this test's config, they are not. So should I remove completely or do you want to keep it with `wantPath: "\Fizz"`?

js_test.go
```diff
@@ -61,26 +80,29 @@ func createConfFile(t *testing.T, content []byte) string {
 	return fName
 }
 
+func shutdownJSServerAndRemoveStorage(t *testing.T, s *server.Server) {
+	var sd string
```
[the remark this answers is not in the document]
kozlovic: Thanks, will add (also to the other package, one per package unfortunately..)

nats_test.go
kozlovic: Ok, as agreed on Slack, just removing this line.

PR #899, opened 2022-02-06
Spelling fixes, mostly comments

philpennock:
The new `gospel` tool, <https://github.com/kortschak/gospel>, uses hunspell libraries but pre-registers as acceptable words every symbol from the Go source, massively reducing the noise and making comment spell-checking a tractable problem.  It recently gained support for a `.words` file located in the same directory as the `go.mod` file, to define local words.  With this, we can fix real issues too.

This PR reduces the complaints down to 4:
 1. A reference to `syncSubscribers` which I can't figure out
 2. A reference to a `pubArg`
 3. Two references to `splitArgs`.

I made two actual code changes:
 1. Fixing a **non-exported** const type for consistency with all the others: `apiStreamList` -> `apiStreamListT`.
 2. Changing an error message to refer to a field which exist

Lots of typo fixes; references to since-renamed fields; etc.

A reference to `PublishAsynMsg` might have been intended to be `PublishAsyncMsg` but that doesn't exist either, so I removed it.  Similarly, `SubjectIsDelivery` lived briefly but a stale reference was left in a comment, so I removed that.

To reproduce:

    go install github.com/kortschak/gospel@latest
    gospel .

### DIFF
--- a/.words
+++ b/.words
@@ -1,0 +1,144 @@
+1
+## ^^^ size estimate, just needs to be non-zero
+##
+## The .words file is used by gospel (v1.2+), which wraps the Hunspell libraries but populates the dictionary with identifiers from the Go source.
+## <https://github.com/kortschak/gospel>
+##
+## Comment lines are not actually parsed as comments, try to be careful ... but in practice they seem to work?
+##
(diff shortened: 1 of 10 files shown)


derekcollison (approved):
LGTM - Nice..

Should we also update travis to run this test as well?

philpennock:
I think we should fix the four I didn't manage to clean up because I didn't know the semantically right fix, and _then_ update Travis. :)

PR #900, opened 2022-02-06
Updates to KV Watcher.

derekcollison:
1. If the underlying stream was very busy pending might not go to zero, so snapshot and mark initial state done when received >= init pending or delta == 0
2. When a watcher was cancelled with a context we would not signal to allow and range w.Updates() calls to exit.
3. General flow control checks were timing sensitive after the initial checks, needed to be 2x HB interval always, not just beginning.

### DIFF
--- a/kv.go
+++ b/kv.go
@@ -1,4 +1,4 @@
-// Copyright 2021 The NATS Authors
+// Copyright 2021-2022 The NATS Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
@@ -20,6 +20,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 )
 
@@ -306,6 +307,11 @@ func (js *js) CreateKeyValue(cfg *KeyValueConfig) (KeyValue, error) {
 		DenyDelete:        true,
 	}
 
+	// If we are at server version 2.7.2 or above use DiscardNew. We can not use DiscardNew for 2.7.1 or below.
+	if js.nc.serverMinVersion(2, 7, 2) {
+		scfg.Discard = DiscardNew
+	}
+
 	if _, err := js.AddStream(scfg); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


kv.go
```diff
@@ -597,14 +604,30 @@ func (w *watcher) Updates() <-chan KeyValueEntry {
...
 
+// close the update chan.
+func (w *watcher) close() {
+	if w == nil {
+		return
+	}
+	w.mu.Lock()
+	if w.updates != nil {
+		close(w.updates)
+		w.updates = nil
```
kozlovic, line 620: As discussed in a call, the issue with that is that if user later calls w.Updates() then they would possibly range on a nil channel which would block.
That being said, maybe users could check that the channel is not nil after calling `Updates()`, but we discussed that you would want the channel to be closed, but not set to nil, so need to revisit that part.

derekcollison: Yep will fix after a bit.

derekcollison: ok changed, if you could take a look that would be great.

kv.go
```diff
@@ -590,6 +596,7 @@ func (kv *kvs) History(key string, opts ...WatchOpt) ([]KeyValueEntry, error) {
 type watcher struct {
 	updates chan KeyValueEntry
 	sub     *Subscription
+	closed  uint32
```
kozlovic, line 599: When using with atomic ops, it is preferable to move those at the top of the struct. If people don't know and later add, say a boolean, between sub and closed, it will panic on a 32bit release.

See https://github.com/golang/go/issues/599

nats.go
```diff
@@ -3944,6 +3946,13 @@ func (nc *Conn) removeSub(s *Subscription) {
 	if s.pCond != nil {
 		s.pCond.Broadcast()
 	}
+
+	// Check for watchers.
+	if jsi != nil && jsi.w != nil {
```
kozlovic, line 3951: This is not enough to guarantee that waitForMsgs() is not invoking the message callback at this time, or about to do it.

kozlovic (changes requested):

kv.go
derekcollison: Will do.

nats.go
derekcollison: Yea I am reworking this as we speak since the large KINE test failed on send to closed chan.

derekcollison: ok I reworked it, please take another look.

kozlovic (approved):
LGTM

PR #902, opened 2022-02-09
Different approach to address initial pending == 0

kozlovic:
This is undoing changes done in PR https://github.com/nats-io/nats.go/pull/901
and makes the changes dicussed in the comments of that PR.

### DIFF
--- a/js.go
+++ b/js.go
@@ -974,6 +974,14 @@ type jsSub struct {
 	pull     bool
 	dc       bool // Delete JS consumer
 
+	// This is ConsumerInfo's Pending+Consumer.Delivered that we get from the
+	// add consumer response. Note that some versions of the server gather the
+	// consumer info *after* the creation of the consumer, which means that
+	// some messages may have been already delivered. So the sum of the two
+	// is a more accurate representation of the number of messages pending or
+	// in the process of being delivered to the subscription when created.
+	pending uint64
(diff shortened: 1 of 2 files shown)


kozlovic:
@ripienaar @derekcollison This is another approach, which I sort of prefer. The addconsumer then js.Subscribe() with special option looked a bit of a hack and was not too comfortable with that.

@derekcollison Before you complain about use of lock in the watcher :-)
- The history limit is currently hard-coded at 64, so that update callback is not going to fire a lot
- The contention will be gone as soon as Watch() returns, after that, getting the lock in update() callback - since there is no contention - would be less of an issue.

derekcollison (approved):
LGTM

derekcollison (approved):
LGTM

js.go
```diff
@@ -1573,6 +1566,7 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync,
 			// Since the library created the JS consumer, it will delete it on Unsubscribe()/Drain()
 			sub.mu.Lock()
 			sub.jsi.dc = true
+			sub.jsi.pending = info.NumPending + info.Delivered.Consumer
```
ripienaar, line 1569: good, and this PR is much nicer thanks.

PR #903, opened 2022-02-10
Add GetRevision to nats.KeyValue

boxboatmatt:
Signed-off-by: Matthew DeVenny <matt@boxboat.com>

Add `GetRevision` to allow retrieval of historical key value versions by revision number

### DIFF
--- a/kv.go
+++ b/kv.go
@@ -373,10 +373,10 @@ func keyValid(key string) bool {
 // Get returns the latest value for the key.
 func (kv *kvs) Get(key string) (KeyValueEntry, error) {
 	e, err := kv.get(key, kvLatestRevision)
-	if err == ErrKeyDeleted {
-		return nil, ErrKeyNotFound
-	}
 	if err != nil {
+		if err == ErrKeyDeleted {
+			return nil, ErrKeyNotFound
+		}
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


kv.go
```diff
@@ -389,14 +405,25 @@ func (kv *kvs) get(key string) (KeyValueEntry, error) {
...
+	}
+
 	if err != nil {
 		if err == ErrMsgNotFound {
 			err = ErrKeyNotFound
 		}
 		return nil, err
 	}
 
+	if m.Subject != b.String() {
```
derekcollison, line 423: This check is only needed if you do line 413.

derekcollison (commented):
In general LGTM..

kv.go
boxboatmatt: Yeah... would the preference be to duplicate the error checking on 416 under the `if/else` and only check subject there? Also thinking would you want a different error for the case where someone provides a sequence that doesn't match subject?

kozlovic (changes requested):

kv.go
boxboatmatt: Comments addressed

kozlovic (changes requested):

kv.go
```diff
@@ -380,7 +383,20 @@ func (kv *kvs) Get(key string) (KeyValueEntry, error) {
...
-func (kv *kvs) get(key string) (KeyValueEntry, error) {
+// GetRevision returns a specific revision value for the key.
+func (kv *kvs) GetRevision(key string, revision uint64) (KeyValueEntry, error) {
+	e, err := kv.get(key, revision)
+	if err == ErrKeyDeleted {
+		return nil, ErrKeyNotFound
+	}
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
kozlovic: (sorry I missed that in first pass)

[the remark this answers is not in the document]
boxboatmatt: Ahh... that is just a copy paste of the current `Get` method but that would be better.. I can update both methods

kozlovic (approved):
LGTM

PR #905, opened 2022-02-10
Expose ErrNoMessages

mprimi:
Change: make private errNoMessagei (in js.go) into public ErrNoMessages (in nats.go)
Motivation: the JS pull consumer interface Fetch() can throw this error. But there's no way to handle it client-side if the error is not accessible

### DIFF
--- a/context.go
+++ b/context.go
@@ -150,7 +150,7 @@ func (s *Subscription) nextMsgWithContext(ctx context.Context, pullSubInternal, 
 		// If internal and we don't want to wait, signal that there is no
 		// message in the internal queue.
 		if pullSubInternal && !waitIfNoMsg {
-			return nil, errNoMessages
+			return nil, ErrNoMessages
 		}
 	}
 
--- a/js.go
+++ b/js.go
@@ -2335,8 +2335,6 @@ func PullMaxWaiting(n int) SubOpt {
 	})
 }
 
-var errNoMessages = errors.New("nats: no messages")
-
 // Returns if the given message is a user message or not, and if
 // `checkSts` is true, returns appropriate error based on the
 // content of the status (404, etc..)
@@ -2367,7 +2365,7 @@ func checkMsg(msg *Msg, checkSts bool) (usrMsg bool, err error) {
 		err = ErrNoResponders
 	case noMessagesSts:
 		// 404 indicates that there are no messages.
-		err = errNoMessages
+		err = ErrNoMessages
 	case reqTimeoutSts:
 		// Older servers may send a 408 when a request in the server was expired
 		// and interest is still found, which will be the case for our
@@ -2479,7 +2477,7 @@ func (sub *Subscription) Fetch(batch int, opts ...PullOpt) ([]*Msg, error) {
 		// are no messages.
 		msg, err = sub.nextMsgWithContext(ctx, true, false)
 		if err != nil {
-			if err == errNoMessages {
+			if err == ErrNoMessages {
 				err = nil
 			}
 			break
@@ -2534,7 +2532,7 @@ func (sub *Subscription) Fetch(batch int, opts ...PullOpt) ([]*Msg, error) {
 				usrMsg, err = checkMsg(msg, true)
 				if err == nil && usrMsg {
 					msgs = append(msgs, msg)
-				} else if noWait && (err == errNoMessages) && len(msgs) == 0 {
+				} else if noWait && (err == ErrNoMessages) && len(msgs) == 0 {
 					// If we have a 404 for our "no_wait" request and have
 					// not collected any message, then resend request to
 					// wait this time.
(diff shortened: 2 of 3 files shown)


mprimi:
Build error does not look related to my change, and I lack permissions to re-try it.

context.go
```diff
@@ -150,7 +150,7 @@ func (s *Subscription) nextMsgWithContext(ctx context.Context, pullSubInternal,
 		// If internal and we don't want to wait, signal that there is no
 		// message in the internal queue.
 		if pullSubInternal && !waitIfNoMsg {
-			return nil, errNoMessages
+			return nil, ErrNoMessages
```
kozlovic, line 153: No need because the `pullSubInternal` indicates that this is an internal call and that error will not be surfaced to the user.

js.go
```diff
@@ -2534,7 +2532,7 @@ func (sub *Subscription) Fetch(batch int, opts ...PullOpt) ([]*Msg, error) {
 				usrMsg, err = checkMsg(msg, true)
 				if err == nil && usrMsg {
 					msgs = append(msgs, msg)
-				} else if noWait && (err == errNoMessages) && len(msgs) == 0 {
+				} else if noWait && (err == ErrNoMessages) && len(msgs) == 0 {
```
kozlovic, line 2535: Again here we capture. Have you seen a case where we return errNoMessages? I think that it is then a loophole and we should not have this condition. Instead we should have either Timeout or if there is any message returned, the error is cleared (see end of function).

wallyqs:
No Msgs status should not have to be handled by the user, so instead need to look for the codepath where a 404 error is being returned unintentionally.

mprimi:
> Are you doing Fetch(1)?

I wasn't.

While minifying the program to share it with you, the error went away.
Give me a moment to figure out what is going on...

mprimi:
@kozlovic Found the problem, it was my fault. 😓

Multiple goroutines were fetching from the same subscription, without synchronization.
So they were probably creating some invalid state.
With that bug solved, I no longer see this error message.

Apologies for the spurious interrupt!

mprimi:
After a little more thinking, I managed to reproduce that error message with a pretty simple program, attaching it.

(Had to append `.TXT` or GH wouldn't let me attach!
[js-pull-consumer-errnomessages.go.TXT](https://github.com/nats-io/nats.go/files/8044894/js-pull-consumer-errnomessages.go.TXT)

I can reproduce the issue reliably on both `darwin-arm64` and `linux-amd64`, with the following:

```
# 1. Create stream:
nats -s $NATS stream add --subjects=test.benchmark --replicas=3 --max-msgs=-1 --max-msgs-per-subject=-1 --max-bytes=-1 --max-age=-1 --storage=file --retention=limits --discard=old --max-msg-size=-1 --dupe-window=2m --no-allow-rollup --no-deny-delete --no-deny-purge benchmark-stream

# 2. Create consumers (this program):
js-pull-consumer-errnomessages -numClients 24 -fetchTimeout 10 -servers $NATS


# 3. Publish some messages
nats bench --server=$NATS --pub=1 --msgs=10000 test.benchmark
```

Notice that it's sensitive to some combination of `numClients` AND `fetchTimeout`.
If you don't see the error, try increasing `numClients` or decreasing `fetchTimeout`.

Using 24 and 10 reproduces almost immediately locally for me:

```
2022/02/10 15:14:06 [0x1400006e7c0] Fetch error: nats: no messages
2022/02/10 15:14:06 [0x1400006e6a0] Fetch error: nats: no messages
2022/02/10 15:14:06 [0x1400006e7c0] Fetch error: nats: no messages
2022/02/10 15:14:06 [0x1400006e7c0] Fetch error: nats: no messages
2022/02/10 15:14:06 [0x1400006e7c0] Fetch error: nats: no messages
2022/02/10 15:14:06 [0x1400006e7c0] Fetch error: nats: no messages
2022/02/10 15:14:06 [0x1400006e6a0] Fetch error: nats: no messages
2022/02/10 15:14:06 [0x1400006e6a0] Fetch error: nats: no messages
2022/02/10 15:14:06 [0x1400006e880] Fetch error: nats: no messages
```

kozlovic:
Actually, re-opening the PR is not the right thing to do. We should create an issue regarding the fact that we get this error message out of a fetch. Will work on that one next.

kozlovic:
@mprimi I can't reproduce it, but then I see that: `github.com/nats-io/nats.go v1.13.1-0.20211018182449-f2416a8b1483`. That seem like very old. Can you pull main branch and try again?

kozlovic:
@mprimi Just tried with that revision and your test program and setup and still can't reproduce it... will look deeper tomorrow.

kozlovic:
Ok, I had to go back to server v2.7.0 to see the issue!

PR #906, opened 2022-02-11
[CHANGED] PurgeDeletes() will now keep markers that are less than 30min old

kozlovic:
There is a breaking change where PurgeDeletes() accepts now a list
of PurgeOpt, not WatchOpt.

We needed from WatchOpt only the context, and as it standed, it was
bad since user could have passed the IncludeHistory() option to
PurgeDeletes(), which would likely "break" the functionality.

Also, when invoking PurgeDeletes(), the delete markers than are older than
a default of 30 minutes will be deleted, however more recent ones will
be kept. The data is always removed, even if a marker is not.

The user can change the 30 minutes threshold using a new purge
option called DeleteMarkersOlderThan(duration). If set to -1,
it restores the old behavior of deleting delete/purge markers,
regardless of their age.

### DIFF
--- a/kv.go
+++ b/kv.go
@@ -68,7 +68,7 @@ type KeyValue interface {
 	// Bucket returns the current bucket name.
 	Bucket() string
 	// PurgeDeletes will remove all current delete markers.
-	PurgeDeletes(opts ...WatchOpt) error
+	PurgeDeletes(opts ...PurgeOpt) error
 	// Status retrieves the status and configuration of a bucket
 	Status() (KeyValueStatus, error)
 }
@@ -147,6 +147,34 @@ func MetaOnly() WatchOpt {
 		opts.metaOnly = true
 		return nil
 	})
+}
+
+type PurgeOpt interface {
+	configurePurge(opts *purgeOpts) error
+}
+
+type purgeOpts struct {
+	dmthr time.Duration // Delete markers threshold
+	ctx   context.Context
+}
+
+// DeleteMarkersOlderThan indicates that delete or purge markers older than that
+// will be deleted as part of PurgeDeletes() operation, otherwise, only the data
+// will be removed but markers that are recent will be kept.
+// Note that if no option is specified, the default is 30 minutes. You can set
+// this option to a negative value to instruct to always remove the markers,
+// regardless of their age.
+type DeleteMarkersOlderThan time.Duration
+
+func (ttl DeleteMarkersOlderThan) configurePurge(opts *purgeOpts) error {
+	opts.dmthr = time.Duration(ttl)
+	return nil
+}
+
+// For nats.Context() support.
+func (ctx ContextOpt) configurePurge(opts *purgeOpts) error {
+	opts.ctx = ctx
+	return nil
 }
 
 // KeyValueConfig is for configuring a KeyValue store.
@@ -536,14 +564,42 @@ func (kv *kvs) delete(key string, purge bool) error {
 	return err
 }
 
+const kvDefaultPurgeDeletesMarkerThreshold = 30 * time.Minute
+
 // PurgeDeletes will remove all current delete markers.
 // This is a maintenance option if there is a larger buildup of delete markers.
-func (kv *kvs) PurgeDeletes(opts ...WatchOpt) error {
-	watcher, err := kv.WatchAll(opts...)
+// See DeleteMarkersOlderThan() option for more information.
+func (kv *kvs) PurgeDeletes(opts ...PurgeOpt) error {
+	var o purgeOpts
+	for _, opt := range opts {
+		if opt != nil {
+			if err := opt.configurePurge(&o); err != nil {
+				return err
+			}
+		}
+	}
+	// Transfer possible context purge option to the watcher. This is the
+	// only option that matters for the PurgeDeletes() feature.
+	var wopts []WatchOpt
+	if o.ctx != nil {
+		wopts = append(wopts, Context(o.ctx))
+	}
+	watcher, err := kv.WatchAll(wopts...)
 	if err != nil {
 		return err
 	}
 	defer watcher.Stop()
+
+	var limit time.Time
+	olderThan := o.dmthr
+	// Negative value is used to instruct to always remove markers, regardless
+	// of age. If set to 0 (or not set), use our default value.
+	if olderThan == 0 {
+		olderThan = kvDefaultPurgeDeletesMarkerThreshold
+	}
+	if olderThan > 0 {
+		limit = time.Now().Add(-olderThan)
+	}
 
 	var deleteMarkers []KeyValueEntry
 	for entry := range watcher.Updates() {
@@ -564,8 +620,11 @@ func (kv *kvs) PurgeDeletes(opts ...WatchOpt) error {
 		b.WriteString(kv.pre)
 		b.WriteString(entry.Key())
 		pr.Subject = b.String()
-		err := kv.js.purgeStream(kv.stream, &pr)
-		if err != nil {
+		pr.Keep = 0
+		if olderThan > 0 && entry.Created().After(limit) {
+			pr.Keep = 1
+		}
+		if err := kv.js.purgeStream(kv.stream, &pr); err != nil {
 			return err
 		}
 		b.Reset()
--- a/test/kv_test.go
+++ b/test/kv_test.go
@@ -420,13 +420,65 @@ func TestKeyValueDeleteTombstones(t *testing.T) {
 	}
 
 	// Now cleanup.
-	err = kv.PurgeDeletes()
+	err = kv.PurgeDeletes(nats.DeleteMarkersOlderThan(-1))
 	expectOk(t, err)
 
 	si, err := js.StreamInfo("KV_KVS")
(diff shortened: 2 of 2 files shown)


kv.go
```diff
@@ -69,6 +69,9 @@ type KeyValue interface {
 	Bucket() string
 	// PurgeDeletes will remove all current delete markers.
 	PurgeDeletes(opts ...WatchOpt) error
```
derekcollison, line 71: What do we use the WatchOpt for mostly here? Just context?

kv.go
```diff
@@ -69,6 +69,9 @@ type KeyValue interface {
...
+	// PurgeDeletesOlderThan will remove delete markers that have been present
+	// for longer than the `olderThan` duration.
+	PurgeDeletesOlderThan(olderThan time.Duration, opts ...WatchOpt) error
```
derekcollison, line 74: We should shoehorn under PurgeDeletes IMO.

derekcollison (commented):
I like it, we should place under normal PurgeDeletes though IMO.

We can change over the WatchOpts possibly since I think we only use them for context here.

kv.go
kozlovic: We pass whatever watch options that Watch() can accept, including context, but yeah. Don't know if that made sense or not, but that's how it is right now.

kv.go
kozlovic: Would like to, but weird to have the duration value passed as a WatchOpt. What do you suggest then?

derekcollison: Creating new Opt that shadows what you could have used reasonably from the watcher which I think is only context IIRC>

ripienaar: Just a FYI If it's too broad with too many options you can end up with nonsensical combinations sent to some commands. @scottf highlighted a number of these issues in the past and we narrowed down what watchopts do to avoid that.

kozlovic: @ripienaar To make sure I understand what you are saying: we currently pass watch options, which then could allow the user to pass non-sensical options, for instance ignore deletes in which case the functionality would simply not work.

I would not mind replacing WathOpt here with PurgeOpt or something else, but we are all ok to break this? v1.3.0 was released with the PurgeDeletes(opts ...WatchOpt), that is, this is not only in main.

derekcollison: As long as we support context which is the only one that would make sense here I think that is correct option.

ripienaar: Yea since we have this one size fits all function we pass options in that we use to implement various aspects of the api surface with too broad allowed options on the user facing function you end up with things that just don’t work - not to mention confusion for users. 

I thought we fixed this but can’t find an example now - or Scotts email

kozlovic: Would someone, maybe @wallyqs, want to have a go at trying to rework this PR? I find the use of the func options a bit much and can't see how to avoid the stuttering, for instance PurgeDeletes(nats.PurgeDeletesOlderThan()). Of course, the option name should likely be different, but I hope you get the point: I think we will start having issues picking an option functions that:
- are unique enough so that they do not cause conflict with existing or future option function names
- avoid stuttering with the API call that uses them.

In case above, say we shorten the option function name to OlderThan(), it is not clear that this function is for KV purge options, but also would prevent use of that name for an option for other kind of APIs. Think of something more common, say Subject() that could be an option for purging a stream per subject, but could easily also be needed for some kind of other actions, which then would have to be qualified (so that the name is unique - otherwise won't compile), which leads to stuttering.

derekcollison: My overall concern was in making sure we did not delete a tombstone that had not yet been delivered to active watchers.

Thinking out loud, maybe first step is to take the existing function and not delete tombstones that are too young? We could still delete the values but keep the tombstone if it is not at least N minutes old?

kozlovic: To make sure I understand, suppose we have these keys layout:
```
Key: "foo" - Values: 1, 2, DEL/PURGE
Key: "bar" - Values: 1, DEL/PURGE
```
If "foo"'s DEL/PURGE is say time t-1hour, and "bar"'s DEL/PURGE is t-1minute, a "PurgeDeletes()" API would result on:
```
Key: "foo" - Values: <nothing>
Key: "bar" - Values: DEL/PURGE
```
Everything would be removed from "foo", for "bar", only values (regardless of how "old" they are), we just keep DEL/PURGE that are not old enough. Correct?

kozlovic: Btw, I just realized that the current PurgeDeletes() may be flawed if this layout exists (assuming it is possible)
```
Key: "foo" - Values: 1, 2, DEL/PURGE, 3, 4, DEL/PURGE
```
Would the watcher give us 2 del/purge entries for the same key, and we would try to purge the stream by subject twice?

Also, what would be the stream purge request for case where you want to purge a key of its values, but keep the del/purge record? Is it a combination of subject and sequence? If a sequence is given, since it is stream's sequence and messages on various keys (subjects) can interleave, would that work?

derekcollison: Hmm, that might be possible, but I think when we do the PurgeDeletes we get last by subject IIRC so we would just get the last DEL/PURGE.

kozlovic: I see, so we create a watcher for ">" and don't include history, so we get the last message for each subject/key. Good.

Now the question regarding the PurgeDeletes that should remove data but possibly not the DEL/PURGE record itself (if too young). Is it a matter of doing a purge per-subject along with sequence of this record? My question does the sequence works per-subject (even though it is the stream sequence) or is it going to remove all messages under this sequence number, possibly deleting valid keys, etc.. I am planning to write a test case with foo and bar keys having interleaved (stream sequence wise) sequences and have one of the keep that is supposed to keep the DEL/PURGE, while other not and see if that would work.

wallyqs: Not sure if this is what we are looking for still but [in this commit](https://github.com/wallyqs/nats.go/commit/3e0d109b17a3ba155c4134bdd8fc0a9efa57df12) added an `PurgeOpt` that only configures `OlderThan` and `Context` that is then passed to `Watch` for cancellation, plus to avoid the option name being repetitive the option is called `OlderThan` and of `time.Duration` type so that could be used in other contexts potentially if needed.

https://github.com/wallyqs/nats.go/commit/3e0d109b17a3ba155c4134bdd8fc0a9efa57df12

derekcollison: @kozlovic the extended purge options allow purge by subject of course but also allow to specify a "keep" number, so in the case where the tombstone is too young I would suggest a purge of the subject with a keep 1, which will be the tombstone.

kozlovic: @wallyqs Yes, this is exactly what we would have to do if adding a new option (my comment, though, was that nats.OlderThan() is not really telling what it is for since the function itself is not a receiver of a given object. So if in the future you need an option that has the same semantic, you would need to find a new name. One way is to prefix with the "context", like PurgeDeletes, but then it makes the code look like it stutters: PurgeDeletes(nats.PurgeDeletesOlderThan).

Anyway, I think that would work, but I am looking at Derek's approach to see if it would work. By the way, the threshold to decide if we keep the tombstone or not during a PurgeDeletes could now become a KeyValueConfig struct's field. I will prototype that, and then we can decide which direction we go.

derekcollison: I like the idea of threshold being part of KeyValueConfig with sane default, maybe 30minutes by default?

kozlovic: That was going to be my next question: so you want to change the default behavior that from now on, the delete/purge markers less than 30min olds will not be removed, unless the user explicitly lower that value. In other words, this new option, if not set, will default to 30min.

derekcollison: I think so yes. They can set it, and we do clean up the history of each key regardless.

derekcollison: Just leave the tombstone with { keep : 1 } if too young.

kozlovic: I have updated the PR (push forced). Let me know if this close to what you had in mind.

derekcollison (approved):
LGTM!

kozlovic:
@derekcollison Yes, but we have a problem. This would work (ability to set the value) only in case where user calls CreateKeyValue(). For KeyValue() (lookup and bind), that won't work. So either we make it a kvs setter, or we pull Wally's PR (remove WatchOpt and add PurgeOpt) but then add the new logic with Keep:1.

derekcollison:
Good point. I would say for now remove option to set and we can see how things play out if folks really want to change it. Those tombstones are small so as long as we remove the history itself and keep them should be fine with default of 30mins to start.

kozlovic:
Ok, sounds good. I will keep it internally so that I can override for tests, but otherwise will not expose it.

kozlovic:
@derekcollison Ok, I have updated the PR by removing the option and move a test to nats package so I can override the 30min for the test. I also changed the PR's title to indicate the change in behavior (especially that it is not configurable atm).

derekcollison (approved):
LGTM

kozlovic:
@derekcollison @wallyqs Ok, last iteration I promise :-) I have taken Wally's changes (although I had to remove `type purgeOptFn` and its `configurePurge` receiver since they were not used - staticcheck would fail).
So this is still keeping the markers that are less than 30 minutes old by default, but can be changed through option. Also, removes use of WatchOpt for PurgeDeletes() which could have been mis-used.

PR #910, opened 2022-02-15
[FIXED] KV use of JS prefix

kozlovic:
If the user creates a JS context with a custom prefix, this needs
to be used for the subject of the Put() and Delete() operations.
This is addressing the architecture design ADR-19.

### DIFF
--- a/kv.go
+++ b/kv.go
@@ -291,6 +291,8 @@ func (js *js) KeyValue(bucket string) (KeyValue, error) {
 		stream: stream,
 		pre:    fmt.Sprintf(kvSubjectsPreTmpl, bucket),
 		js:     js,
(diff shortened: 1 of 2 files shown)


wallyqs (approved):
LGTM

matthiashanel (changes requested):
Update is missing

I checked all locations of `b.WriteString(kv.pre)`.

PurgeDeletes uses it but goes through the js api, hence does not count. (as js uses the prefix) 
Same applies to Watch.

matthiashanel (approved):
Never Mind, update also goes through js.
LGTM

kozlovic:
@matthiashanel Actually, you were right. Adding the test showed that the update would fail, so we need to use the prefix there too. I added the test for that.

matthiashanel (approved):
LGTM

PR #911, opened 2022-02-16
Add ClientTrace struct for JS tracing

wallyqs:
This is an alternative implementation to enable tracing of requests in the JetStream context #849, using the same idiom of a struct with callbacks as the `httptrace` package (https://golang.google.cn/pkg/net/http/httptrace/#ClientTrace). 

```go
	js, err := nc.JetStream(&ClientTrace{
		RequestSent: func(subj string, payload []byte) {
			log.Printf(">>> %s: %s", subj, string(payload))
		},
		ResponseReceived: func(subj string, payload []byte, hdr Header) {
			log.Printf("<<< %s: %s", subj, string(payload))
		},
	})
```

### DIFF
--- a/js.go
+++ b/js.go
@@ -223,7 +223,7 @@ type jsOpts struct {
 	// the domain that produced the pre
 	domain string
 	// enables protocol tracing
-	trace       TraceCB
+	ctrace      ClientTrace
 	shouldTrace bool
 }
 
@@ -263,26 +263,16 @@ func (opt jsOptFn) configureJSContext(opts *jsOpts) error {
 	return opt(opts)
 }
 
-// TraceOperation indicates the direction of traffic flow to TraceCB
-type TraceOperation int
+// ClientTrace can be used to trace API interactions for the JetStream Context.
+type ClientTrace struct {
+	RequestSent      func(subj string, payload []byte)
+	ResponseReceived func(subj string, payload []byte, hdr Header)
+}
 
-const (
-	// TraceSent indicate the payload is being sent to subj
-	TraceSent TraceOperation = 0
-	// TraceReceived indicate the payload is being received on subj
-	TraceReceived TraceOperation = 1
-)
-
-// TraceCB is called to trace API interactions for the JetStream Context
-type TraceCB func(op TraceOperation, subj string, payload []byte, hdr Header)
-
-// TraceFunc enables tracing of JetStream API interactions
-func TraceFunc(cb TraceCB) JSOpt {
-	return jsOptFn(func(js *jsOpts) error {
-		js.trace = cb
-		js.shouldTrace = true
-		return nil
-	})
+func (ct ClientTrace) configureJSContext(js *jsOpts) error {
+	js.ctrace = ct
+	js.shouldTrace = true
+	return nil
 }
 
 // Domain changes the domain part of JetStream API prefix.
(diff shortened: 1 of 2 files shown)


wallyqs:
If this is ok then will make the change for the natscli, it is a small change: https://github.com/nats-io/natscli/blob/6accd61459f59cbb26e6e14cc24b3606ffc7b426/cli/util.go#L448-L454

js.go
```diff
@@ -267,16 +267,16 @@ const (
 	TraceReceived TraceOperation = 1
 )
 
-// TraceCB is called to trace API interactions for the JetStream Context
-type TraceCB func(op TraceOperation, subj string, payload []byte, hdr Header)
+// ClientTrace can be used to trace API interactions for the JetStream Context.
+type ClientTrace struct {
+	RequestSent      func(subj string, payload []byte, hdr Header)
+	ResponseReceived func(subj string, payload []byte, hdr Header)
```
wallyqs, line 273: We could expand this to make them more granular for example, to only trace consumer lookups or for when adding retries make it possible to log an event for when there is a backoff, etc...

js.go
```diff
@@ -267,16 +267,16 @@ const (
 	TraceReceived TraceOperation = 1
 )
 
-// TraceCB is called to trace API interactions for the JetStream Context
-type TraceCB func(op TraceOperation, subj string, payload []byte, hdr Header)
+// ClientTrace can be used to trace API interactions for the JetStream Context.
+type ClientTrace struct {
+	RequestSent      func(subj string, payload []byte, hdr Header)
```
kozlovic, line 272: Well, since we always call RequestSent() with nil as the last argument, why make this function accept hdr? RequestSent and RequestReceived don't have to have the same signature, right?

js.go
```diff
@@ -1491,7 +1491,10 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync,
 		}
 
 		if js.opts.shouldTrace {
-			js.opts.trace(TraceSent, ccSubj, j, nil)
+			ctrace := js.opts.ctrace
+			if ctrace.RequestSent != nil {
+				ctrace.RequestSent(ccSubj, j, nil)
```
kozlovic, line 1496: The original approach had a single function (with operation type to distinguish), that's why we were passing nil as the last argument. Here, you have no reason to pass nil, just change the signature?

js.go
```diff
@@ -267,16 +267,16 @@ const (
 	TraceReceived TraceOperation = 1
 )
 
-// TraceCB is called to trace API interactions for the JetStream Context
-type TraceCB func(op TraceOperation, subj string, payload []byte, hdr Header)
```
kozlovic, line 271: If we remove this, then why not remove TraceOperation (with TraceSent and TraceReceived) above?

kozlovic (commented):
I like the approach, but of course this is again breaking changes (we had just one in previous PR for the KV purge options). I guess we mentioned that the JS/KV API may change, but hopefully we start to be more stable soon.

js.go
wallyqs: Yes can be removed, we haven't tagged a release of the client that included these so should be ok.

js.go
wallyqs: Good point, not needed for requests to the JS API so removed the last parameter.

js.go
wallyqs: Thanks just updated.

js.go
kozlovic: Oh wow... we are due for a release soon then ;-). But then that's good, so it is a bit less of an issue then.

kozlovic (approved):
LGTM

PR #915, opened 2022-02-18
[FIXED] Websocket: discovered urls would not have "wss://" scheme

### LINKED ISSUE #914: Websocket: "invalid websocket connection" when using `wss://`, host resolves to some IPs
The library would try with `https://<host>:port` but then would with resolved IPs without knowing that it is supposed to be https. The request would go out and get a response error saying that the client is sending http to https server.

kozlovic:
If the user originally connects with "wss://" but has not configured
any secure related option (Options.Secure/TLSConfig), then when
adding implicit URLs (from the server INFO) the library would add
the "ws://" prefix to the received URL.
When trying to connect to those URLs, the library would not try
to make a TLS handshake and would get an error back that would
result in a "websocket invalid connection".

Resolves #914

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -1435,7 +1435,7 @@ func (nc *Conn) setupServerPool() error {
 
 	// Check for Scheme hint to move to TLS mode.
 	for _, srv := range nc.srvPool {
-		if srv.url.Scheme == tlsScheme {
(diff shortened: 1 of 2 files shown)


kozlovic:
@ColinSullivan1 @wallyqs PING?

wallyqs (approved):
LGTM

PR #916, opened 2022-02-18
[IMPROVED] Connect error when given an invalid user creds file

kozlovic:
The failure to connect would be "no nkey seed found". It will now
be more explicit:
```
error signing nonce: unable to extract key pair from file "myfile.creds": no nkey seed found
```


derekcollison (approved):
LGTM

wallyqs (approved):
LGTM

PR #923, opened 2022-03-09
[ADDED] ConnectedUrlRedacted() that does not return the password

ripienaar:

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -1803,6 +1803,21 @@ func (nc *Conn) ConnectedUrl() string {
 	return nc.current.url.String()
 }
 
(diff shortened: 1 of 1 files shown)


nats.go
```diff
@@ -1803,6 +1803,21 @@ func (nc *Conn) ConnectedUrl() string {
 	return nc.current.url.String()
 }
 
+// ConnectedUrlRedacted reports the connected server's URL with passwords redacted
```
ripienaar, line 1806: if you connect with like `nats://user:password@foo:4222` `ConnectedUrl()` returns the password in clear, this just an easier way to get something safely logable

PR #927, opened 2022-03-16
Allow JetStream Publish retries iff ErrNoResponders was returned.

derekcollison:
This is to avoid small blips from leader changes surfacing to the end application.


derekcollison:
Overly pedantic check, wasn't needed. Removed and running again now.

PR #928, opened 2022-03-16
Add `nats: ` prefix to stream info errors to match consumer info

derekcollison:
Make same as consumer by adding prefix.


kozlovic (approved):
LGTM, assuming that we don't have a test that check for error without the 'nats: ' prefix :-)

derekcollison:
Just in the server from the other day ;)

wallyqs (approved):

PR #929, opened 2022-03-16
Allow KV and Obj to specify placement options.

derekcollison:

### DIFF
--- a/kv.go
+++ b/kv.go
@@ -188,6 +188,7 @@ type KeyValueConfig struct {
 	MaxBytes     int64
 	Storage      StorageType
 	Replicas     int
+	Placement    *Placement
 }
 
 // Used to watch all keys.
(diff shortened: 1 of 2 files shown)


kv.go
```diff
@@ -347,6 +348,7 @@ func (js *js) CreateKeyValue(cfg *KeyValueConfig) (KeyValue, error) {
 		MaxMsgSize:        maxMsgSize,
 		Storage:           cfg.Storage,
 		Replicas:          replicas,
+		Placement:         cfg.Placement,
```
kozlovic, line 351: Question: would a stream info return the same user-provided-placement information provided here, or will the server alter that based on where the stream is placed?
The reason I ask is that for the v2.7.2/discard policy upgrade process done below, we do a reflect.DeepEqual() based on this stream config and the stream config we get from the js.StreamInfo() call.

kozlovic (approved):
LGTM, but have a question about stream info's config.

kv.go
derekcollison: If it is specified and the system can not accomodate it will fail.

I am adding in ability to "move" assets though now in server, but that update should be stamped etc so we should be good.

PR #930, opened 2022-03-16
Allow JetStream Publish retries iff ErrNoResponders was returned.

derekcollison:
This is to avoid small blips from leader changes surfacing to the end application.


test/js_test.go
```diff
@@ -6266,3 +6267,80 @@ func TestJetStreamSubscribeContextCancel(t *testing.T) {
...
+			sds := fmt.Sprintf(server.JSApiStreamLeaderStepDownT, "TEST")
+			for time.Now().Before(expires) {
+				time.Sleep(1 * time.Second)
+				si, err := js.StreamInfo("TEST")
+				expectOk(t, err)
+				_, err = nc.Request(sds, nil, time.Second)
+				expectOk(t, err)
+
+				// Wait on new leader.
+				checkFor(t, time.Second, 10*time.Millisecond, func() error {
```
kozlovic, line 6307: Got already a flapper here...

test/js_test.go
```diff
@@ -6266,3 +6267,80 @@ func TestJetStreamSubscribeContextCancel(t *testing.T) {
...
+		// Published every 10ms
+		go func() {
+			defer wg.Done()
+			nc, js := jsClient(t, randServer())
+			defer nc.Close()
+
+			for time.Now().Before(expires) {
+				time.Sleep(10 * time.Millisecond)
+				_, err := js.Publish("foo", []byte("OK"))
+				expectOk(t, err)
```
kozlovic, line 6327: And here: `js_test.go:6327: Unexpected error: nats: timeout`

kozlovic (changes requested):
This test had 2 failures out of the 4 runs, so may want to see if that can be improved.

test/js_test.go
derekcollison: Will look at this one.

js.go
```diff
@@ -422,11 +433,23 @@ func (js *js) PublishMsg(m *Msg, opts ...PubOpt) (*PubAck, error) {
 	}
 
 	if err != nil {
-		if err == ErrNoResponders {
-			err = ErrNoStreamResponse
+		for r := 0; err == ErrNoResponders && r < o.rnum; r++ {
+			// To protect against small blips in leadership changes etc, if we get a no responders here retry.
+			time.Sleep(o.rwait)
+			if o.ttl > 0 {
+				resp, err = js.nc.RequestMsg(m, time.Duration(o.ttl))
```
kozlovic, line 440: Note that the call below uses the context that is created at the beginning, so it is likely that the deadline is the original one, whereas here you make each request with the original timeout, so maybe that should be reduced by the time already spent? If not, there will be some difference in behavior between context/timeout.

wallyqs: I think this might be fine, I like how the context controls the max duration of the retries logic. For example:

```go
ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
defer cancel()
_, err := js.Publish(subj, msg, nats.Context(ctx), nats.RetryWait(500*time.Millisecond), nats.RetryAttempts(30))
```

Above would mean to attempt to publish for 10 seconds max, although the retries would potentially take 15s but the context still sets the deadline.  Also means that possible to have infinite retries, so would retry as needed until the context expires:

```go
ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
defer cancel()
_, err := js.Publish(subj, msg, nats.Context(ctx), nats.RetryWait(500*time.Millisecond), nats.RetryAttempts(-1))
```

js.go
```diff
@@ -422,11 +433,23 @@ func (js *js) PublishMsg(m *Msg, opts ...PubOpt) (*PubAck, error) {
 	}
 
 	if err != nil {
-		if err == ErrNoResponders {
-			err = ErrNoStreamResponse
+		for r := 0; err == ErrNoResponders && r < o.rnum; r++ {
```
wallyqs, line 436:
```suggestion
		for r := 0; err == ErrNoResponders && (r < o.rnum || o.rnum < 0); r++ {
```

That change would make it possible to allow infinite retries, in order to let retries happen as needed while context is still active:

```go
ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
defer cancel()
_, err := js.Publish(subj, msg, nats.Context(ctx), nats.RetryWait(500*time.Millisecond), nats.RetryAttempts(-1))
```

js.go
derekcollison: I will do a rough estimation here subtracting out o.rwait and bailing if local ttl in the loop goes negative. Will post separate CL for ease of review but can squash before merging.

js.go
```diff
@@ -422,11 +433,23 @@ func (js *js) PublishMsg(m *Msg, opts ...PubOpt) (*PubAck, error) {
...
+			// To protect against small blips in leadership changes etc, if we get a no responders here retry.
+			time.Sleep(o.rwait)
```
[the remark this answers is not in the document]
derekcollison: Good call, done in separate CL below, will squash at end.

js.go
derekcollison: Should we make sure a ctx is present to do that one?
I will make the change but we could get into an infinite loop if the stream is truly gone and their is no context provided to the Publish call.

derekcollison: Change made below.

kozlovic (approved):
LGTM

wallyqs (approved):
LGTM!

PR #933, opened 2022-03-22
Added BackOff helper method to set backoff through subOpts

mfaizanse:
- Added BackOff helper method to set backoff through subOpts

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -624,7 +624,7 @@ func ExampleSubOpt() {
 	// Start delivering messages with delay based on BackOff array of time durations.
 	js.Subscribe("foo", func(msg *nats.Msg) {
 		fmt.Printf("Received a message: %s\n", string(msg.Data))
-	}, nats.BackOff([]time.Duration{50 * time.Millisecond, 250 * time.Millisecond}))
+	}, nats.ManualAck(), nats.MaxDeliver(2), nats.BackOff([]time.Duration{50 * time.Millisecond, 250 * time.Millisecond}))
 }
 
 func ExampleMaxWait() {
--- a/example_test.go
+++ b/example_test.go
@@ -620,6 +620,11 @@ func ExampleSubOpt() {
 	js.Subscribe("foo", func(msg *nats.Msg) {
 		fmt.Printf("Received a message: %s\n", string(msg.Data))
 	}, nats.StartTime(time.Now().Add(-2*time.Hour)))
+
+	// Start delivering messages with delay based on BackOff array of time durations.
+	js.Subscribe("foo", func(msg *nats.Msg) {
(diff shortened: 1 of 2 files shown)


derekcollison:
What is the goal of this change? For true push I am worried that this could collide with flow (or lack thereof) and cause memory bloat in the client app.

Could be useful maybe for when you do a pull request however but best to understand what the exact intent is here.

mfaizanse:
The NAK delay and backOff is implemented in these PRs: [nats-server](https://github.com/nats-io/nats-server/pull/2812) and [nats-go](https://github.com/nats-io/nats.go/pull/894/commits).

Currently, in the API if we create a consumer then we can set the `BackOff` but if we create a subscriber then we don't have the Opts method to set the `BackOff`. The goal of this PR is to enable the user to set the `BackOff` when creating through a subscriber like this:
```
// Create a subscriber
	_, err = js.Subscribe(
		subject,
		callback,
		nats.Durable("consumer1"),
		nats.ManualAck(),
		nats.AckExplicit(),
		nats.BackOff([]time.Duration{50 * time.Millisecond, 250 * time.Millisecond})
	)
```

example_test.go
```diff
@@ -620,6 +620,11 @@ func ExampleSubOpt() {
 	js.Subscribe("foo", func(msg *nats.Msg) {
 		fmt.Printf("Received a message: %s\n", string(msg.Data))
 	}, nats.StartTime(time.Now().Add(-2*time.Hour)))
+
+	// Start delivering messages with delay based on BackOff array of time durations.
+	js.Subscribe("foo", func(msg *nats.Msg) {
+		fmt.Printf("Received a message: %s\n", string(msg.Data))
+	}, nats.BackOff([]time.Duration{50 * time.Millisecond, 250 * time.Millisecond}))
```
kozlovic, line 627: This complies, but would actually not "work" because you would need to set the MaxDeliver to at least 2 (or more), but also add `nats.ManualAck()` because otherwise the callback would auto-ack anyway, so it is unlikely that there would be redeliveries...

So instead (or at least in addition to) having this here, I would update, say TestJetStreamSubscribe_AckPolicy test to use the nats.BackOff() new option. In this test (at the end of it), we use backoff but with the AddConsumer directly.

kozlovic (changes requested):
Would add a test and either remove from example or "fix" it as described in the comments.

@derekcollison We (I think I did) added the features (Nak and Backoff) but I simply forgot to add the option. So I think @mfaizanse is just fixing my omission here :-)

kozlovic:
@derekcollison I have approved this PR, let me know if you still have concerns: we have added the BackOff list consumer configuration in the client, so not sure why is that an issue to make it an option for the js.Subscribe() calls...

PR #934, opened 2022-03-25
Fix race with async INFO and header APIs

wallyqs:
`RequestMsg` and `PublishMsg` were not checking for headers support under the lock so there was a race condition with async INFO protocols.

### DIFF
--- a/test/headers_test.go
+++ b/test/headers_test.go
@@ -110,6 +110,11 @@ func TestRequestMsg(t *testing.T) {
 	if _, err = nc.RequestMsg(nil, time.Second); err != nats.ErrInvalidMsg {
 		t.Errorf("Unexpected error: %v", err)
 	}
+	ctx, cancel := context.WithTimeout(context.Background(), 250*time.Millisecond)
+	defer cancel()
+	if _, err = nc.RequestMsgWithContext(ctx, nil); err != nats.ErrInvalidMsg {
+		t.Errorf("Unexpected error: %v", err)
+	}
 }
 
 func TestRequestMsgRaceAsyncInfo(t *testing.T) {
@@ -144,6 +149,13 @@ func TestRequestMsgRaceAsyncInfo(t *testing.T) {
 	}
 	defer nc.Close()
 
+	// Extra client with old request.
+	nc2, err := nats.Connect(s.ClientURL(), nats.DontRandomize(), nats.UseOldRequestStyle())
+	if err != nil {
+		t.Fatalf("Error connecting to server: %v", err)
+	}
+	defer nc2.Close()
+
 	subject := "headers.test"
 	sub, err := nc.Subscribe(subject, func(m *nats.Msg) {
 		r := nats.NewMsg(m.Reply)
@@ -161,7 +173,8 @@ func TestRequestMsgRaceAsyncInfo(t *testing.T) {
 
 	// Leave some goroutines publishing in parallel while
 	// async protocols are being received.
-	var received int64
+	var received, receivedWithContext int64
+	var receivedOldStyle, receivedOldStyleWithContext int64
 	var producers int = 50
 	for i := 0; i < producers; i++ {
 		go func() {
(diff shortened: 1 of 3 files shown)


kozlovic (changes requested):
Can you have a look at context.go? It seems that RequestMsgWithContext() is also accessing nc.info.Headers directly.

wallyqs:
@kozlovic context APIs had the same issue, now updated.

test/headers_test.go
```diff
@@ -100,6 +103,218 @@ func TestRequestMsg(t *testing.T) {
...
+		t.Errorf("Unexpected error: %v", err)
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), 250*time.Millisecond)
+	defer cancel()
+	if _, err = nc.RequestMsgWithContext(ctx, nil); err != nats.ErrInvalidMsg {
+		t.Errorf("Unexpected error: %v", err)
+	}
+}
+
+func TestRequestMsgRaceAsyncInfo(t *testing.T) {
```
kozlovic, line 120: This test seem to be flapping and is quite complex. What we are trying to capture here is simply the data race. I have written this "simplified" test that from main would produce the data race. Could you have a look?
```
        s1Opts := natsserver.DefaultTestOptions
	s1Opts.Host = "127.0.0.1"
	s1Opts.Port = -1
	s1Opts.Cluster.Name = "CLUSTER"
	s1Opts.Cluster.Host = "127.0.0.1"
	s1Opts.Cluster.Port = -1
	s := natsserver.RunServer(&s1Opts)
	defer s.Shutdown()

	nc, err := nats.Connect(s.ClientURL())
	if err != nil {
		t.Fatalf("Error connecting to server: %v", err)
	}
	defer nc.Close()

	// Extra client with old request.
	nc2, err := nats.Connect(s.ClientURL(), nats.UseOldRequestStyle())
	if err != nil {
		t.Fatalf("Error connecting to server: %v", err)
	}
	defer nc2.Close()

	subject := "headers.test"
	if _, err := nc.Subscribe(subject, func(m *nats.Msg) {
		r := nats.NewMsg(m.Reply)
		r.Header["Hdr-Test"] = []string{"bar"}
		r.Data = []byte("+OK")
		m.RespondMsg(r)
	}); err != nil {
		t.Fatalf("subscribe failed: %v", err)
	}
	nc.Flush()

	wg := sync.WaitGroup{}
	wg.Add(1)
	ch := make(chan struct{})
	go func() {
		defer wg.Done()

		s2Opts := natsserver.DefaultTestOptions
		s2Opts.Host = "127.0.0.1"
		s2Opts.Port = -1
		s2Opts.Cluster.Name = "CLUSTER"
		s2Opts.Cluster.Host = "127.0.0.1"
		s2Opts.Cluster.Port = -1
		s2Opts.Routes = server.RoutesFromStr(fmt.Sprintf("nats://127.0.0.1:%d", s.ClusterAddr().Port))
		for {
			s := natsserver.RunServer(&s2Opts)
			s.Shutdown()
			select {
			case <-ch:
				return
			default:
			}
		}
	}()

	msg := nats.NewMsg(subject)
	msg.Header["Hdr-Test"] = []string{"quux"}
	for i := 0; i < 100; i++ {
		nc.RequestMsg(msg, time.Second)
		ctx, cancel := context.WithTimeout(context.Background(), time.Second)
		nc.RequestMsgWithContext(ctx, msg)
		cancel()

		// Check with old style requests as well.
		nc2.RequestMsg(msg, time.Second)
		ctx2, cancel2 := context.WithTimeout(context.Background(), time.Second)
		nc2.RequestMsgWithContext(ctx2, msg)
		cancel2()
	}

	close(ch)
	wg.Wait()
```

kozlovic (changes requested):

test/headers_test.go
wallyqs: Thanks! This is much simpler

kozlovic (approved):
LGTM

PR #935, opened 2022-03-25
Enhance client to do reconnect on max connections from server

tbeets:
The client will attempt reconnect when `maximum connections exceeded` error is sent by server immediately prior to a hard disconnect rather than close without reconnect attempt(s).

This is additive to existing re-connection on stale connection from server, auth and perm errors, and read loop errors (such as network disconnect).

### DIFF
--- a/nats_test.go
+++ b/nats_test.go
@@ -276,6 +276,69 @@ var testServers = []string{
 	"nats://localhost:1228",
 }
 
+func TestMaxConnectionsReconnect(t *testing.T) {
+	// Join cluster on first server
+	routes := []*url.URL{&url.URL{Scheme: "nats", Host: fmt.Sprintf("127.0.0.1:%d", TEST_PORT)}}
+
+	// Start first server
+	s1Opts := natsserver.DefaultTestOptions
+	s1Opts.Port = -1
+	s1Opts.MaxConn = 2
+	s1Opts.Cluster = server.ClusterOpts{Name: "test", Host: "127.0.0.1", Port: TEST_PORT}
+	s1Opts.Routes = routes
+	s1 := RunServerWithOptions(&s1Opts)
+	defer s1.Shutdown()
+
+	// Start second server
+	s2Opts := natsserver.DefaultTestOptions
+	s2Opts.Port = -1
+	s2Opts.MaxConn = 2
+	s2Opts.Cluster = server.ClusterOpts{Name: "test", Host: "127.0.0.1", Port: TEST_PORT + 1}
+	s2Opts.Routes = routes
+	s2 := RunServerWithOptions(&s2Opts)
+	defer s2.Shutdown()
+
+	// Only explicitly connect to first server
+	var opts = Options{
+		Url:            s1.ClientURL(),
+		AllowReconnect: true,
+		MaxReconnect:   2,
+		ReconnectWait:  10 * time.Millisecond,
+		Timeout:        200 * time.Millisecond,
+	}
+
+	// Create two connections (the current max) for first server
+	nc1, _ := opts.Connect()
+	defer nc1.Close()
+	nc1.Flush()
+
+	nc2, _ := opts.Connect()
+	defer nc2.Close()
+	nc2.Flush()
+
+	if s1.NumClients() != 2 {
+		t.Fatalf("Expected 2 client connections to first server. Got %d\n", s1.NumClients())
+	}
+
+	if s2.NumClients() > 0 {
+		t.Fatalf("Expected 0 client connections to second server. Got %d\n", s2.NumClients())
+	}
+
+	// Kick one of our two server connections off first server. One client should reconnect to second server.
+	newS1Opts := s1Opts
+	newS1Opts.MaxConn = 1
+	err := s1.ReloadOptions(&newS1Opts)
+	if err != nil {
+		t.Fatalf("Unexpected error changing max_connections [%s]", err)
+	}
+
+	time.Sleep(200 * time.Millisecond)
+
+	if s2.NumClients() <= 0 || s1.NumClients() > 1 {
+		t.Fatalf("Expected client reconnection to second server.\n")
+	}
+}
+
 func TestSimplifiedURLs(t *testing.T) {
 	for _, test := range []struct {
 		name     string
--- a/nats_test.go
+++ b/nats_test.go
@@ -277,15 +277,12 @@ var testServers = []string{
 }
 
 func TestMaxConnectionsReconnect(t *testing.T) {
-	// Join cluster on first server
-	routes := []*url.URL{&url.URL{Scheme: "nats", Host: fmt.Sprintf("127.0.0.1:%d", TEST_PORT)}}
 
 	// Start first server
 	s1Opts := natsserver.DefaultTestOptions
 	s1Opts.Port = -1
 	s1Opts.MaxConn = 2
-	s1Opts.Cluster = server.ClusterOpts{Name: "test", Host: "127.0.0.1", Port: TEST_PORT}
-	s1Opts.Routes = routes
+	s1Opts.Cluster = server.ClusterOpts{Name: "test", Host: "127.0.0.1", Port: -1}
 	s1 := RunServerWithOptions(&s1Opts)
 	defer s1.Shutdown()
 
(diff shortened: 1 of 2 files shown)


nats_test.go
```diff
@@ -276,6 +276,69 @@ var testServers = []string{
 	"nats://localhost:1228",
 }
 
+func TestMaxConnectionsReconnect(t *testing.T) {
+	// Join cluster on first server
+	routes := []*url.URL{&url.URL{Scheme: "nats", Host: fmt.Sprintf("127.0.0.1:%d", TEST_PORT)}}
```
kozlovic, line 281: Would remove this (see below)

nats_test.go
```diff
@@ -276,6 +276,69 @@ var testServers = []string{
...
+
+	// Start first server
+	s1Opts := natsserver.DefaultTestOptions
+	s1Opts.Port = -1
+	s1Opts.MaxConn = 2
+	s1Opts.Cluster = server.ClusterOpts{Name: "test", Host: "127.0.0.1", Port: TEST_PORT}
```
kozlovic, line 287: Change port to `-1`

nats_test.go
```diff
@@ -276,6 +276,69 @@ var testServers = []string{
...
+	s1Opts.Routes = routes
```
kozlovic, line 288: Remove this line (see below)

nats_test.go
```diff
@@ -276,6 +276,69 @@ var testServers = []string{
...
+	s1 := RunServerWithOptions(&s1Opts)
+	defer s1.Shutdown()
+
+	// Start second server
+	s2Opts := natsserver.DefaultTestOptions
+	s2Opts.Port = -1
+	s2Opts.MaxConn = 2
+	s2Opts.Cluster = server.ClusterOpts{Name: "test", Host: "127.0.0.1", Port: TEST_PORT + 1}
```
kozlovic, line 296: Replace port with `-1`

nats_test.go
```diff
@@ -276,6 +276,69 @@ var testServers = []string{
...
+	s2Opts.Routes = routes
```
kozlovic, line 297: Replace this with:
```
s2Opts.Routes = server.RoutesFromStr(fmt.Sprintf("nats://127.0.0.1:%d", s1Opts.Cluster.Port))
```
Once s1 starts, it will update s1Opts.Cluster.Port with the actual port being used. You simply have s2 connects to s1, but that's ok. The original was equivalent since s1 was pointing to itself. We are trying to move away from hardcoded ports when possible.

nats_test.go
```diff
@@ -276,6 +276,69 @@ var testServers = []string{
...
+	nc1, _ := opts.Connect()
+	defer nc1.Close()
+	nc1.Flush()
+
+	nc2, _ := opts.Connect()
+	defer nc2.Close()
+	nc2.Flush()
+
+	if s1.NumClients() != 2 {
+		t.Fatalf("Expected 2 client connections to first server. Got %d\n", s1.NumClients())
```
kozlovic, line 320: Don't need the `\n` at the end.

nats_test.go
```diff
@@ -276,6 +276,69 @@ var testServers = []string{
...
+	}
+
+	if s2.NumClients() > 0 {
+		t.Fatalf("Expected 0 client connections to second server. Got %d\n", s2.NumClients())
```
kozlovic, line 324: Don't need the `\n` at the end.

nats_test.go
```diff
@@ -276,6 +276,69 @@ var testServers = []string{
...
+	newS1Opts.MaxConn = 1
+	err := s1.ReloadOptions(&newS1Opts)
+	if err != nil {
+		t.Fatalf("Unexpected error changing max_connections [%s]", err)
+	}
+
+	time.Sleep(200 * time.Millisecond)
+
+	if s2.NumClients() <= 0 || s1.NumClients() > 1 {
+		t.Fatalf("Expected client reconnection to second server.\n")
```
kozlovic, line 338: Don't need the `\n` at the end (and no need for punctuation).

kozlovic (changes requested):
For some reasons, the same tests are failing in push/PR.. locally they pass for me, wonder if this is just a coincidence or caused by this PR. Once you make the changes I requested, we will see if the same tests fail again.

kozlovic:
@tbeets Discard my comment about same test failing in push/pr since you have only a PR run.. there were 2 tests (but different), one in 1.17 and one in 1.16 (the code coverage run). I am going to recycle to see if those were flappers or not.

kozlovic:
hum... got this test failure again:
```
=== RUN   TestJetStreamClusterStreamLeaderChangeClientErr
    js_test.go:6334: Unexpected error: nats: raft: not leader
--- FAIL: TestJetStreamClusterStreamLeaderChangeClientErr (6.89s)
```

nats_test.go
```diff
@@ -276,6 +276,69 @@ var testServers = []string{
...
+	s2Opts := natsserver.DefaultTestOptions
+	s2Opts.Port = -1
+	s2Opts.MaxConn = 2
+	s2Opts.Cluster = server.ClusterOpts{Name: "test", Host: "127.0.0.1", Port: TEST_PORT + 1}
+	s2Opts.Routes = routes
+	s2 := RunServerWithOptions(&s2Opts)
+	defer s2.Shutdown()
+
+	// Only explicitly connect to first server
+	var opts = Options{
```
wallyqs, line 302: `opts` here will become shared state for the two clients connections, it would be better to change into functional options instead just in case: 
```
	opts := []Option{
		MaxReconnect(2),
		ReconnectWait(10*time.Millisecond),
		Timeout(200*time.Millisecond),
	}

	// Create two connections (the current max) for first server.
	nc1, _ := nats.Connect(s1.ClientURL(), opts...)
	defer nc1.Close()
	nc1.Flush()

	nc2, _ := nats.Connect(s1.ClientURL(), opts...)
	defer nc2.Close()
	nc2.Flush()
```

kozlovic: True..

tbeets: Great recommendations @kozlovic and @wallyqs .  Tips from the best.

I have made all the changes and added the commit to the PR.

nats_test.go
```diff
@@ -276,6 +276,63 @@ var testServers = []string{
...
+
+	// Kick one of our two server connections off first server. One client should reconnect to second server
+	newS1Opts := s1Opts
+	newS1Opts.MaxConn = 1
+	err := s1.ReloadOptions(&newS1Opts)
+	if err != nil {
+		t.Fatalf("Unexpected error changing max_connections [%s]", err)
+	}
+
+	time.Sleep(200 * time.Millisecond)
```
wallyqs, line 329: To avoid the sleep here, you could maybe use the reconnect / disconnect handlers in the options like this:

```go
	reconnectCh := make(chan struct{})
	errCh := make(chan error, 2)
	opts := []Option{
		MaxReconnects(2),
		ReconnectWait(10*time.Millisecond),
		Timeout(200*time.Millisecond),
		ErrorHandler(func(_ *Conn, _ *Subscription, err error) {
			fmt.Println("ERROR: ", err)
		}),
		DisconnectErrHandler(func(_ *Conn, err error) {
			if err != nil {
				errCh <- err
			}
		}),
		ReconnectHandler(func(_ *Conn){
			reconnectCh <- struct{}{}
		}),
	}
```

Then replace the sleep with a couple of channels that check for the disconnect error and reconnect event:

```go
	select {
	case err := <-errCh:
		if err != ErrMaxConnectionsExceeded {
			t.Fatalf("Unexpected error %v", err)
		}
	case <-time.After(2*time.Second):
		t.Fatal("Timed out waiting for disconnect event")
	}

	select {
	case <-reconnectCh:
	case <-time.After(2*time.Second):
		t.Fatal("Timed out waiting for reconnect event")
	}
```

kozlovic: Very good recommendation: you will both check that the error was what you expected *and* remove any flapping condition due to 200ms being maybe not enough some time or too much..

tbeets: Good suggestion @wallyqs .  Those hard-coded sleeps do seem like ticking time bombs (or at least wasted time).

tbeets: Change made and added to PR.

kozlovic (approved):
LGTM

wallyqs (approved):
LGTM!

PR #941, opened 2022-03-29
Set async pub ack inflight pending default to 4K, add StallWait option.

wallyqs:
This makes async publishing stall wait customizable via the `StallWait` option, also sets the default stall pending default to 4K.

### DIFF
--- a/js.go
+++ b/js.go
@@ -187,6 +187,9 @@ const (
 
 	// Default number of retries
 	DefaultPubRetryAttempts = 2
+
+	// defaultAsyncPubAckInflight is the number of async pub acks inflight.
+	defaultAsyncPubAckInflight = 4000
 )
 
 // Types of control messages, so far heartbeat and flow control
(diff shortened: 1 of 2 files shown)


js.go
```diff
@@ -187,6 +187,9 @@ const (
 
 	// Default number of retries
 	DefaultPubRetryAttempts = 2
+
+	// defaultAsyncPubAckInflight is the number of async pub acks inflight.
+	defaultAsyncPubAckInflight = 4 * 1024
```
kozlovic, line 192: FYI: For server change and max ack pending, I was going for 1024, but then realize that we maybe need to stop the kilo byte semantic for just numbers. So I set it to 1000 instead. But will see if Derek asks me to set to 1024 there, so don't update just yet :-)

derekcollison: I do everything old school and power of 2, but let's do what we think is right going forward.

kozlovic (approved):
@wallyqs I have merged the changes in the server for the consumer default max ack pending to 1,000 (not 1,024 - but was 20,000 so already that one was not a 1024 multiple). If we feel like it will be less of a "departure" from our default limits, you can keep 4*1024 instead of 4,000. Really up to you.

PR #947, opened 2022-04-04
[FIXED] JetStream: check stream and consumer names

kozlovic:
Many APIs were not properly checking that the stream and/or
consumer names were provided and valid (no "." in them), which
could lead to requests being sent on a subject that the server
has no interest on and no response was therefore sent to the
client library, which would fail with a timeout/context deadline.


jsm.go
```diff
@@ -295,6 +295,19 @@ type consumerDeleteResponse struct {
...
 }
 
+func checkStreamAndConsumerName(stream, consumer string) error {
+	if stream == _EMPTY_ {
+		return ErrStreamNameRequired
+	}
+	if consumer == _EMPTY_ {
+		return ErrConsumerNameRequired
+	}
+	if strings.Contains(consumer, ".") {
```
derekcollison, line 305: Should we check stream as well?

nats.go
```diff
@@ -145,6 +145,7 @@ var (
 	ErrNotJSMessage                 = errors.New("nats: not a jetstream message")
 	ErrInvalidStreamName            = errors.New("nats: invalid stream name")
 	ErrInvalidDurableName           = errors.New("nats: invalid durable name")
+	ErrInvalidConsumerName          = errors.New("nats: invalid consumer name")
```
derekcollison, line 148: Add stream too?

kozlovic: We already have such error, since we check on AddStream, etc..

kozlovic:
@derekcollison Made the changes (quite a bit more.. force pushed since almost nothing in common with first push). Also added all tests in TestJetStreamManagement.

derekcollison (approved):
LGTM

PR #948, opened 2022-04-05
Add test that disables reconnect buf

wallyqs:
Add test case to document usage of disabling reconnect buffer.


kozlovic (approved):
LGTM

kozlovic:
@wallyqs But as you can see fails on lack of go fmt, so please fix before merging.

wallyqs:
Thanks @kozlovic, fixed before merging.

PR #950, opened 2022-04-07
Add a standard subscription [ci-skip]

ColinSullivan1:
Add a service subject that isn't load balanced for example/demonstration purposes.

Signed-off-by: Colin Sullivan <colin@synadia.com>

### DIFF
--- a/examples/nats-echo/main.go
+++ b/examples/nats-echo/main.go
@@ -44,7 +44,7 @@ func showUsageAndExit(exitcode int) {
 }
 
 func printMsg(m *nats.Msg, i int) {
-	log.Printf("[#%d] Echoing to [%s]: %q", i, m.Reply, m.Data)
+	log.Printf("[#%d] Echoing from [%s] to [%s]: %q", i, m.Subject, m.Reply, m.Data)
 }
 
 func main() {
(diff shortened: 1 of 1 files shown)


examples/nats-echo/main.go
```diff
@@ -115,6 +115,14 @@ func main() {
 				nc.Publish(msg.Reply, msg.Data)
 			}
 		}
+	}
+
+	nc.QueueSubscribe(subj, "echo", func(msg *nats.Msg) {
+		handleMsg(msg)
+	})
+	allSubj := subj + ".all"
```
derekcollison, line 123: Let's make this status.

derekcollison: And for this one we just should send back our geo I think? Maybe if you want add in an identifier at startup to customize?

derekcollison (changes requested):

examples/nats-echo/main.go
ColinSullivan1: Fixed; it's a status and there is an identifier now.

tbeets:
LGTM

tbeets (approved):
LGTM

philpennock (approved):

Issue #880, opened 2022-01-09
timer.go.I'm wondering if it's an nats issue.


sgof1d:
go 1.17.5, 1.17.6 . nats.go 
github.com/nats-io/nats.go v1.13.1-0.20211018182449-f2416a8b1483

some time the time.Timer returned by globalTimerPool.Get(a Duration) was fired already
so request()  will fail with "nats timeout" in 10us or less.

the issue disappears if I disable pool.

kozlovic:
@sgof1d Would you have a test case that reproduce this? The pool implementation does the right thing, I believe, in that when putting it back to the pool it stops the timer and if it can't (meaning it had fired), then it consumes from the channel:
```
func (tp *timerPool) Put(t *time.Timer) {
	if !t.Stop() {
		select {
		case <-t.C:
		default:
		}
	}

	tp.p.Put(t)
}
```

sgof1d:
yes, It seems that pool does thing right.
reproduce rate is about 10%. if I enable pool.
reproduce rae is 0%, if I disable pool.
the test case is :
clients send websocket packet to api gateway
gateway call nats.request() in go routines.
backend server will response to these requests.

the issue happens in  gateway.

I had tested that: 
if affirm timer.C is REALLY empty before Get() return, the issue disappear.

so I'm wondering something is wrong in timer.Stop(), the go standard lib.

kozlovic:
I think I have seen some issue reports around that. Will try to reproduce with using simple timer/pool but outside of NATS and see... out of curiosity, which platform are you running on?

moredure:
https://github.com/golang/go/issues/47859

sgof1d:
> I think I have seen some issue reports around that. Will try to reproduce with using simple timer/pool but outside of NATS and see... out of curiosity, which platform are you running on?

alpine running in container.

wallyqs:
Closing since looks like regression from Go that has been fixed since Go 1.16.

PR #951, opened 2022-04-07
Remove globalTimerPool global variable

jipperinbham:
We occasionally see weird behavior when using Request with timeouts where it appeared as though the timer had already fired even though it did not appear to be the case. Looking through the Request functions, it seemed pragmatic to rely on the standard library's `context` instead. The other use of `globalTimerPool` in `NextMsg` and `FlushTimeout` appeared to also be able to use `context.WithTimeout` which made it no longer necessary to have the `globalTimerPool` variable.

Looking through some open issues, it appeared as though https://github.com/nats-io/nats.go/issues/880 could be addressed with this change.


wallyqs:
Thanks for the contribution. I think this might be preventing the issue because `context` is not using a pool for those timers and avoiding what looks like were `timer` regressions in go 1.16, 1.7 that have been fixed since then? (https://github.com/golang/go/issues/47859)
Internally `context.WithTimeout` uses `AfterFunc` so there would be a goroutine per request, not sure about changing that into being the default yet... 
Also have you tried with switching to `RequestWithContext` instead?

PR #954, opened 2022-04-09
Add bson encoder for mongo driver serialized structures

Ja7ad:
# abstract
BSON is a binary serialization format used to store documents and make remote procedure calls in MongoDB.

# why nats required to bson encoder?
Mongodb has a dedicated serialization structure called bson that's used by most golang projects.
If Nats supported the BSON encoder, developers wouldn't have to recode data into BSON for submission into the database.


kozlovic:
@Ja7ad Thank you for your contribution. However, we have internally said that the encoded connection in core NATS that was introduced long ago, was maybe not the best choice. So we don't think that we want to augment it with a new driver/dependency. Users can have a library that wraps NATS and do their own encoding/decoding during publish/subscribe.

I will leave this opened a little to see if my colleagues concur. Should we decide to instead accept the change, there would be changes needed that I will not go into now if we ultimately decide to reject the PR. Thank you for your understanding.

wallyqs:
Agree with @kozlovic, we should hold off on adding new encoders for now (or maybe rework them into using generics from Go 1.18). As a first step we could create a different package where these encoders reside instead of main nats.go and then deprecate them in nats.go, but to do this think we need to change the way some of the async error callbacks are pushed internally for the encoded connections: https://github.com/nats-io/nats.go/blob/main/enc.go#L210-L212

Ja7ad:
> @Ja7ad Thank you for your contribution. However, we have internally said that the encoded connection in core NATS that was introduced long ago, was maybe not the best choice. So we don't think that we want to augment it with a new driver/dependency. Users can have a library that wraps NATS and do their own encoding/decoding during publish/subscribe.
> 
> I will leave this opened a little to see if my colleagues concur. Should we decide to instead accept the change, there would be changes needed that I will not go into now if we ultimately decide to reject the PR. Thank you for your understanding.

Thank you for review, some projects go to use nosql database (mongodb), for mongo required serialized data into bson encoder.

This encoder is useful, but you think hold off add new encoder, this PR can be draft in my repository.

PR #964, opened 2022-04-27
feat: add UpsertStream function

ludusrusso:


wallyqs:
I think this would be a good addition to the management API, alternatives would be to have an option or make `UpdateStream` work in an upsert way maybe.

PR #966, opened 2022-04-28
fix: handler nats server leadership changed error

ludusrusso:
This is a trial to fix the case described here: https://github.com/nats-io/nats.go/discussions/965

Not sure how to reproduce the issue, I'll try later to create a kind environment for testing!

/cc @wallyqs

### DIFF
--- a/js.go
+++ b/js.go
@@ -2453,6 +2453,9 @@ func checkMsg(msg *Msg, checkSts bool) (usrMsg bool, err error) {
 		// implementation. Regardless, ignore 408 errors until receiving at least
 		// one message.
 		err = ErrTimeout
+	case leadershipChangedSts:
+		// 409 indicates that the consumer leadership changed no the server. The client should ignore 409 errors.
+		err = ErrTimeout
 	default:
 		err = fmt.Errorf("nats: %s", msg.Header.Get(descrHdr))
 	}
--- a/nats.go
+++ b/nats.go
@@ -3306,19 +3306,20 @@ func NewMsg(subject string) *Msg {
 }
 
 const (
-	hdrLine            = "NATS/1.0\r\n"
-	crlf               = "\r\n"
-	hdrPreEnd          = len(hdrLine) - len(crlf)
-	statusHdr          = "Status"
-	descrHdr           = "Description"
-	lastConsumerSeqHdr = "Nats-Last-Consumer"
-	lastStreamSeqHdr   = "Nats-Last-Stream"
-	consumerStalledHdr = "Nats-Consumer-Stalled"
-	noResponders       = "503"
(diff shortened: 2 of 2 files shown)


ludusrusso:
@wallyqs I really don't understand how to fix the build error, can you help me?

wallyqs:
Saw that `TestJetStreamPullSubscribeOptions/max_request_batch` seems to be failing for some reason, will check

ludusrusso:
> Saw that `TestJetStreamPullSubscribeOptions/max_request_batch` seems to be failing for some reason, will check

It seems that the code I've changed generate a Timeout error in the test, this maybe means that in the test happens a Leadership Change that now is handled, but I don't understand exactly where in the code.

js.go
```diff
@@ -2453,6 +2453,9 @@ func checkMsg(msg *Msg, checkSts bool) (usrMsg bool, err error) {
 		// implementation. Regardless, ignore 408 errors until receiving at least
 		// one message.
 		err = ErrTimeout
+	case leadershipChangedSts:
+		// 409 indicates that the consumer leadership changed no the server. The client should ignore 409 errors.
+		err = ErrTimeout
```
kozlovic, line 2458: We are revisiting the handling of the ErrTimeout in Fetch at the moment, so not sure if this makes sense here. If the leadership changed while the library was waiting to get a message (but did not get any), your change would cause Fetch() to go back to sub.nextMsgWithContext() without resending a new request, which means that the library would just sit in the sub's next until timeout and would never receive a message even if some became available. This error from the server means that the request is invalid, so a new request need to be sent.

nats.go
```diff
@@ -3306,19 +3306,20 @@ func NewMsg(subject string) *Msg {
...
+	hdrPreEnd            = len(hdrLine) - len(crlf)
+	statusHdr            = "Status"
+	descrHdr             = "Description"
+	lastConsumerSeqHdr   = "Nats-Last-Consumer"
+	lastStreamSeqHdr     = "Nats-Last-Stream"
+	consumerStalledHdr   = "Nats-Consumer-Stalled"
+	noResponders         = "503"
+	noMessagesSts        = "404"
+	reqTimeoutSts        = "408"
+	leadershipChangedSts = "409"
```
kozlovic, line 3320: I would refrain doing that, since server tends to use the same code for different errors (case in point 408 has several descriptions at the moment). So status 409 is not necessarily leadership change. For instance trying to do a "next message" to a push consumer would return "Consumer is push based", or doing a a Fetch() with a batch higher that MaxRequestBatch would return "Exceeded MaxRequestBatch of <x>", etc...

kozlovic (changes requested):
I would wait a bit until we resolve #967

PR #967, opened 2022-05-02
js: handle 408 fetch requests pending status

wallyqs:
For fetch batches of n > 1, sometimes the client would receive a 408 Requests Pending status messages for Fetch requests (initial request with no wait + expires time) which resulted in the client skipping these errors until reaching the deadline.
In this PR we add special handling for this status so that the client makes a request without no_wait after getting a 408 status with requests pending.

### DIFF
--- a/js.go
+++ b/js.go
@@ -2414,12 +2414,20 @@ func PullMaxWaiting(n int) SubOpt {
 	})
 }
 
-var errNoMessages = errors.New("nats: no messages")
+var (
+	// errNoMessages is an error that a Fetch request using no_wait can receive to signal
+	// that there are no more messages available.
+	errNoMessages = errors.New("nats: no messages")
+
+	// errRequestsPending is an error that represents a sub.Fetch requests that was using
+	// no_wait and expires time got discarded by the server.
+	errRequestsPending = errors.New("nats: requests pending")
+)
 
 // Returns if the given message is a user message or not, and if
 // `checkSts` is true, returns appropriate error based on the
 // content of the status (404, etc..)
-func checkMsg(msg *Msg, checkSts bool) (usrMsg bool, err error) {
+func checkMsg(msg *Msg, checkSts, isNoWait bool) (usrMsg bool, err error) {
 	// Assume user message
 	usrMsg = true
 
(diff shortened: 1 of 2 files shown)


derekcollison:
@jnmoyne use case was batch == 2 IIRC?

js.go
```diff
@@ -2558,7 +2573,8 @@ func (sub *Subscription) Fetch(batch int, opts ...PullOpt) ([]*Msg, error) {
 		// are no messages.
 		msg, err = sub.nextMsgWithContext(ctx, true, false)
 		if err != nil {
-			if err == errNoMessages {
+			// We skip both 404 and any type of 408 errors at this stage.
+			if err == errNoMessages || err == errRequestsPending {
```
kozlovic, line 2577: This will have no effect. We just called sub.nextMsgWithContext(), which is a low-level NATS call, but since we indicate that this is an internal pull call, it will return errNoMessages if there was no message and not go into a wait. You can check that function for details, but there is no way this function returns errRequestsPending since I don't see that you added that outside of checkMsg().

js.go
```diff
@@ -2448,11 +2456,18 @@ func checkMsg(msg *Msg, checkSts bool) (usrMsg bool, err error) {
...
 		err = errNoMessages
 	case reqTimeoutSts:
-		// Older servers may send a 408 when a request in the server was expired
-		// and interest is still found, which will be the case for our
-		// implementation. Regardless, ignore 408 errors until receiving at least
-		// one message.
-		err = ErrTimeout
+		// In case of a fetch request with no wait request and expires time,
+		// it will be a 408 error but with a Requests Pending description.
+		desc := msg.Header.Get(descrHdr)
```
kozlovic, line 2461: From looking at server code, there are several possible descriptions for 408 (so far..): "Interest Expired", "Requests Pending", "Request Canceled" and "Request Timeout".
We need to think if we need to behave any differently for those error status or not. I am not sure. Depending on that answer, we can decide if we need to distinguish description or not.. (again, there are more than that one, and some may be added in the future).

kozlovic (commented):
We probably need to rationalize what those 408 various responses mean from the server perspective and how to react to those. I am not clear on that.

js.go
wallyqs: thanks for the review, removed the check since it had no effect

js.go
wallyqs: Not sure about the rest of the status at the moment but I think one way to simplify this could be to refactor the code to skip 408s unless making the initial expires + no wait batch request.

derekcollison: All of those mean the request is no longer valid and no additional messages will be sent for that request.

PR #972, opened 2022-05-05
spelling fix & dictionary updates

philpennock:
Fix a spelling mistake which made it through to go docs.

gospel 1.13.0 is now out, and it added features which let us remove some old words (eg, pluralizations of types we define), but it also errored out on seeing a comment in .words which used to be acceptable.  So remove comments from .words and introduce .words.readme instead.

We lose the section-introducer comments.  Which is a shame.

### DIFF
--- a/.words
+++ b/.words
@@ -1,26 +1,9 @@
 1
-## ^^^ size estimate, just needs to be non-zero
-##
-## The .words file is used by gospel (v1.2+), which wraps the Hunspell libraries but populates the dictionary with identifiers from the Go source.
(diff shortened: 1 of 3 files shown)


kozlovic (approved):
LGTM

philpennock:
For clarity for future code archaeology, I was misleading above because I wrote in haste.  gospel 1.13.0 is out.  There have been many releases since I last touched this repo, and one of the _earlier_ ones included the fixes which let me remove some words.  Really the relevant point was just to document which version this commit was known to be compatible with.

Issue #859, opened 2021-11-09
Websockets add path for proxy 


bfoxstudio:
client connect from Inernet o  wss://host.com:port1/proxy_nginx_path
nginx redirect to ws://host_locla_company_network:port2/proxy_path

I made a mistake. we need fix nats.go client library
https://github.com/nats-io/nats.go/blob/main/ws.go#L563
	**ustr += u.Path**
...

kozlovic:
@bfoxstudio Again, since URLs that are discovered will not have the path, I wonder if this should be an option and then added to the http's URL (unless of course the provided URL has already a path, then use that one instead). What do you think?

runner-mei:
I had the same problem!  I fix this problem by:



````golang
	// For http request, we need the passed URL to contain either http or https scheme.
	scheme := "http"
	if tlsRequired {
		scheme = "https"
	}
	ustr := fmt.Sprintf("%s://%s", scheme, u.Host)
	u, err = url.Parse(ustr)
	if err != nil {
		return err
	}
````
change to

`````golang
	// For http request, we need the passed URL to contain either http or https scheme.
	u.Scheme = "http"
	if tlsRequired {
		u.Scheme = "https"
	}
`````

**Comment by runner-mei** [comment]
@bfoxstudio I deplpy nats-server behind the nginx;

nginx will route request to nats-server by the url path.

**Comment by shlomi-wexler-apolicy** [comment]
> I had the same problem! I fix this problem by:
> 
> ```go
> 	// For http request, we need the passed URL to contain either http or https scheme.
> 	scheme := "http"
> 	if tlsRequired {
> 		scheme = "https"
> 	}
> 	ustr := fmt.Sprintf("%s://%s", scheme, u.Host)
> 	u, err = url.Parse(ustr)
> 	if err != nil {
> 		return err
> 	}
> ```
> 
> change to
> 
> ```go
> 	// For http request, we need the passed URL to contain either http or https scheme.
> 	u.Scheme = "http"
> 	if tlsRequired {
> 		u.Scheme = "https"
> 	}
> ```

We encountered the same issue and I think there are a few lines of codes missing in the solution. This is the minor change I ended up doing to include the url path:

```
scheme := "http"
if tlsRequired {
	scheme = "https"
}
ustr := fmt.Sprintf("%s://%s%s", scheme, u.Host, u.Path)
u, err = url.Parse(ustr)
if err != nil {
	return err
} 
```

kozlovic:
[...]

PR #974, opened 2022-05-10
Add WS ProxyPath option

ido-gold-apolicy:
This is a PR to fix the issue described here: https://github.com/nats-io/nats.go/issues/859

### DIFF
--- a/ws_test.go
+++ b/ws_test.go
@@ -1113,8 +1113,7 @@ func TestWSNoDeadlockOnAuthFailure(t *testing.T) {
 }
 
 func TestWSProxyPath(t *testing.T) {
-	const proxyPath = "/proxy1"
-	var proxyCalled bool
+	const proxyPath = "proxy1"
 
 	// Listen to a random port
 	l, err := net.Listen("tcp", ":0")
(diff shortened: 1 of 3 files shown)


ws_test.go
```diff
@@ -1108,3 +1111,36 @@ func TestWSNoDeadlockOnAuthFailure(t *testing.T) {
 
 	tm.Stop()
 }
+
+func TestWSProxyPath(t *testing.T) {
+	const proxyPath = "/proxy1"
```
kozlovic, line 1116: Instead of commenting line by line, I feel it is easier to show you how I would like the test to look like. The main difference with yours is that we are going to check with a prefix "/" and without, and use a go channel to check that the proxy was invoked. (note that you don't need to run the NATS Server since the connection is actually not happening)
```
        const proxyPath = "proxy1"

	// Listen to a random port
	l, err := net.Listen("tcp", ":0")
	if err != nil {
		t.Fatalf("Error in listen: %v", err)
	}
	defer l.Close()

	proxyPort := l.Addr().(*net.TCPAddr).Port

	ch := make(chan struct{}, 1)
	proxySrv := &http.Server{
		Handler: http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
			if r.URL.Path == "/"+proxyPath {
				ch <- struct{}{}
			}
		}),
	}
	defer proxySrv.Shutdown(context.Background())
	go proxySrv.Serve(l)

	for _, test := range []struct {
		name string
		path string
	}{
		{"without slash", proxyPath},
		{"with slash", "/" + proxyPath},
	} {
		t.Run(test.name, func(t *testing.T) {
			url := fmt.Sprintf("ws://127.0.0.1:%d", proxyPort)
			nc, err := Connect(url, ProxyPath(test.path))
			if err == nil {
				nc.Close()
				t.Fatal("Did not expect to connect")
			}
			select {
			case <-ch:
				// OK:
			case <-time.After(time.Second):
				t.Fatal("Proxy was not reached")
			}
		})
	}
```

kozlovic (changes requested):
Thank you for the contribution. I would make some changes to the test.

PR #980, opened 2022-05-25
[ADDED] Stream RePublish and some ConsumerConfig new fields

kozlovic:
Namely for pull consumers: ability to override the replica count or storage type:
```go
// Generally inherited by parent stream and other markers, now can be configured directly.
Replicas int `json:"num_replicas"`
// Force memory storage.
MemoryStorage bool `json:"mem_storage,omitempty"`
```

For the stream, this new StreamConfig option:
```go
// Allow republish of the message after being sequenced and stored.
RePublish *SubjectMapping `json:"republish,omitempty"`
```
Where SubjectMapping is:
```go
// SubjectMapping allows a source subject to be mapped to a destination subject for republishing.
type SubjectMapping struct {
	Source      string `json:"src,omitempty"`
	Destination string `json:"dest"`
}
```

### DIFF
--- a/js.go
+++ b/js.go
@@ -373,6 +373,13 @@ const (
 	MsgRollup              = "Nats-Rollup"
 )
 
+// Headers for republished messages.
+const (
+	JSStream       = "Nats-Stream"
+	JSSequence     = "Nats-Sequence"
+	JSLastSequence = "Nats-Last-Sequence"
+)
+
 // MsgSize is a header that will be part of a consumer's delivered message if HeadersOnly requested.
 const MsgSize = "Nats-Msg-Size"
 
@@ -961,8 +968,6 @@ func (d nakDelay) configureAck(opts *ackOpts) error {
 type ConsumerConfig struct {
 	Durable         string          `json:"durable_name,omitempty"`
 	Description     string          `json:"description,omitempty"`
-	DeliverSubject  string          `json:"deliver_subject,omitempty"`
-	DeliverGroup    string          `json:"deliver_group,omitempty"`
 	DeliverPolicy   DeliverPolicy   `json:"deliver_policy"`
 	OptStartSeq     uint64          `json:"opt_start_seq,omitempty"`
 	OptStartTime    *time.Time      `json:"opt_start_time,omitempty"`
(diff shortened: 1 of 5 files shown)


wallyqs (approved):
LGTM

js.go
```diff
@@ -981,11 +986,21 @@ type ConsumerConfig struct {
 	HeadersOnly     bool            `json:"headers_only,omitempty"`
 
 	// Pull based options.
-	MaxRequestBatch   int           `json:"max_batch,omitempty"`
-	MaxRequestExpires time.Duration `json:"max_expires,omitempty"`
+	MaxRequestBatch    int           `json:"max_batch,omitempty"`
+	MaxRequestExpires  time.Duration `json:"max_expires,omitempty"`
+	MaxRequestMaxBytes int           `json:"max_bytes,omitempty"`
```
derekcollison, line 991: Wonder if we should leave this one out?

jsm.go
```diff
@@ -100,6 +100,15 @@ type StreamConfig struct {
...
 	DenyPurge         bool            `json:"deny_purge,omitempty"`
 	AllowRollup       bool            `json:"allow_rollup_hdrs,omitempty"`
+
+	// Allow republish of the message after being sequenced and stored.
+	RePublish *SubjectMapping `json:"republish,omitempty"`
+}
+
+// SubjectMapping allows a source subject to be mapped to a destination subject for republishing.
+type SubjectMapping struct {
+	Source      string `json:"src,omitempty"`
```
derekcollison, line 110: I should make the change in server to not require and default to all here.

js.go
kozlovic: True. Without ability to send a pull request with max bytes, there is little value. I can remove.

jsm.go
kozlovic: Ok. What does that mean for this PR though?

derekcollison: More a comment to myself. ;)

js.go
kozlovic: Done. Sorry for the "git push force", it is because I wanted to remove the field from the git comment itself.

derekcollison (approved):
LGTM

PR #987, opened 2022-06-02
Fix sending ack when AckNonePolicy is set

### LINKED ISSUE #986: Subscribe sends ack even if AckPolicy is set to "none"
Currently, there is no way to prevent sending ack for consumers with `AckNonePolicy`, other than using `ManualAck()`. Auto Ack should not publish acknowledgement if `AckPolicy` is set to `none`.

piotrpio:
This fixes #986


derekcollison (approved):
LGTM!

aricart (approved):
LGTM

wallyqs (approved):
LGTM

PR #989, opened 2022-06-06
Ordered will force R1 and memory storage

derekcollison:
For ordered push semantics no need to have file based, so force R1 and memory based.


piotrpio (approved):
LGTM

PR #991, opened 2022-06-07
Align RePublish with server and add HeadersOnly, also add in AllowDirect option

derekcollison:

### DIFF
--- a/jsm.go
+++ b/jsm.go
@@ -1,4 +1,4 @@
-// Copyright 2021 The NATS Authors
+// Copyright 2021-2022 The NATS Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


jsm.go
```diff
@@ -100,15 +100,18 @@ type StreamConfig struct {
 	DenyDelete        bool            `json:"deny_delete,omitempty"`
 	DenyPurge         bool            `json:"deny_purge,omitempty"`
 	AllowRollup       bool            `json:"allow_rollup_hdrs,omitempty"`
+	AllowDirect       bool            `json:"allow_direct,omitempty"`
 
 	// Allow republish of the message after being sequenced and stored.
-	RePublish *SubjectMapping `json:"republish,omitempty"`
+	RePublish *RePublish `json:"republish,omitempty"`
```
piotrpio, line 106: Do we want to introduce breaking change here for the sake of changing the struct name?

jsm.go
```diff
@@ -100,15 +100,18 @@ type StreamConfig struct {
 	DenyDelete        bool            `json:"deny_delete,omitempty"`
 	DenyPurge         bool            `json:"deny_purge,omitempty"`
 	AllowRollup       bool            `json:"allow_rollup_hdrs,omitempty"`
+	AllowDirect       bool            `json:"allow_direct,omitempty"`
```
piotrpio, line 103: I wonder whether exposing this option for a stream won't give the impression that we already support this feature in the client (and from what I see, we don't). Maybe it would be better if it goes together with actually sending direct requests + checking server compatibility?

jsm.go
derekcollison: Was not released, was experimental so yes we can do that at this point I believe.

jsm.go
derekcollison: Possibly, but you can do the direct gets with nc.Request() with this enabled in the server..

jsm.go
piotrpio: That's fair,  thanks.

PR #993, opened 2022-06-07
Use native time.Time.Equal method for equality check

### LINKED ISSUE #992: Equality check on consumer time values use pointers
https://github.com/nats-io/nats.go/blob/fe748b369b4d29dd402e2692b4b586f15db9d144/js.go#L1265

This performs an equality check on time using pointers rather than dereferencing and using the native `.Equal(..)` method. See https://go.dev/play/p/TiNHPSpN9Da for example.

bruth:
Fixes #992


piotrpio (approved):
LGTM - thanks for the contribution!

PR #994, opened 2022-06-07
Add additional note to JetStream.PullSubscribe on durable semantics

bruth:
Based on user confusion on what the `durable` argument should be for an ephemeral consumer. The `Subscribe` method notes calls out `nats.Durable`, but not the argument itself.


wallyqs (approved):

derekcollison (approved):
LGTM

PR #996, opened 2022-06-10
Expose TLS connection state when possible

ripienaar:
I would like to be able to report things like ciphers,
tls or not, verified or not etc in nats account info
where its shows other connection properties but had
no way to get at this information

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -168,6 +168,7 @@ var (
 	ErrStreamNameAlreadyInUse       = errors.New("nats: stream name already in use")
 	ErrMaxConnectionsExceeded       = errors.New("nats: server maximum connections exceeded")
 	ErrBadRequest                   = errors.New("nats: bad request")
+	ErrConnectionNotTLS             = errors.New("nats: connection is not tls")
 )
 
 func init() {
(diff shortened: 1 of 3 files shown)


nats.go
```diff
@@ -1807,6 +1808,24 @@ func (nc *Conn) makeTLSConn() error {
 	return nil
 }
 
+// TLSConnectionState retrieves the state of the TLS connection to the server
+func (nc *Conn) TLSConnectionState() (tls.ConnectionState, error) {
```
derekcollison, line 1812: Should this be passed by value here? Is that the recommended approach?

derekcollison: Just looked seems so.

derekcollison (approved):
LGTM

nats.go
ripienaar: yeah i find it awkward, i considered calling the go internal and returning a pointer which seemed a bit more idiomatic. So I am not sure if this is best, happy to do that pointer dance if you prefer

derekcollison: Their API does same so we are good IMO.

PR #998, opened 2022-06-13
Change DefaultSubPendingMsgsLimit comment to reflect actual value

bruth:
This is just for consistency.. I don't seeing it help with _understanding_ per se, but it could help in the edge case situation where the pending number is being reported and its greater than 512k which could be confusing.

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -4446,8 +4446,8 @@ func (s *Subscription) ClearMaxPending() error {
 
 // Pending Limits
 const (
-	// DefaultSubPendingMsgsLimit will be 524,288 msgs.
-	DefaultSubPendingMsgsLimit = 512 * 1024
+	// DefaultSubPendingMsgsLimit will be 500k msgs.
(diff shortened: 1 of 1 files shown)


wallyqs (approved):
LGTM

nats.go
```diff
@@ -4446,7 +4446,7 @@ func (s *Subscription) ClearMaxPending() error {
 
 // Pending Limits
 const (
-	// DefaultSubPendingMsgsLimit will be 512k msgs.
+	// DefaultSubPendingMsgsLimit will be 524,288 msgs.
```
derekcollison, line 4449: Why not 512Kb?

bruth: I guess I assumed using Kb as unit in a non-storage context was not common. Happy to change if you think otherwise.. but when I see a `b` I immediately associate storage rather than a discrete number.

bruth: Added another commit.. like so?

piotrpio (approved):
LGTM

Issue #1003, opened 2022-06-24
Can't get large file


ZergsLaw:
Hello guys!

I have a problem, if I try to download large file from ObjectStore by nats.go lib, it's not working, because you have added a hardcode timeout
```r.SetReadDeadline(time.Now().Add(2 * time.Second))```

1) nats-server version - 2.8.2
2) nats.go version - v1.16.0

Platform:
linux/macos/std-docker

#### Steps or code to reproduce the issue:
```
fVideo, err := os.Open(movFilePath)
assert.NoError(err)
t.Cleanup(func() { assert.NoError(fVideo.Close()) })

id := uuid.Must(uuid.NewV4())
meta := &nats.ObjectMeta{
	Name: id.String(),
}

store, err := conn.JetStream.ObjectStore("post_attachments")
assert.NoError(err)

_, err = store.Put(meta, fVideo, nats.Context(ctx))
assert.NoError(err)

file, err := store.Get(id.String(), nats.Context(ctx))
assert.NoError(err)

buf, err := io.ReadAll(file)
assert.NoError(err)
```
#### Expected result:
getting file
#### Actual result:
getting error

derekcollison:
Is this reproduce-able with 2.8.4 server? Does increasing the timeout help solve the problem and allow the test to pass above?

Also how large is the asset and how far away (RTT) is the client application from the NATS system and the store asset?

Thanks..

ZergsLaw:
Hi, yeah, on 2.8.4 same situation

I made fix, and it's working, but I'm not sure, mb my fix isn't correct

ZergsLaw:
https://github.com/nats-io/nats.go/pull/1005 - I'm not sure, mb it is not correct, but with fix it's working

ZergsLaw:
But for file with size 3gb+ not working anyway :(

derekcollison:
So for 3Gb+ your fix is not working regardless yes?

What is RTT between your app and the asset? `nats rtt`?

derekcollison:
What in the example above is the context? Can you share the code that creates that?

ZergsLaw:
Yeah, for 3Gb+ is not working anyway, I'm getting lock

What do you mean? (about RTT), I'm using single node nats with js (I can share test cfg)

about ctx

If I set ctx with my fix it's working (only for not big file size less than 3Gb)

```ctx, cancel := context.WithTimeout(context.Background, time.Second * 10)```

derekcollison:
[...]

PR #1004, opened 2022-06-24
bug: fix hardcode deadline

ZergsLaw:
#1003


derekcollison (commented):
I noticed this has been closed. Just curious as to why?

PR #1005, opened 2022-06-24
bug: fix hardcode deadline

ZergsLaw:
#1003

### DIFF
--- a/object.go
+++ b/object.go
@@ -965,7 +965,16 @@ func (o *objResult) Read(p []byte) (n int, err error) {
 	}
 
 	r := o.r.(net.Conn)
-	r.SetReadDeadline(time.Now().Add(2 * time.Second))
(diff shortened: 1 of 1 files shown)


object.go
```diff
@@ -965,7 +965,16 @@ func (o *objResult) Read(p []byte) (n int, err error) {
 	}
 
 	r := o.r.(net.Conn)
-	r.SetReadDeadline(time.Now().Add(2 * time.Second))
+	deadline, ok := o.ctx.Deadline()
+	if !ok {
+		deadline = time.Now().Add(2 * time.Second)
+	}
+
+	err = r.SetReadDeadline(deadline)
```
derekcollison, line 969: This is an internal pipe IIRC.. So I wonder if you are getting a slow consumer internally in your app from the NATS server you are connected to, or the NATS.go client is deeming this a slow consumer?

PR #1010, opened 2022-06-29
Added support for Subject reporting on StreamInfo requests

### LINKED ISSUE #1009: StreamInfo request doesn't allow reporting of subjects
- `StreamInfoRequest` needs `SubjectsFilter` (string)
- `StreamState` needs `NumSubjects` (uint64) and `Subjects` (map[string]uint64)

aricart:
[FEAT] added `SubjectFilter` to `StreamInfoRequest`. The field takes a subject/wildcard that instruct the request to provide information on matching subjects in `StreamState`.
[FEAT] added `NumSubjects` to `StreamState` enabling report for the number of subjects in the stream
[FEAT] added `Subjects` to `StreamState` a map of subject to the number of messages with the subject which gets populated if the `StreamInfoRequest` specifies a `SubjectFilter`

FIX #1009

### DIFF
--- a/jsm.go
+++ b/jsm.go
@@ -638,9 +638,12 @@ func (js *js) AddStream(cfg *StreamConfig, opts ...JSOpt) (*StreamInfo, error) {
 }
 
 type (
-	// StreamInfoRequest contains additional option to return details about messages deleted from a stream
+	// StreamInfoRequest contains additional option to return
 	StreamInfoRequest struct {
+		// DeletedDetails when true includes information about deleted messages
 		DeletedDetails bool `json:"deleted_details,omitempty"`
+		// SubjectsFilter when set, returns information on the matched subjects
+		SubjectsFilter string `json:"subjects_filter,omitempty"`
 	}
 	streamInfoResponse = streamCreateResponse
 )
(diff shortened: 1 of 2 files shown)


jsm.go
```diff
@@ -638,9 +638,12 @@ func (js *js) AddStream(cfg *StreamConfig, opts ...JSOpt) (*StreamInfo, error) {
 }
 
 type (
-	// StreamInfoRequest contains additional option to return details about messages deleted from a stream
+	// StreamInfoRequest contains additional option to return
 	StreamInfoRequest struct {
+		// DeletedDetails when true includes information about deleted messages
 		DeletedDetails bool `json:"deleted_details,omitempty"`
+		// SubjectsFilter When set, returns information on the matched subjects - can be a wildcard or wild-carded subject
```
derekcollison, line 645: when set (lower) and wildcarded

go.mod
```diff
@@ -3,6 +3,9 @@ module github.com/nats-io/nats.go
 go 1.16
 
 require (
+	github.com/golang/protobuf v1.5.2
```
kozlovic, line 6: Undo changes to go.mod/sum. We use a go_test.mod for tests.

jsm.go
```diff
@@ -638,9 +638,12 @@ func (js *js) AddStream(cfg *StreamConfig, opts ...JSOpt) (*StreamInfo, error) {
 }
 
 type (
-	// StreamInfoRequest contains additional option to return details about messages deleted from a stream
+	// StreamInfoRequest contains additional option to return
 	StreamInfoRequest struct {
+		// DeletedDetails when true includes information about deleted messages
 		DeletedDetails bool `json:"deleted_details,omitempty"`
+		// SubjectsFilter when set, returns information on the matched subjects - can be a wildcard or wildcarded subject
```
kozlovic, line 645: Not sure about the difference between "a wildcard or a wildcarded subject"?

test/js_test.go
```diff
@@ -1702,6 +1702,42 @@ func TestPurgeStream(t *testing.T) {
...
+		Subjects: []string{"foo.*"},
+	})
+	if err != nil {
+		t.Fatalf("Unexpected error: %v", err)
+	}
+
+	if _, err := js.Publish("foo.A", []byte("")); err != nil {
+		t.Fatalf("Unexpected error during publish: %v", err)
+	}
+
```
kozlovic, line 1723: Having another publish on foo.B would help ensure that the library sends the proper filter, or alternatively, listen on the request subject to check that we are doing the right thing.
If you do publish on foo.B, you will need to adjust the rest of the test for the counts (NumSubjects should be 2), etc..

kozlovic (changes requested):

piotrpio (changes requested):

jsm.go
aricart: ">" vs "a.>" - I could just say subject

test/js_test.go
aricart: I committed this by mistake along with the gomods. This was testing to see something Derek mentioned that I understood incorrectly.

jsm.go
kozlovic: ">" is still a subject, so I would not make the distinction.

test/js_test.go
```diff
@@ -1702,6 +1702,42 @@ func TestPurgeStream(t *testing.T) {
 	}
 }
 
+func TestStreamInfoSubjectInfo(t *testing.T) {
```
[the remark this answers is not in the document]
aricart: Renaming

piotrpio (approved):

piotrpio:
LGTM

kozlovic (approved):
LGTM

PR #1011, opened 2022-07-01
Consistent error value of context timeout on subscription Fetch() interface.

wdhongtw:
When using [Subscription.Fetch()](https://github.com/nats-io/nats.go/blob/main/js.go#L2510) with  `nats.Context(ctx)` option,
if context timeout receiving any message, `Fetch` will return error.

There are two possible value of error here, `nats.ErrTimeout` and `context.DeadlineExceeded`, based on some timing in the process/gorouting. If **context timeout before subscription object trying to fetch message**, `nats.ErrTimeout` is returned. Otherwise, `context.DeadlineExceeded` is returned.

- <https://github.com/nats-io/nats.go/blob/main/js.go#L2580>
- <https://github.com/nats-io/nats.go/blob/main/js.go#L2595>

This PR make the error value of fetch timeout predicable, if Fetch is called with `nats.Context(ctx)`, timeout always return
`context.DeadlineExceeded`; if called with `nats.MaxWait(timeout)`, timeout always return `nats.ErrTimeout`.


js_test.go
```diff
@@ -939,6 +941,21 @@ func TestJetStreamExpiredPullRequests(t *testing.T) {
...
 		}
 	}
+
+	t.Run("MaxWait timeout should return nats error", func(t *testing.T) {
+		_, err := sub.Fetch(1, MaxWait(12*time.Nanosecond))
+		if !errors.Is(err, ErrTimeout) {
+			t.Fatalf("Expect ErrTimeout, got err=%#v", err)
+		}
+	})
+	t.Run("Context timeout should return context error", func(t *testing.T) {
```
wallyqs, line 951: I think this test case already covered here: https://github.com/nats-io/nats.go/blob/main/test/js_test.go#L6719-L6726

js_test.go
```diff
@@ -939,6 +941,21 @@ func TestJetStreamExpiredPullRequests(t *testing.T) {
 			t.Fatalf("Expected error and wait for 250ms, got err=%v and dur=%v", err, dur)
 		}
 	}
+
+	t.Run("MaxWait timeout should return nats error", func(t *testing.T) {
```
wallyqs, line 945: Could you move this test block to be as part of this other test? https://github.com/nats-io/nats.go/blob/main/test/js_test.go#L6869

wallyqs (changes requested):
Thanks for the fix! Just one request to move the tests to be under test/js_test.go under the `TestJetStreamPullSubscribeFetchContext` tests.

js_test.go
wdhongtw: Sure :D

js_test.go
wdhongtw: :D

wallyqs (approved):
LGTM!

wdhongtw:
I accidentally leave a empty line in `js_test.go`, so I force push the branch again.  :D

PR #1013, opened 2022-07-12
docs: mention defaults in option docs

costela:
Currently finding out the defaults when using the code is a bit cumbersome: most IDEs will give you the docs for the option (e.g. `nats.MaxReconnects`), which currently doesn't mention the defaults. So you have to go to the full docs and click around before getting to the constants.

I propose adding the current defaults directly to the method/struct docs, so they are easier to find when coding, without "breaking the flow".

This has the downside of needing more care to be kept up-to-date, but I would argue the trade-off is worth it.

WDYT?


kozlovic (approved):
LGTM. Thanks!

PR #1015, opened 2022-07-14
flusher errors now do not fail silently

GeorgeEngland:
The flusher could fail the write and the user would not be able to handle nor get alerted to this error.

Although this is still poor since the error occurs asynchronously, at least the user will be alerted to the error having happened.

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -3095,6 +3095,9 @@ func (nc *Conn) flusher() {
 				if nc.err == nil {
 					nc.err = err
 				}
+				if nc.Opts.AsyncErrorCB != nil {
+					nc.ach.push(func() { nc.Opts.AsyncErrorCB(nc, nil, err) })
+				}
 			}
 		}
 		nc.mu.Unlock()


kozlovic (approved):
Thank you for the contribution.

My first reaction would be: is this really necessary? The reason we don't do any reporting here is because if there is a failure to write to the socket, it is likely that the error will be detected in the readloop (or lack of heartbeats, etc..) and those errors are already handled.

That being said, since I don't see this change breaking any existing test, I would think that it is ok to approve the change. Will see if other maintainers agree with that or not.

wallyqs (approved):
LGTM

GeorgeEngland:
> Thank you for the contribution.
> 
> My first reaction would be: is this really necessary? The reason we don't do any reporting here is because if there is a failure to write to the socket, it is likely that the error will be detected in the readloop (or lack of heartbeats, etc..) and those errors are already handled.
> 
> That being said, since I don't see this change breaking any existing test, I would think that it is ok to approve the change. Will see if other maintainers agree with that or not.

What about a full TCP buffer?  The default timeout of the writer is 2s, so i assume a full buffer could cause a silent failure?

kozlovic:
I don't think we set any deadline so it would just block, but not fail.

GeorgeEngland:
@kozlovic Yes I think you do - the Write method for the timeoutWriter uses a timeout deadline.

nats.go:L5345 

```
// Write implements the io.Writer interface.
func (tw *timeoutWriter) Write(p []byte) (int, error) {
	if tw.err != nil {
		return 0, tw.err
	}

	var n int
	tw.conn.SetWriteDeadline(time.Now().Add(tw.timeout))
	n, tw.err = tw.conn.Write(p)
	tw.conn.SetWriteDeadline(time.Time{})
	return n, tw.err
}
```

GeorgeEngland:
@kozlovic Secondly, the underlying connection also has a timeout of 2 seconds by default

L2014 nats.go:

```
 func (nc *Conn) processConnectInit() error {

	// Set our deadline for the whole connect process
	nc.conn.SetDeadline(time.Now().Add(nc.Opts.Timeout))
	defer nc.conn.SetDeadline(time.Time{})
```
the nc.Opts.Timeout is set to 2 seconds if it is set to 0

PR #1018, opened 2022-07-20
Add simplified JetStream API with pull consumer

piotrpio:
This is a simplified API for JetStream. See `jetstream/README.md` for details.


ripienaar:
Be great if this was also aware of the jetstream error codes

piotrpio:
@ripienaar I added printing error codes together with error description I'm handling a handful of JetStream error codes (see `errors.go`) to be able to return a specific error (e.g. `ErrStreamNotFound`). The rest of the API errors I think we can leave as they are (just print error code + description and return), as it is usually pretty hard to recover from those errors anyway. I also made `APIError` exported so a user can check the error codes with `errors.As()` if necessary. 
What do you think?

ripienaar:
I think just having access to the error code is enough along with some helpers as you suggest. 

The errors can be generated as there is a json file in server repo but probsbly not suitable for this repo. 

I think @aricart has some ideas and expose it in the node code. 

For me just having access to the code is enough.

js.go
```diff
@@ -27,6 +27,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/nats-io/nats.go/parser"
```
wallyqs, line 30: how about moving this one to be part of an `internal/parser` package?

jsv2/kv/kv.go
```diff
@@ -0,0 +1,434 @@
...
+}
+
+// Errors
+var (
+	ErrKeyValueConfigRequired = errors.New("nats: config required")
+	ErrInvalidBucketName      = errors.New("nats: invalid bucket name")
+	ErrInvalidKey             = errors.New("nats: invalid key")
+	ErrBucketNotFound         = errors.New("nats: bucket not found")
+	ErrBadBucket              = errors.New("nats: bucket not valid key-value store")
+	ErrKeyNotFound            = errors.New("nats: key not found")
```
wallyqs, line 234: This ErrKeyNotFound might benefit from some error wrapping too, maybe have something like a `KeyValueError` struct and mention that `errors.Is` is preferred for some of the new checks.

jsv2/kv/bucket.go
```diff
@@ -0,0 +1,398 @@
...
+	} else {
+		m, err = kv.stream.GetMsg(ctx, revision)
+		if err == nil && m.Subject != b.String() {
+			return nil, ErrKeyNotFound
+		}
+	}
+
+	if err != nil {
+		if err == jetstream.ErrMsgNotFound {
+			err = ErrKeyNotFound
```
wallyqs, line 197: We often do this for some errors:
```
if err == jetstream.ErrMsgNotFound {
			err = ErrKeyNotFound
```
I think maybe a sign that error should have ben wrapped instead?  Think it would help have `errors.Is(err, ErrKeyNotFound)` but then be able to distinguish from reason being that it was because key was deleted, got 404 from API or because revision number value does not match the retrieved so transformed into key not found error (like a few lines above).

piotrpio: Makes a ton of sense. I'll scan the `jsv2` codebase and see where it makes sense. BTW, the errors implementation in this PR will change quite a bit, as right now I don't yet have a common type to represent both client and API errors (like we have in `js.go`).

jsv2/kv/kv.go
piotrpio: I decided to add proper KV support in a separate PR, to be able to make it at least a little bit more granular.

jsv2/jetstream/errors.go
```diff
@@ -0,0 +1,189 @@
...
+	// ErrTooManyStalledMsgs is returned when too many outstanding async messages are waiting for ack
+	ErrTooManyStalledMsgs = &jsError{message: "stalled with too many outstanding async published messages"}
+
+	// ErrInvalidOption is returned when there is a collision between options
+	ErrInvalidOption = &jsError{message: "invalid jetstream option"}
+)
+
+// Error prints the JetStream API error code and description
+func (e *APIError) Error() string {
+	return fmt.Sprintf("nats: API error %d: %s", e.ErrorCode, e.Description)
```
wallyqs, line 150: Thinking now that maybe we should add the `Code` as well here besides the error code since that can give more of the semantic meaning right away, like a bad KV update becomes would be a bad request with `code=400` and  `err_code=10071 description='wrong last sequence: 1'`

derekcollison:
Might be good to have you and @wallyqs  and me do a 1hr review of this PR possibly next week if you think its ready.

piotrpio:
@derekcollison sure, let me set something up

examples/jsv2/js-ordered-messages/main.go
```diff
@@ -0,0 +1,82 @@
...
+	s, err := js.CreateStream(ctx, jetstream.StreamConfig{
+		Name:     "TEST_STREAM",
+		Subjects: []string{"FOO.*"},
+	})
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	cons, err := s.OrderedConsumer(ctx, jetstream.OrderedConsumerConfig{
+		MaxResetAttempts: 5,
```
aricart, line 48: not so sure about this - what if I have an ordered consumer that is long-lived?

aricart (approved):
LGTM

aricart (approved):
LGTM

examples/jsv2/js-ordered-messages/main.go
aricart: Yes I understood that - but my point is that, unless the recreates are bounded to some sort of time fail, it will eventually hit that number and then it goes, even if these recreations span weeks in between.

piotrpio: That's not the case, as the recreate counter is created per consumer reset:
```go
for i := 0; ; i++ {
			if c.cfg.MaxResetAttempts > 0 && i == c.cfg.MaxResetAttempts {
				return fmt.Errorf("%w: maximum number of delete attempts reached: %s", ErrOrderedConsumerReset, err)
			}
```

So in that case, only if 5 consecutive retires within a single reset fail, we fail.

tpihl:
Just a question. I didn’t see anything in readme about subject filtering in consumer (or even better, multiple adjustable subject filtering). 

Also, what’s the idea to consume while only knowing the publish subject?

and lastly, how will this support/enhance access rights (between accounts for example) to allow someone to consume a filtered part of a stream (so they can create the consumer, modify it, remove it, all while only consume that part of the stream. Will this now be feasible?

T

piotrpio:
Hey @tpihl, thanks for asking

>  I didn’t see anything in readme about subject filtering in consumer (or even better, multiple adjustable subject filtering).

You would apply subject filters in `ConsumerConfig` when creating the consumer. Multiple subject filters will be added in the next minor `nats-server` release, therefore. Until then, for single filter subjects you will have to provided a `FilterSubjects` slice with 1 element. We removed singular `FilterSubject` field  from the config as it will be deprecated in favor of more general `FilterSubjects` approach with any number of filters.
```go
cons, err := s.AddConsumer(ctx, jetstream.ConsumerConfig{
	Name:               "cons",
	FilterSubjects: []string{"FOO.A"},
})
```

> Also, what’s the idea to consume while only knowing the publish subject?

Hm, right now you do have to know the stream name to create/get the consumer. This is actually required by NATS server and we wanted to have a more CRUD-like approach to managing streams/consumers, and consumers are always directly related to a stream.
That being said - it might be a good idea to add a `StreamNameBySubject()` method (this is actually what's used in `js.Subscribe()` in the old API) to enable easily getting a stream name based on the provided subject. This is what it looks like in the old implementation - https://github.com/nats-io/nats.go/blob/2c5733a5ace24a50f378740a119ecd63622d9ae1/jsm.go#L1533
Thanks for the idea!

> and lastly, how will this support/enhance access rights (between accounts for example) to allow someone to consume a filtered part of a stream (so they can create the consumer, modify it, remove it, all while only consume that part of the stream. Will this now be feasible?

So, if I understand your point correctly, you will be able to create a stream with interest on the exported subject from another account and simply use a consumer on that stream.

Again, thanks for these questions - I will update the `README.md` (and examples) with some details on subject filtering and I'll probably add a `StreamNameBySubject()` method in separate PR. Let me know if I answered your questions and if you need any more clarification.

jetstream/README.md
```diff
@@ -0,0 +1,525 @@
...
+
+    // Get 10 messages from the consumer
+    msgs, _ := c.Fetch(10)
+    var msg jetstream.Msg
+    for msg := range msgs.Messages() {
+        msg.Ack()
+        fmt.Printf("Received a JetStream message: %s\n", string(msg.Data()))
+    }
+    if msgs.Error() {
+        fmt.Println("Error duting Fetch(): ", msgs.Error())
```
wallyqs, line 96:
```suggestion
        fmt.Println("Error during Fetch(): ", msgs.Error())
```

wallyqs (approved):
LGTM

piotrpio:
Closing this PR in favor of https://github.com/nats-io/nats.go/pull/1273, which has cleaned up commit history

PR #1020, opened 2022-07-25
[ADDED] JetStream: nats.DirectGet() and nats.DirectGetNext() options

kozlovic:
The AllowDirect boolean must be set in the stream configuration.
When that is the case, then all servers (leader and replicas) are
part of a DQ group and can respond to a "DIRECT.GET" request.

The KeyValue store makes use of the direct get when connecting
to a stream with the AllowDirect option.


jsm.go
```diff
@@ -48,6 +49,14 @@ type JetStreamManager interface {
 	// GetMsg retrieves a raw stream message stored in JetStream by sequence number.
 	GetMsg(name string, seq uint64, opts ...JSOpt) (*RawStreamMsg, error)
```
derekcollison, line 50: Do we need to get the stream somewhere for this call? If so can we overload instead of introducing DirectGetMsg?

kozlovic: > Do we need to get the stream somewhere for this call? 

Sorry, I am not sure what you mean here.

We could add the DirectGetMsgRequest as a JSOpt (like we do for stream info and purge), but it still would look weird since in GetMsg() "seq" is a mandatory param, so one would have to pass seq==0 if passing a DirectGetMsgRequest option.

derekcollison: I think to make the call we have we at least need stream name. I was not sure if we had more such that we could remember the state of directAllowed. 

I might opt for here to have direct be a subOpt. KV is where it is critical path, so this not as important IMO.

kozlovic: > I think to make the call we have we at least need stream name. I was not sure if we had more such that we could remember the state of directAllowed.

I think I understand, but GetMsg() is not looking up the stream, so there is no state to remember if AllowDirect is set or not. The comment in server code was that if one does a "direct get" to a stream that is not set for allow direct, it would get a timeout or no responder.

>  I might opt for here to have direct be a subOpt.

But then it means something like: `GetMsg("myStream", 0, &nats.DirectGetMsgRequest{NextFor: "foo"})`. I could remove "Seq" from the option request and say that if it is provided in the GetMsg() call (2nd param) then use this one to construct the request? (would have to have a non exported request struct WITH the sequence for the marshal'ing though). That looks a bit ugly.

> KV is where it is critical path, so this not as important IMO.

Well, as you can see, in KV I do call the new API, so if it is GetMsg() with option, I would still need to pass the option to that call, so not sure if that makes any difference. Meaning KV is using the API...

derekcollison: We could add new opt, nats.DirectGet, or nats.AllowDirect or something, to make it simpler.

Would you push down that path or just keep GetMsgDirect?

kozlovic: Would be better to not have new API for sure.

So looking a bit more into it, I see that GetMsg() is actually doing this:
```
func (js *js) GetMsg(name string, seq uint64, opts ...JSOpt) (*RawStreamMsg, error) {
	return js.getMsg(name, &apiMsgGetRequest{Seq: seq}, opts...)
}

// Low level getMsg
func (js *js) getMsg(name string, mreq *apiMsgGetRequest, opts ...JSOpt) (*RawStreamMsg, error) {
	o, cancel, err := getJSContextOpts(js.opts, opts...)
...
```
and `apiMsgGetRequest` is this:
```
type apiMsgGetRequest struct {
	Seq     uint64 `json:"seq,omitempty"`
	LastFor string `json:"last_by_subj,omitempty"`
}
```
So, I could add the NextFor in there, but we would need an option `nats.DirectGet(nextFor string)`, but with `nextFor` being possibly empty, but still needed to trigger the "direct" get API, as opposed to regular one.

For GetLastMsg(), a simple `nats.DirectGet()` option would have been enough (like a switch option).

So we are not far, but not there yet I think... Maybe @wallyqs may have a good idea here?

derekcollison: The management path is not hot path, so I would say let's try to make an option. Under covers directGet() should exist and KV can use that directly..

aricart: On the JavaScript client this is implemented like this:

https://github.com/nats-io/nats.deno/blob/main/nats-base-client/jsmdirect_api.ts

The `DirectMsgRequest` type is really an union of the possible JSON option combinations - the compiler will reject if the shape of the JSON doesn't line up with one of the types:

export type DirectMsgRequest =
  | SeqMsgRequest
  | LastForMsgRequest
  | NextMsgRequest;

export type NextMsgRequest = {
  seq: number;
  next_by_subj: string;
};

export interface SeqMsgRequest {
  seq: number;
}

export interface LastForMsgRequest {
  "last_by_subj": string;
}

The permutation of the options complicates the API for other languages (like Go), I think instead should have the 3 function that takes the required arguments for the particular combination. That assigns a name for the query they are doing, and helps them out.

kv.go
```diff
@@ -399,26 +400,28 @@ func (js *js) CreateKeyValue(cfg *KeyValueConfig) (KeyValue, error) {
 		Duplicates:        duplicateWindow,
 		MaxMsgs:           -1,
 		MaxConsumers:      -1,
+		AllowDirect:       true,
```
ripienaar, line 403: I think this should be an option as we discussed during the server PRs there are concerns around read-after-write when this is in use.  Generally read-after-write less important than we first thought so probably it will be fine, but we need to provide the option for users to retain current usage patterns

derekcollison: Should not be an option imo from client perspective.

derekcollison:
Where are we on this, should we re-review?

jsm.go
wallyqs: I think that to avoid adding another API, we could make a new type that implements `configureJSContext` then that way it would be possible to call like this for the direct version:

```go
js.GetMsg("foo", 1, &nats.GetMsgRequest{Direct: true, NextFor: "bar"}) // Direct: false if not direct needed
```

kozlovic: But then for a direct and LastFor, where seq is not supposed to be present, it will be:
```
js.GetMsg("foo", 0, &nats.GetMsgRequest{Direct: true, LastFor: "bar"})
```
Notice that user will have to pass 0 for the sequence. If that is ok with everybody, we can do that.

wallyqs: Maybe for `LastFor` we suggest using `GetLastMsg` instead and only allow `&nats.GetMsgRequest{Direct: true}` option?

kozlovic:
@derekcollison I will update based on @wallyqs recommendations and ping here when it is ready for re-review.

kozlovic:
@derekcollison @wallyqs Added options. You can review individual commit to see what I changed, or maybe easier to review the total change (Go to Files Changed and make sure that you look at "Changes from all commits").

derekcollison (approved):
LGTM

wallyqs (approved):
LGTM

codegangsta:
🎉

PR #1022, opened 2022-07-29
[FIXED] Ephemeral PullConsumer's Fetch() would fail with "no responders"

bruth:
This was a bug @piotrpio found where passed consumer name (assuming a durable) was being used to define the "next message" subject. This resets it if an ephemeral was created using the server-generated name.

### DIFF
--- a/js.go
+++ b/js.go
@@ -1662,7 +1662,9 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync,
 			// after the AddConsumer returns.
 			if consumer == _EMPTY_ {
 				sub.jsi.consumer = info.Name
-				sub.jsi.nms = fmt.Sprintf(js.apiSubj(apiRequestNextT), stream, info.Name)
+				if isPullMode {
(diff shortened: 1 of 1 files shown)


js.go
```diff
@@ -1662,6 +1662,7 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync,
 			// after the AddConsumer returns.
 			if consumer == _EMPTY_ {
 				sub.jsi.consumer = info.Name
+				sub.jsi.nms = fmt.Sprintf(js.apiSubj(apiRequestNextT), stream, info.Name)
```
kozlovic, line 1665: We should probably do that under condition of `if isPullMode {}`.

kozlovic (changes requested):

js.go
wallyqs: +1 setting the next message subject here would not have side effects for push consumers but better to guard it to be under the check for pull mode just in case.

bruth: Added the guard.

wallyqs (approved):
LGTM

kozlovic (approved):
LGTM

bruth:
Thanks!

PR #1024, opened 2022-08-01
ConsumerInfo fix nil pointer dereference

Sergey-Belyakov:
fix nil pointer dereference when jsi is not inited


derekcollison (approved):
LGTM - Thanks!

PR #1025, opened 2022-08-02
[ADDED] SecureDeleteMsg method to JetStreamManager

piotrpio:
This addesses https://github.com/nats-io/nats-architecture-and-design/issues/118

As mentioned in the above issue, the old `DeleteMsg()` now sets `no_erase` to `true` by default (for better performance), and the new `SecureDeleteMsg()` can be used to completely overwrite the deleted message on server.

### DIFF
--- a/jsm.go
+++ b/jsm.go
@@ -58,8 +58,12 @@ type JetStreamManager interface {
 	// The stream must have been created/updated with the AllowDirect boolean.
 	GetLastMsg(name, subject string, opts ...JSOpt) (*RawStreamMsg, error)
 
-	// DeleteMsg erases a message from a stream.
+	// DeleteMsg deletes a message from a stream. The message is marked as erased, but its value is not overwritten.
 	DeleteMsg(name string, seq uint64, opts ...JSOpt) error
+
+	// SecureDeleteMsg deletes a message from a stream. The deleted message is overwritten with random data
+	// As a result, this operation is slower than DeleteMsg()
+	SecureDeleteMsg(name string, seq uint64, opts ...JSOpt) error
 
 	// AddConsumer adds a consumer to a stream.
 	AddConsumer(stream string, cfg *ConsumerConfig, opts ...JSOpt) (*ConsumerInfo, error)
@@ -1012,7 +1016,8 @@ func convertDirectGetMsgResponseToMsg(name string, r *Msg) (*RawStreamMsg, error
 }
 
 type msgDeleteRequest struct {
-	Seq uint64 `json:"seq"`
+	Seq     uint64 `json:"seq"`
+	NoErase bool   `json:"no_erase,omitempty"`
 }
 
 // msgDeleteResponse is the response for a Stream delete request.
@@ -1022,6 +1027,7 @@ type msgDeleteResponse struct {
 }
 
 // DeleteMsg deletes a message from a stream.
+// The message is marked as erased, but not overwritten
 func (js *js) DeleteMsg(name string, seq uint64, opts ...JSOpt) error {
 	o, cancel, err := getJSContextOpts(js.opts, opts...)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


jsm.go
```diff
@@ -1008,17 +1013,34 @@ func (js *js) DeleteMsg(name string, seq uint64, opts ...JSOpt) error {
 		defer cancel()
 	}
 
-	if name == _EMPTY_ {
-		return ErrStreamNameRequired
+	return js.deleteMsg(o.ctx, name, msgDeleteRequest{Seq: seq, NoErase: true})
+}
+
+// SecureDeleteMsg deletes a message from a stream.
+// The deleted message is overwritten with random data
```
wallyqs, line 1020: maybe mention here that it is slower as a result?

jsm.go
```diff
@@ -1008,17 +1013,34 @@ func (js *js) DeleteMsg(name string, seq uint64, opts ...JSOpt) error {
...
+func (js *js) SecureDeleteMsg(name string, seq uint64, opts ...JSOpt) error {
```
kozlovic, line 1021: If we go the option route, the implementation will simply have to set NoErase to opts.noErase (or whatever boolean you chose).

jsm.go
```diff
@@ -1008,17 +1013,34 @@ func (js *js) DeleteMsg(name string, seq uint64, opts ...JSOpt) error {
...
+	}
+	if cancel != nil {
+		defer cancel()
 	}
 
-	req, err := json.Marshal(&msgDeleteRequest{Seq: seq})
+	return js.deleteMsg(o.ctx, name, msgDeleteRequest{Seq: seq})
+}
+
+func (js *js) deleteMsg(ctx context.Context, stream string, req msgDeleteRequest) error {
```
kozlovic, line 1033: If you go with different APIs, I would make req a pointer to `msgDeleteRequest` so we don't have the copy of this struct when invoked from the Delete or SecureDelete APIs.

test/js_test.go
```diff
@@ -2116,6 +2116,113 @@ func TestJetStreamManagement_DeleteMsg(t *testing.T) {
 	}
 }
 
+func TestJetStreamManagement_SecureDeleteMsg(t *testing.T) {
+	s := RunBasicJetStreamServer()
```
kozlovic, line 2120: Since really we don't know - from the client perspective - the difference between a delete and secure delete, what I would do is simply make sure that the client's "delete message request" contains the noErase boolean when it is a simple delete and not (or set to false) when using the secure version. This is done by creating a NATS subscription on the delete API subject and inspecting the content (see TestJetStreamConsumerConfigReplicasAndMemStorage for example)

kozlovic (changes requested):

test/js_test.go
piotrpio: Good idea, I'll add that.

jsm.go
```diff
@@ -58,9 +58,12 @@ type JetStreamManager interface {
 	// The stream must have been created/updated with the AllowDirect boolean.
 	GetLastMsg(name, subject string, opts ...JSOpt) (*RawStreamMsg, error)
 
-	// DeleteMsg erases a message from a stream.
+	// DeleteMsg deletes a message from a stream. The message is marked as erased, but its value is not overwritten.
 	DeleteMsg(name string, seq uint64, opts ...JSOpt) error
 
+	// SecureDeleteMsg deletes a message from a stream. The message value is overwritten with random data.
+	SecureDeleteMsg(name string, seq uint64, opts ...JSOpt) error
```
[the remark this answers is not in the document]
piotrpio: I thought about that, but I feel that in this case (having really only 2 options), 2 separate methods are a bit more straightforward to the user.  `JSOpt` is already confusingly large bag of options and finding the one particular option which is applicable for a specific method is a hard job, even with good documentation. 

Providing 2 methods on the other hand, should eliminate any confusion. If a user wants to delete message, they will stumble on 2 methods with relevant names: `DeleteMsg()`, which looks like a good starting point and `SecureDeleteMsg()`. Both have clear documentation and are not configurable (which is a good thing IMO).

WDYT?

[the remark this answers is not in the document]
kozlovic: I started with adding a new API for the "direct get" but was told that it was better to add options instead. I do agree - however - that the JSOpt can be a bit confusing.
To sump up, I don't have objections per-se to add a new API (actually, in the C client I have `js_DeleteMsg()` and `js_EraseMsg()` APIs), just felt that the will was to go with options instead of new APIs. If other reviewers in the list are OK with new API, I am ok too.

kozlovic (approved):
LGTM, but you could use checkStreamName() instead of just checking for empty string.

jsm.go
```diff
@@ -1008,17 +1014,34 @@ func (js *js) DeleteMsg(name string, seq uint64, opts ...JSOpt) error {
...
+	if cancel != nil {
+		defer cancel()
 	}
 
-	req, err := json.Marshal(&msgDeleteRequest{Seq: seq})
+	return js.deleteMsg(o.ctx, name, &msgDeleteRequest{Seq: seq})
+}
+
+func (js *js) deleteMsg(ctx context.Context, stream string, req *msgDeleteRequest) error {
+	if stream == _EMPTY_ {
```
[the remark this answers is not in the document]
piotrpio: done

kozlovic:
@piotrpio Do you need new review on this? Looks like there are some force-push'es made, so not sure if things have changed since last review.

piotrpio:
@kozlovic Thanks, I just needed to rebase to resolve conflicts, and now some (not added in this PR) test is failing, I'm looking into it.

kozlovic:
That is maybe a flapper, let's recycle the test.

kozlovic:
If this is ready, maybe we get another LGTM from @wallyqs  and I can merge?

wallyqs (approved):
LGTM

PR #1026, opened 2022-08-02
[CHANGED] Update account info schema

piotrpio:
Related issue: https://github.com/nats-io/nats-architecture-and-design/issues/120

### DIFF
--- a/jsm.go
+++ b/jsm.go
@@ -178,12 +178,17 @@ type apiPagedRequest struct {
 
 // AccountInfo contains info about the JetStream usage from the current account.
 type AccountInfo struct {
+	Tier
+	Domain string          `json:"domain"`
+	API    APIStats        `json:"api"`
+	Tiers  map[string]Tier `json:"tier"`
+}
+
+type Tier struct {
 	Memory    uint64        `json:"memory"`
 	Store     uint64        `json:"storage"`
 	Streams   int           `json:"streams"`
 	Consumers int           `json:"consumers"`
-	Domain    string        `json:"domain"`
-	API       APIStats      `json:"api"`
 	Limits    AccountLimits `json:"limits"`
 }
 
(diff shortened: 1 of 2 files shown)


jsm.go
```diff
@@ -178,12 +178,17 @@ type apiPagedRequest struct {
 
 // AccountInfo contains info about the JetStream usage from the current account.
 type AccountInfo struct {
+	Tier
+	Domain string          `json:"domain"`
+	API    APIStats        `json:"api"`
+	Tiers  map[string]Tier `json:"tier"`
```
kozlovic, line 184: I don't recall why @matthiashanel had it be a map of `Tier` instead of `*Tier`, so that is expected that you have the same here, however, the json tag is marked as "omitempty", so we should have same here. 
PS: `domain` is omitempty too, but it was like that in client before, still, should we modify it as part of the PR?

kozlovic (changes requested):

jsm.go
piotrpio: For response objects, `omitempty` does not make much sense. It will only have effect when marshaling structs to JSON (the zero-value field will not end up in the resulting JSON), but on `Unmarshal()` it has no effect.
Server has it, because server actually marshals the structs - client does not do that, it just reads the response.

Also, I don't see much difference between using  `map[string]Tier` and `map[string]*Tier`. JSON unmarshaler will handle both the same way, and the zero value for a map is `nil`, regardless of the element type. Memory-wise, I don't thing it really matters for this particular API.

kozlovic: You are right, until we have a need to send AccountInfo (marshal it), then there is no need for the omitempty. As for the Tier vs *Tier, that is fine too, I did not ask to change that.

Jarema (approved):
LGTM

PR #1030, opened 2022-08-03
Support Direct Gets by subject

codegangsta:
This PR brings in client support for https://github.com/nats-io/nats-server/pull/3325, which adds server support for direct gets by a particular subject. This should now allow direct gets to work with KVs properly

### DIFF
--- a/go_test.mod
+++ b/go_test.mod
@@ -4,7 +4,7 @@ go 1.17
 
 require (
 	github.com/golang/protobuf v1.4.2
-	github.com/nats-io/nats-server/v2 v2.8.5-0.20220729163007-8aee7d5e51d4
+	github.com/nats-io/nats-server/v2 v2.8.5-0.20220803150712-d7847c97c116
 	github.com/nats-io/nkeys v0.3.0
 	github.com/nats-io/nuid v1.0.1
 	google.golang.org/protobuf v1.23.0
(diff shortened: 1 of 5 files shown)


test/js_test.go
```diff
@@ -7409,6 +7409,19 @@ func TestJetStreamDirectGetMsg(t *testing.T) {
...
+	// Test direct get by subject by trying to get 'bar' directly
+	r, err = js.GetLastMsg("DGM", "bar", nats.DirectGet())
+	if err != nil {
+		t.Fatalf("Error getting message: %v", err)
+	}
+	if r.Subject != "bar" {
+		t.Fatalf("expected subject to be 'bar', got: %v", r.Subject)
+	}
+	if string(r.Data) != "d" {
+		t.Fatalf("Error getting message: %v", err)
```
kozlovic, line 7422: Remove that one I would assume.

go_test.mod
```diff
@@ -14,6 +14,7 @@ require (
 	github.com/klauspost/compress v1.15.9 // indirect
 	github.com/minio/highwayhash v1.0.2 // indirect
 	github.com/nats-io/jwt/v2 v2.3.0 // indirect
+	go.uber.org/automaxprocs v1.5.1 // indirect
```
kozlovic, line 17: Weird.. I check out your branch and then did a `go mod tidy -modfile=go_test.mod` and that was removed, and go.sum is updated, so I would recommend you do that.

kozlovic (changes requested):

go_test.mod
codegangsta: Oh. yeah weird that it was brought in. Thanks for the catch

kozlovic:
@codegangsta Also, unrelated to your PR, but since you updated the server version, and due to @bruth recent change in the server, this test fails:
```
=== RUN   TestJetStreamManagement/update_consumer
    js_test.go:1417: Error on update: max waiting can not be updated
```
would need to be updated in order for the Travis CI to pass. Could you do it?

codegangsta:
Sure I'll take a look

kozlovic:
Meaning that server does not support it, so the test should not try to modify or at least not expect it to be updated :-)

codegangsta:
@kozlovic looks like this may be a flaky test?

```
=== RUN   TestHotSpotReconnect
    cluster_test.go:341: Connected clients to server: nats://127.0.0.1:1226 out of range: 23
```

I'll run it again

codegangsta:
@kozlovic all passed!

kozlovic (approved):
LGTM

PR #1031, opened 2022-08-03
[ADDED] KeyValue: RePublish option in KeyValueConfiguration

kozlovic:
Ability to create a KV with the re-publish feature.

### DIFF
--- a/kv.go
+++ b/kv.go
@@ -225,6 +225,7 @@ type KeyValueConfig struct {
 	Storage      StorageType
 	Replicas     int
 	Placement    *Placement
+	RePublish    *RePublish
 }
 
 // Used to watch all keys.
@@ -401,6 +402,7 @@ func (js *js) CreateKeyValue(cfg *KeyValueConfig) (KeyValue, error) {
 		MaxMsgs:           -1,
 		MaxConsumers:      -1,
 		AllowDirect:       true,
+		RePublish:         cfg.RePublish,
 	}
 
 	// If we are at server version 2.7.2 or above use DiscardNew. We can not use DiscardNew for 2.7.1 or below.
--- a/kv_test.go
+++ b/kv_test.go
@@ -14,7 +14,9 @@
 package nats
 
 import (
+	"fmt"
 	"testing"
+	"time"
 )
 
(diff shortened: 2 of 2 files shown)


kozlovic:
@bruth Wanted to add you as a reviewer, but for some reason I could not find you in the list. Maybe you need to be added into some GH lists.

codegangsta (approved):
Nice and easy. LGTM 👍

kv.go
```diff
@@ -401,6 +402,7 @@ func (js *js) CreateKeyValue(cfg *KeyValueConfig) (KeyValue, error) {
 		MaxMsgs:           -1,
 		MaxConsumers:      -1,
 		AllowDirect:       true,
+		RePublish:         cfg.RePublish,
```
kozlovic, line 405: Humm... I did not test the situation where we call that on an existing stream. We may then have the issue I was discussing during the call at line 425 (where we compare the config that we got from a lookup vs what we are trying to set...). Will see if I can add a test to simulate the upgrade.

kozlovic: Well, maybe that's ok then that the CreateKeyValue fails, the user should then administratively update the stream (if at all possible, not sure if we support updating this). Still will add a test.

derekcollison (approved):
LGTM

kozlovic:
@codegangsta @derekcollison I think that the stream cannot be updated with RePublish, meaning it may not reject the update per-se (as it did with pull consumer MaxWaiting), but it is actually not *updating* the stream and therefore the feature is not enabled. For this PR maybe I keep it that it would fail if the stream is originally not created with RePublish (don't try to auto-update). Will then need to add a test in server to demonstrate the issue, or @derekcollison if you think that it should be supported or should work, I may have done something wrong in my client test...

derekcollison:
I am ok with rejecting, just need to formalize the rejection in the server code. If need be we can add update support at a later time.

kozlovic:
Ok, so will update this PR with test that verifies that 2nd create with republish fails as expected, then will go to server to add test that ensure that we reject update properly (started to work on a test and update is not rejected, but republish update is clearly not applied).

derekcollison:
Perfect thanks.

PR #1032, opened 2022-08-04
[FIXED] JetStream: Fail Ack() (and the likes) for AckNone consumer

### LINKED ISSUE #1027: Attempting to ACK ordered consumer messages may fail forever
## Ordered consumer may get stuck forever trying to ACK

nats-server: 2.9.0-beta.17
nats.go: v1.16.0

Not relevant, but reproduces on macOS and Linux

At high level:
 - Create stream with R=3, and publish some data
 - Create an Ordered consumer
 - Attempt to ACK every message delivered
 - Restart the whole cluster while consuming

Test that reproduces is available here: https://github.com/nats-io/nats-server/pull/3292

#### Expected result:

ACKs are not required in this case, but the client lets me attempt to do them anyway.
A few ideas of how this could be improved:

 - Return an error when trying to ACK an ordered consumer message
 - Ignore (noop) ACKs for ordered consumer messages
 - Document clearly that application using ordered consumer should not ACK

#### Actual result:
Under normal circumstances, everything works fine.
However if the ephemeral consumer under the Ordered consumer gets destroyed, then the client will forever try to ACK using a request topic that nobody is listening to (and get `503` in response).

If an application is trying to be strict and always ACK before moving to the next message, then this application will be stuck forever

kozlovic:
Resolves #1027

### DIFF
--- a/js.go
+++ b/js.go
@@ -1102,6 +1102,7 @@ type jsSub struct {
 	deliver  string
 	pull     bool
 	dc       bool // Delete JS consumer
+	ackNone  bool
 
 	// This is ConsumerInfo's Pending+Consumer.Delivered that we get from the
 	// add consumer response. Note that some versions of the server gather the
@@ -1576,6 +1577,7 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync,
 		nms:      nms,
 		psubj:    subj,
 		cancel:   cancel,
+		ackNone:  o.cfg.AckPolicy == AckNonePolicy,
 	}
 
 	// Auto acknowledge unless manual ack is set or policy is set to AckNonePolicy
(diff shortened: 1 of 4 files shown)


kozlovic:
@mprimi I think that this would have let you know that you should not have been calling `Ack[Sync]()`for a `AckNone` ack policy (the OrderedConsumer is creating a consumer with AckNone).

wallyqs (approved):
LGTM

derekcollison (commented):
What is the problem we are trying to solve? The server can handle an ack for an AckNone consumer.

kozlovic:
> What is the problem we are trying to solve? The server can handle an ack for an AckNone consumer.

Two problems:
- Sending unnecessary ACKs to the server(s)
- Having users incorrectly Ack'ing and trying again and again after getting an "unexpected" error. The context was that Marco had a test using OrderedConsumer and was calling AckSync(), which was working fine until in some cases, when bringing the cluster down and back again, the AckSync() would start to return an error. He was told to ack again in case of error, well, in that case the ack would always fail because the ordered consumer had reset, so the message's ack subject on hand was no longer valid and not known by any of the server, so AckSync() would return an error such as "no responders available".

derekcollison (approved):
LGTM

PR #1039, opened 2022-08-10
[UPDATED] Align StreamSourceInfo schema with server

### LINKED ISSUE #1038: Stream source info does not expose all fields sent by the server
https://github.com/nats-io/nats.go/blob/c157d64783d91fd2084064bd5c07a10dc4bba2d7/jsm.go#L726-L730

does not expose all the fields available

https://github.com/nats-io/nats-server/blob/a4bf4e87f670643540d1d28d1645242234d714d6/server/stream.go#L152-L158

piotrpio:
Resolves #1038

### DIFF
--- a/jsm.go
+++ b/jsm.go
@@ -154,8 +154,8 @@ type ExternalStream struct {
 	DeliverPrefix string `json:"deliver"`
 }
 
-// apiError is included in all API responses if there was an error.
-type apiError struct {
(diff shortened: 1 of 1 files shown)


kozlovic (approved):
LGTM

wallyqs (approved):
LGTM. I was not sure about exposing APIError here since only being used for this type publicly but better to have some parity with the types from the server instead.

PR #1041, opened 2022-08-10
Temporary code change so that server PR can run tests

kozlovic:
I would like to merge that to main so I can have a server PR that pulls nats.go main and fixes the date formatting. Then I will have another nats.go PR to remove the temporary code and point to the latest from server main.


wallyqs (approved):
LGTM

PR #1043, opened 2022-08-11
Add MaxRequestMaxBytes consumer option

bruth:


test/js_test.go
```diff
@@ -5602,6 +5604,19 @@ func testJetStreamFetchOptions(t *testing.T, srvs ...*jsServer) {
...
 
+	t.Run("max request max bytes", func(t *testing.T) {
+		defer js.PurgeStream(subject)
+
+		sub, err := js.PullSubscribe(subject, "max-request-max-bytes", nats.MaxRequestMaxBytes(100))
+		if err != nil {
+			t.Fatal(err)
+		}
+		defer sub.Unsubscribe()
+		if _, err := sub.Fetch(10); err == nil || !strings.Contains(err.Error(), "MaxRequestMaxBytes of 100") {
```
kozlovic, line 5615: Looks like there is no message at this point and so this fails with a timeout and not what you expect. I know that this test can be tricky because those t.Run() may influence the other ones, but you could try to send at least 1 message that is big enough before creating the pull subscribe to see if you get the expected error.

kozlovic (changes requested):

test/js_test.go
bruth: Thanks. I realized this was only partially implemented because the fetch request wasn't including the `MaxBytes` option. I added a pull option which can be used to set this in the fetch request. This is now rejected regardless if a message is available.

kozlovic (approved):
LGTM

PR #1044, opened 2022-08-11
[IMPROVED] Add JetStream error codes, extract ErrConsumerNameAlreadyInUse

### LINKED ISSUE #977: Expose error constant for `consumer name already in use`
## Defect

`nats.go` v1.14.0

N/A

Try to add a consumer when a consumer with the same name exists
```
js.AddConsumer("some stream", consumerConfig)
js.AddConsumer("some stream", consumerConfig)
```

The error is expected, as the consumer exists! 
However, one has to do **manual string comparison** because an error constant is not provided for this case.

Here's an example [from this repo itself](https://github.com/nats-io/nats.go/blob/main/js.go#L1580-L1581):
```
				(strings.Contains(cinfo.Error.Description, `consumer already exists`) ||
					strings.Contains(cinfo.Error.Description, `consumer name already in use`)) {
```

#### Expected result:
This is what I wish I could do:
```
  consumerInfo, err := js.AddConsumer(streamName, consumerConfig)
  if err == nats.ErrConsumerExists {
    // consumer exists - do what is appropriate
  }
```

#### Actual result:
This is what I'm forced to do (brittle and ugly):
```
  consumerInfo, err := js.AddConsumer(streamName, consumerConfig)
  if err != nil && fmt.Sprintf("%s", err) == "consumer name already in use" {
    // consumer exists - do what is appropriate
  }
```

piotrpio:
Resolves #977 

Additional changes in this PR:
1. Make `*APIError` implement `error` interface
2. Return JetStream error codes in error message
3. Verify error messages returned by server instead of parsing error messages as strings
4. Remove `ErrInvalidDurable` and `checkDurName()` in favor of `ErrInvalidConsumerName` and `checkConsumerName()`
5. Use `ErrJetStreamNotEnabledForAccount` error with error code when not enabled for an account instead of general `ErrJetStreamNotEnabled`


piotrpio:
One thing I'm not certain about is removing `ErrInvalidDurableName` error, as this is a breaking change. Not sure what the impact would be, although I'm not a fan of having both `ErrInvalidDurableName` `and `ErrInvalidConsumerName` used interchangeably.

wallyqs (commented):
Nice work @piotrpio 👍, left some comments

jsm.go
```diff
@@ -384,10 +414,10 @@ func (js *js) DeleteConsumer(stream, consumer string, opts ...JSOpt) error {
 	}
 
 	if resp.Error != nil {
-		if resp.Error.Code == 404 {
+		if resp.Error.ErrorCode == ConsumerNotFound {
```
wallyqs, line 417: Wonder if we could make these `ErrConsumerNotFound` plain error types, instead conform to the Error interface and maybe a JetStreamAPIError interface and stamp them with their error codes there, something like:

```
ErrConsumerNotFound             = &apiError{ErrorCode: ...}
```

js.go
```diff
@@ -1228,7 +1228,10 @@ func (js *js) ChanQueueSubscribe(subj, queue string, ch chan *Msg, opts ...SubOp
 // See important note in Subscribe()
 func (js *js) PullSubscribe(subj, durable string, opts ...SubOpt) (*Subscription, error) {
 	mch := make(chan *Msg, js.nc.Opts.SubChanLen)
-	return js.subscribe(subj, _EMPTY_, nil, mch, true, true, append(opts, Durable(durable)))
+	if durable != "" {
+		opts = append(opts, Durable(durable))
+	}
```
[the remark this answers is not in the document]
piotrpio: No, pull ephemerals are already supported. This change is introduced because `nats.Durable()` should not accept empty consumer name (or else `checkConsumerName()` will fail)

jsm.go
```diff
@@ -268,6 +285,30 @@ type consumerResponse struct {
 
 // AddConsumer will add a JetStream consumer.
 func (js *js) AddConsumer(stream string, cfg *ConsumerConfig, opts ...JSOpt) (*ConsumerInfo, error) {
+	if cfg != nil && cfg.Durable != _EMPTY_ {
+		consInfo, err := js.ConsumerInfo(stream, cfg.Durable)
+		if err != nil && !errors.Is(err, ErrConsumerNotFound) && !errors.Is(err, ErrStreamNotFound) {
+			return nil, err
+		}
+		if consInfo != nil {
+			return nil, fmt.Errorf("creating consumer %q on stream %q: %w", cfg.Durable, stream, ErrConsumerNameAlreadyInUse)
```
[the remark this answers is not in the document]
piotrpio: Hmm, idempotency check is a good idea. As to `nats: ` prefix, I'll just reverse the order of the message and put `ErrConsumerNameAlreadyInUse` at the beginning.

jsm.go
```diff
@@ -268,6 +285,30 @@ type consumerResponse struct {
 
 // AddConsumer will add a JetStream consumer.
 func (js *js) AddConsumer(stream string, cfg *ConsumerConfig, opts ...JSOpt) (*ConsumerInfo, error) {
+	if cfg != nil && cfg.Durable != _EMPTY_ {
+		consInfo, err := js.ConsumerInfo(stream, cfg.Durable)
```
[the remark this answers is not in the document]
piotrpio: Yeah, I would lean towards leaving that, as without that `AddConsumer()` I feel does not work as the name suggests (it will simply update a consumer if it already exists).

jsm.go
```diff
@@ -219,6 +219,26 @@ type accountInfoResponse struct {
...
+	JetStreamNotEnabledForAccount ErrorCode = 10039
+
+	StreamNotFound  ErrorCode = 10059
+	StreamNameInUse ErrorCode = 10058
+
+	ConsumerNotFound      ErrorCode = 10014
+	ConsumerNameExists    ErrorCode = 10013
+	ConsumerAlreadyExists ErrorCode = 10105
+
+	MessageNotFound ErrorCode = 10037
```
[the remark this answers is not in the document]
piotrpio: I was thinking about extracting those codes together with errors to a separate package, but moving already defined errors would be a breaking change, so I decided against it. I think the format with `ErrorCode` prefix makes sense, I'll change that.

jsm.go
piotrpio: Good call. I have an idea on how to make the implementation a bit cleaner, but that would require changing the way errors are compared:
1. As you mentioned, change the defined error types to `APIError`, like so:
`ErrConsumerNotFound             = &APIError{ErrorCode: 10014, Code: 404, Description: "nats: consumer not found"}`
2. Implement custom `Is()` method on `APIError`, comparing error codes:
```
func (e *APIError) Is(target error) bool {
	err, ok := target.(*APIError)
	if !ok {
		return false
	}
	return err.ErrorCode == e.ErrorCode && err.Code == e.Code
}
```
3. Simplify client method implementations so that we do not have to compare the error codes manually at all, e.g.:
```
	if info.Error != nil {
		return nil, info.Error
	}
```
instead of
```
	if info.Error != nil {
		if info.Error.ErrorCode == ConsumerNotFound {
			return nil, ErrConsumerNotFound
		}
		if info.Error.ErrorCode == StreamNotFound {
			return nil, ErrStreamNotFound
		}
		return nil, info.Error
	}
```

However, that means user would not be able to check errors by simple comparison (`err == ErrConsumerNotFound`), but instead rely on `errors.Is(err, ErrConsumerNotFound)` (which usually should be the preferred way to go).

The other approach would be to leave the client method implementations untouched and simply change the variable definitions to be `&APIError{}` instead of `errors.New()`.

What do you think?

piotrpio: Also, what do you mean by implementing `JetStreamAPIError` interface (i.e. what would it do). The `error` interface is implemented as part of this PR.

wallyqs: I put together an example of what we could do via an interface here: https://github.com/nats-io/nats.go/pull/1047

jsm.go
```diff
@@ -244,13 +264,10 @@ func (js *js) AccountInfo(opts ...JSOpt) (*AccountInfo, error) {
 		return nil, err
 	}
 	if info.Error != nil {
-		var err error
-		if strings.Contains(info.Error.Description, "not enabled for") {
-			err = ErrJetStreamNotEnabled
-		} else {
-			err = errors.New(info.Error.Description)
+		if info.Error.ErrorCode == JetStreamNotEnabledForAccount {
+			return nil, ErrJetStreamNotEnabled
```
wallyqs, line 268: In the server not enabled for account and not enabled looks like they are two different errors actually... maybe we should start distinguishing them too and use `ErrJetStreamNotEnabledForAccount`?

https://github.com/nats-io/nats-server/blob/52c4872666a26acdf475cb89ae9b5f367324fadd/server/jetstream_errors_generated.go#L472-L474

wallyqs: Another factor why maybe we should clean it up is that `JetStreamNotEnabledForAccount` is a hard error since means that account can never use JetStream in current setup, whereas things are a bit less certain in `ErrJetStreamNotEnabled`, it can be either a temporary error due to JS not ready yet.

piotrpio: Yes, distinguishing these errors is a good idea.

wallyqs (approved):
LGTM!

kozlovic (approved):
LGTM

PR #1045, opened 2022-08-11
kv: fix typo on error type

wallyqs:


kozlovic (commented):
This was released (since v1.13.0) so is that not dangerous to change that exported variable? We can change, but that may break some apps/tests, no?

wallyqs:
@kozlovic the server is not using it so looks fine... but agree, no hurry to merge, I thought I should open just in case there is a window to fix still

kozlovic (approved):
LGTM. I guess for the next release v1.17.0, we could put that in the "[CHANGED]" section of the release notes.

wallyqs:
I think we can postpone merging this one for v1.17.0 and instead include it as part eventual rework how the KeyValue errors are handled so that they are APIError aware.

PR #1046, opened 2022-08-12
Add UserJWTAndSeed helper function

bruth:
The motivation for this is to support situations when the JWT and seed are passed as environment variables or other means that don't rely on a file, such as environments only supporting environment variables.


bruth:
Appears the failure occurred on an unrelated test.

kozlovic (approved):
LGTM

PR #1047, opened 2022-08-12
js: Add JetStreamAPIError with more api response details

wallyqs:
Follow up from #1044 

Example of adding a `JetStreamAPIError` interface that exposes from details about the API response:

```
type JetStreamAPIError interface {
	Code() int
	ErrorCode() int
	Description() string
	Error() string
}
```

To keep it backwards compatible and keep old checks working, instead of defining JetStream errors like this:

```
ErrJetStreamNotEnabled          = errors.New("nats: jetstream not enabled")
```

They can be defined like the following:

```
ErrJetStreamNotEnabled JetStreamAPIError = &jsAPIError{errorCode: 10039}
```


jsm.go
```diff
@@ -161,6 +161,37 @@ type APIError struct {
 	Description string `json:"description,omitempty"`
 }
 
+// JetStreamAPIError is an error result from making a request to the
+// JetStream API.
+type JetStreamAPIError interface {
+	Code() int
+	ErrorCode() int
+	Description() string
+	Error() string
```
wallyqs, line 170: could add some other conditions to be able signal the client that something is `Temporary()` here too, for example when there are transient errors such as `JSClusterNotAvailable`

wallyqs:
Another idea is to make a more general `JetStreamError` but that can still unwrap the APIError: https://github.com/nats-io/nats.go/compare/js-api-err2?expand=1

piotrpio:
I think I like the second approach (form `js-api-err2` branch) a little bit more. It is a bit easier to distinguish API error from client JetStream error if the API error is stored on 1 struct field, not mixed together with `message`. Plus, it makes defining `jsError` from `APIError` more concise.

piotrpio:
1. Converted all JetStream-related errors to `jsError`
2. Switch from comparing error codes to `errors.Is()` in JetStream methods
3. Moved errors to a separate file
4. Removed several unused errors

jserrors.go
```diff
@@ -0,0 +1,185 @@
...
+}
+
+func (err *jsError) Error() string {
+	if err.apiErr != nil && err.apiErr.Description != "" {
+		return err.apiErr.Error()
+	}
+	return fmt.Sprintf("nats: %s", err.message)
+}
+
+func (err *jsError) Unwrap() error {
```
wallyqs, line 179: More info about features of wrapping and `Unwrap` can be found here: 

- https://pkg.go.dev/errors#pkg-functions
- https://github.com/golang/go/wiki/ErrorValueFAQ

kozlovic (approved):
LGTM

PR #1057, opened 2022-08-30
link not allowed in put meta

scottf:
* Prevented links in meta data for puts
* refactored for common publish meta functionality
* Converted all loose errors to vars.
* Updated unit test to check and cover errors.

### DIFF
--- a/object.go
+++ b/object.go
@@ -133,6 +133,7 @@ var (
 	ErrObjectAlreadyExists  = errors.New("nats: an object already exists with that name")
 	ErrNameRequired         = errors.New("nats: name is required")
 	ErrNeeds262             = errors.New("nats: object-store requires at least server version 2.6.2")
+	ErrLinkNotAllowed       = errors.New("nats: a link is not allowed with an object")
 )
 
 // ObjectStoreConfig is the config for the object store.
@@ -311,6 +312,14 @@ func (obs *obs) Put(meta *ObjectMeta, r io.Reader, opts ...ObjectOpt) (*ObjectIn
 		return nil, ErrBadObjectMeta
 	}
 
+	if meta.Opts == nil {
+		meta.Opts = &ObjectMetaOptions{ChunkSize: objDefaultChunkSize}
+	} else if meta.Opts.Link != nil {
+		return nil, ErrLinkNotAllowed
+	} else if meta.Opts.ChunkSize <= 0 {
+		meta.Opts.ChunkSize = objDefaultChunkSize
+	}
+
 	var o objOpts
 	for _, opt := range opts {
 		if opt != nil {
@@ -356,13 +365,8 @@ func (obs *obs) Put(meta *ObjectMeta, r io.Reader, opts ...ObjectOpt) (*ObjectIn
 		return nil, err
 	}
 
-	chunkSize := objDefaultChunkSize
-	if meta.Opts != nil && meta.Opts.ChunkSize > 0 {
-		chunkSize = meta.Opts.ChunkSize
-	}
-
 	m, h := NewMsg(chunkSubj), sha256.New()
-	chunk, sent, total := make([]byte, chunkSize), 0, uint64(0)
+	chunk, sent, total := make([]byte, meta.Opts.ChunkSize), 0, uint64(0)
 
 	// set up the info object. The chunk upload sets the size and digest
 	info := &ObjectInfo{Bucket: obs.name, NUID: newnuid, ObjectMeta: *meta}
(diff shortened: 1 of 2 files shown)


piotrpio:
Tests are failing
```
=== RUN   TestObjectLinks
    /Users/piotrpiotrowski/go/nats.go/test/object_test.go:544: Unexpected error: nats: a link is not allowed with an object
--- FAIL: TestObjectLinks (0.03s)
=== RUN   TestObjectList
    /Users/piotrpiotrowski/go/nats.go/test/object_test.go:703: Unexpected error: nats: a link is not allowed with an object
--- FAIL: TestObjectList (0.03s)
```

object.go
```diff
@@ -133,6 +133,7 @@ var (
 	ErrObjectAlreadyExists  = errors.New("nats: an object already exists with that name")
 	ErrNameRequired         = errors.New("nats: name is required")
 	ErrNeeds262             = errors.New("nats: object-store requires at least server version 2.6.2")
+	ErrLinkNotAllowed       = errors.New("nats: a link is not allowed with an object")
```
piotrpio, line 136: I'm not sure about this error message (lacks context), maybe something like this would be cleaner:
```go
ErrLinkNotAllowed       = errors.New("nats: link cannot be set when putting the object in bucket")
```

piotrpio (changes requested):

object.go
scottf: Changing

scottf:
> Tests are failing
I see why. AddLink /Add Bucket Link call put with no data. They don't have to. I'll fix.

object.go
```diff
@@ -630,6 +634,30 @@ func (obs *obs) Delete(name string) error {
 	return obs.js.purgeStream(obs.stream, &StreamPurgeRequest{Subject: chunkSubj})
 }
 
+func publishMeta(info *ObjectInfo, js JetStreamContext) error {
+	// Prepare the meta message
+	metaSubj := fmt.Sprintf(objMetaPreTmpl, info.Bucket, encodeName(info.ObjectMeta.Name))
+	mm := NewMsg(metaSubj)
+	mm.Header.Set(MsgRollup, MsgRollupSubject)
+
+	info.ModTime = time.Time{} // We don't store this, just do it after publish
```
piotrpio, line 643: This is unnecessary - `time.Time` is a struct, and a zero value for a struct is an empty struct (in this case `time.Time{}`. 

BTW, I wonder whether that's valid - empty time value will get marshaled to `0001-01-01T00:00:00Z` - is `ModTime` required in the meta message or could we simply add `omitempty` on the struct tag in `ObjectInfo{}`? (this will make `json.Marshal()` not add this field in output JSON if it's empty)

object.go
```diff
@@ -630,6 +634,30 @@ func (obs *obs) Delete(name string) error {
...
+	data, err := json.Marshal(info)
+	if err != nil {
+		return err
+	}
+
+	// Publish the meta message.
+	mm.Data = data
+	_, err = js.PublishMsg(mm)
```
piotrpio, line 651: For short calls like this, when we really only care about the error, I would suggest to inline the call to `PublishMsg()`:
```go
if _, err := js.PublishMsg(mm); err != nil {
    return err
}
```

object.go
```diff
@@ -664,9 +692,15 @@ func (obs *obs) AddLink(name string, obj *ObjectInfo) (*ObjectInfo, error) {
 		Name: name,
 		Opts: &ObjectMetaOptions{Link: &ObjectLink{Bucket: obj.Bucket, Name: obj.Name}},
 	}
+	info := &ObjectInfo{Bucket: obs.name, NUID: nuid.Next(), ModTime: time.Now().UTC(), ObjectMeta: *meta}
 
 	// put the link object
-	return obs.Put(meta, nil)
+	err = publishMeta(info, obs.js)
```
piotrpio, line 698: Similar to previous comment:
```go
if err := publishMeta(info, obs.js); err != nil {
    return err
}
```

This applies to the other call to `publishMeta()` in this file as well.

object.go
scottf: Fixing

piotrpio (approved):
LGTM

object.go
scottf: fixed

piotrpio:
The build is failing (staticcheck):
```
test/object_test.go:38:2: this value of obs is never used (SA4006)

test/object_test.go:41:2: this value of obs is never used (SA4006)

test/object_test.go:715:2: this value of lch is never used (SA4006)
```

Other than that, looks good.

PR #1060, opened 2022-08-31
[FIXED] Paging in stream and consumer names listing

### LINKED ISSUE #1059: Paging does not work when listing stream/consumer names over page limit
`ConsumerNames()` and `StreamNames()` send empty request to the server (without offset provided). Because of that, if there are over 1024 (server limit) streams or consumers per stream, these methods return invalid results .

piotrpio:
Resolves #1059


Jarema (approved):
LGTM

PR #1062, opened 2022-08-31
[ADDED] Filter streams by subject

### LINKED ISSUE #1061: Add filtering by subject when listing streams and stream names
`StreamsInfo()` and `StreamNames()` should support setting the `filter` field on request.

piotrpio:
Resolves #1061

### DIFF
--- a/js.go
+++ b/js.go
@@ -239,6 +239,8 @@ type jsOpts struct {
 	purgeOpts *StreamPurgeRequest
 	// streamInfoOpts contains optional stream info options
 	streamInfoOpts *StreamInfoRequest
+	// streamListSubject is used for subject filtering when listing streams / stream names
+	streamListSubject string
 	// For direct get message requests
 	directGet bool
 	// For direct get next message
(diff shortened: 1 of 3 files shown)


js.go
```diff
@@ -359,6 +361,17 @@ func DirectGetNext(subject string) JSOpt {
 	})
 }
 
+// StreamListFilter is an option that can be used to configure `StreamsInfo()` and `StreamNames()` requests.
+// It allows filtering the retured streams by subject associated with each stream.
+// Wildcards can be used. For example, `StreamListFilter(FOO.>.A) will return
```
wallyqs, line 366: `FOO.>.A` is an invalid subject I think, maybe switch to the asterisk wildcard variant:
```suggestion
// Wildcards can be used. For example, `StreamListFilter(FOO.*.A) will return
```

piotrpio: Thanks for catching that - fixed

wallyqs (approved):
LGTM!

PR #1063, opened 2022-08-31
Allow `AckAll` for pull-based consumers

neilalexander:
If you are the only client pulling on a given consumer, the efficiency gains of fetching in batches is largely lost if you have to acknowledge every message you received individually afterwards.

It would seem intuitive to be able to acknowledge only the most recent message in the batch instead and to have that do the right thing.

### DIFF
--- a/js.go
+++ b/js.go
@@ -1390,7 +1390,7 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync,
 	// Some checks for pull subscribers
 	if isPullMode {
 		// Check for bad ack policy
-		if o.cfg.AckPolicy == AckNonePolicy || o.cfg.AckPolicy == AckAllPolicy {
+		if o.cfg.AckPolicy == AckNonePolicy {
 			return nil, fmt.Errorf("nats: invalid ack mode for pull consumers: %s", o.cfg.AckPolicy)
 		}
 		// No deliver subject should be provided


ripienaar:
@aricart same as push+qsubs which we do support.

kozlovic:
> same as push+qsubs which we do support

@ripienaar Then I think we should not, meaning we should reject an AckAll for a QueueSubscribe() because indeed there may be more than one member and I think it is bad that one member acks messages from other members.

ripienaar:
@kozlovic since 8 months the server allows this - agree its risky and probably a bad idea.

kozlovic:
Indeed, seems like the server all since this PR: https://github.com/nats-io/nats-server/pull/2776, so I guess restriction should be removed in client libraries too... yes, I think it is a bad idea.

neilalexander:
I can't really comment on the queue subscriptions as not using those, but in the case of one pull subscriber per consumer (as is the pattern that we are using in our application here), `AckAll` is a pretty big net performance win.

Have done quite a bit of testing with this PR and seems to work fine, although wondering if there might be some sensible place to put a comment warning about the potential for multiple clients to tread on each other.

aricart:
@ripienaar @kozlovic While we are at it, I think we also have the restriction that pull consumers must ack, but wondering if we should allow AckNone - this would allow other scenarios like ordered pull consumer etc.

ripienaar:
For the moment, server would not allow that.

derekcollison:
AckAll was put in to support emulated push in clients with efficient ephemeral pull consumers under covers.

kozlovic (approved):
LGTM. Since the server accepts it, I guess there is no reason to prevent the user to using this ack mode for pull consumers...

PR #1069, opened 2022-09-06
Mirrors can answer direct gets potentially so removing error condition.

derekcollison:

### DIFF
--- a/jsm.go
+++ b/jsm.go
@@ -912,7 +912,7 @@ func (js *js) getMsg(name string, mreq *apiMsgGetRequest, opts ...JSOpt) (*RawSt
 
 	var apiSubj string
 
-	doDirectGetLastBySubject := o.directGet && mreq.LastFor != ""
+	doDirectGetLastBySubject := o.directGet && mreq.LastFor != _EMPTY_
 
 	if doDirectGetLastBySubject {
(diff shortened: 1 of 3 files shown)


wallyqs (approved):
LGTM

kozlovic (approved):

jsm.go
```diff
@@ -1006,9 +1006,10 @@ func convertDirectGetMsgResponseToMsg(name string, r *Msg) (*RawStreamMsg, error
 	if stream == _EMPTY_ {
 		return nil, fmt.Errorf("nats: missing stream header")
 	}
-	if stream != name {
-		return nil, fmt.Errorf("nats: response stream header is '%s', not '%s'", stream, name)
-	}
+
+	// Mirrors can now answer direct gets, so removing check for name equality.
```
aricart, line 1010: Why would a stream name check be useful? the client is making a request, so presumably, the answer is from the stream (or mirror). Is there some nuance here?

kozlovic: I added that based on the feature first added in the server with a test that went along and had the check for the stream name: https://github.com/nats-io/nats-server/blob/c4b5ca7cff77c9e41e5fba00240c05a335764a22/server/jetstream_cluster_test.go#L5087

PR #1071, opened 2022-09-08
Set memory storage to true for ephemeral consumer

goku321:
Signed-off-by: Deepak <sah.sslpu@gmail.com>

When creating ephemeral consumer, storage type is inherited from the stream which is not correct. Ephemerals should be memory only.

### DIFF
--- a/js.go
+++ b/js.go
@@ -1562,6 +1562,11 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync,
 		if cfg.MaxAckPending == 0 && ch != nil && cfg.AckPolicy != AckNonePolicy {
 			cfg.MaxAckPending = cap(ch)
 		}
+
+		// Set memory storage to true if it's an ephemeral.
+		if !isDurable {
+			cfg.MemoryStorage = true
+		}
 		// Create request here.
 		ccreq = &createConsumerRequest{
 			Stream: stream,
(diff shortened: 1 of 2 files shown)


ripienaar:
Epehemeral consumers do survive server restart and their state is preserved cross boots within their inactive threshold.

So they cant be memory - but they are always R1 and have some other optimisations.

derekcollison:
With 2.9 you can ask a consumer to be memory only. In consumer config.

```
	// Force memory storage.
	MemoryStorage bool `json:"mem_storage,omitempty"`
```

goku321:
@derekcollison so when doing a `PullSubscribe` , we first create a consumer and pass its name. But, when we directly use `PullSubscribe` method with an empty consumer name (for ephemeral) it would still create file based consumer.

derekcollison:
Which client version are you using?

Can you share a code snippet of creating the consumer?

goku321:
Here's an [example](https://gist.github.com/goku321/2e00f898d83ad6405c19cf2d7454851e) to create a pull based ephemeral consumer. This will create a file based consumer and there's no way to force memory storage when directly using PullSubscribe. It would be good to have this feature and we wouldn't have to worry about consumer names.

One workaround is to create a consumer using `js.AddConsumer` (this will also require to set the `Durable` field) and then passing the consumer name to `PullSubscribe`:
```
cons, _ := js.AddConsumer("stream", &nats.ConsumerConfig{
		MemoryStorage: true,
		AckPolicy: nats.AckExplicitPolicy,
		Replicas: 1,
		Durable: "consumer1",
	})

sub, _  := js.PullSubscribe("subject", cons.Name)
```
I'm not sure if it's similar to an ephemeral.

derekcollison:
Just leave off Durable name, should work with 2.9 server. And you can override InactiveThreshold as you see fit, default is 5s IIRC.

PR #1072, opened 2022-09-09
[CHANGED] StreamInfo() will now return all subjects when requested

jnmoyne:
If a subject filter is specified in the StreamInfoRequest{} option,
then all matching subjects will be returned and not be capped to
the server limit of 100,000. It is internally using pagination
that was added in the server PR: https://github.com/nats-io/nats-server/pull/3454

### DIFF
--- a/jsm.go
+++ b/jsm.go
@@ -178,7 +178,7 @@ type apiPaged struct {
 // apiPagedRequest includes parameters allowing specific pages to be requested
 // from APIs responding with apiPaged.
 type apiPagedRequest struct {
-	Offset int `json:"offset"`
+	Offset int `json:"offset,omitempty"`
 }
 
 // AccountInfo contains info about the JetStream usage from the current account.
@@ -696,12 +696,17 @@ func (js *js) AddStream(cfg *StreamConfig, opts ...JSOpt) (*StreamInfo, error) {
 type (
 	// StreamInfoRequest contains additional option to return
 	StreamInfoRequest struct {
+		apiPagedRequest
 		// DeletedDetails when true includes information about deleted messages
 		DeletedDetails bool `json:"deleted_details,omitempty"`
 		// SubjectsFilter when set, returns information on the matched subjects
 		SubjectsFilter string `json:"subjects_filter,omitempty"`
 	}
-	streamInfoResponse = streamCreateResponse
+	streamInfoResponse = struct {
+		apiResponse
+		apiPaged
+		*StreamInfo
+	}
 )
 
 func (js *js) StreamInfo(stream string, opts ...JSOpt) (*StreamInfo, error) {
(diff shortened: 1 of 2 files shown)


jsm.go
```diff
@@ -682,12 +685,17 @@ func (js *js) AddStream(cfg *StreamConfig, opts ...JSOpt) (*StreamInfo, error) {
 type (
 	// StreamInfoRequest contains additional option to return
 	StreamInfoRequest struct {
+		apiPagedRequest
```
kozlovic, line 688: I would make this public (I know, that means modifying some other structs that embeds this struct). If we do so, then we may not need a new API (StreamContainedSubjects), but pass &nats.StreamInfoRequest{} as an option to js.StreamInfo(). See test TestStreamInfoSubjectInfo() for an example. The new field "Offset" would be set by the user but then it would be responsibility to the user to call StreamInfo() again with the new offset.

kozlovic (changes requested):
At the very least you need to update go_test.mod to the latest version of the server so you can benefit from the feature you added in the server. But I think we can get away with reusing StreamInfo() as opposed to add a new function.

go_test.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/nats-io/nats.go
 
-go 1.17
+go 1.16
```
kozlovic, line 3: Revert that change.

jsm.go
```diff
@@ -37,6 +37,9 @@ type JetStreamManager interface {
 	// StreamInfo retrieves information from a stream.
 	StreamInfo(stream string, opts ...JSOpt) (*StreamInfo, error)
 
+	// StreamContainedSubjects queries the stream for the subjects it holds with optional filter
+	StreamContainedSubjects(stream string, filter ...string) (map[string]uint64, error)
```
kozlovic, line 41: Since the code rejects more than 1 filter, not sure why you make it `...string`, and that prevents the use of `opts ...JSOpts` that you should be adding instead. The filter can then technically be made an option.

jsm.go
```diff
@@ -727,6 +735,78 @@ func (js *js) StreamInfo(stream string, opts ...JSOpt) (*StreamInfo, error) {
 	return resp.StreamInfo, nil
 }
 
+func (js *js) StreamContainedSubjects(stream string, filter ...string) (map[string]uint64, error) {
+	if err := checkStreamName(stream); err != nil {
```
kozlovic, line 739: You need to pass JSOpt and do like StreamInfo in that use will be able to override the ttl for the request (especially an API like this that can take time to collect all results). The filter could then be made an option.

jsm.go
```diff
@@ -727,6 +735,78 @@ func (js *js) StreamInfo(stream string, opts ...JSOpt) (*StreamInfo, error) {
...
+		return nil, err
+	}
+
+	if len(filter) > 1 {
```
kozlovic, line 743: See comment above, if you don't support more than 1, then why make it a `...string`?

jsm.go
```diff
@@ -727,6 +735,78 @@ func (js *js) StreamInfo(stream string, opts ...JSOpt) (*StreamInfo, error) {
...
+
+	f := ">"
+	if len(filter) == 1 && filter[0] != "" {
+		f = filter[0]
+	}
+
+	var i int
+	var subjectMessagesMap map[string]uint64 = nil
+
+	o, cancel, err := getJSContextOpts(js.opts)
```
kozlovic, line 755: you would then add `, opts...` for opts that are passed as arguments.

kozlovic (changes requested):

jsm.go
jnmoyne: I copied the signature and function name from jsm.go (https://github.com/nats-io/jsm.go/blob/31b0f1e3966a7125454efd34ad99a3267c204e28/manager.go#L564). Not sure why `...string`, just assumed it was done with an eye to allowing more than one filter in the future?

In any case yes it's much better to have the filter in the options

jsm.go
```diff
@@ -727,6 +735,74 @@ func (js *js) StreamInfo(stream string, opts ...JSOpt) (*StreamInfo, error) {
...
+	var subjectMessagesMap map[string]uint64 = nil
+
+	if o.streamInfoOpts != nil && o.streamInfoOpts.SubjectsFilter == _EMPTY_ {
+		o.streamInfoOpts.SubjectsFilter = ">"
+	}
+
+	for {
+		var req []byte
+
+		o.streamInfoOpts.Offset = i
```
kozlovic, line 760: What if the user did not provide a filter, then o.streamInfoOpts will be nil and this will panic. You need to have a stack variable declared about (around line 753) and maybe set the SubjectsFilter based on the user provided one (if any). Something like this:
```
sio := streamInfoOpts{}
if o.streamInfoOpts != nil && ... {
   sio.SubjectFilter = ">"
}
for {
   ...
   sio.Offset = i
   if req, err = json.Marshal(&sio); err != nil {
        return nil, err
   }
...
```
Hope that makes sense.

kozlovic (changes requested):

jsm.go
kozlovic: Actually I meant to override sio.SubjectFilter to anything that is passed by the user, so likely initialize sio with ">" and then override.
```
sio := streamInfoOpts{SubjectsFilter: ">"}
if o.streamInfoOpts != nil && o.streamInfoOpts.SubjectsFilter != _EMPTY_ {
  sio.SubjectsFilter = o.streamInfoOpts.SubjectsFilter
}
```
This is assuming that this API does not need any of the possible streamInfoOpts except for filter. Otherwise adapt the above.

jsm.go
jnmoyne: Ok so I changed my mind after all about exposing a separate function and instead merged the paging functionality into the existing StreamInfo(): if you pass a value in SubjectFilters in the streamInfoOpts it returns the subject details for the subjects that match (for all the matching subjects and doing the pagination if needed). In the end I think it's the better option/best of both worlds (no new API and as a user I don't need to worry about pagination)

jsm.go
jnmoyne: Yes, sorry I missed that, it's now replaced by the new version of StreamInfo in this last commit

jsm.go
```diff
@@ -701,30 +706,63 @@ func (js *js) StreamInfo(stream string, opts ...JSOpt) (*StreamInfo, error) {
...
+
+	siOpts := StreamInfoRequest{}
 	if o.streamInfoOpts != nil {
-		if req, err = json.Marshal(o.streamInfoOpts); err != nil {
+		siOpts.DeletedDetails = o.streamInfoOpts.DeletedDetails
+		siOpts.SubjectsFilter = o.streamInfoOpts.SubjectsFilter
+	}
+
+	for {
+		siOpts.Offset = i
```
kozlovic, line 721: This is changing the behavior in that it makes all StreamInfo request being a non empty request with an Offset (since I don't think this is omitempty). I would set a boolean in the "if" condition above that at least do the offset setting and marshal ONLY if the o.streamInfoOpts is not nil, or maybe rewrite in a way that you deal with simple info where no subject details is asked first, then deal with the cases where subjects are returned.

jsm.go
```diff
@@ -701,30 +706,63 @@ func (js *js) StreamInfo(stream string, opts ...JSOpt) (*StreamInfo, error) {
...
+		// for backwards compatibility
+		var total int
+		if resp.Total != 0 {
+			total = resp.Total
+		} else {
+			total = len(resp.State.Subjects)
+		}
+
+		if subjectMessagesMap == nil {
+			subjectMessagesMap = make(map[string]uint64, total)
```
kozlovic, line 754: Technically, here even if user did not ask for details, you are now creating this map and returning as a resp.StreamInfo.State.Subjects (empty, but non nil map). More reasons maybe to treat the two situations differently, or at the very least do the creation only if map is nil AND the aforementioned boolean is set (indicating that we were asking for subject details)

kozlovic (changes requested):

kozlovic (approved):

PR #1073, opened 2022-09-10
Some Updates

derekcollison:
As we prep post 2.9 release.


wallyqs (approved):
LGTM

nats.go
```diff
@@ -624,11 +625,30 @@ type Msg struct {
...
+	if m == nil || msg == nil {
+		return false
+	}
+	if m.Subject != msg.Subject || m.Reply != msg.Reply {
+		return false
+	}
+	if !bytes.Equal(m.Data, msg.Data) {
+		return false
+	}
+	return reflect.DeepEqual(m.Header, msg.Header)
```
piotrpio, line 649: What would you think about implementing `Equal()` method for `Header` instead of using `reflect.DeepEqual()`?? It's quite a lot faster and a bit cleaner IMO:
```go
func (h Header) Equal(h2 Header) bool {
	if len(h) != len(h2) {
		return false
	}
	for k, v1 := range h {
		v2, ok := h2[k]
		if !ok {
			return false
		}
		for i, header := range v1 {
			if header != v2[i] {
				return false
			}
		}
	}
	return true
}
```

piotrpio (changes requested):

nats.go
derekcollison: Yep we can do that.

derekcollison: ok done.

piotrpio: I was thinking about extracting header comparison to separate `Equal()` method, as a utility, but as long as we do not need it, your approach is also fine.

piotrpio (approved):
LGTM

go_test.sum
```diff
@@ -22,6 +22,8 @@ github.com/nats-io/jwt/v2 v2.3.0 h1:z2mA1a7tIf5ShggOFlR1oBPgd6hGqcDYsISxZByUzdI=
 github.com/nats-io/jwt/v2 v2.3.0/go.mod h1:0tqz9Hlu6bCBFLWAASKhE5vUA4c24L9KPUUgvwumE/k=
 github.com/nats-io/nats-server/v2 v2.8.5-0.20220803150712-d7847c97c116 h1:NoZ5jkLgMNijnDh96QENq4M06AF34GXlvaYtHGXm/Jk=
 github.com/nats-io/nats-server/v2 v2.8.5-0.20220803150712-d7847c97c116/go.mod h1:3Yg3ApyQxPlAs1KKHKV5pobV5VtZk+TtOiUJx/iqkkg=
+github.com/nats-io/nats-server/v2 v2.9.0 h1:DLWu+7/VgGOoChcDKytnUZPAmudpv7o/MhKmNrnH1RE=
```
kozlovic, line 25: Seeing that there are only additions, I wonder if you ran `go mod tidy -modfile=go_test.mod`?

nats.go
```diff
@@ -2908,7 +2941,14 @@ func (nc *Conn) processMsg(data []byte) {
...
 
 	// FIXME(dlc): Should we recycle these containers?
-	m := &Msg{Header: h, Data: msgPayload, Subject: subj, Reply: reply, Sub: sub}
+	m := &Msg{
+		Subject: subj,
+		Reply:   reply,
+		Header:  h,
+		Data:    msgPayload,
+		Sub:     sub,
+		wsz:     len(data) + len(subj) + len(reply),
```
kozlovic, line 2950: If this is going to be used for maxbytes in pull consumer, then I think that you need to also add the header length (that the server uses to count towards max bytes). The header length should be nc.ps.ma.hdr as shown just above.

test/js_test.go
```diff
@@ -7587,6 +7587,8 @@ func TestJetStreamSubscribeContextCancel(t *testing.T) {
 }
 
 func TestJetStreamClusterStreamLeaderChangeClientErr(t *testing.T) {
+	t.Skip("The 2.9 server changed behavior making this test fail now")
```
kozlovic, line 7590: What is failing? And does that mean that the test is no longer relevant or that there is a breaking change? I don't remember changes in server that would explain that.

kozlovic (changes requested):

nats.go
wallyqs: I thought `data` here would have accounted both headers + data?

kozlovic: Right, sorry about that.

go_test.sum
derekcollison: I thought I did but will check again.

test/js_test.go
derekcollison: Remember we were over zealous in responding from non-leaders in the server and changed that behavior to make some of the server tests pass consistently.

This test now does have timeouts for calls to StreamInfo.

go_test.sum
derekcollison: fixed.

kozlovic (approved):
LGTM

PR #1074, opened 2022-09-12
[ADDED] Listing buckets and bucket names

### LINKED ISSUE #1008: Object Store API: list buckets
## Feature Request

List buckets (Object Store API)

#### Use Case:

Create buckets, store things, come back later, or from a different client.
 
Which buckets exist?

#### Proposed Change:

Similar to list streams:
```
StreamNames(opts ...JSOpt) <-chan string
StreamsInfo(opts ...JSOpt) <-chan *StreamInfo
```
But for buckets, e.g.:
```
BucketNames(opts ...JSOpt) <-chan string
BucketsInfo(opts ...JSOpt) <-chan ObjectStore
```

Anyone creating buckets dynamically without a set-in-stone naming scheme

#### Alternative Approaches

Here's the ugly workaround I came up with. 

```
  bucketNames := []string{}
  for streamName := range js.StreamNames() {
    if strings.HasPrefix(streamName, "OBJ_") {
      bucketName := strings.Replace(streamName, "OBJ_", "", 1)
      bucketNames = append(bucketNames, bucketName)
    }
  }
```

⚠️ One downside of this approach, for anyone else going down this road: expired buckets will still show up in the list. Because the underlying stream does not get deleted.

piotrpio:
Resolves #1008

### DIFF
--- a/object.go
+++ b/object.go
@@ -44,6 +44,10 @@ type ObjectStoreManager interface {
 	CreateObjectStore(cfg *ObjectStoreConfig) (ObjectStore, error)
 	// DeleteObjectStore will delete the underlying stream for the named object.
 	DeleteObjectStore(bucket string) error
+	// BucketNames is used to retrieve a list of bucket names
+	BucketNames(opts ...ObjectOpt) <-chan string
+	// BucketsInfo is used to retrieve a list of buckets
+	BucketsInfo(opts ...ObjectOpt) <-chan ObjectStore
 }
 
 // ObjectStore is a blob store capable of storing large objects efficiently in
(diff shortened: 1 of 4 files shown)


derekcollison (commented):
They Buckets or Objects?

piotrpio:
@derekcollison not sure I get it. These methods list buckets only (as per #1008)

derekcollison:
Apologies, so this is to list all object stores in an account, got it.

piotrpio:
Yep, I think if that looks good it might be a good idea to do the same thing for KV buckets

object.go
```diff
@@ -44,6 +44,10 @@ type ObjectStoreManager interface {
 	CreateObjectStore(cfg *ObjectStoreConfig) (ObjectStore, error)
 	// DeleteObjectStore will delete the underlying stream for the named object.
 	DeleteObjectStore(bucket string) error
+	// BucketNames is used to retrieve a list of bucket names
+	BucketNames(opts ...ObjectOpt) <-chan string
```
derekcollison, line 48: Maybe StoreNames and Stores?

derekcollison:
Naming stuff is hard ;)

object.go
piotrpio: Hm, makes sense as we mostly use `ObjectStore` instead of `Bucket` elsewhere. What about `ObjectStoreNames` and `ObjectStores`? Just to make it more obvious when using `JetStreamContext` interface.

derekcollison: We need to keep StreamsInfo but let's make alias for Streams() and that will be consistent imo.

derekcollison (approved):
LGTM

PR #1078, opened 2022-09-14
Add option to set memory storage to true for a consumer

goku321:
This allows forcing memory storage through sub options while subscribing.

### DIFF
--- a/js.go
+++ b/js.go
@@ -1357,6 +1357,9 @@ func checkConfig(s, u *ConsumerConfig) error {
 	if u.Replicas > 0 && u.Replicas != s.Replicas {
 		return makeErr("replicas", u.Replicas, s.Replicas)
 	}
+	if u.MemoryStorage && !s.MemoryStorage {
+		return makeErr("memory storage", u.MemoryStorage, s.MemoryStorage)
+	}
 	return nil
 }
 
--- a/js.go
+++ b/js.go
@@ -2488,8 +2488,8 @@ func ConsumerReplicas(replicas int) SubOpt {
 	})
 }
 
-// SetMemoryStorage sets the memory storage to true for a consumer.
-func SetMemoryStorage() SubOpt {
+// ConsumerMemoryStorage sets the memory storage to true for a consumer.
+func ConsumerMemoryStorage() SubOpt {
 	return subOptFn(func(opts *subOpts) error {
 		opts.cfg.MemoryStorage = true
 		return nil
--- a/js.go
+++ b/js.go
@@ -2485,6 +2485,14 @@ func ConsumerReplicas(replicas int) SubOpt {
 	})
 }
 
+// SetMemoryStorage sets the memory storage to true for a consumer.
+func SetMemoryStorage() SubOpt {
+	return subOptFn(func(opts *subOpts) error {
+		opts.cfg.MemoryStorage = true
+		return nil
+	})
+}
+
 func (sub *Subscription) ConsumerInfo() (*ConsumerInfo, error) {
 	sub.mu.Lock()
 	// TODO(dlc) - Better way to mark especially if we attach.
--- a/js_test.go
+++ b/js_test.go
@@ -1096,3 +1096,64 @@ func TestJetStreamConvertDirectMsgResponseToMsg(t *testing.T) {
 		t.Fatalf("Wrong header: %v", r.Header)
 	}
 }
+
+func TestJetStreamSetMemoryStorage(t *testing.T) {
+	opts := natsserver.DefaultTestOptions
+	opts.Port = -1
+	opts.JetStream = true
+	s := natsserver.RunServer(&opts)
+	defer shutdownJSServerAndRemoveStorage(t, s)
+
+	nc, js := jsClient(t, s)
+	defer nc.Close()
+
+	if _, err := js.AddStream(&StreamConfig{Name: "STR", Subjects: []string{"foo"}}); err != nil {
+		t.Fatalf("Error adding stream: %v", err)
+	}
+
+	// Pull ephemeral consumer with memory storage.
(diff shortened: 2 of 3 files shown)


js.go
```diff
@@ -1357,6 +1357,9 @@ func checkConfig(s, u *ConsumerConfig) error {
 	if u.Replicas > 0 && u.Replicas != s.Replicas {
 		return makeErr("replicas", u.Replicas, s.Replicas)
 	}
+	if u.MemoryStorage && !s.MemoryStorage {
```
piotrpio, line 1360: I think we should be verifying it both ways, i.e. `true -> false` and `false -> true`. Right now check will pass even if `u.MemoryStorage == false` and `s.MemoryStorage == true`.
```suggestion
	if u.MemoryStorage != !s.MemoryStorage {
```

js_test.go
```diff
@@ -1096,3 +1096,64 @@ func TestJetStreamConvertDirectMsgResponseToMsg(t *testing.T) {
 		t.Fatalf("Wrong header: %v", r.Header)
 	}
 }
+
+func TestJetStreamSetMemoryStorage(t *testing.T) {
+	opts := natsserver.DefaultTestOptions
+	opts.Port = -1
+	opts.JetStream = true
+	s := natsserver.RunServer(&opts)
```
piotrpio, line 1104: You can just use `s := RunBasicJetStreamServer()` instead.

js.go
```diff
@@ -2485,6 +2488,14 @@ func ConsumerReplicas(replicas int) SubOpt {
 	})
 }
 
+// SetMemoryStorage sets the memory storage to true for a consumer.
+func SetMemoryStorage() SubOpt {
```
piotrpio, line 2492: To be more in line with other sub options, I would suggest changing the name to `MemoryStorage()`

piotrpio (changes requested):

js.go
kozlovic: No we can't unfortunately, because we don't have a "3 states" value: true, false, not-specified. The point of this function is to check the user *intent* vs what is already configured in the server. 
We have no way to express that the user does not want MemoryStorage, so if you call js.Subscribe() without specifying anything and the JS consumer happens to be configured on the server with MemoryStorage, you don't want to call to fail. Otherwise we would need an option that says "NoMemoryStorage", etc..

So this function for boolean is not great, but the best we can do is check when the intent was to set a boolean to true, then only we can check that it matches the server setting.

js.go
kozlovic: Yes. Also wonder when the name is "too generic", is going to be a problem if one day we need the same name for a different option? Meaning this is a SubOpt, but what if we need an option one day for JSOpt (or any other). For that reason, should we be more precise in the naming? Something like "ConsumerMemoryStorage()"? (not good, but you get the idea).

js_test.go
```diff
@@ -1096,3 +1096,64 @@ func TestJetStreamConvertDirectMsgResponseToMsg(t *testing.T) {
...
+		t.Fatalf("Error adding stream: %v", err)
+	}
+
+	// Pull ephemeral consumer with memory storage.
+	sub, err := js.PullSubscribe("foo", "", SetMemoryStorage())
+	if err != nil {
+		t.Fatalf("Error on subscribe: %v", err)
+	}
+
+	consInfo, err := sub.ConsumerInfo()
```
kozlovic, line 1120: Instead of repeating this code for the 3 types of consumers, I would have made it a small helper function inline:
```
checkCons := func(sub *Subscription) {
    t.Helper()
    consInfo, err := sub.ConsumerInfo()
    ...
}
```
then you would invoke it after each sub creation.

js_test.go
```diff
@@ -1096,3 +1096,64 @@ func TestJetStreamConvertDirectMsgResponseToMsg(t *testing.T) {
...
+	}
+
+	consInfo, err = sub.ConsumerInfo()
+	if err != nil {
+		t.Fatalf("Error getting consumer info: %v", err)
+	}
+
+	if !consInfo.Config.MemoryStorage {
+		t.Fatalf("Expected memory storage to be %v, got %+v", true, consInfo.Config.MemoryStorage)
+	}
```
kozlovic, line 1158: What you may want to add is a test where a durable is created with MemoryStorage and call a js.SubscribeSync() with Durable(<of that name>) but without the memory storage option in the subscribe call and verifies that it actually works. That would catch the change the Piotr was suggesting that would - in my opinion - break things.

kozlovic (changes requested):

kozlovic:
@goku321 Do you think you could get the changes in soon? We are trying to release today or tomorrow (Friday) and since this is an addition it would be better to go in the 1.17.0 than in a 1.17.1 :-). The test could be changed/updated later, however it would be nice to change the name of the option now as @piotrpio suggested. Thanks!

js.go
goku321: I named it "SetMemoryStorage" because the name "MemoryStorage" is already being used. Should I name it "ConsumerMemoryStorage" as @kozlovic suggested?

goku321:
Sure @kozlovic I can change the name now and will update the tests later :)

js.go
kozlovic: Yes please, go with ConsumerMemoryStorage() then, and this proves my point about possible name clashes unfortunately...

kozlovic:
@goku321 Yes, please update the PR with the option name change and tests can be updated later on. Thanks!

kozlovic (approved):
LGTM

kozlovic:
@piotrpio The changes look good to me. Not sure if you want chime in or approve and merge the PR.

piotrpio (approved):
LGTM as well

PR #1080, opened 2022-09-14
Adjust consumer creation to nats-server v2.9.0

### LINKED ISSUE #1077: Adjust consumer create to server version 2.9.0
Related issue: https://github.com/nats-io/nats-architecture-and-design/issues/148

piotrpio:
Resolves #1077

### DIFF
--- a/js.go
+++ b/js.go
@@ -122,9 +122,19 @@ const (
 	apiAccountInfo = "INFO"
 
 	// apiConsumerCreateT is used to create consumers.
-	apiConsumerCreateT = "CONSUMER.CREATE.%s"
+	// it accepts stream name and consumer name.
+	apiConsumerCreateT = "CONSUMER.CREATE.%s.%s"
+
+	// apiConsumerCreateT is used to create consumers.
+	// it accepts stream name, consumer name and filter subject
+	apiConsumerCreateWithFilterSubjectT = "CONSUMER.CREATE.%s.%s.%s"
+
+	// apiLegacyConsumerCreateT is used to create consumers.
+	// this is a legacy endpoint to support creating ephemerals before nats-server v2.9.0.
+	apiLegacyConsumerCreateT = "CONSUMER.CREATE.%s"
 
 	// apiDurableCreateT is used to create durable consumers.
+	// this is a legacy endpoint to support creating durable consumers before nats-server v2.9.0.
 	apiDurableCreateT = "CONSUMER.DURABLE.CREATE.%s.%s"
 
 	// apiConsumerInfoT is used to create consumers.
@@ -1031,6 +1041,7 @@ func (d nakDelay) configureAck(opts *ackOpts) error {
 // ConsumerConfig is the configuration of a JetStream consumer.
 type ConsumerConfig struct {
 	Durable         string          `json:"durable_name,omitempty"`
+	Name            string          `json:"name,omitempty"`
 	Description     string          `json:"description,omitempty"`
 	DeliverPolicy   DeliverPolicy   `json:"deliver_policy"`
 	OptStartSeq     uint64          `json:"opt_start_seq,omitempty"`
@@ -1621,95 +1632,59 @@ func (js *js) subscribe(subj, queue string, cb MsgHandler, ch chan *Msg, isSync,
 
 	// If we are creating or updating let's process that request.
 	if shouldCreate {
-		j, err := json.Marshal(ccreq)
+		info, err := js.upsertConsumer(stream, cfg.Durable, ccreq.Config)
 		if err != nil {
-			cleanUpSub()
-			return nil, err
-		}
-
-		var ccSubj string
-		if isDurable {
-			ccSubj = js.apiSubj(fmt.Sprintf(apiDurableCreateT, stream, cfg.Durable))
-		} else {
-			ccSubj = js.apiSubj(fmt.Sprintf(apiConsumerCreateT, stream))
-		}
-
-		if js.opts.shouldTrace {
-			ctrace := js.opts.ctrace
-			if ctrace.RequestSent != nil {
-				ctrace.RequestSent(ccSubj, j)
+			var apiErr *APIError
+			if ok := errors.As(err, &apiErr); !ok {
+				cleanUpSub()
+				return nil, err
 			}
-		}
-		resp, err := nc.Request(ccSubj, j, js.opts.wait)
-		if err != nil {
-			cleanUpSub()
-			if err == ErrNoResponders {
-				err = ErrJetStreamNotEnabled
+			if consumer == _EMPTY_ ||
+				(apiErr.ErrorCode != JSErrCodeConsumerAlreadyExists && apiErr.ErrorCode != JSErrCodeConsumerNameExists) {
+				cleanUpSub()
+				if errors.Is(apiErr, ErrStreamNotFound) {
+					return nil, ErrStreamNotFound
+				}
+				return nil, err
 			}
-			return nil, err
-		}
-		if js.opts.shouldTrace {
-			ctrace := js.opts.ctrace
-			if ctrace.ResponseReceived != nil {
-				ctrace.ResponseReceived(ccSubj, resp.Data, resp.Header)
-			}
-		}
-
-		var cinfo consumerResponse
-		err = json.Unmarshal(resp.Data, &cinfo)
-		if err != nil {
-			cleanUpSub()
-			return nil, err
-		}
-		info = cinfo.ConsumerInfo
-
-		if cinfo.Error != nil {
 			// We will not be using this sub here if we were push based.
 			if !isPullMode {
 				cleanUpSub()
 			}
-			if consumer != _EMPTY_ &&
-				(cinfo.Error.ErrorCode == JSErrCodeConsumerAlreadyExists || cinfo.Error.ErrorCode == JSErrCodeConsumerNameExists) {
 
-				info, err = js.ConsumerInfo(stream, consumer)
-				if err != nil {
-					return nil, err
-				}
-				deliver, err = processConsInfo(info, o.cfg, isPullMode, subj, queue)
-				if err != nil {
-					return nil, err
-				}
+			info, err = js.ConsumerInfo(stream, consumer)
+			if err != nil {
+				return nil, err
+			}
+			deliver, err = processConsInfo(info, o.cfg, isPullMode, subj, queue)
+			if err != nil {
+				return nil, err
+			}
 
-				if !isPullMode {
-					// We can't reuse the channel, so if one was passed, we need to create a new one.
-					if isSync {
-						ch = make(chan *Msg, cap(ch))
-					} else if ch != nil {
-						// User provided (ChanSubscription), simply try to drain it.
-						for done := false; !done; {
-							select {
-							case <-ch:
-							default:
-								done = true
-							}
+			if !isPullMode {
+				// We can't reuse the channel, so if one was passed, we need to create a new one.
+				if isSync {
+					ch = make(chan *Msg, cap(ch))
+				} else if ch != nil {
+					// User provided (ChanSubscription), simply try to drain it.
+					for done := false; !done; {
+						select {
+						case <-ch:
+						default:
+							done = true
 						}
 					}
-					jsi.deliver = deliver
-					jsi.hbi = info.Config.Heartbeat
+				}
+				jsi.deliver = deliver
+				jsi.hbi = info.Config.Heartbeat
 
-					// Recreate the subscription here.
-					sub, err = nc.subscribe(jsi.deliver, queue, cb, ch, isSync, jsi)
-					if err != nil {
-						return nil, err
-					}
-					hasFC = info.Config.FlowControl
-					hasHeartbeats = info.Config.Heartbeat > 0
+				// Recreate the subscription here.
+				sub, err = nc.subscribe(jsi.deliver, queue, cb, ch, isSync, jsi)
+				if err != nil {
+					return nil, err
 				}
-			} else {
-				if errors.Is(cinfo.Error, ErrStreamNotFound) {
-					return nil, ErrStreamNotFound
-				}
-				return nil, cinfo.Error
+				hasFC = info.Config.FlowControl
+				hasHeartbeats = info.Config.Heartbeat > 0
 			}
 		} else {
 			// Since the library created the JS consumer, it will delete it on Unsubscribe()/Drain()
@@ -1960,7 +1935,7 @@ func (sub *Subscription) resetOrderedConsumer(sseq uint64) {
 		cfg.DeliverPolicy = DeliverByStartSequencePolicy
 		cfg.OptStartSeq = sseq
 
-		ccSubj := fmt.Sprintf(apiConsumerCreateT, jsi.stream)
+		ccSubj := fmt.Sprintf(apiLegacyConsumerCreateT, jsi.stream)
 		j, err := json.Marshal(jsi.ccreq)
 		js := jsi.js
 		sub.mu.Unlock()
--- a/js.go
+++ b/js.go
@@ -255,6 +255,9 @@ type jsOpts struct {
 	directGet bool
 	// For direct get next message
 	directNextFor string
+
+	// featureFlags are used to enable/disable specific JetStream features
+	featureFlags featureFlags
 }
 
 const (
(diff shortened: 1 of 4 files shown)


jserrors.go
```diff
@@ -98,8 +98,14 @@ var (
 	// ErrCantAckIfConsumerAckNone is returned when attempting to ack a message for consumer with AckNone policy set.
 	ErrCantAckIfConsumerAckNone JetStreamError = &jsError{message: "cannot acknowledge a message for a consumer with AckNone policy"}
 
-	// DEPRECATED: ErrInvalidDurableName is no longer returned and will be removed in future releases
-	// Use ErrInvalidConsumerName instead
+	// ErrRequireServerVersion is returned when using a feature which requires a higher server version.
+	ErrRequireServerVersion JetStreamError = &jsError{message: "invalid server version"}
+
+	// ErrConsumerNameMismtch is returned user provides both consumer name and durable name and they are not equal.
+	ErrConsumerNameMismtch JetStreamError = &jsError{message: "consumer name and durable name do not match"}
```
wallyqs, line 105: typo
```suggestion
	ErrConsumerNameMismatch JetStreamError = &jsError{message: "consumer name and durable name do not match"}
```

test/js_test.go
```diff
@@ -1346,51 +1346,166 @@ func TestJetStreamManagement(t *testing.T) {
...
+				t.Fatalf("ConsumerInfo is not correct %+v", ci)
+			}
+		})
 
-		// configs are the same
-		if _, err = js.AddConsumer("foo", &nats.ConsumerConfig{Durable: "dlc", AckPolicy: nats.AckExplicitPolicy}); err != nil {
-			t.Fatalf("Expected no error; got: %v", err)
-		}
-	})
+		t.Run("ephemeral consumer without name", func(t *testing.T) {
```
wallyqs, line 1440: how about mentioning that this is for the legacy form in the test?
```suggestion
		t.Run("legacy ephemeral consumer without name", func(t *testing.T) {
```

wallyqs (approved):
LGTM

jsm.go
```diff
@@ -259,34 +259,57 @@ type consumerResponse struct {
 
 // AddConsumer will add a JetStream consumer.
 func (js *js) AddConsumer(stream string, cfg *ConsumerConfig, opts ...JSOpt) (*ConsumerInfo, error) {
-	if cfg != nil && cfg.Durable != _EMPTY_ {
-		consInfo, err := js.ConsumerInfo(stream, cfg.Durable)
+	if cfg == nil {
+		cfg = &ConsumerConfig{}
+	}
+	if cfg.Name != _EMPTY_ && !js.nc.serverMinVersion(2, 9, 0) {
```
kozlovic, line 265: I implemented the C client and did not do the version check. Derek commented on that PR that this is not necessary: https://github.com/nats-io/nats.c/pull/580

jsm.go
```diff
@@ -304,13 +327,21 @@ func (js *js) upsertConsumer(stream string, cfg *ConsumerConfig, opts ...JSOpt)
...
-	if cfg != nil && cfg.Durable != _EMPTY_ {
-		if err := checkConsumerName(cfg.Durable); err != nil {
-			return nil, err
+	if consumerName == _EMPTY_ {
+		// if consumer name is empty, use the legacy ephemeral endpoint
+		ccSubj = fmt.Sprintf(apiLegacyConsumerCreateT, stream)
+	} else if err := checkConsumerName(consumerName); err != nil {
+		return nil, err
+	} else if js.nc.serverMinVersion(2, 9, 0) {
+		// if above server version 2.9.0, use the endpoints with consumer name
```
kozlovic, line 336: If not checking the server name, what I think should be done is just in checking if the new config Name is specified. If it is, then that signals the intent to use the new API endpoints. If user specifies Name, but connects to an older server, they will get a timeout: this is not ideal, but I don't think checking server version is reliable anyway (could be connected to one 2.9.0, but server accepting the request be older version, or vice versa).

jserrors.go
```diff
@@ -98,8 +98,14 @@ var (
 	// ErrCantAckIfConsumerAckNone is returned when attempting to ack a message for consumer with AckNone policy set.
 	ErrCantAckIfConsumerAckNone JetStreamError = &jsError{message: "cannot acknowledge a message for a consumer with AckNone policy"}
 
-	// DEPRECATED: ErrInvalidDurableName is no longer returned and will be removed in future releases
-	// Use ErrInvalidConsumerName instead
+	// ErrRequireServerVersion is returned when using a feature which requires a higher server version.
```
kozlovic, line 101: May not be required, see below.

jserrors.go
```diff
@@ -98,8 +98,14 @@ var (
...
+	ErrRequireServerVersion JetStreamError = &jsError{message: "invalid server version"}
+
+	// ErrConsumerNameMismatch is returned user provides both consumer name and durable name and they are not equal.
+	ErrConsumerNameMismatch JetStreamError = &jsError{message: "consumer name and durable name do not match"}
```
kozlovic, line 105: May not be required, see below.

jsm.go
```diff
@@ -259,34 +259,57 @@ type consumerResponse struct {
...
 func (js *js) AddConsumer(stream string, cfg *ConsumerConfig, opts ...JSOpt) (*ConsumerInfo, error) {
-	if cfg != nil && cfg.Durable != _EMPTY_ {
-		consInfo, err := js.ConsumerInfo(stream, cfg.Durable)
+	if cfg == nil {
+		cfg = &ConsumerConfig{}
+	}
+	if cfg.Name != _EMPTY_ && !js.nc.serverMinVersion(2, 9, 0) {
+		return nil, fmt.Errorf("%w: %s", ErrRequireServerVersion, "consumer name requires at least server version 2.9.0")
+	}
+	if cfg.Name != _EMPTY_ && cfg.Durable != _EMPTY_ && cfg.Name != cfg.Durable {
```
kozlovic, line 268: The server already sends an error if sending to the new API endpoint but with a durable in the request that does not match the name in the subject. So not sure we have to do this in the client.

kozlovic: I just updated the C PR with a test that uses different name and durable and verified that server sends an error message already.

kozlovic (changes requested):

jsm.go
piotrpio: I wasn't sure whether we should also add client side validation for this case. I guess you're right, I'll remove it.

kozlovic: It is always a choice between "saving on unnecessary network round-trip" and adding "duplicated" checks in the client library. Since server is doing the check already, I would remove from client.

jsm.go
piotrpio: Makes sense, I'll change it.

jsm.go
```diff
@@ -259,34 +259,57 @@ type consumerResponse struct {
 
 // AddConsumer will add a JetStream consumer.
 func (js *js) AddConsumer(stream string, cfg *ConsumerConfig, opts ...JSOpt) (*ConsumerInfo, error) {
-	if cfg != nil && cfg.Durable != _EMPTY_ {
-		consInfo, err := js.ConsumerInfo(stream, cfg.Durable)
+	if cfg == nil {
+		cfg = &ConsumerConfig{}
+	}
+	if cfg.Name != _EMPTY_ && !js.nc.serverMinVersion(2, 9, 0) {
+		return nil, fmt.Errorf("%w: %s", ErrRequireServerVersion, "consumer name requires at least server version 2.9.0")
```
wallyqs, line 266: not for this PR maybe but we should be doing this style of error wrapping with the `ErrRequireServerVersion` in other places too like: https://github.com/nats-io/nats.go/blob/cc189da40f83a04fa519b36767d496f435b79a85/kv.go#L309

jsm.go
piotrpio: So, just to clarify - when user provides only `Durable`, we should use the old endpoints (`CONSUMER.DURABLE.CREATE`)?
If so, that makes it problematic for e.g. `Subscribe("foo", nats.Durable("cons"))`, as basically it will always use old API subject, unless we add another `SubOpt` for just setting consumer name.

kozlovic: We should then possibly set the Name to the Durable before calling js.AddConsumer(). But then, I agree that we have the situation where this is not a user choice and so connecting to older server would be a problem...

piotrpio: So from what I understand, we have following options:
1. Not verify the server version compatibility and call the new API if `Name` is set - problem here is, `Subscribe()` will always call the old API when combined with `Bind()`/`Durable()` options
2. Do as above, but set `Name` to the value of `Durable` in `Subscribe()` - that means however that we would always be using new API, even for older servers (that's not an option)
3. Verify server version when choosing the right subject in `upsertConsumer()` - here, we might have an issue of connecting to an older server, as you described in the original comment.

Looking at those, I would still lean towards option 3 - I could strip version check in `AddConsumer()`/`UpdateConsumer()`, but leave the version check when choosing the right subject in `upsertConsuper()`. What do you think?

jsm.go
piotrpio: done!

jsm.go
wallyqs: If we want to start testing server compatibility across client versions maybe we could add some additional tests with a different go module to run the older server with new client, something like this maybe: https://github.com/nats-io/nats.go/commit/55ccb7b1f2193e0e9febbf8a23595705b4b7e4a5?diff=unified

scottf: @aricart and I both check the server version so we can use the new endpoints if the user sets only durable.

piotrpio: @wallyqs I like the idea, but I'm a bit whether it won't introduce a horrible mess in the future, with each breaking change on the server requiring us to create separate `go.mod` files to point to the right server version. 

I suppose it might be a good idea to have e.g. a separate repository running clients tests on multiple server versions. That's something we briefly discusses with @Jarema some time ago, we can revisit this.

kozlovic: It would not have been much of an issue if we were passing the ConsumerConfig as an option (instead of individual options that update a ConsumerConfig, although the advantage of options the way we do is that we *could* (but not really do) handle "set" vs "not set" for boolean properties). If the ConsumerConfig was provided, then it would be up to the user to set the ConsumerConfig.Name or not, and that would have been the only driver to select new API or not. Of course, we could add a ConsumerName() SubOpt to do that.

Since the rest of the group seem to be using the server version to send to new API for a durable which consumer config's name is not set, we could do that (and I will update the C client to match that behavior).

piotrpio: `ConsumerName()` is an option for sure, but then we would also have to add some alternative version of `Bind()` which sets stream and name, rather than stream and durable (figuring out a name for this option would be fun for sure).

kozlovic: Go with the server version check then if all agree that this is the best approach we have.

wallyqs: We can still make it possible if needed to pass the ConsumerConfig to configure Subscribe without options by implementing the SubOpt interface, that would avoid having to keep up with ConsumerConfig latest changes.  Something like this would work:

```go
func (config *ConsumerConfig) configureSubscribe(opts *subOpts) error {
// Implements SubOpt interface...
}

// Can use ConsumerConfig struct to configure without extra options.
_, err := js.Subscribe("foo", cb, &nats.ConsumerConfig{
	Durable: "bar",
	Replicas: 1,
})
```

kozlovic (approved):
LGTM

jsm.go
piotrpio: It's a nice option to have (we can consider adding it of course), but I still believe picking the right subject based on server version is useful - if we add it, then the new API will be used by the client as soon as both server and client are updated to the new version (transparently, without needing to change the application code).

ripienaar: The pure version orientated approach means users do not get a chance to say they need time to update infrastructure like ACLs and so forth to start using it.

piotrpio: Hm, not sure how big of an issue that would be vs forcing the user to intentionally start using the new API by changing their application code (IMO this transition should ideally be seamless, as the user of client library I don't need to care about the API subjects). What do you think?

ripienaar: There is a class of user who care deeply to lock down the APIs according to needs. Also a class of user who have spent considerable time in navigating the API subjects for cross account use via imports and exports. 

A major design feature of these API subjects is to enable that lock down. Or to be selectively imported to manage permissions and restrictions. 

The Venn diagrams of users likely to pay for NATS and those who care deeply for subject security probsbly has quite a lot of overlap. 

So you might not care, but I think we should consider if introducing new features in the most user hostile way possible is perhaps not the right thing - especially considering the users most likely affected.

aricart: The issue with the ACL is possibly real, but in reality, the same logic can be applied the other way. If they deploy the new client but don't update the servers, they can update their ACLs and then deploy the servers.

The client possibly could have a way of rejecting the use of the new API (I do for test purposes). But at some point, the clients become too complex and too flexible.

scottf:
If anyone is interested, this is what I do and I think Alberto does pretty much the same:
```
1. If name is provided and server is not 2.9.0 or later, error
2. if name and durable are provided they must match
3. consumer name is either name or durable, whichever is provided. Both can be blank
if consumer name is blank against any server version, use CONSUMER.CREATE.{{stream}}
else if server is 2.9.0 or later
   If filter subject is blank or '>' use CONSUMER.CREATE.{{stream}}.{{consumerName}}
   else use CONSUMER.CREATE.{{stream}}.{{consumerName}}.{{filterSubject}}
else, since consumer name is not blank (so durable) and server is older than 2.9.0, use CONSUMER.DURABLE.CREATE.{stream}.{consumerDurableName}
```

jsm.go
ripienaar: The question is how do users know? Do we have effective communication channels to tell people about these changes? Release notes and blog post are notoriously ineffective - especially as ours tend to be enormous. 

So do we feel users are served well enough by the communication and warning we give them about upcoming changes?

In another world these would be considered  breaking changes and tooling and just human behaviour is aware of major changes. These are not just new features. They majorly change existing code simply because it happens to point at another version server.

As much as I loathe the go major version change behaviour this does demonstrate the utility of that.

aricart (approved):
LGTM

js.go
```diff
@@ -294,6 +297,20 @@ func (opt jsOptFn) configureJSContext(opts *jsOpts) error {
...
 }
 
+type featureFlags struct {
+	useDurableConsumerCreate bool
+}
+
+// UseLegacyDurableConsumers makes JetStream use the legacy (pre nats-server v2.9.0) subjects for consumer creation.
+// If this option is used when creating JetStremContext, $JS.API.CONSUMER.DURABLE.CREATE.<stream>.<consumer> will be used
+// to create a consumer with Durable provided, rather than $JS.API.CONSUMER.CREATE.<stream>.<consumer>.
+func UseLegacyDurableConsumers() JSOpt {
```
ripienaar, line 307: Seems like a good compromise to me, allows people to opt out of a feature should they be in a situation where server updates isnt possible.

Of course in other projects this is opt in -> opt out -> remove flag over 3 releases. But seems this is the best we can do at present.

aricart: The main issue with this is that the default client behavior is still to use the new feature. So really they have to change code, so they could also hold upgrading the server or the client. If there's a flag, my guess is it should be opt-in, rather than opt-out.

ripienaar: My preference would be an opt-in also.  

To your point "they could also hold upgrading the server"....this is just not the case, our default answer to any support question is "upgrade the server", users might not be in a position to cange code etc, hence why I prefer opt-in.

wallyqs: I guess the way these type of changes are rolled out is to wait for a few releases before making the default if there is no alternative, I like the feature flags approach from the PR, maybe we could have eventually an option for the JetStream context to always use latest features instead or to enable a subset of them.

```go
nc.JetStream(nats.UseLatestJSFeatures()) // all opt-in
nc.JetStream(nats.UseLegacyDurableConsumers) // opt-out 
nc.JetStream(&nats.JetStreamFeatures{LegacyDurableConsumers: true, AllowDirect: true})
```

piotrpio: `UseLatestJSFeatures()` would be a nice addition in the future for sure. I would probably lean towards merging this PR as is (with having the opt-out option if needed, to be deprecated and removed in future releases), not to change the behavior vs other clients right now. If we decide we want to have future features like this opt-in, we need to be consistent across all clients. As some of the clients are already released without a feature flag, I don't think it's good to have this discrepancy.

wallyqs (approved):
LGTM

PR #1082, opened 2022-09-14
Add natsProtoErr that can be used with errors.Is

### LINKED ISSUE #1081: Casing issue in library prevents `ErrorIs` from working correctly
## Defect
There is a case-mismatch between nats.go and nats-server that breaks `ErrorIs` from working.
See more discussion at https://natsio.slack.com/archives/C069GSYFP/p1663083344148659

nats.go: v1.16.1-0.20220810192301-fb5ca2cbc995
nats-server: v2.8.5-0.20220831210747-62f91c1dd2cf // 2.9.0-RC.16
MacOS 12.5.1

```
package mytest

import (
	"testing"
	"time"

	server "github.com/nats-io/nats-server/v2/server"
	nats "github.com/nats-io/nats.go"
	"github.com/stretchr/testify/require"
)

// Timeout for request/response.
const Timeout = 2 * time.Second

func startServer(t *testing.T) (s *server.Server) {
	t.Helper()

	s = server.New(&server.Options{
		Host:             "127.0.0.1",
		Port:             server.RANDOM_PORT,
		NoLog:            true,
		NoSigs:           true,
		Authorization: "abc",
		MaxControlLine:   2048,
	})

	go server.Run(s) // nolint:errcheck

	if !s.ReadyForConnections(Timeout) {
		t.Fatal("starting nats server: timeout")
	}
	return
}

func Test_requests(t *testing.T) {

	var server = startServer(t)
	defer server.Shutdown()

	var serverAddress = "nats://" + server.Addr().String() // use server address

	_, err := nats.Connect(serverAddress)
	require.Error(t, err)
	require.ErrorIs(t, err, nats.ErrAuthorization)
}```

#### Expected result:
test passes

#### Actual result:
test fails with 
        	Error:      	Target error should be in err chain:
        	            	expected: "nats: authorization violation"
        	            	in chain: "nats: Authorization Violation"

### PR DESCRIPTION
**Pull request by wallyqs**
The server can send both `-ERR 'Authorization Violation'` or `-ERR 'authorization violation'` depending on some conditions so in the client side we usually try to normalize the behavior by changing to lowercase when checking for an error.
One exception was during the initial connect where the client instead returned the raw error protocol message.

Since this behavior has been in the wild for many years, instead of returning the raw error or the proper error type like `ErrAuthorization` to keep things backwards compatible just in case, this adds a new `natsProtoErr` that can be used with `errors.Is` to check whether the raw protocol message is equivalent to other errors.

Fixes #1081

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -2323,6 +2323,19 @@ func normalizeErr(line string) string {
 	return s
 }
 
+// natsProtoErr represents an -ERR protocol message sent by the server.
+type natsProtoErr struct {
+	description string
+}
+
+func (nerr *natsProtoErr) Error() string {
+	return fmt.Sprintf("nats: %s", nerr.description)
+}
+
+func (nerr *natsProtoErr) Is(err error) bool {
+	return strings.ToLower(nerr.Error()) == err.Error()
+}
+
 // Send a connect protocol message to the server, issue user/password if
 // applicable. Will wait for a flush to return from the server for error
 // processing.
@@ -2377,8 +2390,7 @@ func (nc *Conn) sendConnect() error {
 				// in doReconnect()).
 				nc.processAuthError(authErr)
 			}
-
-			return errors.New("nats: " + proto)
+			return &natsProtoErr{proto}
 		}
 
 		// Notify that we got an unexpected protocol.
--- a/test/auth_test.go
+++ b/test/auth_test.go
@@ -14,6 +14,7 @@
 package test
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 	"sync/atomic"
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: nats.go [hunk 1]
```diff
@@ -2377,8 +2390,7 @@ func (nc *Conn) sendConnect() error {
 				// in doReconnect()).
 				nc.processAuthError(authErr)
 			}
-
-			return errors.New("nats: " + proto)
+			return &natsProtoErr{proto}
```
kozlovic, line 2393: Sure, but have you tried adding a `return authErr` in the "if" statement above? Does that work? Point is that we may not need to do any conversions for the known errors handled in checkAuthError. Your changes is still valid I guess for all other cases?

kozlovic: And by "does it work", I mean the user code could then do `if err == nats.ErrAuthorization` without even have to use the errors.Is().

wallyqs: It does work but then I saw that we also tested for uppercase literal error in other tests like: https://github.com/nats-io/nats.go/blob/main/test/cluster_test.go#L192-L194
and recommendation from Go team is to start using `errors.Is` anyway instead of checking directly to check for wrapped errors as well.

kozlovic (approved):
LGTM

piotrpio (approved):
LGTM

PR #1088, opened 2022-09-19
[ADDED] Listing KeyValue store and KeyValue store names

### LINKED ISSUE #1075: KV API: list buckets and bucket names
Add `KeyValueStores()` and `KeyValueStoreNames()` for listing KV buckets (similarly to #1008)

piotrpio:
Resolves #1075

### DIFF
--- a/kv.go
+++ b/kv.go
@@ -969,25 +969,16 @@ func (kv *kvs) Status() (KeyValueStatus, error) {
 // KeyValueStoreNames is used to retrieve a list of key value store names
 func (js *js) KeyValueStoreNames() <-chan string {
 	ch := make(chan string)
-	ctx, cancel := context.WithTimeout(context.Background(), defaultRequestWait)
 	l := &streamLister{js: js}
 	l.js.opts.streamListSubject = fmt.Sprintf(kvSubjectsTmpl, "*")
-	l.js.opts.ctx = ctx
 	go func() {
-		if cancel != nil {
-			defer cancel()
-		}
 		defer close(ch)
 		for l.Next() {
 			for _, info := range l.Page() {
 				if !strings.HasPrefix(info.Config.Name, "KV_") {
 					continue
 				}
-				select {
-				case ch <- info.Config.Name:
-				case <-ctx.Done():
-					return
-				}
+				ch <- info.Config.Name
 			}
 		}
 	}()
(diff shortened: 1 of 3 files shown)


kv.go
```diff
@@ -980,3 +965,73 @@ func (kv *kvs) Status() (KeyValueStatus, error) {
 
 	return &KeyValueBucketStatus{nfo: nfo, bucket: kv.name}, nil
 }
+
+// KeyValueStoreNames is used to retrieve a list of key value store names
+func (js *js) KeyValueStoreNames() <-chan string {
+	ch := make(chan string)
+	ctx, cancel := context.WithTimeout(context.Background(), defaultRequestWait)
```
wallyqs, line 972: maybe derive this timeout from the ones that are part of the js context? 
https://github.com/nats-io/nats.go/blob/715a5917c806d2858f6fc80663fbf0910ad58ee3/js.go#L234-L238

piotrpio: Looking at this, it seems I can drop creating the context altogether, as `l.Next()` has its own context (either from `js.opts.ctx` or from `js.opts.wait`). I changed it, take a look if it looks all right.

kv.go
```diff
@@ -980,3 +965,55 @@ func (kv *kvs) Status() (KeyValueStatus, error) {
 
 	return &KeyValueBucketStatus{nfo: nfo, bucket: kv.name}, nil
 }
+
+// KeyValueStoreNames is used to retrieve a list of key value store names
+func (js *js) KeyValueStoreNames() <-chan string {
+	ch := make(chan string)
+	l := &streamLister{js: js}
+	l.js.opts.streamListSubject = fmt.Sprintf(kvSubjectsTmpl, "*")
```
wallyqs, line 973: Do you know if this subject will work with the JS Prefix? Or whether it needs something like `js.apiSubj()` to make it so?

piotrpio: This is the filter subject for `STREAM.LIST` api and points to `$KV.%s.>`. AFAIK, `CreateKeyValue()` creates a stream with interest in this subject (`%s` is the name), so there is no need for `$JS` prefix. Or do I misunderstand it?

wallyqs: ahh right, ok then there is no need for the prefix there.

wallyqs (approved):
LGTM

PR #1095, opened 2022-09-27
js: Fix typo in AccountInfo JSON to handle Tiers

wallyqs:
Looking at the schema seems it should have been `tiers`: https://github.com/nats-io/nats-architecture-and-design/issues/120


kozlovic (approved):
LGTM. (hopefully this matters only for serialization, which we are not doing in the client for that one).

piotrpio (approved):
LGTM

PR #1097, opened 2022-10-03
[CHANGED] Ordered consumer recreate on missing heartbeat

piotrpio:
Addresses: https://github.com/nats-io/nats-architecture-and-design/issues/162
Resolves: https://github.com/nats-io/nats.go/issues/1094

This PR changes:
- Ordered consumer will now be recreated on missing heartbeats (failed activity check)
- Timeout or no responders on resetting ordered consumer is no longer considered terminal error - operation will be retried on subsequent failed heartbeat checks

### DIFF
--- a/js.go
+++ b/js.go
@@ -2040,11 +2040,17 @@ func (sub *Subscription) activityCheck() {
 	sub.mu.Unlock()
 
 	if !active {
-		nc.mu.Lock()
-		if errCB := nc.Opts.AsyncErrorCB; errCB != nil {
-			nc.ach.push(func() { errCB(nc, sub, ErrConsumerNotActive) })
+		if !jsi.ordered || nc.Status() != CONNECTED {
+			nc.mu.Lock()
+			if errCB := nc.Opts.AsyncErrorCB; errCB != nil {
+				nc.ach.push(func() { errCB(nc, sub, ErrConsumerNotActive) })
+			}
+			nc.mu.Unlock()
+			return
 		}
-		nc.mu.Unlock()
+		sub.mu.Lock()
+		sub.resetOrderedConsumer(jsi.sseq + 1)
+		sub.mu.Unlock()
 	}
 }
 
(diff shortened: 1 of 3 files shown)


js.go
```diff
@@ -1966,22 +1966,23 @@ func (sub *Subscription) resetOrderedConsumer(sseq uint64) {
...
 		resp, err := nc.Request(js.apiSubj(ccSubj), j, js.opts.wait)
 		if err != nil {
-			if err == ErrNoResponders {
-				err = ErrJetStreamNotEnabled
+			if errors.Is(err, ErrNoResponders) || errors.Is(err, ErrTimeout) {
+				// if creating consumer failed, retry
+				return
 			}
-			pushErr(err)
+			pushErr(fmt.Errorf("%w: recreating ordered consumer", err))
```
kozlovic, line 1973: What is the `%w` doing? Also, since you seem to have this fmt.Errorf() almost every time you call pushErr(), should you have that part of pushErr() itself? (I know there is a check/pushErr in case of request marshaling failing, but that is very unlikely, and still it would be a failure recreating the ordered consumer anyway).

kozlovic (changes requested):

js.go
piotrpio: `%w` is the error wrapping semantic. Thanks to it, if you call e.g. `errors.Is(err, ErrStreamNotFound)` on this error it will unwrap correctly and return true for `fmt.Errorf("%w: recreating ordered consumer", ErrStreamNotFound)`.

As to your second point, you're right, I'll change it.

piotrpio: Additionally, I added the `recreating ordered consumer` part to at least give some insight to the user as to where an error happened (async error callback does not give any information on where the error comes from, so this is better than nothing).
 In the near future I'll be working on the async errors revamp (at least for ordered consumers) to be able to identify them and react more easily)

PR #1099, opened 2022-10-05
[CHANGED] Listing Key Value stores and object stores to return status instead of interface

piotrpio:
This PR addresses https://github.com/nats-io/nats-architecture-and-design/issues/154

This PR changes the signatures of `KeyValueStores()` and `ObjectStores()` methods. For KV, this is not a breaking change since this functionality is yet to be released. For object store, **this is a breaking change** since these methods were released as part of v1.17.0.

### DIFF
--- a/kv.go
+++ b/kv.go
@@ -37,8 +37,8 @@ type KeyValueManager interface {
 	DeleteKeyValue(bucket string) error
 	// KeyValueStoreNames is used to retrieve a list of key value store names
 	KeyValueStoreNames() <-chan string
-	// KeyValueStores is used to retrieve a list of key value stores
-	KeyValueStores() <-chan KeyValue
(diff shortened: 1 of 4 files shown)


kozlovic (approved):
Ok with the change, but since this is a breaking change, I would recommend that we either release v1.18.0 "soon", or we start to post PR against a "dev" branch as opposed to main, so that we keep breaking changes out of main in case we need to provide a patch to existing version before being able to release the next minor release.

I don't think that there is anything wrong releasing v1.18.0 "just" after v1.17.0, especially that most of commits on main right now are [ADDED] or [CHANGED]. So if you think that the next release is going to be v1.18.0 (and not 1.17.1), then go ahead and merge. If not, let's create the "dev" branch and switch this PR against that branch.

ripienaar:
Object store is still experimental so I think it’s fine

PR #1112, opened 2022-10-19
Added support for KV mirrors and sources.

derekcollison:
Also updated state for proper Put() and Get() semantics with mirrors and across domains, e.g. leafnodes.

In practice when a mirror is across a domain, should be named the same as origin. That allows an app to run anywhere without anything special in terms of domains when binding to the KV itself.

### DIFF
--- a/test/kv_test.go
+++ b/test/kv_test.go
@@ -863,3 +863,117 @@ func TestListKeyValueStores(t *testing.T) {
 		})
 	}
 }
+
+func TestKeyValueMirrorCrossDomains(t *testing.T) {
+	conf := createConfFile(t, []byte(`
+		server_name: HUB
+		listen: 127.0.0.1:-1
+		jetstream: { domain: HUB }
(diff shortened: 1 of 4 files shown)


test/kv_test.go
```diff
@@ -863,3 +863,117 @@ func TestListKeyValueStores(t *testing.T) {
...
+	// Create main KV on HUB
+	nc, js := jsClient(t, s)
+	defer nc.Close()
+
+	kv, err := js.CreateKeyValue(&nats.KeyValueConfig{Bucket: "TEST"})
+	expectOk(t, err)
+
+	_, err = kv.PutString("name", "derek")
+	expectOk(t, err)
+	_, err = kv.PutString("age", "22")
```
kozlovic, line 899: you wish ;-)

kozlovic (approved):
LGTM

test/kv_test.go
derekcollison: Ouch! :)

PR #1116, opened 2022-11-01
Use BindStream for watcher so it works with mirrors.

derekcollison:
Updated so connecting to mirror would allow watchers to work.


kozlovic (approved):
LGTM. That's what I had to do in the C client...

aricart (commented):
LGTM

PR #1123, opened 2022-11-05
Release v1.19.1

piotrpio:
### Added
- JetStream:
  - Expose `StreamNameBySubject` method on `JetStreamContext` interface for looking up stream by subject (#1114)
  - Return stream alternates when fetching stream info (#1118) 

### Changed
- KV:
  - Use BindStream for watcher so it works with mirrors (#1116) 

### Fixed
- KV:
  - Delete key not working correctly across mirrors (#1115)


Jarema (approved):
LGTM

PR #1125, opened 2022-11-08
Extract ErrConsumerDeleted in Fetch()

### LINKED ISSUE #1121: Errors returned from server should be lowered cased if passed to application
I encountered this with invalidating pull requests. The header description is properly title case, but the Go client should lower case it if it passes the string up to the application.

piotrpio:
Resolves #1121

### DIFF
--- a/test/js_test.go
+++ b/test/js_test.go
@@ -956,7 +956,7 @@ func TestJetStreamSubscribe(t *testing.T) {
 	}
 }
 
-func TestPullSubsctibeConsumerDeleted(t *testing.T) {
+func TestPullSubscribeConsumerDeleted(t *testing.T) {
 	s := RunBasicJetStreamServer()
 	defer shutdownJSServerAndRemoveStorage(t, s)
 


test/js_test.go
```diff
@@ -956,6 +956,63 @@ func TestJetStreamSubscribe(t *testing.T) {
 	}
 }
 
+func TestPullSubsctibeConsumerDeleted(t *testing.T) {
```
wallyqs, line 959:
```suggestion
func TestPullSubscribeConsumerDeleted(t *testing.T) {
```

wallyqs (approved):
LGTM!

js.go
```diff
@@ -2575,6 +2575,12 @@ func checkMsg(msg *Msg, checkSts, isNoWait bool) (usrMsg bool, err error) {
 			// one message when making requests without no_wait.
 			err = ErrTimeout
 		}
+	case jetStream409Sts:
+		if strings.Contains(strings.ToLower(string(msg.Header.Get(descrHdr))), "consumer deleted") {
```
ripienaar, line 2579: so how are we handling this client facing breaking change?

piotrpio: This status was introduced last week in the server (notifying about consumer being deleted) and we only depend on "consumer deleted" string, so think we can treat this as a new feature: https://github.com/nats-io/nats-server/pull/3605

WDYT?

ripienaar: I meantion the nats.go release being 3 days ago to highlight that the next nats.go release is probably not tomorrow - meaning the delta between current users on a current server and this change langing in a release is likely to be quite big, relatively speaking.  Thus increasing the risk that this will impact real users.

piotrpio: You're right about that. To shorten the delta we could issue a patch release sooner (e.g. tomorrow), minimizing the impact.

PR #1133, opened 2022-11-15
[ADDED] ConnectHandler invoked on initial successful connect

### LINKED ISSUE #1124: Add connect handler
## Feature Request

When clients have `RetryOnFailedConnect=true` set, it would be useful to have a connect handler so we know if the client connected or not. See this discussion:

https://github.com/nats-io/nats.go/discussions/1122

#### Use Case:

Clients that have `RetryOnFailedConnect=true`. We can use this information to know when the client is connected and it's safe to do NATS requests, etc.

#### Proposed Change:

Add a ConnectHandler (similar to Reconnect, Disconnect, and Closed handlers).

Developers writing clients who need to know if we connected.

#### Alternative Approaches

polling nc.Status()

piotrpio:
Resolves #1124

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -375,6 +375,12 @@ type Options struct {
 	// DisconnectedCB will not be called if DisconnectedErrCB is set
 	DisconnectedErrCB ConnErrHandler
 
+	// ConnectedCB sets the connected handler called when the initial connection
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

wallyqs (approved):
LGTM!

PR #1136, opened 2022-11-17
Add ErrKeyExists that is compatible with errors.Is on kv.Create

wallyqs:
Follow up from (https://github.com/nats-io/nats.go/pull/1135)
This changes the returned error on `kv.Create` now to be wrapping the APIError and show error information relevant to the `kv.Create` API as well:

```
nats: wrong last sequence: 1: key exists
```

instead of:

```
nats: wrong last sequence: 1
```

The wrapped error makes it possible to make checks like this as well:

```go
_, err = kv.Create("key", []byte("1"))
if !errors.Is(err, ErrKeyExists) { ... }
```

And also still possible to get the original `APIError` if needed:

```go
_, err = kv.Create("key", []byte("1"))
var aerr *APIError
if errors.As(err, &aerr) { ... }
```

### DIFF
--- a/kv.go
+++ b/kv.go
@@ -635,9 +635,9 @@ func (kv *kvs) Create(key string, value []byte) (revision uint64, err error) {
 
 	// Check if the expected last subject sequence is not zero which implies
 	// the key already exists.
-	var aerr *APIError
-	if errors.Is(err, ErrKeyExists) && errors.As(err, &aerr) {
-		return 0, fmt.Errorf("%w: %s", ErrKeyExists, aerr.Description)
+	if errors.Is(err, ErrKeyExists) {
+		jserr := ErrKeyExists.(*jsError)
+		return 0, fmt.Errorf("%w: %s", err, jserr.message)
 	}
 
 	return 0, err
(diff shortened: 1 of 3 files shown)


kv.go
```diff
@@ -297,7 +297,14 @@ var (
 	ErrKeyDeleted             = errors.New("nats: key was deleted")
 	ErrHistoryToLarge         = errors.New("nats: history limited to a max of 64")
 	ErrNoKeysFound            = errors.New("nats: no keys found")
-	ErrKeyExists              = errors.New("nats: key exists")
+)
+
+// KeyValueError is an error result for KeyValue operations which internally
+// are JetStreamError as well so this is just an alias.
+type KeyValueError = JetStreamError
```
wallyqs, line 304: Not sure we need this, we could just define `ErrKeyExists` as a `JetStreamError` probably

piotrpio: I would opt for keeping it a `JetStreamError` as there may be many `JetStreamErrors` returned from KV operations anyway (e.g. `ErrStreamNotFound` etc.). Unless we want to only return `KeyValueErrors` from kv operations, I suppose we could keep it all as `JetStreamError`. Plus, there is nothing wrong with having the same error for object store, where `ErrKeyExists` would also make sense.

wallyqs: Agree, I removed it for now.

bruth (approved):
This looks solid to me and retains backwards compatibility except for the, hopefully rare/no longer used practice of, doing `if err.Error() == "..." { ... }`.

PR #1147, opened 2022-11-29
Add SkipTLSHandshake optional function to CustomDialer

wallyqs:
Follow up from https://github.com/nats-io/nats.go/pull/1146

This adds a small interface `SkipTLSHandshake() bool` that when implemented a `CustomDialer` may opt into skipping the handshake if not needed by the dialer implementation.

/cc @oderwat

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"net"
 	"time"
 
 	"github.com/nats-io/nats.go"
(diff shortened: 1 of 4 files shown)


piotrpio (commented):
This looks good. Would be nice to have documentation on this though. As `skipTLSDialer` will not be visible in godoc, I would suggest to document this either on `CustomDialer` interface or `CustomDialer` field on `Options`.

piotrpio:
@wallyqs Also, `staticcheck` complains about:
```
ws_test.go:900:2: this value of nc is never used (SA4006)
```

example_test.go
```diff
@@ -44,6 +45,40 @@ func ExampleConnect() {
...
+	//
+	//  func (sd *skipTLSDialer) Dial(network, address string) (net.Conn, error) {
+	//  	    return sd.dialer.Dial(network, address)
+	//  }
+	//
+	//  func (sd *skipTLSDialer) SkipTLSHandshake() bool {
+	//  	    return true
+	//  }
+	//
+	sd := &skipTLSDialer{dialer: &net.Dialer{Timeout: 2 * time.Second}}
```
piotrpio, line 77:
```suggestion
	return sd.skipTLS
}

func ExampleCustomDialer() {
	// Given the following CustomDialer implementation:
	//
	//  type skipTLSDialer struct {
	//  	    dialer  *net.Dialer
	//  	    skipTLS bool
	//  }
	//
	//  func (sd *skipTLSDialer) Dial(network, address string) (net.Conn, error) {
	//  	    return sd.dialer.Dial(network, address)
	//  }
	//
	//  func (sd *skipTLSDialer) SkipTLSHandshake() bool {
	//  	    return true
	//  }
	//
	sd := &skipTLSDialer{dialer: &net.Dialer{Timeout: 2 * time.Second}, skipTLS: true}
```

piotrpio (approved):
LGTM! One check still failing (field `skipTLS` unused), I've added a suggestion.

oderwat:
I want to thank all of you that worked on this and got it upstream so quickly!

Issue #1149, opened 2022-12-01
Replace deprecated calls to io/ioutil


piotrpio:
`io/ioutil` has been deprecated since go 1.16. We should replace its usages across the codebase with counterparts from package `os`. Each deprecated function's documentation points to the up-to-date counterpart.

To find all deprecated references, run:
```
GOFLAGS="-mod=mod -modfile=go_test.mod" staticcheck ./... 
```

cesarvspr:
Working on this:)

cesarvspr:
https://github.com/nats-io/nats.go/pull/1150

cesarvspr:
I think this issue can be closed :)

PR #1150, opened 2022-12-01
chore: remove deprecated usage of "io/ioutil"

cesarvspr:


cesarvspr:
closes https://github.com/nats-io/nats.go/issues/1149

go.mod
```diff
@@ -3,6 +3,9 @@ module github.com/nats-io/nats.go
 go 1.16
 
 require (
+	github.com/golang/protobuf v1.5.2
+	github.com/nats-io/nats-server/v2 v2.9.8
```
wallyqs, line 7: thanks @cesarvspr for the PR! You can avoid changing both `go.mod` and `go.sum` by running the tests as follows:
```
go test -modfile=go_test.mod ./...
```

piotrpio (commented):
Thanks for the contribution! As per @wallyqs comment, please clean up `go.mod` and `go.sum` (they should not contain test-only dependencies), then LGTM 👍

go.mod
cesarvspr: This works, thanks! Any idea why [tests are intermittent](https://github.com/nats-io/nats.go/runs/9834152071) on the pipeline? @wallyqs

piotrpio (approved):
LGTM!

PR #1151, opened 2022-12-02
Provide a way to handle 409 Leadership Changed errors in client

sata-form3:
Without this change, it makes it hard for clients to handle consumer errors relating to leadership changes differently from any other error.

Follows same approach as ErrConsumerDeleted.

### DIFF
--- a/js.go
+++ b/js.go
@@ -2580,6 +2580,11 @@ func checkMsg(msg *Msg, checkSts, isNoWait bool) (usrMsg bool, err error) {
 			err = ErrConsumerDeleted
 			break
 		}
+
+		if strings.Contains(strings.ToLower(string(msg.Header.Get(descrHdr))), "leadership change") {
+			err = ErrConsumerLeadershipChanged
+			break
+		}
 		fallthrough
 	default:
 		err = fmt.Errorf("nats: %s", msg.Header.Get(descrHdr))
--- a/jserrors.go
+++ b/jserrors.go
@@ -102,7 +102,7 @@ var (
 	ErrConsumerDeleted JetStreamError = &jsError{message: "consumer deleted"}
 
 	// ErrConsumerLeadershipChanged is returned when pending requests are no longer valid after leadership has changed
-	ErrConsumerLeadershipChanged JetStreamError = &jsError{message: "leadership changed"}
+	ErrConsumerLeadershipChanged JetStreamError = &jsError{message: "Leadership Changed"}
 
 	// DEPRECATED: ErrInvalidDurableName is no longer returned and will be removed in future releases.
 	// Use ErrInvalidConsumerName instead.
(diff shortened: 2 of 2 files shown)


piotrpio (approved):
LGTM!

js.go
```diff
@@ -2580,6 +2580,11 @@ func checkMsg(msg *Msg, checkSts, isNoWait bool) (usrMsg bool, err error) {
 			err = ErrConsumerDeleted
 			break
 		}
+
+		if strings.Contains(strings.ToLower(string(msg.Header.Get(descrHdr))), "leadership change") {
+			err = ErrConsumerLeadershipChanged
+			break
+		}
```
wallyqs, line 2587: Thanks, we need to revise this codepath eventually to bubble up these errors in a way that is more predictable... These errors are JetStream errors without an error code so maybe not well represented by the `APIError` type, what I'm thinking is that we could wrap this as a new `statusError` type so that below when return the error with an unexpected case.

There are other 409 status codes that we are not catching that may end up in a similar way, probably `NATS/1.0 409 Message Size Exceeds MaxBytes` will end up as "nats: Message Size Exceeds MaxBytes" for example.
Have a wip on trying to addressing some of this in a different PR: https://github.com/nats-io/nats.go/commit/aa8123d2e61317385e8b8c1cdcc65e14b52e78de

wallyqs (approved):
LGTM, thanks for the PR and raising the issue!

js.go
sata-form3: > Thanks, we need to revise this codepath eventually to bubble up these errors in a way that is more predictable... These errors are JetStream errors without an error code so maybe not well represented by the `APIError` type, what I'm thinking is that we could wrap this as a new `statusError` type so that below when return the error with an unexpected case.
> 
> There are other 409 status codes that we are not catching that may end up in a similar way, probably `NATS/1.0 409 Message Size Exceeds MaxBytes` will end up as "nats: Message Size Exceeds MaxBytes" for example. Have a wip on trying to addressing some of this in a different PR: [aa8123d](https://github.com/nats-io/nats.go/commit/aa8123d2e61317385e8b8c1cdcc65e14b52e78de)

Yeah, this PR is just a stop gap measure.

```go
hdr := []byte("NATS/1.0 409 Leadership Change\r\n\r\n")
	for reply := range o.prm {
		o.outq.send(newJSPubMsg(reply, _EMPTY_, _EMPTY_, hdr, nil, nil, 0))
	}
	o.prm = nil
```

[Link](https://github.com/nats-io/nats-server/blob/c4c876129356b785f3883d3a5e8a3b5d9511b979/server/consumer.go#L1913-L1917)

At the moment, the message header and body are just defined as byte slices. I would say this isn't very sufficient structure enough when it comes to managing changes to the protocol or supporting a better error handling without inventing a new binary data protocol. Which is of course possible, but it comes as other costs, such as implementing it in all languages there is a client library for the nats libraries to use.

It would be perhaps good to start using a serialising format which would support backwards and forward compatibility, nested data structures and has library support for a number of languages already, such as Protobuffs, Avro or whatever that fits Nats eco system best. 
Such an approach would even give the possibility of having the error codes explicitly published thus making it very easy for clients to know what errors exist to map them to canonical error types in the language the client library is written in.

sata-form3:
I'm not authorized to press the merge button, can someone do it please? :)

PR #1155, opened 2022-12-05
[IMPROVED] Increase wait time in tests to prevent failing in CI

piotrpio:


derekcollison:
Should not take that long tbh.. Are these failures new?

derekcollison:
Also let's flip travis to go 1.19 and go 1.18.

piotrpio:
@derekcollison will do

PR #1160, opened 2022-12-12
[ADDED] Service api improvements

piotrpio:
Changes to service API:
- comply with ADR
- change package name (`services` -> `service`)
- add custom callback handlers: `DoneHandler` and `ErrorHandler`
- add wrapper around `nats.Msg` (`Request`) to enable adding helper methods
- change `Service` to a struct (instead of `interface`)
- add documentation and examples
- add more tests
- general cleanup

Git diff unfortunately displays this diff as added/removed files, because of the changed package/dir name (I wanted to make it singular).


service/service.go
```diff
@@ -31,38 +31,28 @@ import (
...
+	// Service represents a configured NATS service.
+	// It should be created using [Add] in order to configure the appropriate NATS subscriptions
+	// for request handler and monitoring.
+	Service struct {
+		// Config contains a configuration of the service
+		Config
 
-	// Service is an interface for service management.
-	// It exposes methods to stop/reset a service, as well as get information on a service.
-	Service interface {
```
derekcollison: Why change away from an interface?

service/example_package_test.go
```diff
@@ -0,0 +1,70 @@
...
+			// service handler
+			Handler: incrementHandler,
+			// a unique subject serving as a service endpoint
+			Subject: "numbers.increment",
+		},
+	}
+	// Multiple instances of the servcice with the same name can be created.
+	// Requests to a service with the same name will be load-balanced.
+	for i := 0; i < 5; i++ {
+		svc, err := Add(nc, config)
```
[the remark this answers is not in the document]
piotrpio: It's in separate package now, so the way you actually use it is by calling `service.Add()`, which I think is really intuitive and pretty much a standard in Go (e.g. `list.New()` or `context.WithTimeout()`, `http.Server()` etc.).

In that sense, `service.AddService()` seems redundant.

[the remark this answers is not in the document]
derekcollison: I thought you would hang it off alive connection. nc.AddService. This is how we register interest with a connection, and that is what a service it imo. Do you now make them pass in a nats.Connection?

service/service.go
piotrpio: I was wondering whether to better have interface or not, but my general reasoning is:
- we don't really need an interface here - unlike in JetStream simplification, where we'll probably want to support Push and Pull consumers separately (and possibly ordered consumer as well) on one interface, here we have 1 implementation
- by not using interfaces, we enable other users to add their own interfaces on top of `Service()` however they want, and we do not break anything for them if we want to add more methods
- core NATS client does not rely on any interface, so while it's not 100% consistent (`JesStreamContext` is an interface), we do not break conventions

service/example_package_test.go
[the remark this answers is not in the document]
piotrpio: Yes - I want to separate services to a separate, importable package - if user does not want to use service framework, the API is less cluttered and they do not see methods they do not need. After importing `service` package and typing `service.` in IDE, you'll see all you need to start using interfaces. It's self-documenting and easier to use.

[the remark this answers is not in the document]
derekcollison: We should discuss, we had stated we did not want this as a separate package IIRC because we wanted this to be the default way to do micro-services etc.

Let's schedule some time to discuss before this gets merged.

service/service.go
derekcollison: I think lately folks have pushed us more towards interfaces. I agree we should optimize for extensibility. 

@ripienaar What are your thoughts?

service/example_package_test.go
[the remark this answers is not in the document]
piotrpio: Example of adding a new service looks sth like this:
```go
	nc, err := nats.Connect("127.0.0.1:4222")
	if err != nil {
		log.Fatal(err)
	}

	config := Config{
		Name:     "EchoService",
		Endpoint: Endpoint{Subject: "echo", Handler: echoHandler},
	}

	srv, err := service.Add(nc, config)
	if err != nil {
		log.Fatal(err)
	}
```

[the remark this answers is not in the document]
piotrpio: Sure, let's discuss this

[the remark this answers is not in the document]
wallyqs: The part where `nc.AddService` may come in handy is when you have multiple services on top of the same connection and you want to be able to use the error handling on a per service basis, since `nc.Opts.AsyncErrorCB` belongs to the connection, in current PR it seems that only done Service would be able to use the error handler.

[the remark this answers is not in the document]
piotrpio: Current service implementation also has `ErrorHandler`, which extends `AsyncErrorCB`.

service/service.go
ripienaar: If you want people to wrap it or augment with additional abilities without changing code then interfaces help.  So for KV for example its an interface and we can wrap a cache, codec, e2e encryption or whatever we like - those just wrap the interface and any other calling code doesnt need to change.

So if that kind of thing isnt a goal here then you can be more fluid about it. 

Regarding core nats package - Msg is a great example of someting that really should have been an interface. Rather than the way things are now with a Msg struct with a ton of methods on it thats just never used by core nats

service/example_package_test.go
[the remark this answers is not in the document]
ripienaar: This is the kind of package that will sprawl with dependencies like observability and more over time.  Might be worth keeping in a seperate package to avoid those being in nats.go core

derekcollison:
I suggest we have it as a separate package, but similar to JetStream its constructor is in the core pkg.

So nc.AddService().

I also think its best that service goes back to an interface.

/cc @ripienaar @aricart

service/service.go
```diff
@@ -0,0 +1,493 @@
...
+	if nc.Opts.AsyncErrorCB != nil {
+		nc.SetErrorHandler(func(c *nats.Conn, s *nats.Subscription, err error) {
+			if config.ErrorHandler != nil {
+				config.ErrorHandler(svc, &NATSError{
+					Description: err.Error(),
+					Subject:     s.Subject,
+				})
+			}
+			svc.Stop()
+			svc.natsHandlers.asyncErr(c, s, err)
```
wallyqs, line 293: wondering about possible race conditions from grabbing `nc.Opts.AsyncErrorCB` without the lock, `SetErrorHandler` does grab the lock in the client internally that part should be ok.
So the approach is to apply reconfigure the error handler and closing over the original error callback, then internally in the client when the async error callback dispatcher calls it, it will execute the svc error handler and then the nats error handler serially.

service/service.go
```diff
@@ -0,0 +1,493 @@
...
+	} else {
+		nc.SetClosedHandler(func(c *nats.Conn) {
+			svc.Stop()
+		})
+	}
+
+	svc.natsHandlers.asyncErr = nc.Opts.AsyncErrorCB
+	if nc.Opts.AsyncErrorCB != nil {
+		nc.SetErrorHandler(func(c *nats.Conn, s *nats.Subscription, err error) {
+			if config.ErrorHandler != nil {
```
wallyqs, line 286: I think here you need to check whether the `s *nats.Subscription` belongs to one of the subscriptions that were mounted by the Service (reqSub, etc...), as the error handler would apply to any other subscription from the client.

service/service.go
```diff
@@ -0,0 +1,493 @@
...
+				config.ErrorHandler(svc, &NATSError{
+					Description: err.Error(),
+					Subject:     s.Subject,
+				})
+			}
+			svc.Stop()
```
wallyqs, line 292: should this one be run in a goroutine? What could be interesting here is to check whether the error is a SlowConsumer error for example to have it drain, and maybe let the service resubscribe once its buffers have caught up.

service/example_package_test.go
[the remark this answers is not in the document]
Jarema: @ripienaar You have a good point here, but unfortunately, separate packages have the same `go.mod`.

We would need a multi-module setup. That brings it's own problems and complexity (especially with incompatible versions of dependencies).

service/service.go
aricart: interface = can change internal details or have alternative implementations = win.

piotrpio:
@derekcollison I agree that changing `Service` back to an interface may be useful, so I'll do that.

As to having constructor as a method on `*nats.Conn` (`nc.AddService`), that's harder because we  need to deal with cyclic dependencies (which go does not allow). The whole structure of what we want to achieve suggests that core nats connection can be created without any knowledge of a service, but service cannot be created without knowledge of core nats connection. To me, that classifies it as a separate package, dependent on core nats. 

Creating a wrapper around an existing type, even reusing some kind of connection in it, by passing the connection as an argument is not uncommon in go, neither in standard library nor in other popular libraries. Some examples:
- `crypto/tls` takes `net.Conn` as argument to create new TLS connection:

```go
func Client(conn net.Conn, config *Config) *Conn
```

- `net/smtp` package uses `net.Conn` to create a new smtp client: 

```go
func NewClient(conn net.Conn, host string) (*Client, error)
```

- `gorilla/websocket` (https://github.com/gorilla/websocket), the most popular go websocket library, also uses `net.Conn` to create a websocket connection: 

```go
func NewClient(netConn net.Conn, u *url.URL, requestHeader http.Header, readBufSize, writeBufSize int) (c *Conn, response *http.Response, err error)
```

Ultimately, we do a similar thing - we create a new layer on top on `nats.Conn` in order to enhance capabilities.

wallyqs:
@piotrpio that's a good point on cyclic dependencies, it will not be possible to separate concerns into two packages and also have something like `nc.AddService(&service.Config{})` since the queue subscription async handler has to be concrete type like `func(msg *Msg)`.  If we had way to be able to create async callbacks in the core client without using concrete types (e.g. an interface `Msg` type), then you would be able to abstract most of the client and subscription behavior into smaller interfaces as needed and then be able to separate into two packages.

piotrpio:
@wallyqs yeah, I was thinking about relying on interfaces in `service` package, but because of the concrete types (message handlers, but also `*nats.Subscription` being returned by `Subscribe()`), that will not be possible...

derekcollison:
ok we can discuss, but feel it should be `nc.AddService`

wallyqs:
to get to `nc.AddService` and still be able to have another package, we could add a few extra small interfaces to the service package that represent the needed behavior from the client, then implement the internals of `service` package based on those: https://github.com/nats-io/nats.go/commit/b169849d659c3f824861ef97533f493abaab6972

piotrpio:
@wallyqs Yeah, I tried that approach yesterday and while it's doable, I don't like it for a couple of reasons:

- Even though the interface implementation would be hidden in core nats client, we would still need to expose those extra interfaces in `services` package. For experimental phase that would be OK I assume, but we don't know how long this will be beta and when nats client will support interfaces by default (which would allow us to remove them eventually). I don't like exporting interfaces which make sense only for one, internal use case.

- I'm worried about the development and maintanance of services if we basically split the logic across 2 packages. In the future API will be getting new features, and a lot of them will require adding and implementing new methods for this interface in core nats package. I'm worried that this will become very cumbersome to maintain + hard to read and understand if some of the newer folks will want to contribute.

- The interfaces you proposed in your POC are minimal, but in reality we would have to add at least `Subscribe()`, but also some methods to be able to modify (extend and restore) callbacks for core client (cancel handler and async error handler). The logic for those handlers would probably have to reside in `nats` package.

service/service.go
piotrpio: I'm not sure whether it should run in a goroutine... It is added in async handler anyway, I would probably keep it synchronous within one callback.

As to detecting e.g. slow consumer error (maybe others), I would do that in a separate PR if that's ok, just to make this one works properly and then add improvements.

service/service.go
piotrpio: Good catch - done.

service/service.go
piotrpio: You're right, that may be a problem. What do you think about adding new methods on connection to get the handlers with a lock (similar to what we have for `Set...`):
```go
func (nc *Conn) ErrorHandler() ErrHandler {
	if nc == nil {
		return nil
	}
	nc.mu.Lock()
	defer nc.mu.Unlock()
	return nc.Opts.AsyncErrorCB
}
```

micro/service.go
```diff
@@ -0,0 +1,554 @@
...
+		Name:    config.Name,
+		ID:      id,
+		Version: config.Version,
+	}
+
+	infoHandler := func(req *Request) {
+		response, _ := json.Marshal(svc.Info())
+		if err := req.Respond(response); err != nil {
+			if err := req.Error("500", fmt.Sprintf("Error handling INFO request: %s", err)); err != nil && config.ErrorHandler != nil {
+				go config.ErrorHandler(svc, &NATSError{req.Subject, err.Error()})
```
wallyqs, line 261: This would spin up as many go routines as errors appear, in the nats.go these are serialized over a channel to be able to have some control over these: https://github.com/nats-io/nats.go/blob/main/nats.go#L2769-L2792

micro/service.go
```diff
@@ -0,0 +1,554 @@
...
+			}
+		}
+	}
+
+	pingHandler := func(req *Request) {
+		response, _ := json.Marshal(ping)
+		if err := req.Respond(response); err != nil {
+			if err := req.Error("500", fmt.Sprintf("Error handling PING request: %s", err)); err != nil && config.ErrorHandler != nil {
+				go config.ErrorHandler(svc, &NATSError{req.Subject, err.Error()})
```
wallyqs, line 270: maybe could use `err error` as part of the state from `NATSError`? That way later on could be able to use `errors.Is(err, ErrSlowConsumer)` and handle that case for example

micro/service.go
piotrpio: Changed implementation to use an error dispatcher per-service

micro/service.go
piotrpio: I would leave it as it is for now, `ErrSlowConsumer` would be handled internally anyway. We could add it later on if needed.

micro/service.go
```diff
@@ -0,0 +1,592 @@
...
+		return "SCHEMA"
+	default:
+		return ""
+	}
+}
+
+// AddService adds a microservice.
+// It will enable internal common services (PING, STATS, INFO and SCHEMA) as well as
+// the actual service handler on the subject provided in config.Endpoint
+// A service name and Endpoint configuration are required to add a service.
```
Jarema, line 220:
```suggestion
// A service name, version and Endpoint configuration are required to add a service.
```

Jarema (approved):
LGTM!

PR #1169, opened 2022-12-22
Add note on disabling reconnect buffer

wallyqs:
Always forget whether `nats.ReconnectBufSize(0)` or `nats.ReconnectBufSize(-1)` disables it so adding to the docs.

- `nats.ReconnectBufSize(0)`: Uses default reconnect buf size
-  `nats.ReconnectBufSize(-1)`: Disables and throws outbound buffer limit error when publishing while disconnected


bruth (approved):

piotrpio (approved):
LGTM!

PR #1172, opened 2022-12-24
Create a generic "nats" error

sylr:
This allows to catch all NATS error client side with:

```go
if errors.Is(err, nats.ErrNats) {
    ...
}
```


wallyqs:
Thanks for the contribution, there some cleanup that we need to do for the errors to better take advantage of error wrapping and the error interface.  My opinion here is that it should be a `nats.Error` interface type somewhat like the [net.Error type](https://pkg.go.dev/net#Error) (minus `Temporary()`) kind of like this: 

```go
	ErrConnectionClosed Error = &natsErr{msg: "connection closed"}
	ErrDrainTimeout     Error = &natsErr{msg: "draining connection timed out", timeout: true}
	ErrTimeout          Error = &natsErr{msg: "timeout", timeout: true}
```

sylr:
go1.20 comes with multi error wrapping so you could combine multiple errors such as ErrNats and ErrTimeout in subsequent and relevant errors.

wallyqs:
@sylr yes! I'm looking forward to that change and how we could adopt, btw what do you think is the main use case to be able to match against all errors from the package? Seems would be mostly for debugging or logging since the generic error does not give enough info about how to handle the error?

sylr:
It's pretty hard to know which errors could be returned by nats sdk's methods so a generic catchable error can be used to implement a generic fallback if nats errors. In the end you could end up implementing more specific error handling based on the full error but I think a generic one is a good way to start.

PR #1176, opened 2022-12-30
[FIXED] Pass opts to ConsumerInfo call in AddConsumer

### LINKED ISSUE #1161: JetStreamContext AddConsumer calls ConsumerInfo without passing context
## Defect

`nats.go 1.21.0`
`nats-server 2.9.9`

server is running using `nats:2.9.9-alpine` docker

When using the `AddConsumer` with a custom context (by passing `nats.Context(...)` as jsOpts), with a durable name, which means it needs to check it's config against the server. This can be seen at this code:
https://github.com/nats-io/nats.go/blob/6c6add8d63597f84bee75d37bb1520e01552a02d/jsm.go#L312-L326

But the call to `js.ConsumerInfo` isn't passed the context, which results in it using the defaults of the configured jetstream. As a result, I can't control manually the context and cancel it, even thought the function call looks like it should do it.

```go
	consumer, err := js.AddConsumer("A", &nats.ConsumerConfig{
		Durable:        consumerName,
		FilterSubject:  subject,
	}, nats.Context(ctx))
```

#### Expected result:

As long as I don't cancel the context, the client should error with `context deadline reached`.

#### Actual result:

If for 5 seconds (the default timeout interval) the server didn't respond, the context is cancel and `AddConsumer` fails.

piotrpio:
Resolves #1161

### DIFF
--- a/jsm.go
+++ b/jsm.go
@@ -310,7 +310,7 @@ func (js *js) AddConsumer(stream string, cfg *ConsumerConfig, opts ...JSOpt) (*C
 		consumerName = cfg.Durable
 	}
 	if consumerName != _EMPTY_ {
-		consInfo, err := js.ConsumerInfo(stream, consumerName)
+		consInfo, err := js.ConsumerInfo(stream, consumerName, opts...)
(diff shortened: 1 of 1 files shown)


Jarema (approved):
LGTM

PR #1179, opened 2023-01-04
Implement RequestMsgWithContext

DavidS-om:
When trying to inject headers (e.g. for OpenTelemetry), this method is missing. If this approach is acceptable, I can add some testing. I'd appreciate if someone more familiar with the code base could take the tests off my hands tho.

### DIFF
--- a/context.go
+++ b/context.go
@@ -215,6 +215,38 @@ func (nc *Conn) FlushWithContext(ctx context.Context) error {
 	return err
 }
 
+// RequestMsgWithContext will create an Inbox and perform a Request
+// using the provided cancellation context with the Inbox reply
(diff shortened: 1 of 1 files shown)


DavidS-om:
An implementation alternative could be `func (c *EncodedConn) RequestWithContext(ctx context.Context, subject string, hdr Header, v interface{}, vPtr interface{}) error` and build the `Msg` instance in the function. That might be cleaner.

wallyqs:
Thanks @DavidS-om for the PR, we try to not extend further the `encoders` package functionality since it might get deprecated soon. I'd recommend for now instead use `RequestMsgWithContext` and do the serializing outside of nats.

PR #1181, opened 2023-01-04
Check whether credentials are missing before using them

wallyqs:
On `nats.Connect` only the last error when attempting to connect will be reported, so a failure to connect with a bad configuration will result in `no servers available` in case there are multiple servers discovered or configured. In case of using credentials, this will hide issues like not configuring correctly the path to the credentials and just getting an Authorization Error for example as well.
To improve usability we can check whether the creds have been configured correctly when processing the option rather than waiting to connect (have stumbled upon this issue a couple of times so this should improve usability).

Also started to adopt wrapping errors syntax to check for the underlying io error.


bruth (approved):
Looks great!

PR #1185, opened 2023-01-10
[OPTIMIZE] optimize add consumer

jnmoyne:
Optimize add consumer to not re-define the consumer if the consumer config is equal to the current consumer config


piotrpio (approved):
LGTM!

PR #1186, opened 2023-01-10
Add context to microservice api

eventgraph:
In trying to move existing services to use the `micro` package the lack of `context` has made that difficult.  From logging, cancellation, error handling, tracing, etc `context` has made handling cross cutting concerns much easier.

In the examples given there isn't much need for it other than logging.  I moved those logs to use the upcoming `slog` package with its built in support for get/set on context.

Note the `go.mod` was changed to include `slog` but pulled in a bunch of NATs other dependencies that aren't directly called out on the `go.mod` file.  Not sure the right way to handle this.


micro/example_test.go
```diff
@@ -43,56 +47,81 @@ func ExampleAddService() {
...
+		os.Exit(1)
 	}
 	defer nc.Close()
 
 	config := micro.Config{
 		Name: "EchoService",
 		Endpoint: micro.Endpoint{
 			Subject: "echo",
-			Handler: micro.HandlerFunc(func(micro.Request) {}),
+			Handler: micro.HandlerFunc(func(context.Context, micro.Request) {}),
```
wallyqs, line 98: I don't think it might be necessary to be changing these, what could be done is take advantage of the interface func here and have a wrapper for the handler, maybe something like:

```go 
func WithLoggingContext(ctx context.Context, h micro.Handler) micro.Handler {
	return micro.HandlerFunc(func(req micro.Request) {
             logger := slog.FromContext(ctx) // do context things
             logger.Info(...)
             h(req)
	})
}
```

What I do see is that you want to be able to inspect whether the response was successful or not after responding, maybe there could be a way to inspect via the interface whether the internal `respondError` for the request.

go.mod
```diff
@@ -3,8 +3,19 @@ module github.com/nats-io/nats.go
 go 1.18
 
 require (
+	github.com/golang/protobuf v1.5.2
+	github.com/nats-io/nats-server/v2 v2.9.11
 	github.com/nats-io/nkeys v0.3.0
 	github.com/nats-io/nuid v1.0.1
+	golang.org/x/exp v0.0.0-20230108222341-4b8118a2686a
```
wallyqs, line 10: there is a different `go_test.mod` where you can add test dependencies and can be called via `go test -modfile=go_test.mod`

micro/example_test.go
eventgraph: Good point, happy to change. I wasn't sure about the details of how it should be handled, just didn't want to make an issue with at least putting some effort into a version that passes the current tests.

eventgraph: How does this help if you need to use the context from within the handler?

go.mod
eventgraph: fixed

micro/example_test.go
wallyqs: Something like this?
```go
func WithContext(h func(context.Context, micro.Request)) micro.Handler {
	return micro.HandlerFunc(func(req micro.Request) {
		h(ctx, req)
	})
}
```

wallyqs: You could do something like this instead:
```go
func WithContext(ctx context.Context, h func(context.Context, micro.Request)) micro.Handler {
	return micro.HandlerFunc(func(req micro.Request) {
		h(ctx, req)
	})
}

func ExampleAddService_Custom() {
	nc, err := nats.Connect("127.0.0.1:4222")
	if err != nil {
		log.Fatal(err)
	}
	defer nc.Close()

	echoHandler := func(ctx context.Context, req micro.Request) {
		log.Println("With Context:", ctx.Value("logging"))
		req.Respond(req.Data())
	}

	ctx := context.Background()
	ctx = context.WithValue(ctx, "logging", "example")

	config := micro.Config{
		Name:        "EchoService",
		Version:     "1.0.0",
		Description: "Send back what you receive",
		Endpoint: micro.Endpoint{
			Subject: "echo",
			Handler: WithContext(ctx, echoHandler),
		},
}
```

go_test.mod
```diff
@@ -8,6 +8,7 @@ require (
 	github.com/nats-io/nkeys v0.3.0
 	github.com/nats-io/nuid v1.0.1
 	google.golang.org/protobuf v1.23.0
+	golang.org/x/exp v0.0.0-20230108222341-4b8118a2686a
```
wallyqs, line 11: calling `go mod tidy -modfile=go_test.mod` may help here with the `go_test.sum`

micro/example_test.go
piotrpio: Agreed, I think leveraging `Handler` interface to pass basically anything you want to a handler (context, just logger etc.) would be the way to go here. 

I was wondering whether we could mimic what HTTP does with context embedded in `http.Request` and then retrieving it using `req.Context()`, but I am not convinced (as @eventgraph mentioned, that does not sit well with how you would use context in go).

eventgraph: @piotrpio we don't really care where it lives, just that I have a context to work with.  Context as the first argument is the norm, the addition on `http.Request` I think was the only way to keep the interface the same without breakage as Context didn't come about for a while.  Since this is a new alpha contract my hope is you can fix it to be "right".

eventgraph: Would you like me to close this PR and make a new Issue / Feature Request?

piotrpio: Hey, sorry for the delayed response. I created an issue (https://github.com/nats-io/nats.go/issues/1196) and will implement soon. I'm closing this PR for now.

PR #1188, opened 2023-01-13
[CHANGED] Service schemas

piotrpio:
- updated service schemas configuration according to ADR
- fixed empty fields in STATS response

### DIFF
--- a/micro/service.go
+++ b/micro/service.go
@@ -17,6 +17,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net/url"
 	"regexp"
 	"strings"
 	"sync"
@@ -74,6 +75,7 @@ type (
 
 	endpointOpts struct {
 		subject string
+		schema  *Schema
 	}
 
 	// ErrHandler is a function used to configure a custom error handler for a service,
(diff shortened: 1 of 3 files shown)


micro/service.go
```diff
@@ -482,6 +497,9 @@ func (c *Config) valid() error {
 	if !semVerRegexp.MatchString(c.Version) {
 		return fmt.Errorf("%w: version: version should not be empty should match the SemVer format", ErrConfigValidation)
 	}
+	if _, err := url.Parse(c.APIURL); err != nil {
```
aricart, line 500: I know we call it as api URL, but here we are making it so that it must be - imho this value should just be passthrough.

aricart (approved):
LGTM - just a minor nit

ripienaar (approved):
LGTM, will defer to yous on the URL thing, I think a URL needs to be a URL so like the current approach

micro/service.go
piotrpio: I think if it's `api_url`, we should validate... `url.Parse` in go a rather loose validation, allowing no schema, relative urls etc., so it should give some flexibility but disallow clearly invalid strings (newlines, urls starting with `:` etc.)

PR #1190, opened 2023-01-17
js: use custom inbox prefix in async reply

### LINKED ISSUE #1189: Unused NATS Inbox Prefix
## Defect

 - [ ] Included a [Minimal, Complete, and Verifiable example] (https://stackoverflow.com/help/mcve)
 
 Hello,
 
 When using Inbox Prefix, some requests are not using Inbox Prefix. It generates error like this as we're not allowing user to subscribe to _INBOX.>, only _INBOX.PREFIX.>:
 
 ```
nats: Permissions Violation for Subscription to "_INBOX.1bl31l.*"
nats: Permissions Violation for Subscription to _INBOX.rzC14y.*
nats: Permissions Violation for Subscription to "_INBOX.I2Ytl7.*"
```

It looks like this code is using a const InboxPrefix and not the custom one: https://github.com/nats-io/nats.go/blob/32b9daa3d2eb1d3f04ae7cd30b12bef1d4587baa/js.go#L655 Maybe it's the issue.

- Nats.go: 1.22.1
- Nats Server: 2.9.0

Ubuntu 20.04 / Containerd

Currently, we don't know exactly what part of code generates this violation. NATS Conn is initialised using:

```go
natsConn, err := nats.Connect(
    url,
    nats.CustomInboxPrefix("_INBOX.PREFIX")),
    [...]
)
```

#### Expected result:

All requests to Nats are using Custom Inbox Prefix.

#### Actual result:

Some queries are not using Custom Inbox Prefix.

guyguy333:
This PR uses Custom Inbox prefix in JS async reply. It should fix #1189 

Signed-off-by: Guillaume Delbergue <guillaume.delbergue@hiventive.com>

### DIFF
--- a/js.go
+++ b/js.go
@@ -652,7 +652,11 @@ func (js *js) newAsyncReply() string {
 		for i := 0; i < aReplyTokensize; i++ {
 			b[i] = rdigits[int(b[i]%base)]
 		}
-		js.rpre = fmt.Sprintf("%s%s.", InboxPrefix, b[:aReplyTokensize])
+		inboxPrefix := InboxPrefix
+		if js.nc.Opts.InboxPrefix != _EMPTY_ {
+			inboxPrefix = js.nc.Opts.InboxPrefix + "."
+		}
+		js.rpre = fmt.Sprintf("%s%s.", inboxPrefix, b[:aReplyTokensize])
 		sub, err := js.nc.Subscribe(fmt.Sprintf("%s*", js.rpre), js.handleAsyncReply)
 		if err != nil {
 			js.mu.Unlock()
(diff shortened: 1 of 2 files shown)


wallyqs (commented):
LGTM, could you cherry pick or include the test code from this commit into your branch? https://github.com/wallyqs/nats.go/commit/31c48cd0a9a8c9d5ba86d0ce09a9996e2feb1c79

guyguy333:
> LGTM, could you cherry pick or include the test code from this commit into your branch? [wallyqs@31c48cd](https://github.com/wallyqs/nats.go/commit/31c48cd0a9a8c9d5ba86d0ce09a9996e2feb1c79)

Sure, I added your commit. Thanks

wallyqs (approved):

PR #1200, opened 2023-01-28
[ADDED] Support for stream subject transform

jnmoyne:
Adds support for all stream subject transform functionalities and thereby enabled KV bucket sourcing


jsm.go
```diff
@@ -136,6 +139,12 @@ type StreamConfig struct {
 	MirrorDirect bool `json:"mirror_direct"`
 }
 
+// SubjectTransformConfig is for applying a subject transform (to matching messages) before doing anything else when a new message is received
```
piotrpio, line 142:
```suggestion
// SubjectTransformConfig is for applying a subject transform (to matching messages) before doing anything else when a new message is received.
```

kv.go
```diff
@@ -414,14 +414,22 @@ func (js *js) CreateKeyValue(cfg *KeyValueConfig) (KeyValue, error) {
...
+			var sourceBucketName string
+			if strings.HasPrefix(ss.Name, kvBucketNamePre) {
+				sourceBucketName = ss.Name[len(kvBucketNamePre):]
+			} else {
+				sourceBucketName = ss.Name
 				ss.Name = fmt.Sprintf(kvBucketNameTmpl, ss.Name)
 			}
+
+			if ss.External == nil || sourceBucketName != cfg.Bucket {
+				ss.FilterSubject = fmt.Sprintf("$KV.%s.>", sourceBucketName)
```
piotrpio, line 427:
```suggestion
				ss.FilterSubject = fmt.Sprintf(kvSubjectsTmpl, sourceBucketName)
```

kv_test.go
```diff
@@ -251,3 +251,51 @@ func TestKeyValueCreate(t *testing.T) {
...
+		t.Fatalf("Error creating key: %v", err)
+	}
+
+	kvC, err := js.CreateKeyValue(&KeyValueConfig{Bucket: "C", Sources: []*StreamSource{{Name: "A"}, {Name: "B"}}})
+	if err != nil {
+		t.Fatalf("Error creating kv: %v", err)
+	}
+
+	// Wait half a second to make sure it has time to populate the stream from it's sources
+	time.Sleep(500 * time.Millisecond)
```
piotrpio, line 292: Always waiting 500ms in a test is kind of a lot... It would be good to either decrease this (if possible) or at least replace this with some function which periodically checks for the result (with a timeout), something like `waitFor()` func in drain_test.go

js_test.go
```diff
@@ -1268,3 +1268,54 @@ func TestStreamNameBySubject(t *testing.T) {
...
+
+	m, err := sub.NextMsg(time.Second)
+	if err != nil {
+		t.Fatalf("Unexpected error: %v", err)
+	}
+
+	if m.Subject != "fromtest.transformed.test" {
+		t.Fatalf("the subject of the message doesn't match the expected fromtest.transformed.test: %s", m.Subject)
+	}
+
```
piotrpio, line 1320: Remove newline

js_test.go
```diff
@@ -1268,3 +1268,54 @@ func TestStreamNameBySubject(t *testing.T) {
...
 }
+
+func TestJetStreamTransform(t *testing.T) {
+	s := RunBasicJetStreamServer()
+	defer shutdownJSServerAndRemoveStorage(t, s)
+
+	nc, js := jsClient(t, s)
+	defer nc.Close()
+
+	var err error
```
piotrpio, line 1279: No need for that, just do:
```
_, err := js.AddStream(&StreamConfig{
		Name:             "ORIGIN",
		Subjects:         []string{"test"},
		SubjectTransform: &SubjectTransformConfig{Source: ">", Destination: "transformed.>"},
		Storage:          MemoryStorage,
	})
```

js_test.go
jnmoyne: Done

js_test.go
jnmoyne: Done

kv_test.go
jnmoyne: Reduced to 100 ms

piotrpio (approved):
LGTM

kv_test.go
```diff
@@ -251,3 +251,51 @@ func TestKeyValueCreate(t *testing.T) {
...
+		t.Fatalf("Error creating key: %v", err)
+	}
+
+	kvC, err := js.CreateKeyValue(&KeyValueConfig{Bucket: "C", Sources: []*StreamSource{{Name: "A"}, {Name: "B"}}})
+	if err != nil {
+		t.Fatalf("Error creating kv: %v", err)
+	}
+
+	// Wait half a second to make sure it has time to populate the stream from it's sources
+	time.Sleep(100 * time.Millisecond)
```
wallyqs, line 292: This looks flaky... why not just wrap below code with the `kvC.Get` with some retries until they work?

test/js_test.go
```diff
@@ -5362,17 +5362,6 @@ func testJetStreamMirror_Source(t *testing.T, nodes ...*jsServer) {
 		}
 	})
 
-	t.Run("bind to stream with wrong subject fails", func(t *testing.T) {
-		_, err := js.SubscribeSync("secret", nats.BindStream("origin"))
```
wallyqs, line 5366: btw why was this one deleted? Is it due to upgrading nats-server version?

js_test.go
```diff
@@ -1268,3 +1268,51 @@ func TestStreamNameBySubject(t *testing.T) {
...
+		Name:     "SOURCING",
+		Sources:  []*StreamSource{{Name: "ORIGIN", FilterSubject: ">", SubjectTransformDest: "fromtest.>"}},
+		Storage:  MemoryStorage,
+	})
+	if err != nil {
+		t.Fatalf("Unexpected error: %v", err)
+	}
+
+	// Create a sync subscription with an in-memory ephemeral consumer.
+	sub, err := js.SubscribeSync(">", ConsumerMemoryStorage(), BindStream("SOURCING"))
```
wallyqs, line 1305: Maybe add what other subjects would work around here besides full wildcard? Or would only full wildcard work because it has to match the FilterSubject?

test/js_test.go
jnmoyne: Yes, this PR is the companion to the new nats-server changes which introduce subject transforms to streams and as part of those changes the checking of the consumer filter having to match the stream's subject listeners has been removed, hence this test is no longer valid. It's been discussed at length between myself and Derek and myself and Tomasz both in the PR comments and in the #eng thread. TLDR is that the old assumption that all the messages in a stream are with subjects that match the current subject listen of the stream is no longer valid (actually stopped being so when we started allowing updating the subjects of an existing stream) and not only is it no longer valid but it's that artificial restriction is already a problem (i.e. people getting confused of how they can create a consumer and subscribe to a stream that doesn't listen to any Core NATS subjects), and would not allow source chaining (which is very much a desired functionality, especially once you have the subject transform functionality in the streams).

kv_test.go
jnmoyne: It's not flaky, just a question of timing. Without it, the test passes like 99% of the time but I did see it fail once. IMHO it's expected as there is a small delay between the stream creation call returning and the stream being populated from it's sources, so I added the sleep to give the stream a little bit of time to come up and get populated with the two messages from the 2 sources and haven't seen it fail since. I actually lowered it further to 20 ms and it still passes every time.

js_test.go
jnmoyne: Sure, I changed it to `fromtest.>` and changed the stream to not listen to any subjects. And you inadvertently make the point I made in the conversation above, with the removal of checking the subject is a subset of the stream's listen FilterSubject the test with no filter subject set in the stream (or set to something that doesn't intersect fromtest.>) now works when using `fromtest.>` in the SubscribeSync while before it would not have.

kv_test.go
jnmoyne: I did implement something like you suggest of checking the number of messages and waiting until it reaches 2 and then doing the gets - see the last commit.
And what happens is that the first size check always returns 0 so it does the sleep of 20ms then the next time around the stream is populated and it goes on with the gets, so the tests with this extra check before sleep always runs the same amount of time as with just the sleep.

test/js_test.go
wallyqs: Thanks for the update, in that case I would prefer we add something like the following replacing this test to clarify the behavior: 

```go
	t.Run("bind to stream with subject not in stream", func(t *testing.T) {
		// The Stream does not have a subject called 'nothing' but client is still able
		// to bind to the origin stream even though it cannot consume messages.
		// After updating the stream with the subject this consumer will be able to
		// match and receive messages.
		sub, err := js.SubscribeSync("nothing", nats.BindStream("origin"))
		if err != nil {
			t.Fatal(err)
		}
		_, err = sub.NextMsg(1 * time.Second)
		if !errors.Is(err, nats.ErrTimeout) {
			t.Fatal("Expected timeout error")
		}

		info, err := sub.ConsumerInfo()
		if err != nil {
			t.Fatal(err)
		}
		got := info.Stream
		expected := "origin"
		if got != expected {
			t.Fatalf("Expected %v, got %v", expected, got)
		}

		got = info.Config.FilterSubject
		expected = "nothing"
		if got != expected {
			t.Fatalf("Expected %v, got %v", expected, got)
		}

		t.Run("can consume after stream update", func(t *testing.T) {
			_, err = js.UpdateStream(&nats.StreamConfig{
				Name: "origin",
				Placement: &nats.Placement{
					Tags: []string{"NODE_0"},
				},
				Storage:  nats.MemoryStorage,
				Replicas: 1,
				Subjects: []string{"origin", "nothing"},
			})
			js.Publish("nothing", []byte("hello world"))

			msg, err := sub.NextMsg(1 * time.Second)
			if err != nil {
				t.Error(err)
			}
			got = msg.Subject
			expected = "nothing"
			if got != expected {
				t.Fatalf("Expected %v, got %v", expected, got)
			}

		})
	})
```

wallyqs: (I will send a patch for this one after merging your PR which have just LGTMd)

wallyqs (approved):

js_test.go
wallyqs: Thanks for the explanation!

jnmoyne:
👍

PR #1204, opened 2023-02-06
Add option to skip host lookup

chen-shmilovich-sysdig:
Sometimes we want NATS to avoid resolving the IP address of a host. For example, when using a proxy that only lets a specific address connect and blocks everything else.
This is why I added a new option in this PR. It lets you skip the hostname lookup.


piotrpio (approved):
LGTM!

PR #1209, opened 2023-02-10
Go 1.20 updates, add build tag for rand.Seed

wallyqs:
Adds Go 1.20 and also deprecates use of rand.Seed in the client for Go +1.20 series since this is now always done by default (https://github.com/golang/go/issues/54880)

### DIFF
--- a/rand.go
+++ b/rand.go
@@ -1,0 +1,29 @@
+// Copyright 2023 The NATS Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


rand.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"math/rand"
+	"time"
+)
+
+func init() {
+	// This is not needed since Go 1.20 because now rand.Seed always happens
+	// by default (uses runtime.fastrand64 instead as source).
+	rand.Seed(time.Now().UnixNano())
```
wallyqs, line 28: we don't need it now probably but from Go 1.19 it was possible to get call into fastrand64 like this: `maphash.Bytes(maphash.MakeSeed(), nil)`

derekcollison: This a new file?

wallyqs: yes a new file to add the build tags so that this does not apply to Go +1.20

derekcollison: ok so just copyright 2023..

derekcollison (approved):
LGTM

PR #1211, opened 2023-02-14
[ADDED] FetchChan method to utilize non-blocking pull subscription requests

### LINKED ISSUE #1201: Add non-blocking Fetch method for JetStream pull subscriptions
Currently using `Fetch()` returns `[]*nats.Msg` and blocks until the provided number of messages is received (`batch`) or at least one message is available on the stream (`no_wait` is set to `true` in the pull request`).

A new method should be added (e.g. `FetchChan()`), returning a `<-chan *nats.Msg`, which sets the `no_wait` parameter to `false` in order to wait for the whole batch (or timeout) before terminating the pull request. The channel can be closed when all messages are received or timeout is reached to enable iterating over the messages using `range`.

piotrpio:
Resolves #1201

### DIFF
--- a/js.go
+++ b/js.go
@@ -2794,7 +2794,8 @@ func (mb *messageBatch) Done() bool {
 // FetchBatch pulls a batch of messages from a stream for a pull consumer.
 // Unlike [Subscription.Fetch], it is non blocking and returns [MessageBatch],
 // allowing to retrieve incoming messages from a channel.
-// The returned channel is always close - it is safe to iterate over it using range.
+// The returned channel is always closed after all messages for a batch have been 
+// delivered by the server - it is safe to iterate over it using range.
 //
 // To avoid using default JetStream timeout as fetch expiry time, use [nats.MaxWait]
 // or [nats.Context] (with deadline set).
--- a/js.go
+++ b/js.go
@@ -2761,26 +2761,38 @@ func (sub *Subscription) Fetch(batch int, opts ...PullOpt) ([]*Msg, error) {
 	return msgs, nil
 }
 
-type FetchResult interface {
+// MessageBatch provides methods to retrieve messages consumed using [Subscribe.FetchBatch].
+type MessageBatch interface {
+	// Messages returns a channel on which messages will be published.
 	Messages() <-chan *Msg
+
+	// Error returns an error encountered when fetching messages.
 	Error() error
+
+	// Done signals end of execution.
+	Done() bool
 }
 
-type fetchResult struct {
+type messageBatch struct {
 	msgs chan *Msg
 	err  error
+	done bool
 }
 
-func (fr *fetchResult) Messages() <-chan *Msg {
-	return fr.msgs
+func (mb *messageBatch) Messages() <-chan *Msg {
+	return mb.msgs
 }
 
-func (fr *fetchResult) Error() error {
-	return fr.err
+func (mb *messageBatch) Error() error {
+	return mb.err
 }
 
-// FetchChan pulls a batch of messages from a stream for a pull consumer.
-// Unlike [Subscription.Fetch], it is non blocking and returns [FetchResult],
+func (mb *messageBatch) Done() bool {
+	return mb.done
+}
+
+// FetchBatch pulls a batch of messages from a stream for a pull consumer.
+// Unlike [Subscription.Fetch], it is non blocking and returns [MessageBatch],
 // allowing to retrieve incoming messages from a channel.
 // The returned channel is always close - it is safe to iterate over it using range.
 //
@@ -2788,8 +2800,8 @@ func (fr *fetchResult) Error() error {
 // or [nats.Context] (with deadline set).
 //
 // This method will not return error in case of pull request expiry (even if there are no messages).
-// Any other error encountered when receiving messages will cause FetchChan to stop receiving new messages.
-func (sub *Subscription) FetchChan(batch int, opts ...PullOpt) (FetchResult, error) {
+// Any other error encountered when receiving messages will cause FetchBatch to stop receiving new messages.
+func (sub *Subscription) FetchBatch(batch int, opts ...PullOpt) (MessageBatch, error) {
 	if sub == nil {
 		return nil, ErrBadSubscription
 	}
@@ -2867,7 +2879,7 @@ func (sub *Subscription) FetchChan(batch int, opts ...PullOpt) (FetchResult, err
 	default:
 	}
 
-	result := &fetchResult{
+	result := &messageBatch{
 		msgs: make(chan *Msg, batch),
 	}
 	var msg *Msg
@@ -2893,6 +2905,7 @@ func (sub *Subscription) FetchChan(batch int, opts ...PullOpt) (FetchResult, err
 	}
 	if len(result.msgs) == batch || result.err != nil {
 		close(result.msgs)
+		result.done = true
 		return result, nil
 	}
 
@@ -2914,11 +2927,16 @@ func (sub *Subscription) FetchChan(batch int, opts ...PullOpt) (FetchResult, err
 	reqJSON, err := json.Marshal(req)
 	if err != nil {
 		close(result.msgs)
+		result.done = true
 		result.err = err
 		return result, nil
 	}
 	if err := nc.PublishRequest(nms, rply, reqJSON); err != nil {
+		if len(result.msgs) == 0 {
+			return nil, err
+		}
 		close(result.msgs)
+		result.done = true
 		result.err = err
 		return result, nil
 	}
@@ -2952,6 +2970,7 @@ func (sub *Subscription) FetchChan(batch int, opts ...PullOpt) (FetchResult, err
 			result.err = o.checkCtxErr(err)
 		}
 		close(result.msgs)
+		result.done = true
 	}()
 	return result, nil
 }
--- a/js.go
+++ b/js.go
@@ -2794,7 +2794,7 @@ func (mb *messageBatch) Done() bool {
 // FetchBatch pulls a batch of messages from a stream for a pull consumer.
 // Unlike [Subscription.Fetch], it is non blocking and returns [MessageBatch],
 // allowing to retrieve incoming messages from a channel.
-// The returned channel is always closed after all messages for a batch have been 
+// The returned channel is always closed after all messages for a batch have been
 // delivered by the server - it is safe to iterate over it using range.
 //
 // To avoid using default JetStream timeout as fetch expiry time, use [nats.MaxWait]
--- a/js.go
+++ b/js.go
@@ -1,4 +1,4 @@
-// Copyright 2020-2022 The NATS Authors
+// Copyright 2020-2023 The NATS Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
@@ -2673,16 +2673,6 @@ func (sub *Subscription) Fetch(batch int, opts ...PullOpt) ([]*Msg, error) {
 		return nil, err
 	}
 
-	// Use the deadline of the context to base the expire times.
-	deadline, _ := ctx.Deadline()
-	ttl = time.Until(deadline)
-	checkCtxErr := func(err error) error {
-		if o.ctx == nil && err == context.DeadlineExceeded {
-			return ErrTimeout
-		}
-		return err
-	}
-
 	var (
 		msgs = make([]*Msg, 0, batch)
 		msg  *Msg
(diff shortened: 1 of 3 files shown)


js.go
```diff
@@ -2766,11 +2756,214 @@ func (sub *Subscription) Fetch(batch int, opts ...PullOpt) ([]*Msg, error) {
...
+// Unlike [Subscription.Fetch], it is non blocking and returns [FetchResult],
+// allowing to retrieve incoming messages from a channel.
+// The returned channel is always close - it is safe to iterate over it using range.
+//
+// To avoid using default JetStream timeout as fetch expiry time, use [nats.MaxWait]
+// or [nats.Context] (with deadline set).
+//
+// This method will not return error in case of pull request expiry (even if there are no messages).
+// Any other error encountered when receiving messages will cause FetchChan to stop receiving new messages.
+func (sub *Subscription) FetchChan(batch int, opts ...PullOpt) (FetchResult, error) {
```
wallyqs, line 2792: How about `FetchBatch` and then the interface is a `MessageBatch`? It would be then like `FetchBatch pulls a batch of messages... and returns a MessageBatch`.

js_test.go
```diff
@@ -116,7 +116,6 @@ func TestJetStreamOrderedConsumer(t *testing.T) {
 
 	// Create a sample asset.
 	msg := make([]byte, 1024*1024)
-	//lint:ignore SA1019 crypto/rand.Read is recommended after Go 1.20 but fine for this test.
```
wallyqs, line 119: 👍  this was related to upstream changes in staticcheck

js.go
```diff
@@ -2766,11 +2756,214 @@ func (sub *Subscription) Fetch(batch int, opts ...PullOpt) ([]*Msg, error) {
...
+			}
+			if usrMsg {
+				result.msgs <- msg
+				requestMsgs++
+			}
+		}
+		if err != nil {
+			result.err = o.checkCtxErr(err)
+		}
+		close(result.msgs)
```
wallyqs, line 2954: maybe add a 1s tick here that releases the goroutine one `len(result.msgs)` has reached 0 to ensure messages have been consumed or give up when the client is closed

js.go
```diff
@@ -2766,11 +2756,214 @@ func (sub *Subscription) Fetch(batch int, opts ...PullOpt) ([]*Msg, error) {
...
+		result.err = err
+		return result, nil
+	}
+	cancelContext = false
+	go func() {
+		if cancel != nil {
+			defer cancel()
+		}
+		var requestMsgs int
+		for requestMsgs < requestBatch {
```
wallyqs, line 2931: once the batch has been filled, this loop ends and then closes the `result.msgs` channel, I think this means that the channel cannot be consumed even if some messages were buffered?

test/js_test.go
```diff
@@ -955,6 +955,250 @@ func TestJetStreamSubscribe(t *testing.T) {
...
+
+	_, err := js.AddStream(&nats.StreamConfig{
+		Name:     "TEST",
+		Subjects: []string{"foo"},
+	})
+	if err != nil {
+		t.Fatalf("Unexpected error: %v", err)
+	}
+
+	t.Run("basic fetch", func(t *testing.T) {
```
wallyqs, line 973: need a test where the batch is delivered all the messages, after all have been delivered then try to get the delivered messages

test/js_test.go
```diff
@@ -955,6 +955,250 @@ func TestJetStreamSubscribe(t *testing.T) {
...
+		}
+		if res.Error() == nil && !errors.Is(err, nats.ErrConsumerDeleted) {
+			t.Fatalf("Expected error: %s; got: %s", nats.ErrConsumerDeleted, err)
+		}
+		if len(msgs) != 5 {
+			t.Fatalf("Expected %d messages; got: %d", 5, len(msgs))
+		}
+	})
+
+	t.Run("validation errors", func(t *testing.T) {
```
wallyqs, line 1164: just in case, try to add a test where you have multiple fetch calls under different goroutines concurrently then a single select waiting for all the message channels

js.go
```diff
@@ -2766,11 +2756,214 @@ func (sub *Subscription) Fetch(batch int, opts ...PullOpt) ([]*Msg, error) {
...
+
+	requestBatch := batch - len(result.msgs)
+	req := nextRequest{
+		Expires:  expires,
+		Batch:    requestBatch,
+		MaxBytes: o.maxBytes,
+	}
+	reqJSON, err := json.Marshal(req)
+	if err != nil {
+		close(result.msgs)
```
wallyqs, line 2916: something to note is that just closing would mean that `res.Messages()` channel would result in zero values of messages if called again, but if we nil the channel then `for range batch.Messages()` would block forever.

wallyqs: may need a condition for the result to signal that the result is not done yet, maybe part of the interface`IsDone() bool`

js.go
```diff
@@ -2766,11 +2756,214 @@ func (sub *Subscription) Fetch(batch int, opts ...PullOpt) ([]*Msg, error) {
...
 	if err != nil && len(msgs) == 0 {
-		return nil, checkCtxErr(err)
+		return nil, o.checkCtxErr(err)
 	}
 	return msgs, nil
 }
 
+type FetchResult interface {
+	Messages() <-chan *Msg
+	Error() error
```
wallyqs, line 2766: I think we need a way to stamp that the batch is done, (definition of done being that you cannot consume further messages from it), maybe something like:

```suggestion
	Done() bool
```

or maybe context like?

```go
  Done() <-chan struct{}
```

js.go
```diff
@@ -2766,11 +2756,214 @@ func (sub *Subscription) Fetch(batch int, opts ...PullOpt) ([]*Msg, error) {
...
+	reqJSON, err := json.Marshal(req)
+	if err != nil {
+		close(result.msgs)
+		result.err = err
+		return result, nil
+	}
+	if err := nc.PublishRequest(nms, rply, reqJSON); err != nil {
+		close(result.msgs)
+		result.err = err
+		return result, nil
```
wallyqs, line 2923: can we have no messages at this point? in that case I think better here to `return nil, err` in the case this publish failed (would have been a connection closed error or a sync error that reconnect buffer is full)

js.go
piotrpio: I like that 👍

js.go
piotrpio: Hmm, not sure I get it. When the channel is closed, clients can still consume messages from it... Closing a channel only signals that no more messages will come in.

js.go
piotrpio: It can be consumed, you cannot send new values to it.

js.go
piotrpio: If you drain + close the channel, it will simply have 0 values inside, nothing should be blocking when using `range`. So really, subsequent calls to `res.Messages()` should not have a negative effect: https://go.dev/play/p/SwlO1ZhKgDr

That being said, if you just do `msg := <-res.Messages()` after closing the channel, it will return nil, that is correct. You can check if the channel is closed by doing `msg, ok := <-res.Messages()`, but we could also add a `Done()` method in addition to that, to be explicit.

js.go
piotrpio: Hmm, I would rather go with `Done() bool` - since we are iterating over values, I would like to avoid having to have for loops with `select`, I think boolean is enough - if it's false, you can safely consume messages.

js.go
piotrpio: Yeah, if there are indeed no messages at this point we can return `nil, err` 👍  If there is at least 1 message that was in buffer previously however, I think we should still return `result, nil`, with error in result.

js.go
wallyqs: you're right, it is fine to close the channel actually...

js.go
wallyqs: as you mentioned above this actually not an issue, I misremembered this part of channels usage

js.go
wallyqs: sounds good to me 👍

js.go
piotrpio: Done

test/js_test.go
piotrpio: done

js.go
wallyqs: thanks for the clarification (should have double checked myself first 😓 ), you are also covering with docs that the channel is closed after being consumed so that's good too.

piotrpio: I've added `Done() bool` to the `MessagesBatch` interface.

js.go
wallyqs: nice!

js.go
piotrpio: done

js.go
```diff
@@ -2766,11 +2756,233 @@ func (sub *Subscription) Fetch(batch int, opts ...PullOpt) ([]*Msg, error) {
...
+}
+
+func (mb *messageBatch) Done() bool {
+	return mb.done
+}
+
+// FetchBatch pulls a batch of messages from a stream for a pull consumer.
+// Unlike [Subscription.Fetch], it is non blocking and returns [MessageBatch],
+// allowing to retrieve incoming messages from a channel.
+// The returned channel is always close - it is safe to iterate over it using range.
```
wallyqs, line 2797:
```suggestion
// The returned channel is always closed after all messages for a batch have been 
// delivered by the server - it is safe to iterate over it using range.
```

test/js_test.go
```diff
@@ -955,6 +955,381 @@ func TestJetStreamSubscribe(t *testing.T) {
...
+			if _, err := js.Publish("foo", []byte("msg")); err != nil {
+				t.Fatalf("Unexpected error: %s", err)
+			}
+		}
+		var r1, r2, r3 nats.MessageBatch
+		started := &sync.WaitGroup{}
+		started.Add(3)
+		errs := make(chan error, 3)
+		go func() {
+			r1, err = sub.FetchBatch(10)
```
wallyqs, line 1022: this `err` being shared among goroutines might seem to be causing a data race in the test

```
=== RUN   TestPullSubscribeFetchBatch/multiple_concurrent_fetches
==================
WARNING: DATA RACE
Write at 0x00c0003b64f8 by goroutine 50:
  github.com/nats-io/nats%2ego.(*Subscription).FetchBatch.func2()
     nats-dev/src/github.com/nats-io/nats.go/js.go:2974 +0x444

Previous read at 0x00c0003b64f8 by goroutine 42:
  github.com/nats-io/nats%2ego.(*messageBatch).Done()
      nats-dev/src/github.com/nats-io/nats.go/js.go:2791 +0x30
  github.com/nats-io/nats.go/test.TestPullSubscribeFetchBatch.func2()
      nats-dev/src/github.com/nats-io/nats.go/test/js_test.go:1074 +0xc14
  testing.tRunner()
      /opt/homebrew/Cellar/go/1.19.5/libexec/src/testing/testing.go:1446 +0x188
  testing.(*T).Run.func1()
      /opt/homebrew/Cellar/go/1.19.5/libexec/src/testing/testing.go:1493 +0x40

Goroutine 50 (running) created at:
  github.com/nats-io/nats%2ego.(*Subscription).FetchBatch()
      nats-dev/src/github.com/nats-io/nats.go/js.go:2945 +0xf04
  github.com/nats-io/nats.go/test.TestPullSubscribeFetchBatch.func2.1()
      nats-dev/src/github.com/nats-io/nats.go/test/js_test.go:1022 +0x64

Goroutine 42 (running) created at:
  testing.(*T).Run()
      /opt/homebrew/Cellar/go/1.19.5/libexec/src/testing/testing.go:1493 +0x55c
  github.com/nats-io/nats.go/test.TestPullSubscribeFetchBatch()
      nats-dev/src/github.com/nats-io/nats.go/test/js_test.go:1006 +0x368
  testing.tRunner()
      /opt/homebrew/Cellar/go/1.19.5/libexec/src/testing/testing.go:1446 +0x188
  testing.(*T).Run.func1()
      /opt/homebrew/Cellar/go/1.19.5/libexec/src/testing/testing.go:1493 +0x40
==================
    js_test.go:1075: FetchBatch result channel should be closed after receiving all messages on r1
    testing.go:1319: race detected during execution of test
```

test/js_test.go
```diff
@@ -955,6 +955,381 @@ func TestJetStreamSubscribe(t *testing.T) {
...
+		if r1.Error() != nil {
+			t.Fatalf("Unexpected error: %s", r1.Error())
+		}
+		if r2.Error() != nil {
+			t.Fatalf("Unexpected error: %s", r2.Error())
+		}
+		if r3.Error() != nil {
+			t.Fatalf("Unexpected error: %s", r3.Error())
+		}
+		if !r1.Done() {
```
wallyqs, line 1074: this one seems to be a data race... probably due to marking `bm.done` bool in a different goroutine without locks and then checking it here, maybe it was needed the `Done() <-chan struct{}` approach after all to avoid locks... I'm also good on punting adding `Done()` to the `MessageBatch` interface for now if it becomes problematic.

piotrpio: Yes, my bad. I revise my statement about not liking `Done() <-chan struct{}` 😅 It does not force anyone to use `select` + it can always be used just like a boolean with `len(res.Done()) == 0`. I changed that and fixed the above data race as well.

wallyqs (approved):
LGTM!

Issue #1216, opened 2023-02-22
Recreate Ordered Consumer Fails when R1 Stream's server restarts


caleblloyd:
Recreating an Ordered Consumer is Failing when the `nats-server` hosting an R1 stream restarts.

The specific error returned by the client is:

```
[handleConsumerSequenceMismatch](nats: insufficient resources: recreating ordered consumer)
```

This happens when the client is connected to a NATS Server that is up, but the NATS Server hosting the R1 stream is restarting.  JetStream returns the `insufficient resources` error because it cannot place a peer group for the new consumer.

PR #1217, opened 2023-02-23
[FIX] Retry recreating ordered consumer on insufficient resources error

piotrpio:
Resolves https://github.com/nats-io/nats.go/issues/1216


Jarema (approved):
LGTM

PR #1220, opened 2023-02-24
Revert "[ADDED] Support for stream subject transform"

piotrpio:
Reverts nats-io/nats.go#1200 - this functionality will be added after nats-server@2.10.0 release


Jarema (approved):
LGTM, assuming green CI.

wallyqs:
This is fine too since in the server still locking to v1.23 release: https://github.com/nats-io/nats-server/blob/main/go.mod#L9

PR #1224, opened 2023-02-28
[ADDED] Exponential backoff for server reconnects

piotrpio:


nats.go
```diff
@@ -2559,6 +2572,28 @@ func (nc *Conn) stopPingTimer() {
 	}
 }
 
+// DefaultReconnectBackoffHandler returns a default reconnect exponential backoff interval.
+// Base reconnect wait is 10ms, with x2 multiplier. Max wait time is 2 minutes.
+// 10ms, 20ms, 40ms, 80ms...2m
+// A random jitter is added to the result.
+func DefaultReconnectBackoffHandler(jitter time.Duration) ReconnectDelayHandler {
```
aricart, line 2579: With this feature, I am wondering if some of the other options we have on the clients should be deprecated or go away (at the very least, this should be the default)

piotrpio:
@aricart I am wondering about a couple of things:
1. What should we use as default base/max/multiplier values for reconnect attempts? It should be consistent across the clients, with `ReconnectWait` go and javascript had 2 seconds, which seems to be a bit much. Rust already has backoff with 1ms as the first wait time, which seems a bit low :) In this PR I set base 10ms, multiplier x2 with max 2 minutes, it seems reasonable to me but I wonder what's your take on this.
2. Should we get rid of setting `MaxReconnects` by default? I would assume so, to basically try endlessly with some max interval.

nats.go
```diff
@@ -50,8 +50,7 @@ const (
 	Version                   = "1.24.0"
 	DefaultURL                = "nats://127.0.0.1:4222"
 	DefaultPort               = 4222
-	DefaultMaxReconnect       = 60
-	DefaultReconnectWait      = 2 * time.Second
+	DefaultMaxReconnect       = -1
```
wallyqs, line 53: makes sense to change into a span of time but not sure about changing this var as the default since has been like that from the start... maybe as an option to test out new behavior like `MaxTotalReconnectTime(2*time.Minute)`? We also need to be able to set `max reconnect` to -1 so that it means that we want the client to reconnect forever and never give up reconnecting

aricart: @piotrpio's change is right in line with this.

PR #1227, opened 2023-03-07
Fix possible race on EncodedConnection LastError

dreisel:
accessing the err property of Conn is not thread safe. This uses the getter method with proper locks.


piotrpio (approved):
LGTM, thanks for submitting the PR!

PR #1231, opened 2023-03-12
support metadata for micro

ripienaar:


ripienaar:
Implements https://github.com/nats-io/nats-architecture-and-design/pull/205

micro/service.go
```diff
@@ -90,9 +91,10 @@ type (
 
 	// ServiceIdentity contains fields helping to identity a service instance.
 	ServiceIdentity struct {
-		Name    string `json:"name"`
-		ID      string `json:"id"`
-		Version string `json:"version"`
+		Name     string            `json:"name"`
+		ID       string            `json:"id"`
+		Version  string            `json:"version"`
+		Metadata map[string]string `json:"metadata"`
```
piotrpio, line 97: Maybe we could use `omitempty` for metadata to avoid sending `null` field if it's not provided?

ripienaar: updated it in both places.

aricart: (I actually believe the same)

ripienaar:
@piotrpio tests green now, pls merge when happy.

PR #1232, opened 2023-03-12
Add Deadline header, update micro HandlerFunc signature

actatum:
Add deadline header for propagating request cancellation to server handlers, update HandlerFunc signature to take a context

### DIFF
--- a/jetstream/pull.go
+++ b/jetstream/pull.go
@@ -637,10 +637,9 @@ func (hb *hbMonitor) Reset(dur time.Duration) {
 }
 
 func (s *pullSubscription) Stop() {
-	if atomic.LoadUint32(&s.closed) == 1 {
+	if !atomic.CompareAndSwapUint32(&s.closed, 0, 1) {
 		return
 	}
-	atomic.StoreUint32(&s.closed, 1)
 	close(s.done)
 	if s.consumeOpts.stopAfterMsgsLeft != nil {
 		if s.delivered >= s.consumeOpts.StopAfter {


actatum:
I think another good option to add context support without breaking backwards compatibility would be to mirror what the http.Request type does and embed the context into the request struct.

For example

Request interface:
```go
// Request represents service request available in the service handler.
// It exposes methods to respond to the request, as well as
// getting the request data and headers.
Request interface {
	// Respond sends the response for the request.
	// Additional headers can be passed using [WithHeaders] option.
	Respond([]byte, ...RespondOpt) error

	// RespondJSON marshals the given response value and responds to the request.
	// Additional headers can be passed using [WithHeaders] option.
	RespondJSON(interface{}, ...RespondOpt) error

	// Error prepares and publishes error response from a handler.
	// A response error should be set containing an error code and description.
	// Optionally, data can be set as response payload.
	Error(code, description string, data []byte, opts ...RespondOpt) error

	// Data returns request data.
	Data() []byte

	// Headers returns request headers.
	Headers() Headers

	// Subject returns underlying NATS message subject.
	Subject() string

	// Context returns the associated request context.
	Context() context.Context

	// WithContext returns a new instance of the Request with the given context.
	WithContext(context.Context) Request
}
```

Request implementation:
```go
// request is a default implementation of Request interface
request struct {
	msg          *nats.Msg
	respondError error
        ctx context.Context
}

func (r *request) Context() context.Context {
      return r.ctx
}

func (r *request) WithContext(ctx context.Context) Request {
      r.ctx = ctx
      return r
}
```

piotrpio:
Hey @actatum, thanks for submitting the PR. 

Firstly, we do not want to introduce additional header passed in the request in each `RequestMsgWithContext()` call - this is a core NATS functionality and adding extra headers will make each request larger. If needed, custom headers can be added to a request in the message itself.

Secondly, there is a helper in `micro` called `ContextHandler()` which allows you to set up your service with ctx in handler:
```go
	handler := func(ctx context.Context, req micro.Request) {
		// handler
	}
	config := micro.Config{
		Name:    "test_service",
		Version: "0.1.0",
		Endpoint: &micro.EndpointConfig{
			Subject: "test.func",
			Handler: micro.ContextHandler(ctx, handler),
		},
	}
```

Lastly, we want `ctx` in `micro` to be optional, so changing the signature of `HandlerFunc` is not the way we should go.

PR #1240, opened 2023-03-27
style: more intention-revealing microservice monitoring setup

levb:
###### Summary

Refactored the code for setting up the monitoring verb handlers for microservices. Made the naming more intention-revealing, the code more table-driven.

NOTE: This is my first PR to this repo, I am filing it largely for learning the process.

### DIFF
--- a/micro/service.go
+++ b/micro/service.go
@@ -325,7 +325,7 @@ func (s Verb) String() string {
 // A service name, version and Endpoint configuration are required to add a service.
 // AddService returns a [Service] interface, allowing service management.
 // Each service is assigned a unique ID.
-func AddService(nc *nats.Conn, config Config) (_ Service, err error) {
+func AddService(nc *nats.Conn, config Config) (Service, error) {
 	if err := config.valid(); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


micro/service.go
```diff
@@ -325,7 +325,7 @@ func (s Verb) String() string {
 // A service name, version and Endpoint configuration are required to add a service.
 // AddService returns a [Service] interface, allowing service management.
 // Each service is assigned a unique ID.
-func AddService(nc *nats.Conn, config Config) (Service, error) {
+func AddService(nc *nats.Conn, config Config) (_ Service, err error) {
```
piotrpio, line 328: I'm not sure I like having named returns here - they tend to make it harder to debug/understand code. After your other changes, you have it just for the sake of 2 returns (lines 365 and 394). IMO it would be cleaner to simply remove the defer and named returns and just close `asyncDispatcher` in those 2 lines.

piotrpio (commented):
Really good changes! Just one nit, other than that looks great.

micro/service.go
levb: I personally find using named return parameters/`defer` combo to be an effective and intention-revealing tool (as long as the actual `return` statements are explicit). But in this case fewer lines of code is better? :) Done.

piotrpio (approved):
LGTM

PR #1242, opened 2023-04-01
avoid producing nil metadata

ripienaar:
It is quite hard to validate in a way that allows object or nil using jsonschema so metadata should either be omitempty or always be object.

During initial reviews we made it not omitempty, so this ensures its never nil.  Can also make it omitempty if cleaner.


piotrpio (approved):
LGTM

ripienaar:
Tests are finally green, sorry I missed those failures @piotrpio

PR #1250, opened 2023-04-19
KVerror type wraps APIError

owayss:
Fixes https://github.com/nats-io/kubecon-contribfest/issues/1

Adds KVError type to allow wrapping APIError.

E.g.:
```go
return KVError{apiError: ErrStreamNotFound, message: ErrBucketNotFound}
```

### DIFF
--- a/kv.go
+++ b/kv.go
@@ -283,6 +283,32 @@ type KeyValueEntry interface {
 }
 
 // Errors
+
+// KVError allows wrapping APIError.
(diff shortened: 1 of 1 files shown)


piotrpio (approved):
Hey @owayss , I changed the target branch to `kc-contib-main` so that we can use the new type for kv errors before merging everything to main. 

LGTM!

wallyqs (approved):

PR #1253, opened 2023-04-24
Use TLS from Memory

nh3000-org:
Eliminate the need to store TLS certificates on the file system.


piotrpio:
Hey @nh3000-org, thanks for the PR, so sorry it took so long to review. 

This is a very good idea, I just have 2 suggestions:
1. We recently added `TLSCertCB` and `RootCAsCB` callbacks to `nats.Options` to be able to leverage various ways of setting tls certs. It is the same strategy we are already using with JWT creds (see `nats.UserJWT()`). It would be good if you could change your implementation to be setting the callbacks instead of `TLSConfig` field. For reference, you can take a look how it's done for JWTs in `nats.UserJWTAndSeed()`: https://github.com/nats-io/nats.go/blob/main/nats.go#L1143 and how we're setting `TLSCertCB` in `nats.ClientCert()`: https://github.com/nats-io/nats.go/blob/main/nats.go#L889

2. It would be good to add tests for those new options - you can re-use the certificates from `test/configs/certs`. For reference, see `TestClientCertificate` in `conn_test.go`

nats.go
```diff
@@ -884,6 +911,35 @@ func RootCAs(file ...string) Option {
...
 }
 
+// RootCAsMem is a helper option to provide the RootCAs pool from memory.
+// If Secure is not already set this will set it as well.
+func RootCAsMem(mem []byte) Option {
+
+	return func(o *Options) error {
+		rootCAsCB := func() (*x509.CertPool, error) {
+			pool := x509.NewCertPool()
+			for _, f := range mem {
```
piotrpio, line 921: You're iterating over a byte slice here, and for each byte you're doing `ok := pool.AppendCertsFromPEM(mem)` - that does not make much sense. Probably the signature of this function should be:
```go
func RootCAsMem(rootCAs ...[]byte) Option {
```

And then you can iterate over the certificates.

nats.go
```diff
@@ -911,6 +967,33 @@ func ClientCert(certFile, keyFile string) Option {
 	}
 }
 
+// ClientCertMem is a helper option to provide the client certificate from memory.
+// If Secure is not already set this will set it as well.
+func ClientCertMem(certFile []byte, keyFile []byte) Option {
```
piotrpio, line 972: Invalid param names, those are not files
```suggestion
func ClientCertMem(cert []byte, key []byte) Option {
```

piotrpio (changes requested):

nats.go
```diff
@@ -884,6 +911,35 @@ func RootCAs(file ...string) Option {
 	}
 }
 
+// RootCAsMem is a helper option to provide the RootCAs pool from memory.
+// If Secure is not already set this will set it as well.
+func RootCAsMem(mem []byte) Option {
+
```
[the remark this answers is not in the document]
nh3000-org: Done...

nats.go
nh3000-org: done.

nh3000-org:
I tested on linux and android.

piotrpio:
This still does not compile:
```
# github.com/nats-io/nats.go [github.com/nats-io/nats.go.test]

./nats.go:893:22: undefined: mem
```

nh3000-org:
ok doing a fresh pull and go update, will resolve.  thank you

nh3000-org:
Passed..

nh3000-org:
That's weird. Is the main branch correct? It is saying passed.

nh3000-org:
I think this is ready to go.

nats.go
```diff
@@ -884,6 +884,32 @@ func RootCAs(file ...string) Option {
 	}
 }
 
+// RootCAsMem is a helper option to provide the RootCAs pool from memory.
+// If Secure is not already set this will set it as well.
+func RootCAsMem(rootCAs []byte) Option {
+	return func(o *Options) error {
+		rootCAsCB := func() (*x509.CertPool, error) {
+			pool := x509.NewCertPool()
+			for _, f := range rootCAs {
```
piotrpio, line 893: As I mentioned before, this does not make sense, you are iterating over a byte slice, byte by byte, and growing cert pool for each byte...

Please look at this comment (which you marked as resolved): https://github.com/nats-io/nats.go/pull/1253#discussion_r1249472944

piotrpio (changes requested):

nats.go
nh3000-org: ok, should I caste rootCAs to string here or convert the type to string?


func RootCAsMem(rootCAs string) Option {

or cast it here

for _, f := range string(rootCAs) {

or maybe it does not matter where, just looking for optimum solution.

It can be done at application level or here.

nh3000-org: Correct me if I am wrong, but it must be processed as a byte array to include line endings in the certificate. I tested, and it fails when cast to string, Perhaps there is a different method that I am un-aware of?  Pool does not have an append to add the line endings. Please advise on what to do,

nh3000-org: btw: byte array is way original code worked

piotrpio: That is not what I meant, I do not think you should be casting it to string. You have one argument to function - `rootCAs []byte` - it is a single CA I assume. In `rootCAsCB` you are iterating **over a byte slice**, which means the loop will be executed `len(rootCAs)` times. As a simple example, try putting any `Println` in this for loop, e.g. like this:
```
rootCAsCB := func() (*x509.CertPool, error) {
	pool := x509.NewCertPool()
	for _, f := range rootCAs {
		fmt.Println("Appending cert from PEM")
		ok := pool.AppendCertsFromPEM(rootCAs)
		if !ok {
			return nil, fmt.Errorf("nats: failed to parse root certificate from memory from %q", f)
		}
	}
	return pool, nil
}
``` 
And then execute `TestClientCertificateMem` and see what happens.

nh3000-org:
Agree, it's not the best way, I don't know hoe to preserve line endings in
the certificate without processing as a byte array.  The cert requires the
line endings as part of spec. Could you point me in the right direction.
I'm just trying to help, but ran into the wall on ensuring the cert is in
proper format. I may of missed something here, but think this is only way
to do it.  The original code processes the byte array in the same way.
This problem must be addressed in the context of use. From a pure
programming point of view you are correct, but the context requires a not
so pretty solution.

Thank you for your help.

nats.go
piotrpio: Sorry to only get back to it now. Your `rootCAsCB` is still not correct I think and could be simplified to this:
```go
rootCAsCB := func() (*x509.CertPool, error) {
	pool := x509.NewCertPool()
	ok := pool.AppendCertsFromPEM(rootCAs)
	if !ok {
		return nil, fmt.Errorf("nats: failed to parse root certificate from memory")
	}
	return pool, nil
}
```

There is no need for the for loop. Also, the error message is not correct as `f` is a single byte.

nh3000-org:
Thanks, it worked.  Sorry for being so dense. Thank you for help...I'll
update the pr..

PR #1254, opened 2023-04-25
Move tests from nats package under test/* directory

masumomo:
# OVERVIEW

**WHAT**:
- As titled.

**WHY**:
- In order to prevent `nats.go` from bringing nats-server dependencies to `go.mod`.

### Reference

- This PR relates https://github.com/nats-io/kubecon-contribfest/issues/6

### Not Implemented
- We have some tests that use internal fields, so I left them as it is for now. We will work on it in separated issues.
   - For example, `nc.mu.Lock()` and 'nc.mu.Unlock()' in tests.
   - I added comments `// Need access to the internals for testing.` for those tests. 
- `TestNoRaceObjectContextOpt` sometimes fails, but it's unrelated to this change.
  - As I investigated, it might come from the timing when cancelling the context and shutdown the JetStream.


go.mod
```diff
@@ -3,11 +3,18 @@ module github.com/nats-io/nats.go
 go 1.19
 
 require (
+	github.com/golang/protobuf v1.4.2
```
wallyqs, line 6: @masumomo Thanks for the contribution! 🙌  Could you revert the changes to `go.mod` and `go.sum`? You can instead use the following `go_test.mod` like this for running the tests to avoid including the server into the main `go.mod`:

```
go test ./... --count=1 -p=1 -modfile=go_test.mod
```

masumomo: Hello Wally! Oh, Thank you so much! I realised we have `go_test.mod` in the repo and an instruction on README to run the test 🙏  I'll revert it!

piotrpio:
Hey @masumomo!

I checked out your branch, updated it and fixed remaining issues: https://github.com/nats-io/nats.go/pull/1441. 
So I'll be closing this PR, but your contribution is merged into main (finally, sorry it took so long!).

Cheers, thanks for the help!

PR #1256, opened 2023-04-27
Add option for KeyWatcher to join a queue group

arnarg:
This can be useful when you want to react to events in a key value bucket but want it to be distributed across multiple instances of your application.

As far as I can tell this _does_ now require the user to explicitly stop the watcher so that the durable consumer that will be created (when a queue group is specified without a named consumer) is cleaned up. I explained this in the comment above the option.

### DIFF
--- a/kv.go
+++ b/kv.go
@@ -123,6 +123,8 @@ type watchOpts struct {
 	includeHistory bool
 	// retrieve only the meta data of the entry
 	metaOnly bool
+	// watch as part of a queue group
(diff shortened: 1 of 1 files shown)


arnarg:
I just realized a durable name can not be set on an ordered consumer. Back to the drawing board!

Is this a feature that is possible to support without knowing the internals of KV and subscribing directly to the `$KV.{{bucket}}.>`?

PR #1261, opened 2023-05-11
[CHANGED] Generate consumer name in Subscribe() when name is not provided

piotrpio:
This PR changes the way consumer is created in `Subscribe()` when not using `nats.Durable().
- Previously, client used `CONSUMER.CREATE.<stream_name>` subject to create ephemeral consumer, consumer name being generated on the server.
- Now, consumer name is generated in the client, so that either `CONSUMER.CREATE.<stream_name>.<consumer_name>` or `CONSUMER.CREATE.<stream_name>.<consumer_name>.<filter_subject>` is used


derekcollison:
Be good to have a Name() option for js.Subscribe() as well, don't think that exists.

piotrpio:
@derekcollison agreed, although I'd call it `ConsumerName()` to avoid confusion... I'll add it in a separate PR.

derekcollison:
Thanks!

PR #1263, opened 2023-05-12
Add ConsumerName option, use similar ephemeral names as server

wallyqs:


js.go
```diff
@@ -2474,6 +2476,14 @@ func ConsumerMemoryStorage() SubOpt {
 	})
 }
 
+// ConsumerName sets the name for a consumer.
```
derekcollison, line 2479: If I use this and set it, it will still use long form create yes?

js.go
```diff
@@ -3646,3 +3656,22 @@ func (st *StorageType) UnmarshalJSON(data []byte) error {
 	}
 	return nil
 }
+
+// Length of our hash used for named consumers.
+const nameHashLen = 8
+
+// Computes a hash of 8 characters for the name.
+func getHash(name string) string {
```
derekcollison, line 3664: Since we don't use any other size would collapse into one function with fixed size of 8 for now.

wallyqs: sounds good

wallyqs: updated to a single function

derekcollison (approved):
LGTM

PR #1264, opened 2023-05-12
[CHANGED] Reload TLS certificates on reconnect

piotrpio:
- Use callbacks when reconnecting to get current cert/key and CAs (similar to credentials)
- Send async error on failed connect (previously only protocol auth errors were reported)


nats.go
```diff
@@ -2474,6 +2509,9 @@ func (nc *Conn) sendConnect() error {
 		// Read the rest now...
 		proto, err = nc.readProto()
 		if err != nil {
+			if !nc.initc && nc.Opts.AsyncErrorCB != nil {
```
wallyqs, line 2512: These probably help would help with the visibility of the errors that I was mentioning above 👍, still think that might be better to just check that the callbacks won't work at config time (bad path to client cert) on connect than after connect init has started, otherwise I think last error on connect would be no servers available instead of the config error.

wallyqs (approved):
LGTM

PR #1266, opened 2023-05-16
[ADDED] Connection status event listeners 

piotrpio:
This PR enables receiving connection status change events on given channels. 

```go
dcChan := make(chan nats.Status)
nc.RegisterStatusChangeListener(nats.DISCONNECTED, dcChan)
...
<-dcChan
fmt.Println("disconnected")
```


Jarema (approved):
That looks fine by me.
The control is very granular, with easy to use API and a powerful granular one too.
However, this is pretty opinionated part of the client and I would like @wallyqs to take a look before merging :).

nats.go
```diff
@@ -5455,6 +5456,66 @@ func (nc *Conn) GetClientID() (uint64, error) {
...
+	for _, s := range statuses {
+		nc.RegisterStatusChangeListener(s, ch)
+	}
+	return ch
+}
+
+// RegisterStatusChangeListener registers a channel waiting for a specific status change event.
+// Status change events are non-blocking - if no receiver is waiting for the status change,
+// it will not be sent on the channel. Closed channels are ignored.
+func (nc *Conn) RegisterStatusChangeListener(status Status, ch chan Status) {
```
ripienaar, line 5473: Maybe `RegisterStatusChangeListener(ch chan Status, status ...Status)` so one can easily listen for just a few statusses?

It's a bit weird being the only channel based event listener in this client, but seems like a nice addition and good use of the channel anyway

wallyqs: I like the variadic idea, I was thinking maybe can make `ConnStatusChanged(...Status)` as a variadic with the defaults being `[]Status{CONNECTED, RECONNECTING, DISCONNECTED, CLOSED}` if nothing passed and keep only one (maybe `StatusChanges(...Status)` or `StatusChanged(...Status)`? so that it pairs better with the `Status() Status` API that is already there.

nats.go
```diff
@@ -5455,6 +5456,66 @@ func (nc *Conn) GetClientID() (uint64, error) {
 	return nc.info.CID, nil
 }
 
+// ConnStatusChanged returns a channel on which connection status changes will be reported.
+// Reported statuses: CONNECTED, RECONNECTING, DISCONNECTED, CLOSED.
+func (nc *Conn) ConnStatusChanged() chan Status {
```
wallyqs, line 5461: How about:

```go
func(nc *Conn) StatusChanged(...Status) chan Status 
```

so that it lines up better with the getter for connection status:
```go
func(nc *Conn) Status() Status
```

also ideally this would be a receive only channel here when but closing the channel seems the most straightforward way to unregister these.

wallyqs (commented):
Some comments, but like the idea 👍

nats.go
ripienaar: Yes, I like this.  Might be some value in being able to supply own chan but happy to leave that till a real need arise.

piotrpio: Agreed, I also like that. I was thinking more granular approach with providing own channel per-status might be useful, but I agree that maybe this is something we can add in the future if necessary.

And yeah, receive-only channel would be ideal and that's what I wanted to to do at first, but closing the channel seemed like a good way to unregister.

test/conn_test.go
```diff
@@ -2761,3 +2761,150 @@ func TestRetryOnFailedConnectWithTLSError(t *testing.T) {
...
+		s.Shutdown()
+		s = RunDefaultServer()
+		defer s.Shutdown()
+		if err := nc.Publish("foo", []byte("msg")); err != nil {
+			t.Fatalf("Unexpected error")
+		}
+	})
+}
+
+func TestConnStatusChangedEvents(t *testing.T) {
```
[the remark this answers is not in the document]
piotrpio: It was flapping because CONNECTED status was being sent twice on successful reconnects - once in `sendConnect()` and then again in `doReconnect()`. The latter seems obsolete, so I dropped sending and event at that point.

wallyqs (approved):
LGTM!

PR #1282, opened 2023-06-01
[FIXED] Cleanup reply sub after putting object

piotrpio:
Async reply subscription was created on each `object.Put()` and not cleaned up. This PR closes this subscription after `Put()` completes.

### DIFF
--- a/object.go
+++ b/object.go
@@ -370,14 +370,17 @@ func (obs *obs) Put(meta *ObjectMeta, r io.Reader, opts ...ObjectOpt) (*ObjectIn
 	}
 
 	// Create our own JS context to handle errors etc.
-	js, err := obs.js.nc.JetStream(PublishAsyncErrHandler(func(js JetStream, _ *Msg, err error) { setErr(err) }))
+	jetStream, err := obs.js.nc.JetStream(PublishAsyncErrHandler(func(js JetStream, _ *Msg, err error) { setErr(err) }))
 	if err != nil {
 		return nil, err
 	}
+
+	defer jetStream.(*js).cleanupReplySub()
+
 	purgePartial := func() {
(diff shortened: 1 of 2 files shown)


object.go
```diff
@@ -370,14 +370,17 @@ func (obs *obs) Put(meta *ObjectMeta, r io.Reader, opts ...ObjectOpt) (*ObjectIn
 	}
 
 	// Create our own JS context to handle errors etc.
-	js, err := obs.js.nc.JetStream(PublishAsyncErrHandler(func(js JetStream, _ *Msg, err error) { setErr(err) }))
+	jetStream, err := obs.js.nc.JetStream(PublishAsyncErrHandler(func(js JetStream, _ *Msg, err error) { setErr(err) }))
 	if err != nil {
 		return nil, err
 	}
+
+	defer jetStream.(*js).cleanupReplySub()
```
wallyqs, line 378: qq, why do we have to type assert here?

piotrpio: `nc.JetStream()` returns an interface, not `*js` so I could either add `cleanupReplySub()` as interface method or type assert here - I decided not to clutter the interface with another method (eve unexported), but not sure which approach is better.

wallyqs: ah right... makes sense

wallyqs:
likely unrelated flapper but might be worth looking at: 
```
=== RUN   TestPullSubscribeFetchBatch/cancel_context_during_fetch
    js_test.go:1423: Expected 5 messages; got: 0
--- FAIL: TestPullSubscribeFetchBatch (0.37s)
```

piotrpio:
@wallyqs yeah, from what I saw that's a flapper and it's on the list :)

Issue #1279, opened 2023-05-31
Data race with the 1.26 jetstream.Consume API


alexisvisco:
## Defect

There is a datarace with the Consume API of the jetstream package with the Nats version package 1.26

#### Versions of `nats.go` and the `nats-server` if one was involved:
2.9.3 - nats:2.9.3-alpine

#### OS/Container environment:
Hardware Overview:

  Model Name:	MacBook Pro
  Model Identifier:	MacBookPro18,2
  Model Number:	Z14X000MTFN/A
  Chip:	Apple M1 Max
  Total Number of Cores:	10 (8 performance and 2 efficiency)
  Memory:	64 GB
  System Firmware Version:	8419.80.7
  OS Loader Version:	8419.80.7
  Serial Number (system):	CY3F9G0QMC
  Hardware UUID:	B43CF36A-5AEA-5935-9C7B-70A01DA1C3C2
  Provisioning UDID:	00006001-000A51D101F8801E
  Activation Lock Status:	Disabled

#### Steps or code to reproduce the issue:
```go
package main

import (
	"context"
	"fmt"
	"github.com/nats-io/nats.go"
	"github.com/nats-io/nats.go/jetstream"
	"time"
)

func main() {
	nc, err := nats.Connect("localhost:4223")
	if err != nil {
		panic(err)
	}

	stream, err := jetstream.New(nc)
	if err != nil {
		panic(err)
	}

	_, err = stream.Stream(context.Background(), "hook-stream")
	if err == jetstream.ErrStreamNotFound {
		_, err = stream.CreateStream(context.Background(), jetstream.StreamConfig{
			Name:        "hook-stream",
			Description: "receipt every hooks",
			Subjects:    []string{"hooks.*"},
			Retention:   jetstream.LimitsPolicy,
			MaxAge:      time.Hour,
		})
		if err != nil {
			panic(err)
		}
	} else if err != nil {
		panic(err)
	}

	go func() {
		for {
			time.Sleep(time.Second)
			_, err = stream.Publish(context.Background(), "hooks.hello", []byte("Hello World"))
			if err != nil {
				panic(err)
			}
		}
	}()

	go consumer(err, stream, "hook-stream", "hook-consumer-hello")

	// wait forever
	for {
	}

}

func consumer(err error, stream jetstream.JetStream, sName, cName string) {
	consumer, err := stream.AddConsumer(context.Background(), sName, jetstream.ConsumerConfig{
		Name:          cName,
[...]

PR #1291, opened 2023-06-05
jetstream: fix datarace when using consumer #1279

alexisvisco:
Hi, I opened the issue #1279 and give a try to fix it. It seems that some code using the pending field on the pull struct was not protected at certain point. 

With the help of the stacktrace of the datarace I fixed it.

### DIFF
--- a/jetstream/pull.go
+++ b/jetstream/pull.go
@@ -652,8 +652,7 @@ func (p *pullConsumer) fetch(req *pullRequest) (MessageBatch, error) {
 				}
 				userMsg, err := checkMsg(msg)
 				if err != nil {
-					if !errors.Is(err, nats.ErrTimeout) && !errors.Is(err, ErrNoMessages) && !errors.Is(err,
-						ErrMaxBytesExceeded) {
+					if !errors.Is(err, nats.ErrTimeout) && !errors.Is(err, ErrNoMessages) && !errors.Is(err,ErrMaxBytesExceeded) {
 						res.err = err
 					}
 					res.done = true
--- a/jetstream/pull.go
+++ b/jetstream/pull.go
@@ -505,7 +505,6 @@ func (s *pullSubscription) Next() (Msg, error) {
 }
 
 func (s *pullSubscription) handleStatusMsg(msg *nats.Msg, msgErr error) error {
-	s.TryLock()
 	if !errors.Is(msgErr, nats.ErrTimeout) && !errors.Is(msgErr, ErrMaxBytesExceeded) {
 		if s.consumeOpts.ErrHandler != nil {
 			s.consumeOpts.ErrHandler(s, msgErr)
--- a/jetstream/pull.go
+++ b/jetstream/pull.go
@@ -652,7 +652,8 @@ func (p *pullConsumer) fetch(req *pullRequest) (MessageBatch, error) {
 				}
 				userMsg, err := checkMsg(msg)
 				if err != nil {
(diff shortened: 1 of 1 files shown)


jetstream/pull.go
```diff
@@ -185,16 +185,21 @@ func (p *pullConsumer) Consume(handler MessageHandler, opts ...PullConsumeOpt) (
 			return
 		}
 		defer func() {
-			if sub.pending.msgCount < consumeOpts.ThresholdMessages ||
-				(sub.pending.byteCount < consumeOpts.ThresholdBytes && sub.consumeOpts.MaxBytes != 0) &&
+			sub.Lock()
+			pending := sub.pending
```
alexisvisco, line 189: Copying the value of pending because next we are using `resetPendingMsgs` and it locks in the function itself so I can't create a lock defer unlock in addition we are already in a defer so it might seems overly complicated to defer again.

jetstream/pull.go
```diff
@@ -203,7 +208,12 @@ func (p *pullConsumer) Consume(handler MessageHandler, opts ...PullConsumeOpt) (
 			if msgErr == nil {
 				return
 			}
-			if err := sub.handleStatusMsg(msg, msgErr); err != nil {
+
+			sub.Lock()
+			err := sub.handleStatusMsg(msg, msgErr)
```
alexisvisco, line 213: Can't put the lock inside the `handleStatusMsg` because it is used in `Next` which is globally locked.

jetstream/pull.go
```diff
@@ -642,7 +652,8 @@ func (p *pullConsumer) fetch(req *pullRequest) (MessageBatch, error) {
 				}
 				userMsg, err := checkMsg(msg)
 				if err != nil {
-					if !errors.Is(err, nats.ErrTimeout) && !errors.Is(err, ErrNoMessages) && !errors.Is(err, ErrMaxBytesExceeded) {
+					if !errors.Is(err, nats.ErrTimeout) && !errors.Is(err, ErrNoMessages) && !errors.Is(err,
+						ErrMaxBytesExceeded) {
```
piotrpio, line 656: Moving a single parameter to new line looks a bit ugly - either leave everything in single line or split conditions to separate lines.

jetstream/pull.go
piotrpio: Yeah, unfortunately `Next()` has to be locked as a whole, which is a bit painful considering the logic of handling `pendingMsgs` and status messages is common across `Consume` and `Next` - I'll probably clean that up for the next release, but for now that will work.

piotrpio (commented):
One minor nit, other than that looks great!

jetstream/pull.go
alexisvisco: Yes looks like my formatter is doing something crazy

jetstream/pull.go
alexisvisco: A potential trick (to not refactor the whole logic) is to use `TryLock` but the go team does not advise to use it because it means something is probably wrong. 

https://pkg.go.dev/sync#Mutex.TryLock
. Note that while correct uses of TryLock do exist, they are rare, and use of TryLock is often a sign of a deeper problem in a particular use of mutexes.

alexisvisco:
@piotrpio  the PR looks good, I think we just need to restart the job because travis give us : 
```
Unable to start NATS Server in Go Routine [recovered]
	panic: Unable to start NATS Server in Go Routine
	```

piotrpio (approved):
@alexisvisco , yeah, that's not linked to your PR, I'll be fixing that tests. LGTM!

PR #1295, opened 2023-06-05
Fix incorrect example in jetstream README

bojanz:
The options right below the example are correct, but the example itself uses what I assume are outdated method names.


piotrpio (approved):
LGTM!

PR #1300, opened 2023-06-07
[CHANGED] Simplified JetStream API changes and improvements

piotrpio:
- Fixed typo in `PullThresholdBytes` option
- Removed push consumer only fields from `ConsumerConfig`
- Removed `context.Context` from `PublishAsync` and `PublishMsgAsync`
- Renamed `AddConsumer` to `CreateOrUpdateConsumer` - this change is introduced in anticipation for separation of create and update operations in nats-server. After they'll be introduced, there will be 3 methods:
  - `CreateOrUpdateConsumer()` - current upsert behavior
  - `CreateConsumer()` - will not overwrite consumer
  - `UpdateConsumer()` - will fail if consumer does not exist

### DIFF
--- a/examples/jetstream/js-consume/main.go
+++ b/examples/jetstream/js-consume/main.go
@@ -47,7 +47,7 @@ func main() {
 		log.Fatal(err)
 	}
 
-	cons, err := s.AddConsumer(ctx, jetstream.ConsumerConfig{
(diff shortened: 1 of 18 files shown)


derekcollison:
IIRC we could block PublishAsync based on the outstanding ack window.

PR #1312, opened 2023-06-13
[CHANGED] Simplified API for listing streams and stream names, added option to filter by stream name

piotrpio:
- `ListStreams` and `StreamNames` now allow for iterating over returned channel (channel is always closed)
- `ListStreams` and `StreamNames` now both accept `WithStreamListSubject` for filtering out returned streams


Jarema (approved):
LGTM!

PR #1314, opened 2023-06-13
Fix Race condition in object store with nats context

### LINKED ISSUE #1313: Object Store Get with a nats.Context causes a race condition
## Defect

* `nats.go` version = `v1.27.0`
mac
Perform a `Get` on an `ObjectStore` while providing the `nats.Context` as a `GetObjectOpt` causes a race condition.

If we modify the `TestObjectBasics` test to add a `nats.Context(context.Background()` as an option to the seciont that gets the object back (line 95), the test will have a race condition.

```
func TestObjectBasics(t *testing.T) {
.
.
.
	// Now get the object back.
	result, err := obs.Get("BLOB", nats.Context(context.Background()))
	expectOk(t, err)
	expectOk(t, result.Error())
	defer result.Close()
.
.
.
```

#### Expected result:
The test to pass without any race conditions.

#### Actual result:
The test will pass when running without the `-race` flag, but as soon as we add the `-race` flag, then the test fails due to the race condition.

anthonyjacques20:
The `err` variable can be accessed in separate `goroutines` and that causes a race condition. This is demonstrated in the `object_test` file.

It doesn't look like the `err` in the `processChunk` function needs to be the same error used elsewhere in the `Get` method. By creating an `err` variable in the `processChunk` function, there is no race condition anymore.

Close #1313

### DIFF
--- a/object.go
+++ b/object.go
@@ -622,6 +622,7 @@ func (obs *obs) Get(name string, opts ...GetObjectOpt) (ObjectResult, error) {
 	result.digest = sha256.New()
 
 	processChunk := func(m *Msg) {
+		var err error
 		if ctx != nil {
 			select {
 			case <-ctx.Done():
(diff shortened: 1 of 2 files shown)


wallyqs (approved):
LGTM, thanks for the fix 👍

PR #1315, opened 2023-06-14
[IMPROVED] Replace empty interface with any

piotrpio:


ripienaar:
Are we ready to say no more 1.17 support ?

piotrpio:
@ripienaar I would say yes - go version in `go.mod` should indicate the minimal go version under which the package compiles, for us it's 1.19. We'll of course note this in release notes, but I do not think we should defer using language features introduced 3 minor releases ago.

PR #1317, opened 2023-06-14
Create consumer on OrderedConsumer, add missing tests

piotrpio:
- Previously client was lazily initializing consumer when `OrderedConsumer` was used, but that made `Info()` return an error if invoked before consuming messages
- Added some missing unit tests


Jarema (approved):
LGTM!

tpihl:
@piotrpio, a reflection;

All tests (as I read them) are trying the scenario where the stream AND the consumer is created before the msg(s) is emitted to the stream.

Can that order make a difference, i.e. create stream, emit msg(s) and THEN create consumer. I think this is the much more common scenario, the "start with empty stream" is only the corner-case just when stream was created.

Just a reflection; T

piotrpio:
@tpihl Good point, I will add such tests 👍

PR #1322, opened 2023-06-17
The private inbox of the connection is used.

### LINKED ISSUE #1307: New jetstream interface: Problem using private inbox (was: How to get subjects in a stream?)
How would I replace this when using the new API?

```
sInfo, _ := js.StreamInfo(streamName, &nats.StreamInfoRequest{SubjectsFilter: subject})
_, found := sInfo.State.Subjects[subject]
if !found {
	if listen == 0 {
		return nil, os.ErrNotExist
	}
}
```

oderwat:
Previously, the normal inbox was used, which is a security risk in real applications.

Fixes #1307


piotrpio (approved):
LGTM!

oderwat:
Is there an ETA for this fix going into a release?

PR #1325, opened 2023-06-19
[FIXED] Creating ephemeral consumers on server < 2.9.0

### LINKED ISSUE #1321: Jetstream K/V bucket fails due to consumer name not being durable
## Defect
Provided these code snippets:
```go
	kv, err := jetstream.KeyValue("foo")
	switch {
	case errors.Is(err, nats.ErrBucketNotFound):
		return jetstream.CreateKeyValue(&nats.KeyValueConfig{
			Bucket:      "foo",
			Description: "foobar",
		})

	stuff, err := s.kv.Keys()
	switch {
	// If we find that theres no keys then init is a no-op
	case errors.Is(err, nats.ErrNoKeysFound):
		return nil
	case err != nil:
		return err
	}
```
I receive the following error:
```
nats: consumer expected to be durable but a durable name was not set
```

Client library version: `v1.26.0+`. This bug does NOT exist in `v1.25.0`.

#### Expected result:

No errors when creating accessing k/v buckets.

#### Actual result:

The below error:
```
nats: consumer expected to be durable but a durable name was not set
```

piotrpio:
Resolves #1321

### DIFF
--- a/jsm.go
+++ b/jsm.go
@@ -361,19 +361,28 @@ func (js *js) upsertConsumer(stream, consumerName string, cfg *ConsumerConfig, o
 
 	var ccSubj string
 	if consumerName == _EMPTY_ {
-		// if consumer name is empty, use the legacy ephemeral endpoint
+		// if consumer name is empty (neither Durable nor Name is set), use the legacy ephemeral endpoint
 		ccSubj = fmt.Sprintf(apiLegacyConsumerCreateT, stream)
 	} else if err := checkConsumerName(consumerName); err != nil {
 		return nil, err
-	} else if !js.nc.serverMinVersion(2, 9, 0) || (cfg.Durable != "" && js.opts.featureFlags.useDurableConsumerCreate) {
-		// if server version is lower than 2.9.0 or user set the useDurableConsumerCreate flag, use the legacy DURABLE.CREATE endpoint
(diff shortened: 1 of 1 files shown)


jsm.go
```diff
@@ -361,20 +361,29 @@ func (js *js) upsertConsumer(stream, consumerName string, cfg *ConsumerConfig, o
...
 		ccSubj = fmt.Sprintf(apiLegacyConsumerCreateT, stream)
 	} else if err := checkConsumerName(consumerName); err != nil {
 		return nil, err
-	} else if !js.nc.serverMinVersion(2, 9, 0) || (cfg.Durable != "" && js.opts.featureFlags.useDurableConsumerCreate) {
-		// if server version is lower than 2.9.0 or user set the useDurableConsumerCreate flag, use the legacy DURABLE.CREATE endpoint
-		ccSubj = fmt.Sprintf(apiDurableCreateT, stream, consumerName)
-	} else {
-		// if above server version 2.9.0, use the endpoints with consumer name
-		if cfg.FilterSubject == _EMPTY_ || cfg.FilterSubject == ">" {
+	} else if js.nc.serverMinVersion(2, 9, 0) {
```
ripienaar, line 368: This kind of check is just not useful for these cases. The server version you are connected to do does not tell you about the jetstream API

Jarema (approved):
LGTM!

PR #1327, opened 2023-06-19
[ADDED] HeadersOnly field on OrderedConsumerConfig

### LINKED ISSUE #1323: new jetstream interface does not allow an OrderedConsumer to be HeadersOnly. Why?
Sometimes headers is enough, why force OrderedConsumer to bring body when normal consumer can be HeadersOnly? I think it makes sense to add HeadersOnly to the OrderedConsumerConfig.

/T

piotrpio:
Resolves #1323


Jarema (approved):
LGTM!

PR #1329, opened 2023-06-22
New JetStream client docs: Fixed unit from kB to B in documentation o…

fabiant7t:
The number argument of the consumer's FetchBytes method means bytes, not kilobytes.
Fixed the comment.


piotrpio (commented):
LGTM!

PR #1330, opened 2023-06-22
Revert "[IMPROVED] Replace empty interface with any"

piotrpio:
This reverts change from #1315 before a patch release and will be added again after the release.


Jarema (approved):
LGTM

PR #1332, opened 2023-06-23
[IMPROVED] Replace empty interface with any

piotrpio:
This PR restores changes reverted for patch release in #1330


Jarema (approved):
LGTM

Issue #1333, opened 2023-06-23
Inconsistency in TestParseNum test function


YukiBobier:
In `internal/parser/parse_test.go`, I found an inconsistency in the TestParseNum function.

Specifically, the test case named "parse 0" is as follows:
```go
{
	name:     "parse 0",
	given:    "",
	expected: 0,
},
```

The given string in this case is empty, but according to the name of the test case ("parse 0"), it seems like it should be given the string "0" instead.

This inconsistency can lead to confusion about the purpose of the test case, and may also mean the function is not being tested as thoroughly as possible.

I am planning to fix this issue by modifying the given string in this test case to "0" to match the test name and ensure the function is correctly tested with this input. I will follow up with a pull request addressing this issue.

PR #1334, opened 2023-06-23
[FIXED] Correct input for "parse 0" test case

YukiBobier:
# Resolve issue #1333

This change aligns the input with the name of the test case and ensures that the function is accurately tested with this input.

### DIFF
--- a/internal/parser/parse_test.go
+++ b/internal/parser/parse_test.go
@@ -35,7 +35,7 @@ func TestParseNum(t *testing.T) {
 		},
 		{
 			name:     "parse 0",
-			given:    "",
+			given:    "0",
 			expected: 0,
 		},
 		{


piotrpio (changes requested):
Hey @YukiBobier , thanks for the PR, indeed this may have been confusing.

Just one thing - I think we should still have a test where the input is an empty string, just with better test case name ("parse empty string"). That would mean that we should have 2 test cases instead of 1 - one with "0" as input and other with "" as input.

YukiBobier:
> Hey @YukiBobier , thanks for the PR, indeed this may have been confusing.
> 
> Just one thing - I think we should still have a test where the input is an empty string, just with better test case name ("parse empty string"). That would mean that we should have 2 test cases instead of 1 - one with "0" as input and other with "" as input.

Hi @piotrpio ,

Thanks for your feedback on this pull request.

I wanted to point out that there is already an existing test case that tests the function with an empty string as input. Here's the test case:
https://github.com/nats-io/nats.go/blob/7c2bded3d4788650393b681c5df4a1bf8553bd15/internal/parser/parse_test.go#L46-L50

Given this, I've made the change to have the "parse 0" test case use the string "0" as input, as per the test case name.

If there's any misunderstanding on my part, please let me know.

piotrpio (approved):
@YukiBobier My bad, I missed that there was a test case for empty string already.

LGTM, thanks for cotribution!

YukiBobier:
@piotrpio No problem at all! I'm glad the changes look good to you. Thanks for your approval and for taking the time to review my pull request.

I'm really happy to contribute to this project and look forward to making further contributions.

PR #1337, opened 2023-06-27
[CHANGED] Move NakWithDelay to separate method

### LINKED ISSUE #1299: API Design nak with delay non testable with mocks
## Feature Request

In the new jetstream API there is a change where the `NakWithDelay`  was replaced by the `Nak(jetstream.WithNakDelay(...))`

Prior to the update I was mocking a nats.Message with a custom interface and I was able to check if a nak occurred with a given duration.

Now I can't because the `NakWithDelay` is now a functional parameter option of the `Nak` function and all is private so I can't check what options I pass to the nak.

Is this something that could be changed ? 

#### Use Case:

Testing that a nak with delay contains a certain duration 

#### Proposed Change:

Maybe exposing the ackOpts and his fields ?

Every developers that mocks jetstream.Message and has delay with nak

piotrpio:
This cleans up the API a bit, avoiding having a method with a single option to modify the behavior.

Resolves #1299

### DIFF
--- a/jetstream/message.go
+++ b/jetstream/message.go
@@ -49,7 +49,11 @@ type (
 		DoubleAck(context.Context) error
 		// Nak negatively acknowledges a message
 		// This tells the server to redeliver the message
-		Nak(...NakOpt) error
+		Nak() error
+		// NakWithDelay negatively acknowledges a message
+		// This tells the server to redeliver the message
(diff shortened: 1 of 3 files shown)


Jarema (approved):
LGTM!

PR #1341, opened 2023-07-07
Unset start time for ordered consumer on retry

### LINKED ISSUE #1281: Restarting Ordered Consumer by Start Time
https://github.com/nats-io/nats.go/blob/7c684685694621ba5e98b2471b3a11618d03da6f/js.go#L1983

If we are using an ordered consumer with a particular start time. On the consumer restarting I think it tries to resume at the sequence it last left off at, but this is in conflict with the start time. If the consumer was already successfully running with the deliver policy `DeliverByStartTime` then it should clear the start time option when resuming with `DeliverByStartSequence`

bruth:
If an ordered consumer was configured with a start time, on retry, the copied consumer config would start with sequence but the existing start time opt would conflict causing a config error.

Fix #1281

---

Just noting that I added the `StartTime()` opt to the test which, without clearing start time on retry (the fix), it will demonstrate the behavior and get this message:

```
nats: consumer delivery policy is deliver by start sequence, but optional start time is also set: recreating ordered consumer on connection [4] for subscription on "a"
```

### DIFF
--- a/js.go
+++ b/js.go
@@ -1991,6 +1991,9 @@ func (sub *Subscription) resetOrderedConsumer(sseq uint64) {
 		cfg.DeliverSubject = newDeliver
 		cfg.DeliverPolicy = DeliverByStartSequencePolicy
 		cfg.OptStartSeq = sseq
+		// In case the consumer was created with a start time, we need to clear it
+		// since we are now using a start sequence.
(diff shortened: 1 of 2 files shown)


piotrpio (approved):
LGTM!

PR #1345, opened 2023-07-13
[FIXED] Invalid heartbeat errors in Consume()

### LINKED ISSUE #1343: New JetStream API: calling `Stop()` makes `Next()` block indefinitely or until a missing heartbeat error
## Defect

- nats.go: `v1.27.1`
- nats-server: `v2.9.19`

- OS: `linux`
-  Arch: `amd64`
- Go: `v1.20.5`

1. Use the new JetStream API consumer with `Messages()`
```go
iter, err := cons.Messages(jetstream.PullMaxMessages(1))
if err != nil {
	log.Fatal(err)
}
```
2. Call the `Next()` method in a **separate** goroutine
```go
var wg sync.WaitGroup

wg.Add(1)
go func() {
	defer wg.Done()

	for {
		// Next() blocks until a message is available.
		// After calling Stop() keeps blocking until we get an ErrNoHeartbeat,
		// then next call will return ErrMsgIteratorClosed because of the
		// atomic.LoadUint32(&s.closed) == 1 check in Next().
                // If ReportMissingHeartbeats is false, then it will block indefinitely.
		msg, err := iter.Next()

		if err != nil {
			if errors.Is(err, jetstream.ErrMsgIteratorClosed) {
				fmt.Println("Iterator closed")
				break
			}

			fmt.Println("Next err: ", err)
			// If Stop() was called then the next Next() call will return ErrMsgIteratorClosed.
			continue
		}

		fmt.Println(string(msg.Data()))
		msg.Ack()
	}
}()
```
3. Call the `Stop()` method to stop consuming messages
```go
iter.Stop() // Should make Next() return ErrMsgIteratorClosed
```

In the [example](https://github.com/nats-io/nats.go/blob/main/examples/jetstream/js-messages/main.go) showing the usage of the new JetStream API with `Messages()`, this issue is not apparent because `Next()` is called in the **main** goroutine.

Full example: https://github.com/mdawar/jetstream-api-demos/tree/main/shutdown

#### Expected result:
The `Stop()` method should make `Next()` return an error (Possibly `ErrMsgIteratorClosed`).

#### Actual result:
There are 2 results depending on the configuration:

##### 1. `ReportMissingHeartbeats` is `true` (default)

After calling `Stop()` consuming messages stops (the `done` channel is closed, and `pullMessages` stops fetching messages) but `Next()` keeps on blocking until there's an error `ErrNoHeartbeat` then exiting depends on breaking the loop or returning after the error.

##### 2. `ReportMissingHeartbeats` is `false`

```go
cons.Messages(jetstream.WithMessagesErrOnMissingHeartbeat(false))
```

After calling `Stop()` the `Next()` call will block indefinitely (Process must be killed: `kill -9 PID`).

piotrpio:
Resolves #1343

### DIFF
--- a/jetstream/README.md
+++ b/jetstream/README.md
@@ -413,8 +413,12 @@ and new subscription is created for each execution.
 #### Continuous polling
 
 There are 2 ways to achieve push-like behavior using pull consumers in
-`jetstream` package. Both `Messages()` and `Consume()` methods perform exactly
-the same optimizations and can be used interchangeably.
+`jetstream` package. Both `Messages()` and `Consume()` methods perform similar optimizations
+and for most cases can be used interchangeably.
+
+There is an advantage of using `Messages()` instead of `Consume()` for work-queue scenarios,
+where messages should be fetched one by one, as it allows for finer control over fetching
+single messages on demand.
 
 Subject filtering is achieved by configuring a consumer with a `FilterSubject`
 value.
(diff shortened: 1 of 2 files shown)


Jarema (approved):
LGTM!

PR #1346, opened 2023-07-14
[FIXED] PublishAsyncMaxPending not reset after server disconnect

piotrpio:
This fixes an issue with `PublishAsync()` where pending acks were not being reset on disconnect. This caused the pending acks counter to grow (up to the limit) until connection is closed.


aricart (approved):
LGTM

Jarema (approved):
LGTM!

PR #1364, opened 2023-08-11
Fix deadlock caused by locked mutex

Zamony:
Mutex must be unlocked on return from function.


piotrpio (approved):
LGTM, thank you for the contribution!

PR #1374, opened 2023-08-22
[IMPROVED] Fixed flaky tests in js.Subscribe() and kv

piotrpio:
This PR fixes 2 frequent flaky tests:
- `TestJetStreamSubscribe_AckPolicy()` - in this test `sub.Drain()` was executed before getting consumer info. Even though `nats.Durable()` option was used in `Subscribe()`, the consumer is still deleted if created by library
- `TestKeyValueMirrorCrossDomains()` - there were races in this test and it sometimes failed if values were not propagated to the mirror before attempting to get them

### DIFF
--- a/test/js_test.go
+++ b/test/js_test.go
@@ -4155,12 +4155,11 @@ func TestJetStreamSubscribe_AckPolicy(t *testing.T) {
 			}
 
 			<-ctx.Done()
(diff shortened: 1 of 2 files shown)


wallyqs (approved):
LGTM

PR #1378, opened 2023-08-25
[ADDED] UpdatesOnly() option for object store watchers

piotrpio:
- Added `UpdatesOnly()` option to object store watchers
- Modify KV watcher behavior for `UpdatesOnly()` so that it does not send `nil` update after creating the watcher


aricart (approved):
LGTM

Jarema (approved):
LGTM!

PR #1379, opened 2023-08-26
[ADDED] CreateConsumer and UpdateConsumer methods to work with server 2.10 capabilities

piotrpio:


jnmoyne (approved):
Just small nit about the readme still having an example using CreateOrUpdate, otherwise LGTM!

PR #1384, opened 2023-08-29
[ADDED] Stream and consumer metadata

piotrpio:
Addresses: https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-33.md


Jarema (approved):
LGTM

PR #1395, opened 2023-09-11
feat: implement CreateOrUpdateStream function

samanebi:
After using JetStream on a private project I felt the need for CreateOrUpdateStream like CreateOrUpdateConsumer. This function updates if stream exists, otherwise it creates the stream.

### DIFF
--- a/jetstream/jetstream.go
+++ b/jetstream/jetstream.go
@@ -425,16 +425,10 @@ func (js *jetStream) UpdateStream(ctx context.Context, cfg StreamConfig) (Stream
 func (js *jetStream) CreateOrUpdateStream(ctx context.Context, cfg StreamConfig) (Stream, error) {
 	s, err := js.UpdateStream(ctx, cfg)
 	if err != nil {
-		if err == ErrStreamNotFound {
-			s, err = js.CreateStream(ctx, cfg)
-			if err != nil {
-				return nil, err
-			}
-
-			return s, nil
+		if !errors.Is(err, ErrStreamNotFound) {
+			return nil, err
 		}
-
-		return nil, err
+		return js.CreateStream(ctx, cfg)
 	}
 
 	return s, nil
--- a/jetstream/jetstream.go
+++ b/jetstream/jetstream.go
@@ -69,6 +69,8 @@ type (
 		CreateStream(ctx context.Context, cfg StreamConfig) (Stream, error)
 		// UpdateStream updates an existing stream
 		UpdateStream(ctx context.Context, cfg StreamConfig) (Stream, error)
+		// CreateOrUpdateStream creates a stream with given config. If stream already exists, it will be updated (if possible).
+		CreateOrUpdateStream(ctx context.Context, cfg StreamConfig) (Stream, error)
 		// Stream returns a [Stream] hook for a given stream name
 		Stream(ctx context.Context, stream string) (Stream, error)
 		// StreamNameBySubject returns a stream name stream listening on given subject
@@ -418,6 +420,23 @@ func (js *jetStream) UpdateStream(ctx context.Context, cfg StreamConfig) (Stream
 		name:      cfg.Name,
 		info:      resp.StreamInfo,
 	}, nil
+}
+
+func (js *jetStream) CreateOrUpdateStream(ctx context.Context, cfg StreamConfig) (Stream, error) {
+	var s Stream
+	s, err := js.UpdateStream(ctx, cfg)
+	if err != nil {
+		if err == ErrStreamNotFound {
+			s, err = js.CreateStream(ctx, cfg)
+			if err != nil {
+				return nil, err
+			}
+		}
+
+		return nil, err
+	}
+
+	return s, nil
 }
 
 // Stream returns a [Stream] hook for a given stream name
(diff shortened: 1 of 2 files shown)


piotrpio:
Thanks you for creating the PR! The difference between creating streams and consumers is server side - for consumers, there used to be a single API for both creating and updating a consumer, thus it was a single operation on the client (that will change with the upcoming server release though, and we will be adding `CreateConsumer` and `UpdateConsumer` to the client as well). For streams, there are 2 separate APIs which means we can do safe creates and safe updates, which should usually be the preferred way.

That said, I am not strongly against adding `CreateOrUpdateStream` as convenience method, however that should probably be then reflected in other client libraries as well. Let me get back to you with this.

samanebi:
> Thanks you for creating the PR! The difference between creating streams and consumers is server side - for consumers, there used to be a single API for both creating and updating a consumer, thus it was a single operation on the client (that will change with the upcoming server release though, and we will be adding `CreateConsumer` and `UpdateConsumer` to the client as well). For streams, there are 2 separate APIs which means we can do safe creates and safe updates, which should usually be the preferred way.
> 
> That said, I am not strongly against adding `CreateOrUpdateStream` as convenience method, however that should probably be then reflected in other client libraries as well. Let me get back to you with this.

Thanks :) I would be happy to help to implement this function in other clients

MohsenKasraeifar:
this feature is going to be a very very useful feature and will use it definitely.

jetstream/jetstream.go
```diff
@@ -420,6 +422,24 @@ func (js *jetStream) UpdateStream(ctx context.Context, cfg StreamConfig) (Stream
 	}, nil
 }
 
+func (js *jetStream) CreateOrUpdateStream(ctx context.Context, cfg StreamConfig) (Stream, error) {
+	s, err := js.UpdateStream(ctx, cfg)
+	if err != nil {
+		if err == ErrStreamNotFound {
```
piotrpio, line 428: I would suggest using `errors.Is()` instead of `==` for comparing errors for consistency with the rest of the package as well as safeguarding from error wrapping.

jetstream/test/jetstream_test.go
```diff
@@ -431,6 +431,236 @@ func TestCreateStreamMirrorCrossDomains(t *testing.T) {
...
+				return
+			}
+			if err != nil {
+				t.Fatalf("Unexpected error: %v", err)
+			}
+		})
+	}
+}
+
+func TestCreateOrUpdateStream_CreateMirrorCrossDomains(t *testing.T) {
```
piotrpio, line 517: I don't think it's useful to have this test for `CreateOrUpdateStream` to be honest, we already test the same functionality in `TestCreateStreamMirrorCrossDomains` so it brings no value. Please remove it.

jetstream/test/jetstream_test.go
```diff
@@ -431,6 +431,236 @@ func TestCreateStreamMirrorCrossDomains(t *testing.T) {
 	}
 }
 
+func TestCreateOrUpdate_CreateStream(t *testing.T) {
```
piotrpio, line 434: Can you refactor this to have a single test for this function (i.e. instead of having `TestCreateOrUpdate_CreateStream` and `TestCreateOrUpdate_UpdateStream` to just have `TestCreateOrUpdateStream`.

jetstream/jetstream.go
```diff
@@ -420,6 +422,24 @@ func (js *jetStream) UpdateStream(ctx context.Context, cfg StreamConfig) (Stream
 	}, nil
 }
 
+func (js *jetStream) CreateOrUpdateStream(ctx context.Context, cfg StreamConfig) (Stream, error) {
```
piotrpio, line 425: This method could use some simplification :) You could get rid of excessive nesting and improve readability:
```go
func (js *jetStream) CreateOrUpdateStream(ctx context.Context, cfg StreamConfig) (Stream, error) {
	s, err := js.UpdateStream(ctx, cfg)
	if err != nil {
		if !errors.Is(err, ErrStreamNotFound) {
			return nil, err
		}
		return js.CreateStream(ctx, cfg)
	}

	return s, nil
}
```

piotrpio (changes requested):
The feature looks good and we will add it! I only have a few suggestions.

jetstream/jetstream.go
samanebi: Yes, i fixed this.

samanebi:
All requested changes has been applied 
@piotrpio

jetstream/test/jetstream_test.go
samanebi: yes, mentioned tests has been merged together

jetstream/test/jetstream_test.go
samanebi: ok. this test has been removed

jetstream/jetstream.go
samanebi: I have changed it to `errors.Is()`

samanebi:
> I am wondering when do you plan on merging this PR ? @piotrpio

Is there any update on this? I have done the suggested changes
@piotrpio 
@Jarema 
@derekcollison

piotrpio (approved):
LGTM!

piotrpio:
@samanebi Sorry for the delay, looks good! It'll be a part of the next release.

PR #1397, opened 2023-09-11
[ADDED] Invoke ErrorHandler on preparing connection string when reconnecting

### LINKED ISSUE #1338: JetStream Simplified API continuous polling with `Consume()` and `Messages()`
First thank you for the new JetStream API, it's very easy to use and really provides a better DX.

I have a few questions about continuous polling with `Consume()` and `Messages()` because I had different experiences with both even though the docs say that they're interchangeable:

> Both Messages() and Consume() methods perform exactly the same optimizations and can be used interchangeably.

I'm using NATS for a job queue with workers running on multiple servers and each worker can be set to process multiple jobs concurrently that might take between a couple of seconds to a couple of minutes to finish.

This is how I'm creating the stream and the consumer:

```go
stream, _ := js.CreateStream(ctx, jetstream.StreamConfig{
	Name:      "jobs",
	Subjects:  []string{"jobs.>"},
	Retention: jetstream.WorkQueuePolicy,
})

consumer, _ := stream.CreateOrUpdateConsumer(ctx, jetstream.ConsumerConfig{
	Durable:       "JobsConsumer",
	ReplayPolicy:  jetstream.ReplayInstantPolicy,
	AckPolicy:     jetstream.AckExplicitPolicy,
	AckWait:       5 * time.Minute,
	MaxAckPending: -1,
})
```

With `Consume()`, I have used the code from the [`js-parallel-consume`](https://github.com/nats-io/nats.go/blob/97ccbd884fa75a3a5e86bef0ee4c9374dc0dd2b5/examples/jetstream/js-parallel-consume/main.go#L59) example:
```go
for i := 0; i < 5; i++ {
	cc, err := cons.Consume(
		func(consumeID int) jetstream.MessageHandler {
			return func(msg jetstream.Msg) {
				fmt.Printf("Received msg on consume %d\n", consumeID)
                                time.Sleep(2 * time.Second) // Simulate work (Using 30s for slow workers)
				msg.Ack()
			}
		}(i),
                jetstream.PullMaxMessages(1),
		jetstream.ConsumeErrHandler(func(consumeCtx jetstream.ConsumeContext, err error) {
			fmt.Printf("consume error: %v\n", err)
		}),
	)

	if err != nil {
		log.Fatal(err)
	}

	defer cc.Stop()
}
```

I'm running locally 4 workers each running `Consume()` 5 times as shown in the loop above, 2 of these workers are set to simulate work for 30 seconds each and 2 others for 2 seconds only.

I'm publishing 100 messages for testing this setup:

```go
for i := 0; i < 100; i++ {
	fmt.Println("Publishing job", i)

	_, err := js.Publish(ctx, "jobs.info", []byte(fmt.Sprintf("Job payload: %d", i)))

	if err != nil {
		log.Printf("cannot publish job #%d: %v", i, err)
	}
}
```

My observations:
- The workers consume messages in order of their connection (I don't know if the ID is being used for distribution)
- All the `Consume` callbacks are also run in order on each worker before other workers start consuming messages
- I get `nats: no heartbeat received` errors on the slow workers with `Consume()` with sleep for 30s after the callbacks return
- After these errors the slow workers seem to be pulling messages more than `PullMaxMessages` which is set to `1`, I was expecting these messages to be processed by other idle workers (I don't know if this is a misconfiguration on my side).

I have solved all of these issues by using `Messages()`:

```go
iter, err := consumer.Messages(
	jetstream.PullMaxMessages(1),
	jetstream.WithMessagesErrOnMissingHeartbeat(true), // Added for consistency with Consume()
)
if err != nil {
	log.Fatal(err)
}

defer iter.Stop()

parallelConsumers := 5
sem := make(chan struct{}, parallelConsumers)

for {
	sem <- struct{}{}

	go func() {
		defer func() {
			<-sem
		}()

		msg, err := iter.Next()
		if err != nil {
			fmt.Println("next err: ", err)
			return
		}

		fmt.Printf("Received msg: %s\n", msg.Data())
		time.Sleep(2 * time.Second) // Simulate work
		msg.Ack()
	}()
}
```

Now the workers consume the messages equally, and without heartbeat errors.
And even if I'm running 5 consumers (calling `Next()`) on each worker, the messages are pulled by the available worker goroutines fairly.

So it seems that `Consume()` and `Messages()` do not have the same performance, unless I have been doing something wrong on my side.

My question are:
- Are the `Consume()` callbacks supposed to be long running or not? Heartbeats are missed when they're long running.
- How messages are distributed when using pull consumers? It seems the consumer ID or the order of connection affects this.
- In my `Messages()` example which is worker as intended, is it OK to call `Next` for multiple goroutines? I'm assuming yes because of the mutex used in the code.

Used versions:
- Go: `v1.20.5` (linux/amd64)
- NATS server: `v2.9.19`
- nats.go: `v1.27.1`

Demo repository:
https://github.com/mdawar/jetstream-api-demos/tree/main/workqueue

Thank you.

### LINKED ISSUE #1388: nc.processConnectInit() does not call errorHandler, if it fails
### What version were you using?

2.9.21

### What environment was the server running in?

Mac M1

### Is this defect reproducible?

Yes

### Given the capability you are leveraging, describe your expectation?

I expected the error handler to be called to report error on reconnection.

### Given the expectation, what is the defect you are observing?

https://github.com/nats-io/nats.go/blob/9153b83ed1141cfbef65bd8d7b4fe2e477edd1e5/nats.go#L2733

When `nc.processConnectInit()` fails, it does not call the errorHandler. We had a bug in our code, where we deleted the nats.cred file after initial connection. When nats server will restart, nats client will not be able to reconnect (because we have deleted the creds file).

The issue is that this was not reported in the error handler. Ultimately, we had to debug using breakpoint to find the issue. It be very helpful to the users when `nc.processConnectInit()` fails, report the error using error Handler.

piotrpio:
Resolves #1388

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -2501,6 +2501,9 @@ func (nc *Conn) sendConnect() error {
 	// Construct the CONNECT protocol string
 	cProto, err := nc.connectProto()
 	if err != nil {
+		if !nc.initc && nc.Opts.AsyncErrorCB != nil {
+			nc.ach.push(func() { nc.Opts.AsyncErrorCB(nc, nil, err) })
+		}
 		return err
 	}
 
--- a/nats_test.go
+++ b/nats_test.go
@@ -1575,6 +1575,38 @@ func TestUserCredentialsChainedFile(t *testing.T) {
 	}
 }
 
+func TestReconnectMissingCredentials(t *testing.T) {
+	ts := runTrustServer()
+	defer ts.Shutdown()
+
+	chainedFile := createTmpFile(t, []byte(chained))
+	defer os.Remove(chainedFile)
+
+	url := fmt.Sprintf("nats://127.0.0.1:%d", TEST_PORT)
+	errs := make(chan error, 1)
+	nc, err := Connect(url, UserCredentials(chainedFile), ErrorHandler(func(_ *Conn, _ *Subscription, err error) {
+		errs <- err
+	}))
+	if err != nil {
+		t.Fatalf("Expected to connect, got %v", err)
+	}
+	defer nc.Close()
+	os.Remove(chainedFile)
+	ts.Shutdown()
+
+	ts = runTrustServer()
+	defer ts.Shutdown()
+
+	select {
+	case err := <-errs:
+		if !strings.Contains(err.Error(), "no such file or directory") {
+			t.Fatalf("Expected error about missing creds file, got %q", err)
+		}
+	case <-time.After(5 * time.Second):
+		t.Fatal("Did not get error about missing creds file")
+	}
+}
+
 func TestUserJWTAndSeed(t *testing.T) {
 	if server.VERSION[0] == '1' {
 		t.Skip()


wallyqs (approved):
LGTM

PR #1398, opened 2023-09-12
[ADDED] Default value for FlusherTimeout

piotrpio:
This adds a default value for flusher timeout, preventing deadlock in certain circumstances

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -61,6 +61,7 @@ const (
 	DefaultReconnectBufSize   = 8 * 1024 * 1024 // 8MB
 	RequestChanLen            = 8
 	DefaultDrainTimeout       = 30 * time.Second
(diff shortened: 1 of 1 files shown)


nats.go
```diff
@@ -61,6 +61,7 @@ const (
 	DefaultReconnectBufSize   = 8 * 1024 * 1024 // 8MB
 	RequestChanLen            = 8
 	DefaultDrainTimeout       = 30 * time.Second
+	DefaultFlusherTimeout     = 30 * time.Second
```
wallyqs, line 64: maybe set to `time.Minute` that should be enough I think, and just before the default ping interval `2*time.Minute`

wallyqs (approved):
LGTM

PR #1401, opened 2023-09-14
[ADDED] Customizing queue groups in micro

piotrpio:
Addresses ADR-32 changes from https://github.com/nats-io/nats-architecture-and-design/pull/239


Jarema (approved):
LGTM

PR #1419, opened 2023-09-26
Use TLS from Memory

nh3000-org:
This PR adds using TLS from memory in addition to a file location. It provides a easier deployment for edge/iot devices from a single binary without the need to expose certificates on the file system. 

The use case is to help deploying to a large number of clients using a single binary for initial setup.

Testing added to conn_test.go  at TestClientCertificateMem  to test server and client certificates.  This tests both features being proposed.

Tested on Linux and Android.


nh3000-org:
Please remove this pr I see that the function is provided in TLSConfig.  Thank You.

PR #1420, opened 2023-09-27
[FIXED] Stream create response test for sources being present in the response.

jnmoyne:
Server 2.10 now returns source infos (as well as the config) so you can for example tell which sources where successfully added for the stream, and what their current state is) but 2.9 would only return the source config info inside the stream info.

nats.go 2.10 added a check that all of the sources defined in the config where indeed added to the stream in the stream create transform, which would work with a 2.10 server but not with a 2.9.

This fixes the test by downgrading it to the best you can do with a 2.9 server which is check against the stream config returned in the create response message.


ndeubert (approved):
Fixed my issue. :)

Jarema (approved):
LGTM

PR #1433, opened 2023-10-06
[ADDED] TLSHandshakeFirst option

kozlovic:
This allows to connect to a server that is also configured to perform the TLS handshake first, that is, before sending the INFO protocol.

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -311,6 +311,13 @@ type Options struct {
 	// TLSCertCB is used to fetch and return custom tls certificate.
 	TLSCertCB TLSCertHandler
 
+	// TLSHandshakeFirst is used to instruct the library perform
+	// the TLS handshake right after the connect and before receiving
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

piotrpio (approved):
Looks good, but I think we should wait with merging it until this feature is introduced in the server right?

kozlovic:
> I think we should wait with merging it until this feature is introduced in the server right?

@piotrpio Actually no. The server relies on the client new API for its tests. I made the client not depend on the server for the test (since it is quite simple here), but not the server. If you merge and soon after have to do a release while the server still does not have the feature, then it would have to be stated in the client's release notes that this option is to be used with servers at version x.y.z and above.

piotrpio:
@kozlovic Got it - in that case, I'm merging the PR

PR #1443, opened 2023-10-17
Compatibility tests for micro

piotrpio:
- added compatibility test for service api - complimentary PR in compat-tests: https://github.com/synadia-io/client-compatibility-tests/pull/14
- fixed a bug which caused `asyncDispatcher` to not close properly
- user errors are now reported in stats


Jarema (approved):
LGTM

PR #1449, opened 2023-10-25
Remove ordered consumer on reset and fix recreated consumer name

piotrpio:
- library now attempts to remove consumer befor creating a new one when resetting ordered consumer
- fixed recreated consumer name to be in the same format as the initial one (short hash of nuid instead of full nuid

### DIFF
--- a/js.go
+++ b/js.go
@@ -1241,6 +1241,10 @@ func (sub *Subscription) deleteConsumer() error {
 		sub.mu.Unlock()
 		return nil
 	}
+	if jsi.stream == _EMPTY_ || jsi.consumer == _EMPTY_ {
+		sub.mu.Unlock()
+		return nil
+	}
 	stream, consumer := jsi.stream, jsi.consumer
 	js := jsi.js
 	sub.mu.Unlock()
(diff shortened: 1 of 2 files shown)


js.go
```diff
@@ -2047,7 +2051,26 @@ func (sub *Subscription) resetOrderedConsumer(sseq uint64) {
...
 
-		consName := nuid.Next()
+		// Attempt to delete the existing consumer.
+		// If ErrConsumerNotFound is returned, it means that the consumer was already deleted
+		// by the server, so we can proceed to create a new consumer.
+		err := sub.deleteConsumer()
+		if err != nil && !errors.Is(err, ErrConsumerNotFound) {
+			var apiErr *APIError
+			if errors.Is(err, ErrJetStreamNotEnabled) || errors.Is(err, ErrTimeout) || errors.Is(err, context.DeadlineExceeded) {
+				// if creating consumer failed, retry
```
wallyqs, line 2061: here we abort recreating the ordered consumer if deleting the previous consumer fails? I wonder if we should ignore the delete consumer error but call `pushErr` here and then proceed to recreate the ordered consumer instead as before

derekcollison: The delete call should be async and best effort, I would not even wait for a response.

piotrpio: Good points, done

wallyqs (approved):
LGTM

PR #1457, opened 2023-11-02
Update nkeys dependency to v0.4.6

piotrpio:
Signed-off-by: Piotr Piotrowski <piotr@synadia.com>

PR #1458, opened 2023-11-02
Update nkeys and other deps

wallyqs:
follow up from https://github.com/nats-io/nats.go/pull/1457 to update other deps and cleanup go.sum


piotrpio (approved):
LGTM!

PR #1466, opened 2023-11-10
Add service pkg and deprecate micro pkg

bruth:
Summary of changes:

- New `service` package
- Type aliases and wrapped functions retained in `micro` package including a test to confirm
- Add `service.New` in favor of `service.AddService` to remove naming stutter


bruth:
Test failures so far are unrelated to this change..

PR #1468, opened 2023-11-10
[FIXED] Invalid PullExpiry validation

### LINKED ISSUE #1467: No heartbeat on MessagesContext is not instance of `ErrNoHeartbeat`
### Observed behavior

When calling `Next` on a `jetstream.MessagesContext`, if it returns a `nats: no heartbeat received` it does not return `true` for `errors.Is(err, nats.ErrNoHeartbeat)`.

```go
// .... create consumer
it, err := consumer.Messages(jetstream.PullExpiry(expTime))
if err != nil {
	return err
}
defer it.Stop()

for {
	msg, err := it.Next()
	if err != nil {
                  // This always returns false even if the error matches "nats: no heartbeat received"
		 if errors.Is(err, nats.ErrNoHeartbeat) {
                     fmt.Println("next msg heartbeat error")
                 }
		return err
	}
}

```

### Expected behavior

If the error returned is a heartbeat error it should validate as true when checking if the error is a `nats.ErrNoHeartbeat`

### Server and client version

server: 2.10.3
client: github.com/nats-io/nats.go 1.31.0

### Host environment

**Server**
Managed Kubernetes (AKS) 1.25.11
OS: Ubuntu 22.04
Container runtime: containerd 202307.12.0
Architecture: amd64

**Client**
MacOS 12.5.1
CPU: M1 Pro
Go version: 1.21.3

### Steps to reproduce

- create consumer
- call `consumer.Messages` with `PullExpiry` set to `time.Millisecond`
- call `Next` on the `MessagesContext`
- an error should be returned. check `errors.Is(err, nats.ErrNoHeartbeat)`, it will be false

Using the example in observed behaviour section set `expTime` to `time.Millisecond`

piotrpio:
Fixed invalid option validation where pull expiry could be set to < 1s. This bug was only present in `consumer.Messages()`

Fixes #1467

### DIFF
--- a/jetstream/options.go
+++ b/jetstream/options.go
@@ -122,7 +122,7 @@ type PullExpiry time.Duration
 
 func (exp PullExpiry) configureConsume(opts *consumeOpts) error {
 	expiry := time.Duration(exp)
-	if expiry < 1*time.Second {
+	if expiry < time.Second {
 		return fmt.Errorf("%w: expires value must be at least 1s", ErrInvalidOption)
 	}
 	opts.Expires = expiry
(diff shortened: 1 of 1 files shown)


Jarema (approved):
LGTM

PR #1475, opened 2023-11-20
Update README.md

p4xx07:
This reconnect part does not work if you don't include also the PingInterval

### DIFF
--- a/README.md
+++ b/README.md
@@ -341,6 +341,7 @@ nc.QueueSubscribe("foo", "job_workers", func(_ *Msg) {
 // the connection in reconnecting state if it failed to connect right away.
 nc, err := nats.Connect(nats.DefaultURL,
     nats.RetryOnFailedConnect(true),
+    nats.PingInterval(time.Second),
     nats.MaxReconnects(10),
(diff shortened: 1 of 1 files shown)


Allan-Nava:
+1

README.md
```diff
@@ -341,6 +341,7 @@ nc.QueueSubscribe("foo", "job_workers", func(_ *Msg) {
 // the connection in reconnecting state if it failed to connect right away.
 nc, err := nats.Connect(nats.DefaultURL,
     nats.RetryOnFailedConnect(true),
+    nats.PingInterval(time.Second),
```
wallyqs, line 344: 1 sec for the ping interval is super low, 30s to 2 minutes would be a better duration. Though you should not have to set this so would be good to double check the root cause at what what is happening here

p4xx07: I can bump it up to 30 I agree with you.

piotrpio:
@p4xx07 `PingInterval` has a default value of 2 minutes if you don't provide it, and it should not be necessary to use the option for `RetryOnFailedConnect` to work. Can you maybe create an issue with some more details if you think there is a bug in the library?

PR #1484, opened 2023-12-07
fix: race condition when adding endpoints from goroutine

actatum:
fixes race condition when setting endpoint subscription, appending to list of endpoints and setting endpoint stats.

Description:

When AddEndpoint() calls addEndpoint() the Endpoint is assigned a subscription, the service list of endpoints is appended to, and the Endpoint's stats field is populated. There is no lock around any of this and if the AddEndpoint method is called from another goroutine it can cause a race condition.

This PR includes a test that recreates the race condition and can be verified by removing the lock and unlock calls that have been added to service.go running the test `TestAddEndpoint_RaceCondition` with the `-race` flag on. If you replace the lock/unlock calls the test will pass.

### DIFF
--- a/micro/service.go
+++ b/micro/service.go
@@ -692,6 +692,9 @@ func (s *service) serviceIdentity() ServiceIdentity {
 
 // Info returns information about the service
 func (s *service) Info() Info {
+	s.m.Lock()
+	defer s.m.Unlock()
+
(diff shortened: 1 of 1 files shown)


piotrpio:
@actatum thank you for the PR and sorry for the late review. LGTM!

PR #1490, opened 2023-12-12
[ADDED] ListKeys method for listing kv keys

piotrpio:

### DIFF
--- a/jetstream/kv.go
+++ b/jetstream/kv.go
@@ -67,7 +67,10 @@ type (
 		// WatchAll will invoke the callback for all updates.
 		WatchAll(ctx context.Context, opts ...WatchOpt) (KeyWatcher, error)
 		// Keys will return all keys.
+		// DEPRECATED: Use ListKeys instead to avoid memory issues.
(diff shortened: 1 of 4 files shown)


jetstream/kv.go
```diff
@@ -67,7 +67,10 @@ type (
 		// WatchAll will invoke the callback for all updates.
 		WatchAll(ctx context.Context, opts ...WatchOpt) (KeyWatcher, error)
 		// Keys will return all keys.
+		// DEPRECATED: Use ListKeys instead to avoid memory issues.
```
wallyqs, line 70: which was the memory issue?

derekcollison: Have a use case now with > 600k keys.

wallyqs: ah I see, makes sense to switch to use channels instead

wallyqs (approved):
LGTM

Jarema (approved):
LGTM.
Will add note to ADR that such scalable method should be available in all clients, if any without iterators had similar approach.

PR #1493, opened 2023-12-13
[FIXED] Invalid error returned when object store is not found

### LINKED ISSUE #1491: obj store api returns stream not found instead of bucket not found
### Observed behavior

obj store api returns stream not found instead of bucket not found 

### Expected behavior

bucket not found

### Server and client version

any

### Host environment

_No response_

### Steps to reproduce

_No response_

piotrpio:
Fixes #1491


Jarema (commented):
This is a good change, but still potentially breaking for no big gain, especially considering that the Object Store in the new API retunrs explicit and good error.

piotrpio:
Yeah, agreed. I'll close this PR and only leave the proper error in new API

PR #1497, opened 2023-12-15
[FIXED] Race condition when getting pull subscriptions in ordered consumer

piotrpio:
Fixes concurrent read/write to subscriptions map


Jarema (approved):
LGTM

PR #1511, opened 2024-01-02
[ADDED] Return account reservations on AccountInfo

piotrpio:
This adds reservations to AccountInfo and fixes missing Tiers map in new JetStream API. Reservations on AccountInfo will be available in the next server release.


Jarema (approved):
LGTM

PR #1529, opened 2024-01-19
docs: safe to use with proper config

thorntonmc:
In the examples for establishing secure (TLS) connections to NATS - the README shows using a nats.Secure method - which allows us to pass in a TLS config.

However, the godoc warns against this. @wallyqs informed me that this comment can be removed.


piotrpio (approved):
LGTM! Thank you for the contribution.

PR #1543, opened 2024-02-02
[IMPROVED] Export nested ConsumerManager interface

piotrpio:
This exports `streamConsumerManager` which was used inside `Stream` interface.

This is done so that the interface methods are visible in godoc. Prior to this change, because `streamConsumerManager` is not exported, those methods are available in API but hidden from documentation.


Jarema (approved):
LGTM!

PR #1554, opened 2024-02-12
Add `PausedUntil` to consumer config

neilalexander:
This is to support work on pausing consumers.

Will remain in draft state until we're all happy with the API shape.


piotrpio:
This was added in https://github.com/nats-io/nats.go/pull/1571

PR #1559, opened 2024-02-16
Add Missing Compression Struct Tag

samuelattwood:
Missing struct tag results in `Compression` with a capital C when serialized


piotrpio (approved):
LGTM!

PR #1573, opened 2024-02-28
Adds a keys method to the NATS message header

autodidaddict:
On the surface this is just a convenience function, however, by adding the `Keys` method with this signature, we make the `msg.Header` field a suitable target for injection of open telemetry span context. In other words, after this change, the `Header` field will satisfy the [TextMapCarrier](https://pkg.go.dev/go.opentelemetry.io/otel@v1.24.0/propagation#TextMapCarrier) interface:

```
type TextMapCarrier interface {

	// Get returns the value associated with the passed key.
	Get(key) string
	// Set stores the key-value pair.
	Set(key string, value string)
	// Keys lists the keys stored in this carrier.
	Keys() []string
}
```

This means we'll be able to use NATS message headers to automatically convey OpenTelemetry span context via the `TraceContext`'s `Inject` and `Extract` functions:

```
tctx.Inject(ctx, msg.Header)
...
rehydratedContext := tctx.Extract(ctx, msg.Header)
```

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -3640,6 +3640,17 @@ func (h Header) Get(key string) string {
 	return _EMPTY_
 }
 
+// Returns the list of keys in the message header
+func (h Header) Keys() []string {
+	keys := make([]string, len(h))
+	i := 0
+	for k := range h {
+		keys[i] = k
+		i++
+	}
+	return keys
+}
+
 // Values returns all values associated with the given key.
 // It is case-sensitive.
 func (h Header) Values(key string) []string {
--- a/nats_test.go
+++ b/nats_test.go
@@ -29,6 +29,7 @@ import (
 	"reflect"
 	"regexp"
 	"runtime"
+	"slices"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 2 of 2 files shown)


piotrpio (approved):
LGTM!

wallyqs:
I think you can do instead?

```go
tctx.Inject(ctx, otel.HeaderCarrier(msg.Header))
rehydratedContext := tctx.Extract(ctx, otel.HeaderCarrier(msg.Header))
```

autodidaddict:
that won't work unless `msg.Header` has a `Keys()` function, which is what this PR adds. Am I missing some subtle Go typecasting thing?

wallyqs:
This a small example of how to adapt msg.Header to the otel interface:
```
package main

import (
	"fmt"
	"net/http"
	"github.com/nats-io/nats.go"
	otel "go.opentelemetry.io/otel/propagation"
)

func main() {
	hdr := http.Header{}
	hdr.Set("foo", "1")
	hdr.Set("bar", "2")
	hdr.Set("quux", "2")
	msg := &nats.Msg{}
	msg.Header = nats.Header(hdr)
	nhdr := otel.HeaderCarrier(msg.Header)
	fmt.Println(nhdr.Keys())
	// => [Foo Bar Quux]
}
```

PR #1575, opened 2024-02-29
[IMPROVED] Performance of respToken

### LINKED ISSUE #1560: the cost of respToken in respHandler is much higher than expected
### Observed behavior

In a simple subscription service, about 70k/s qps.
Observed through pprof, the cost of respToken in respHandler is much higher than expected.
I'm not using any wildcard topics, it's all fixed topics.
resptoken

### Expected behavior

less overhead here

### Server and client version

github.com/nats-io/nats-server/v2 v2.8.4
github.com/nats-io/nats.go v1.27.1

This part of the code is the same as the latest version.

### Host environment

Linux  x64 
Centos 7.3

### Steps to reproduce

-

piotrpio:
Using `fmt.Sscanf()` in muxer caused performance overhead for each request sent by the client. It is sufficient to only compare the expected inbox prefix with the response subject and grab the token from the end of the subject (as is done in other clients).

Benchmark results:
1. Using `fmt.Sscanf`: `BenchmarkRespToken-10 2836392 418.7 ns/op 83 B/op 4 allocs/op`
3. Comparing prefixes: `BenchmarkRespToken-10 408571200 2.837 ns/op 0 B/op 0 allocs/op`

Fixes #1560

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -566,7 +566,6 @@ type Conn struct {
 	respSub       string               // The wildcard subject
 	respSubPrefix string               // the wildcard prefix including trailing .
 	respSubLen    int                  // the length of the wildcard prefix excluding trailing .
-	respScanf     string               // The scanf template to extract mux token
 	respMux       *Subscription        // A single response subscription
 	respMap       map[string]chan *Msg // Request map for the response msg channels
 	respRand      *rand.Rand           // Used for generating suffix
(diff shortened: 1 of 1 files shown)


derekcollison (approved):
LGTM

Jarema (approved):
LGTM!

wallyqs (approved):
LGTM 👍

PR #1582, opened 2024-03-12
[IMPROVED] Fetch and FetchBatch for draining and closed subscriptions

piotrpio:

### DIFF
--- a/js.go
+++ b/js.go
@@ -2861,7 +2861,13 @@ func (sub *Subscription) Fetch(batch int, opts ...PullOpt) ([]*Msg, error) {
 	}
 	var hbTimer *time.Timer
 	var hbErr error
-	if err == nil && len(msgs) < batch {
+	sub.mu.Lock()
+	subClosed := sub.closed || sub.draining
+	sub.mu.Unlock()
+	if subClosed {
+		err = errors.Join(ErrBadSubscription, ErrSubscriptionClosed)
+	}
+	if err == nil && len(msgs) < batch && !subClosed {
 		// For batch real size of 1, it does not make sense to set no_wait in
 		// the request.
 		noWait := batch-len(msgs) > 1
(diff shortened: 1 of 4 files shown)


Jarema (approved):
LGTM!

js.go
```diff
@@ -2861,7 +2861,13 @@ func (sub *Subscription) Fetch(batch int, opts ...PullOpt) ([]*Msg, error) {
 	}
 	var hbTimer *time.Timer
 	var hbErr error
-	if err == nil && len(msgs) < batch {
+	sub.mu.Lock()
```
georgian-micsa-form3, line 2864: Maybe extract following logic
```go 
	sub.mu.Lock()
	subClosed := sub.closed || sub.draining
	sub.mu.Unlock()
```

in a new function inside Subscription e.g. `isClosedOrDraining()` or similar

and reuse it in both places: `Fetch` and `FetchBatch`?

test/js_test.go
```diff
@@ -1239,6 +1239,64 @@ func TestPullSubscribeFetchWithHeartbeat(t *testing.T) {
...
+	replySubject := strings.Replace(sub.Subject, "*", "abc", 1)
+	payload := `{"batch":10,"no_wait":true}`
+	if err := nc.PublishRequest(nextSubject, replySubject, []byte(payload)); err != nil {
+		t.Fatalf("Unexpected error: %s", err)
+	}
+	time.Sleep(100 * time.Millisecond)
+
+	// now drain the subscription, messages should be in the buffer
+	sub.Drain()
+	msgs, err := sub.Fetch(100)
```
georgian-micsa-form3, line 1282: Calling `Fetch` immediately after `Drain` returns messages instead of an error. Is this normal behaviour and why?

piotrpio: `Drain` is not blocking, so calling `sub.Drain()` essentially just starts the draining process without waiting for completion. `Fetch()` first checks if there are any messages in the buffer (those messages should be accessible while draining) and returns if there are any - it will never send a new request when draining though.

Only when we're fully drained/closed and there are no more messages waiting in the buffer we're returning an error. This tests simulates this behavior.

js.go
piotrpio: That's such a tiny piece of logic only used in 2 places, I would honestly just leave it inlined for clarity :)

With those kinds of checks I usually wait until I have to repeat it at least 3 times before extracting to a separate method, but that's just preference.

PR #1586, opened 2024-03-19
[IMPROVED] Add test checking if client reconnects after jwt expires

piotrpio:


caleblloyd (approved):
LGTM

PR #1589, opened 2024-03-21
[micro] Add API to read Reply subject from `micro.Request`

### LINKED ISSUE #1579: [micro] Add API to read Reply subject from `micro.Request`
### Proposed change

To not introduce a breaking change we'd need to add a new interface (`RequestV2` ??) that extends the `Request` and add a new method

```go
type RequestV2 interface {
    Request

    Reply() string
}
```
 

### Use case

In some cases you can trigger a microservice but you don't care about the results so maybe you just send a publish instead of a request.

In this case would be nice to know if a request has a reply subject just to avoid to Marshal a response and try to send back it

```go
func MyHandler(r micro.RequestV2) {
    response := handleRequest()

    if r.Reply() == "" {
        return
    }

   err := r.RespondJSON(response)
   if err != nil {
      // do something
   }
} 
```

### Contribution

I'm interested to help with implementation in case this will be accepted

mcosta74:
Allow to check if the request contains a `Reply` subject

Fixes #1579

Signed-off-by: Massimo Costa <costa.massimo@gmail.com>

### DIFF
--- a/micro/request.go
+++ b/micro/request.go
@@ -58,6 +58,9 @@ type (
 
 		// Subject returns underlying NATS message subject.
 		Subject() string
+
+		// Reply returns underlying NATS message reply subject.
+		Reply() string
 	}
 
 	// Headers is a wrapper around [*nats.Header]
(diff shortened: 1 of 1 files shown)


themartorana:
Came here to submit [this same PR](https://github.com/nats-io/nats.go/compare/main...Flyclops:nats.go:microReploy), thanks @mcosta74 for beating me to it. Definitely need this exposed to be able to move from our own proprietary subject router that passes the `nats.Msg` around to NATS `micro`. `.Reply` is the only concrete `nats.Msg` property that hasn't been exposed by the `micro.Request` interface.

piotrpio (approved):
LGTM!

PR #1604, opened 2024-04-05
[IMPROVED] Add note to readme about backwards compatibility

piotrpio:


Jarema (approved):
LGTM!

PR #1616, opened 2024-04-17
[FIXED] KV discard policy in jetstream pkg

### LINKED ISSUE #1615: kv is created with OLD discard policy using jetstream pkg while with nats pkg it gets NEW
### Observed behavior

kv is created with OLD discard policy using jetstream pkg while with nats pkg it gets NEW

### Expected behavior

both pkgs shall have the same resolution

### Server and client version

2.10.9
1.34.1

### Host environment

_No response_

### Steps to reproduce

_No response_

piotrpio:
Fixes  #1615


Jarema (approved):
LGTM

PR #1624, opened 2024-04-27
[ADDED] Force reconnect

piotrpio:
Addresses https://github.com/nats-io/nats-architecture-and-design/issues/259

### DIFF
--- a/test/reconnect_test.go
+++ b/test/reconnect_test.go
@@ -924,7 +924,7 @@ func TestForceReconnect(t *testing.T) {
 	}
 
 	// Force a reconnect
-	err = nc.Reconnect()
+	err = nc.ForceReconnect()
 	if err != nil {
 		t.Fatalf("Unexpected error on reconnect: %v", err)
 	}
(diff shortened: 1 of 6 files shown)


levb:
@piotrpio `Reconnect()` ignores `opts.AllowReconnect`, is this by design? (asking for the C client)

Jarema (approved):
LGTM

test/reconnect_test.go
```diff
@@ -884,21 +884,218 @@ func TestAuthExpiredReconnect(t *testing.T) {
...
-			t.Fatalf("Expected to be in reconnecting state after jwt expires, got %v", s)
+	WaitOnChannel(t, stasusCh, nats.RECONNECTING)
+	WaitOnChannel(t, stasusCh, nats.CONNECTED)
+	nc.Close()
+}
+
+func TestForceReconnect(t *testing.T) {
+	s := RunDefaultServer()
+
+	nc, err := nats.Connect(s.ClientURL(), nats.ReconnectWait(10*time.Second))
```
wallyqs, line 895: Only one behavior that wanted to confirm is whether we get an error on disconnect or not, and it seems the disconnect callback is called but without an error which I think it is good, e.g.

```go
nats.DisconnectErrHandler(func(_ *nats.Conn, err error) {
	t.Logf("ERROR: %v", err) // error is nil here
}),
```

piotrpio: Yes, the callback is invoked without error which I thought is a reasonable behavior.

wallyqs (approved):
LGTM

PR #1630, opened 2024-05-09
[UPDATED] Added json tags to KeyValueConfig struct to behave consistently with the StreamConfig one

pricelessrabbit:
This PR add json tags to the `KeyValueConfig` struct like the ones already present in `StreamConfig`. 
This is useful when working with KV buckets because currently you need a map or support struct to marshal / unmarshal a  bucket config to / from JSON.

- I used the `StreamConfig` tags as reference.
- I replicated the `omitempty` option for the fields that have it in `StreamConfig`, but a check is needed for the other ones.

### DIFF
--- a/jetstream/kv.go
+++ b/jetstream/kv.go
@@ -197,52 +197,52 @@ type (
 		// Bucket is the name of the KeyValue store. Bucket name has to be
 		// unique and can only contain alphanumeric characters, dashes, and
 		// underscores.
-		Bucket string
+		Bucket string `json:"bucket"`
 
 		// Description is an optional description for the KeyValue store.
-		Description string
+		Description string `json:"description,omitempty"`
 
 		// MaxValueSize is the maximum size of a value in bytes. If not
 		// specified, the default is -1 (unlimited).
-		MaxValueSize int32
+		MaxValueSize int32 `json:"max_value_size,omitempty"`
 
 		// History is the number of historical values to keep per key. If not
 		// specified, the default is 1. Max is 64.
-		History uint8
+		History uint8 `json:"history"`
 
 		// TTL is the expiry time for keys. By default, keys do not expire.
-		TTL time.Duration
+		TTL time.Duration `json:"ttl,omitempty"`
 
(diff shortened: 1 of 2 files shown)


jetstream/kv.go
```diff
@@ -197,52 +197,52 @@ type (
...
 
 		// Storage is the type of storage to use for the KeyValue store. If not
 		// specified, the default is FileStorage.
-		Storage StorageType
+		Storage StorageType `json:"storage"`
 
 		// Replicas is the number of replicas to keep for the KeyValue store in
 		// clustered jetstream. Defaults to 1, maximum is 5.
-		Replicas int
+		Replicas int `json:"replicas"`
```
piotrpio, line 226: To be consistent with stream config, this should probably be `num_replicas`

piotrpio (changes requested):
Thanks for the PR! I have one suggestion regarding the tag name.

Other than that, I don't think `omitempty` matters a lot in that case since we never really serialize the config from within the library. If anything, I would probably use `omitempty` to represent optionality of those fields, similarly to how we have it in object store - i.e. only bucket name would not have it.

kv.go
```diff
@@ -249,22 +249,22 @@ func purge() DeleteOpt {
...
-	Replicas     int
-	Placement    *Placement
-	RePublish    *RePublish
-	Mirror       *StreamSource
-	Sources      []*StreamSource
+	Bucket       string          `json:"bucket"`
+	Description  string          `json:"description,omitempty"`
+	MaxValueSize int32           `json:"max_value_size,omitempty"`
+	History      uint8           `json:"history,omitempty"`
+	TTL          time.Duration   `json:"ttl,omitempty,omitempty"`
```
piotrpio, line 256: Sorry about that, I did not notice at first, you have doubled `omitempty` here.

piotrpio (commented):
Looks good, just one more nit I only noticed now.

kv.go
pricelessrabbit: sry, fixed!

piotrpio (approved):
LGTM, thank you for the contribution!

PR #1625, opened 2024-04-30
Added description for the StreamConfig struct in jsm.go

kazmerdome:
Enhanced documentation by adding a description for the StreamConfig struct in jsm.go

PR #1617, opened 2024-04-17
[FIXED] Add discard policy repair logic in CreateKeyValue

piotrpio:
Signed-off-by: Piotr Piotrowski <piotr@synadia.com>

PR #1632, opened 2024-05-17
Release v1.35.0

piotrpio:
### Added
- Service API:
  - Added API to read reply subject from `micro.Request` (#1589)
- Core NATS:
  - Added `ForceReconnect` method on connection for manually triggering reconnect (#1624)

### Fixed
- KeyValue:
  - Fixed invalid `DiscardPolicy` being set when creating a new KeyValue store using the `jetstream` package. Previously created KV instances will be auto-repaired when calling `CreateKeyValue` (#1616, #1617)
- JetStream:
  - Fixed validation for streams, consumers and KV buckets to avoid protocol errors (#1613)
- Core NATS:
  - Fixed `ConnectedCB` not being invoked when using `RetryOnFailedConnect: true` and the initial `nats.Connect` fails (#1619)
 
### Improved
- Core NATS:
  - Remove unnecessary allocations from `readMIMEHeader` (#1609)
- KeyValue:
  - Added json tags to KeyValueConfig struct (#1630)
- Legacy JetStream:
  - Added documentation for `StreamConfig` in jsm.go (#1625)

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ When using or transitioning to Go modules support:
 ```bash
 # Go client latest or explicit version
 go get github.com/nats-io/nats.go/@latest
-go get github.com/nats-io/nats.go/@v1.34.1
+go get github.com/nats-io/nats.go/@v1.35.0
 
 # For latest NATS Server, add /v2 at the end
 go get github.com/nats-io/nats-server/v2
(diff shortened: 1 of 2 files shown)


Jarema (approved):
LGTM

PR #1636, opened 2024-05-23
msgpack encoder

FosenTS:
- added msgpack encoder using github.com/vmihailenco/msgpack/v5 library
- added msgpack encoder tests


ripienaar:
Thanks for the contribution @FosenTS :)

if you are using encoded connections today I would suggest not using them for more things instead write a helper to do that in your own code

FosenTS:
Thanks!

PR #1640, opened 2024-06-05
reproduce panic when creating consumer on wildcard stream

areknoster:
Added test panics on `CreateOrUpdateConsumer`. The issue doesn't happen, when the stream's subject is not wildcard `">"`


areknoster:
This was only for bug report purpose, I'm closing since it's note supposed to be merged in this form.

PR #1645, opened 2024-06-11
[IMPROVED] Ordered consumer creation and initial config settings

piotrpio:
This changes a few things around creating ordered consumers:
- initial ordered consumer creation is now done without retries
- fixed an issue where start seq could be invalid when resetting a consumer which did not receive any messages
- simplified getConsumerConfig()


Jarema (approved):
LGTM,

though would be good to have few runs on client-tests with failground.

piotrpio:
@Jarema While these changes are fine, I did discover a bug when testing with failground. I will create a separate PR for it.

Issue #1545, opened 2023-12-29
Trailing slash on URL leads to connection errors


philpennock:
### Observed behavior

Given a hostname `example.org`, running:
```sh
nats -s tls://example.org server info
```
_succeeds_, but
```sh
nats -s tls://example.org/ server info
```
**fails**; if the hostname points to a single server, I get an error like:
```
nats: error: nats: no servers available for connection
```
but if the hostname points to a cluster, I get an error like:
```
nats: error: dial tcp 192.0.2.1:0: i/o timeout
```

### Expected behavior

Either of two results:

1. The trailing slash on the URL after the hostname is irrelevant, and connection succeeds and is identical
2. The trailing slask on the URL immediately yields a parse error for being an invalid URL format

### Server and client version

```
% nats --version
0.1.1-109-g7226b46
```

Multiple server versions observed with this.

The client version corresponds to a commit made yesterday, 2023-12-28.

PR #1654, opened 2024-06-19
[IMPROVED] Trim trailing slash if set on server address which can cause errors during lookup

josephwoodward:
This change addresses issue https://github.com/nats-io/nats.go/issues/1545, opting for trimming any possible trailing slash as opposed to returning an error to the user.


josephwoodward:
@piotrpio Not a problem!

PR #1657, opened 2024-06-20
Added KeysWithFilter methods to KV interface in kv.go file

Darshan174:


Darshan174:
Closes #1655

Darshan174:
added the proposed changes in nats.go/util/kv.go file

Issue #1656, opened 2024-06-20
panic: interface conversion: jetstream.Consumer is nil, not *jetstream.pullConsumer


zonque:
### Observed behavior

```
panic: interface conversion: jetstream.Consumer is nil, not *jetstream.pullConsumer
goroutine 57 [running]:
github.com/nats-io/nats.go/jetstream.(*orderedConsumer).reset(0x17929c0)
        github.com/nats-io/nats.go@v1.36.0/jetstream/ordered.go:542 +0x328
github.com/nats-io/nats.go/jetstream.(*orderedConsumer).Consume.func2()
        github.com/nats-io/nats.go@v1.36.0/jetstream/ordered.go:150 +0xc0
created by github.com/nats-io/nats.go/jetstream.(*orderedConsumer).Consume in goroutine 54
        github.com/nats-io/nats.go@v1.36.0/jetstream/ordered.go:146 +0x76c
``` 

### Expected behavior

No panic.

### Server and client version

Server version 2.10.16
Client version v1.36.0

### Steps to reproduce

This happens randomly, cannot provide a good reproducer at this point. Will comment on this ticket when I do.

piotrpio:
Hello @zonque, thank you for creating the issue, I'll try to reproduce as well and will come back to you when I have a fix.

piotrpio:
Fixed in #1686

PR #1666, opened 2024-06-25
jetstream: ordered consumer: handle closed subscriptions gracefully

zonque:
In `orderedConsumer.reset()`, `retryWithBackoff()` is returned from with `false, nil` in case the subscription is closed. In this case, `var cons Consumer` is still `nil`, and the code crashes in the attempt to cast it to a `pullConsumer` later on.

Fix this by catching the `nil` case explicitly.

Signed-off-by: Daniel Mack <daniel.mack@holoplot.com>

### DIFF
--- a/jetstream/ordered.go
+++ b/jetstream/ordered.go
@@ -539,6 +539,9 @@ func (c *orderedConsumer) reset() error {
 	if err != nil {
 		return err
 	}
+	if cons == nil {
+		return nats.ErrSubscriptionClosed
+	}
 	c.currentConsumer = cons.(*pullConsumer)
 	return nil
 }


zonque:
Closes #1656

zonque:
This is a regression from 66b92a9b it appears.

zonque:
@piotrpio ping

zonque:
It's tricky to test because the issue is triggered by a race condition between the creation of a consumer and the teardown of a NATS server. Not sure how this could be tested, but this is very clearly a bug, as you can easily identify the code path that leads to a `nil` pointer dereference :shrug:

piotrpio:
Hello @zonque, thanks for finding this and sorry it tool so long to review.

Your solution solves the nil pointer, but I have a few suggestions on how it can be improved:
1. Don't use `nats.ErrSubscriptionClosed` - this is a core NATS error implicating that a NATS subscription was closed - in the context of JetStream consumers it does not make much sense.
2. I don't think just returning an error is enough - the way it works after your change it will trigger an error callback for `Consume()`, which I don't think should be done since the consuming was stopped by the user (either with `Stop()` or `Drain()`). Similarly, for `Messages()`, you'll get an ambiguous `nats.ErrSubscriptionClosed` when calling `Next()`.

My suggestion would be to:
1. introduce a new error variable, e.g. `errOrderedConsumerStopped` (similarly to `errOrderedSequenceMismatch`) which can be internal, we don't have to expose it to the user. We can return this error if `cons == nil`.
2. Handle the error properly in `Consume()` and `Next()`. For `Consume()`, I think we can simply return ([here](https://github.com/nats-io/nats.go/blob/97e6a525e9d861338e89d31e81063a726a718941/jetstream/ordered.go#L150)), for `Next()`, we can return `ErrMsgIteratorClosed` since we're sure there will be no more messages ([here](https://github.com/nats-io/nats.go/blob/97e6a525e9d861338e89d31e81063a726a718941/jetstream/ordered.go#L292) and [here](https://github.com/nats-io/nats.go/blob/97e6a525e9d861338e89d31e81063a726a718941/jetstream/ordered.go#L312))

Again, sorry it took that long. Let me know if you'll be working on that or I should take it off your hands.

zonque:
> Again, sorry it took that long. Let me know if you'll be working on that or I should take it off your hands.

Thanks for the feedback. As you have a better view on the bigger picture, I think it might in fact be easier if you roll your own version of the fix?

piotrpio:
@zonque, sure, I'll take care of this and make sure the fix is included in the next release.

piotrpio:
I created a PR with the fix: https://github.com/nats-io/nats.go/pull/1686

So I'll be closing this one - thanks for your contribution!

PR #1667, opened 2024-06-30
[FIXED] Sync Publish Retry not working

souravagrawal:
This supposed to retry RetryAttempts times if there is "no responder" error and should retry RetryAttempts times before it gives up. So after (RetryAttempts * RetryWait) seconds it should give up if there is "no responder" error on every attempt.


souravagrawal:
This PR fixes https://github.com/nats-io/nats.go/issues/1664

PR #1689, opened 2024-07-31
[FIXED] Change Fetch client timeout to a higher value

piotrpio:


Jarema (commented):
That is an interesting approach, though I wonder if it makes sense - as the actual delay between timeout actually occuring and getting it by the client does not change depending on the timeout duration, just RTT.

Maybe we should agree on the OSS team what offset do we put here.

piotrpio:
Because `Fetch()` and `FetchBatch()` both accept context, it is not feasible to set the internal timeout to a value larger than expiry (which would make things easier). We cannot have a constant value when decrementing, that's why I went with `0.9 * ttl`. This however is problematic with very large user timeouts, when the diff could potentially be minutes. I changed it now to saturate at 5s. For very low timeout values this will still be not very usable, but that's usually not a valid approach anyway.

Jarema (approved):
LGTM

PR #1693, opened 2024-08-05
Fix setting deliver policy in Fetch() for OrderedConsumer

piotrpio:


Jarema (approved):
LGTM

Issue #1678, opened 2024-07-18
`PublishAsync` Retry Mechanism Not Working as Expected


somratdutta:
### Observed behavior

```go
js.PublishAsync("test", []byte("hi"), nats.RetryWait(5*time.Second), nats.RetryAttempts(10))
```

It gives up at first attempt itself after waiting for 5 second.

### Expected behavior

The method should retry publishing the message 10 times, with a 5-second wait between attempts, giving up after 50 seconds if there is an error on each attempt.

### Server and client version

*   NATS Server Version: 2.10.3
*   NATS Client Version: 1.35.0

### Host environment

*   Operating System: Mac OS

### Steps to reproduce

1.  Start a NATS server cluster in Jetstream mode with trace & debug enabled.
2.  Use the following sample Go code to publish a message to a subject that does not exist, with retry options set.

```go

package main

import (
    "github.com/nats-io/nats.go"
    "log"
    "time"
)


func main
()
 {
    nc, err := nats.Connect(nats.DefaultURL)
    if err != nil {
        log.Fatalf("Error connecting to NATS: %v", err)
    }
    defer nc.Close()

    js, err := nc.JetStream()
    if err != nil {
        log.Fatalf("Error getting JetStream context: %v", err)
    }

    subject := "test"
    message := []byte("hi")
    retryWait := 5 * time.Second
    retryAttempts := 10

    log.Printf("Async Publishing to subject %s with %d retry attempts and %v retry wait time", subject, retryAttempts, retryWait)

    // PublishAsync with retry options
    ack, err := js.PublishAsync(subject, message,
        nats.RetryWait(retryWait),
        nats.RetryAttempts(retryAttempts),
    )
    if err != nil {
        log.Fatalf("Async Publish error: %v", err)
    }

    // Simply wait for a long duration to ensure processing
    time.Sleep(time.Duration(retryAttempts)*retryWait + 10*time.Second)
    log.Println("Wait completed")

    if ack.Err() != nil {
        log.Printf("PublishAsync final error: %v", ack.Err())
    } else {
        log.Println("PublishAsync acknowledged successfully")
}
```

**Console Output:**

```
2024/07/18 18:40:33 Async Publishing to subject test with 10 retry attempts and 5s retry wait time
2024/07/18 18:41:33 Wait completed
2024/07/18 18:41:33 PublishAsync final error: 0x14000220000

Process finished with the exit code 0
```

**Server Logs:**

```
.....
[...]

PR #1695, opened 2024-08-08
[IMPROVEMENT] added client retry for jetstream async publish old API

pranavmehta94:
This PR contains following:
1. Added retry logic for jestream async publish using old API
2. retries are controlled by retry wait and retry attempts publish options
3. no responder error is retried
4. pub ack future object is re-used in retries

This PR fixes following issues:

- https://github.com/nats-io/nats.go/issues/1678

### DIFF
--- a/js.go
+++ b/js.go
@@ -633,6 +633,7 @@ type pubAckFuture struct {
 	retries    int
 	maxRetries int
 	retryWait  time.Duration
+	reply      string
 }
 
 func (paf *pubAckFuture) Ok() <-chan *PubAck {
@@ -846,7 +847,6 @@ func (js *js) handleAsyncReply(m *Msg) {
 		cb := js.opts.aecb
 		js.mu.Unlock()
 		if cb != nil {
-			paf.msg.Reply = _EMPTY_
 			cb(paf.js, paf.msg, err)
 		}
 	}
@@ -855,7 +855,6 @@ func (js *js) handleAsyncReply(m *Msg) {
 	if len(m.Data) == 0 && m.Header.Get(statusHdr) == noResponders {
 		if paf.retries < paf.maxRetries {
 			paf.retries++
-			paf.msg.Reply = m.Subject
 			time.AfterFunc(paf.retryWait, func() {
 				js.mu.Lock()
 				paf := js.getPAF(id)
@@ -989,17 +988,18 @@ func (js *js) PublishMsgAsync(m *Msg, opts ...PubOpt) (PubAckFuture, error) {
 		return nil, errors.New("nats: reply subject should be empty")
 	}
 	var id string
+	var reply string
 
 	// register new paf if not retrying
 	if paf == nil {
-		m.Reply = js.newAsyncReply()
+		reply = js.newAsyncReply()
 
-		if m.Reply == _EMPTY_ {
+		if reply == _EMPTY_ {
 			return nil, errors.New("nats: error creating async reply handler")
 		}
 
-		id := m.Reply[js.replyPrefixLen:]
-		paf = &pubAckFuture{msg: m, st: time.Now(), maxRetries: o.rnum, retryWait: o.rwait}
+		id = reply[js.replyPrefixLen:]
+		paf = &pubAckFuture{msg: m, st: time.Now(), maxRetries: o.rnum, retryWait: o.rwait, reply: reply}
 		numPending, maxPending := js.registerPAF(id, paf)
 
 		if maxPending > 0 && numPending > maxPending {
@@ -1011,9 +1011,14 @@ func (js *js) PublishMsgAsync(m *Msg, opts ...PubOpt) (PubAckFuture, error) {
 			}
 		}
 	} else {
-		id = m.Reply[js.replyPrefixLen:]
+		reply = paf.reply
+		id = reply[js.replyPrefixLen:]
 	}
-	if err := js.nc.PublishMsg(m); err != nil {
+	hdr, err := m.headerBytes()
+	if err != nil {
+		return nil, err
+	}
+	if err := js.nc.publish(m.Subject, reply, hdr, m.Data); err != nil {
 		js.clearPAF(id)
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


js.go
```diff
@@ -806,20 +812,23 @@ func (js *js) handleAsyncReply(m *Msg) {
...
-		// Defer here so error is processed and can be checked.
-		defer close(dch)
+
+	closeDch := func() {
+		// Check on anyone one waiting on done status.
+		if js.dch != nil && len(js.pafs) == 0 {
+			dch := js.dch
+			js.dch = nil
+			// Defer here so error is processed and can be checked.
+			defer close(dch)
```
piotrpio, line 830: This defer does not work as expected since it'll be executed at the end of `closeDch`, not the caller. While it would be good to extract this func since it's repeated twice, I do think that we should make sure that `close(dch)` is executed at the very end to first send signals on the `pubAckFuture` and only later to `PublishAsyncComplete()`

js.go
```diff
@@ -894,7 +933,10 @@ func (js *js) PublishAsync(subj string, data []byte, opts ...PubOpt) (PubAckFutu
 const defaultStallWait = 200 * time.Millisecond
 
 func (js *js) PublishMsgAsync(m *Msg, opts ...PubOpt) (PubAckFuture, error) {
-	var o pubOpts
+	var o = pubOpts{
+		rwait: DefaultPubRetryWait,
```
piotrpio, line 937: Since this is an old API many people rely on, I would be hesitant to make the retries default, but rather treat them as opt-in and clearly state it in the doc. I know that this makes it work differently than the new API, but I don't think we should be changing the default behavior in this case.

js.go
```diff
@@ -806,20 +812,23 @@ func (js *js) handleAsyncReply(m *Msg) {
...
+		// Check on anyone one waiting on done status.
+		if js.dch != nil && len(js.pafs) == 0 {
+			dch := js.dch
+			js.dch = nil
+			// Defer here so error is processed and can be checked.
+			defer close(dch)
+		}
 	}
 
 	doErr := func(err error) {
```
piotrpio, line 834: We should clear the reply subject before passing the message in error handler. If user wants to retry manually inside the error handler, having `paf.msg.Reply` set will cause the retry attempt to fail with `nats: reply subject should be empty`

object.go
```diff
@@ -398,7 +398,7 @@ func (obs *obs) Put(meta *ObjectMeta, r io.Reader, opts ...ObjectOpt) (*ObjectIn
 		return nil
 	}
 
-	m, h := NewMsg(chunkSubj), sha256.New()
```
piotrpio, line 401: Why is this change necessary?

test/js_test.go
piotrpio, line 1: We should really have tests for async retries. For reference, see tests in the new API: https://github.com/nats-io/nats.go/blob/main/jetstream/test/publish_test.go#L1402

piotrpio (changes requested):
Thanks for the contribution! I have a few comments.

object.go
pranav-mehta: Earlier, reply field in msg was set to empty string on function exit using defer. After adding retry logic,  go test detected race between setting reply in async handler and emptying it in publish async msg. So, the defer call which emptied the reply inside msg was removed. 
The code was re-using the msg object inside the loop which called publish async msg and the test failed complaining about nats reply being not empty. Hence, the new msg object is being created in every iteration of the loop now

pranav-mehta:
@piotrpio I incorporated the review comments but the CI pipeline now fails in an unrelated test. When I run the tests locally, all the test succeedd

object.go
pranavmehta94: Earlier, reply field in msg was set to empty string on `PublishMsgAsync` exit using defer. After adding retry logic, go test detected race between setting reply in async handler `handleAsyncReply` and emptying it in `PublishMsgAsync`. So, the defer call which emptied the reply inside PublishMsgAsync was removed. The code was re-using the msg object inside the loop which called `PublishMsgAsync` and the test failed complaining about nats reply being not empty. Hence, the new msg object is being created in every iteration of the loop now

pranavmehta94:
[@piotrpio](https://github.com/piotrpio) I incorporated the review comments but the CI pipeline now fails in an unrelated test. When I run the tests locally, all the tests succeed

object.go
piotrpio: This is no good I think. We should not leave `Msg.Reply` set with the generated subject as it makes it impossible to re-publish by the user (as in the object store now). By the way, having `Put()` in object store recreate the message with each iteration means we have increased allocations (especially for large payloads / small chunks).

I may have an idea of how to tackle this problem - we could stop modifying the message passed by the user (which would prevent race) and utilize the low level `nc.publish()`. Here's my initial attempt from your branch: https://gist.github.com/piotrpio/796e5203e2d5be2cabda771e9ec51d54

Possibly this can be improved, but it seems to work.

piotrpio: Ok but even so, with your changes the message will never be restored to its original state right? Even after the message has been processed. If we need to fix it anyway we can also improve by not modifying the message at all I think.

pranavmehta94: I understand that a caller would assume that message would not be modified in anyway after  `PublishMsgAsync ` is called and `PubAckFuture ` is processed. This is the reason why Reply was not unset earlier. I will update the PR by applying your patch

piotrpio: But it was unset before your changes: https://github.com/nats-io/nats.go/pull/1695/files#diff-176a2342c55218b70f165bed6a7731dc83d901f9513382f3c98bbd3d08ce0678L941

After `PubAckFuture` is processed, the message can be re-used by the user and thus should be in its original state:
`/ The message should not be changed *until* the PubAckFuture has been processed.`

Again, your changes are good in general and we should definitely process those but we also need to be very careful when changing any behavior in the old JetStream API.

pranavmehta94: Yes I understand it was unset before these changes as well

pranavmehta94: @piotrpio I have updated the PR by applying your patch with one small patch. Please review the changes

piotrpio (approved):
LGTM, thanks for the contribution!

PR #1705, opened 2024-08-27
[FIXED] Invalid fetch sequence in ordered consumer Fetch and Next after timeout

piotrpio:


Jarema (approved):
LGTM!

PR #1711, opened 2024-09-04
feat(kv): implement KeysWithFilters and ListKeysWithFilters with tests

somratdutta:
- Added `KeysWithFilters` to filter and return a list of matching keys.
- Added `ListKeysWithFilters` to filter keys and return them as an iterable lister.
- Included test cases for both methods to validate filtering logic and key retrieval.

Resolves: #1655


piotrpio (changes requested):
Hello @somratdutta, thank you for the contribution and sorry it took so long to review this. There are a few issues with this PR and it'll be easier to list them in a single comment rather than commenting on individual lines:

1. The filtering should not be done client-side, for a few reasons. Firstly, it is not efficient as you're fetching all keys from the server, even if you're interested in 1% of them based on the provided filter. Secondly, the filter you apply will not work with wildcards, which users should be able to provide. For filtering, we should leverage the `FilterSubjects` field on `ConsumerConfig` and let the server do the work and simply return results to the user.
That presents another issue unfortunately. While you can add a single filter to `Watch()` and use it to filter the keys (instead of `WatchAll()`), there currently is no version of `Watch()` which accepts multiple filters. 

2. We agreed internally that we don't really want to bloat the API with too many new methods, so we settled on adding a single one: `ListKeysFiltered(ctx context.Context, filters ...string) (KeyLister, error)`. This will allow us to add 0 or more keys easily and iterate over results in efficient way (not having to allocate memory for all keys at once if there are many of them). Also in this new method `WatchOpt` can be dropped as it does not really contribute to the result.

I understand that this was not properly outlined in the related issue, apologies for that. Here's my proposal - I'll work on adding a watcher which can accept multiple filters (it should not take longer than 1-2 days I hope) and after it's done you'll be able to rework this PR to use this new API - obviously if you're still willing to.

somratdutta:
Thanks, @piotrpio for the detailed feedback on the PR.
I now get the approach and it makes a lot of sense.
I would love to still work on it as I have previously worked on many NATS-related PRs.
FYI : I am currently working on adding Jetstream support to Nats-Flink connector 😄

piotrpio:
Hello @somratdutta! We've merged the PR for `WatchFiltered` (https://github.com/nats-io/nats.go/pull/1739) so you can continue with this PR now :)

piotrpio:
@somratdutta could you let us know if you'll be working on that? If you don't have time for that, I can continue your work :)

somratdutta:
@piotrpio Thanks for the ping, I was busy with the `connector` work and the festive seasons.shenanigans :)
I will work on it tomorrow.

somratdutta:
@piotrpio I have added the changes, let me know if this is fine.
Thanks.

jetstream/kv.go
```diff
@@ -174,10 +174,17 @@ type (
 		// Deprecated: Use ListKeys instead to avoid memory issues.
 		Keys(ctx context.Context, opts ...WatchOpt) ([]string, error)
 
+		// KeysWithFilters returns a filtered list of keys in the bucket.
+		// It returns a complete slice of matching keys.
+		KeysWithFilters(ctx context.Context, kv KeyValue, filters ...string) ([]string, error)
```
piotrpio, line 179: As per my comment from earlier, we decided to stick with having a single method (at least for now) returning a `KeyLister`, so this one should be removed.

jetstream/kv.go
```diff
@@ -174,10 +174,17 @@ type (
...
+
 		// ListKeys will return KeyLister, allowing to retrieve all keys from
 		// the key value store in a streaming fashion (on a channel).
 		ListKeys(ctx context.Context, opts ...WatchOpt) (KeyLister, error)
 
+		// ListKeysWithFilters returns a KeyLister for filtered keys in the bucket.
+		ListKeysWithFilters(ctx context.Context, kv KeyValue, filters ...string) (<-chan string, error)
```
piotrpio, line 186: This method should return `KeyLister` as noted in the previous comment:
```go
ListKeysFiltered(ctx context.Context, filters ...string) (KeyLister, error)
```

Also, the `kv KeyValue` parameter is not necessary.

jetstream/kv.go
```diff
@@ -1271,6 +1278,55 @@ func (kv *kvs) ListKeys(ctx context.Context, opts ...WatchOpt) (KeyLister, error
...
+			break
+		}
+		keys = append(keys, entry.Key())
+	}
+	return keys, nil
+}
+
+// ListKeysWithFilters returns a channel of keys matching the provided filters using WatchFiltered.
+func (kv *kvs) ListKeysWithFilters(ctx context.Context, _ KeyValue, filters ...string) (<-chan string, error) {
+	if len(filters) == 0 {
```
piotrpio, line 1305: This check is not needed, it should be possible to pass no filters to just get all keys. it will essentially be a way more versatile version of current `ListKeys()`.

jetstream/kv.go
```diff
@@ -1271,6 +1278,55 @@ func (kv *kvs) ListKeys(ctx context.Context, opts ...WatchOpt) (KeyLister, error
...
+		return nil, errors.New("filters cannot be empty")
+	}
+
+	watcher, err := kv.WatchFiltered(ctx, filters, nil)
```
piotrpio, line 1309: No need to pass `nil` for options as this is a variadic arg, you can just skip it.

jetstream/kv.go
```diff
@@ -1271,6 +1278,55 @@ func (kv *kvs) ListKeys(ctx context.Context, opts ...WatchOpt) (KeyLister, error
 	return kl, nil
 }
 
+// KeysWithFilters retrieves keys that match the provided filters using WatchFiltered.
+func (kv *kvs) KeysWithFilters(ctx context.Context, _ KeyValue, filters ...string) ([]string, error) {
```
piotrpio, line 1282: This can be removed.

jetstream/kv.go
```diff
@@ -1271,6 +1278,55 @@ func (kv *kvs) ListKeys(ctx context.Context, opts ...WatchOpt) (KeyLister, error
...
+	if len(filters) == 0 {
+		return nil, errors.New("filters cannot be empty")
+	}
+
+	watcher, err := kv.WatchFiltered(ctx, filters, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	keysCh := make(chan string)
```
piotrpio, line 1314: You should initialize and return `keyLister`

piotrpio (commented):
In general it looks fine, but I have a few comments.

somratdutta:
Hi @piotrpio.
Have worked on your feedback while back, can you verify once. 😄

piotrpio (approved):
LGTM!

PR #1719, opened 2024-09-23
[FIXED] Publish async not closing done and stall channels after failed retries

piotrpio:
This fixes an issue in new JetStream API where if maximum number of retries
in `PublishMsgAsync` are reached and the publish failed, done and stall channels
are not closed.
Additionally, this fixes a potential race issue with modifying user msg when publishing.


Jarema (approved):
LGTM

PR #1737, opened 2024-11-14
[IMPROVED] Add extra option to OrderedConsumerConfig

mwain:
Id like to set these options for ordered consumers.

### DIFF
--- a/jetstream/consumer_config.go
+++ b/jetstream/consumer_config.go
@@ -263,6 +263,11 @@ type (
 		// Maximum number of attempts for the consumer to be recreated in a
 		// single recreation cycle. Defaults to unlimited.
 		MaxResetAttempts int
+
+		// Metadata is a set of application-defined key-value pairs for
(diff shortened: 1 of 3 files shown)


piotrpio:
Hello @mwain, thank you for the contribution! While I agree that metadata may be useful for ordered consumers, I'm not sure about  `MaxRequestMaxBytes` - max bytes buffered client side can be set using options on `Consume` and `Messages` - what's your use case here?

mwain:
Hi @piotrpio, thanks for the review. We used it to cap the bytes being buffered. TBH i didn't see the `PullMaxBytes` option 🙈 

Would you like me to update this to remove `MaxRequestMaxBytes`?

piotrpio:
@mwain That would be great, thanks!

mwain:
@piotrpio Updated

piotrpio (approved):
LGTM, Thank you for the contribution!

PR #1743, opened 2024-12-02
[FIXED] Race in MessageBatch

### LINKED ISSUE #1733: Race in reading error from message batch
### Observed behavior

type messageBatch has a race where the user can just check the error at any time, but there is no mutex on the read. Meanwhile any error like context ending will cause an error to be written.

### Expected behavior

no race

### Server and client version

client 1.37.0

### Host environment

_No response_

### Steps to reproduce

kill context and read error from message batch

piotrpio:
This fixes a race in both new and legacy JetStream API, where for Fetch APIs returning `MessageBatch` there was a race condition when accessing the error.

Fixes #1733


wallyqs (approved):
LGTM

Jarema (approved):
LGTM

PR #1745, opened 2024-12-03
Update installation commands

alexbozhenko:
1. The info about pre go modules is not relevant anymore, so drop it.
2. Extra slashes are not needed before the version suffix, per https://go.dev/ref/mod#go-install. So make it consistent.

Also, not sure if the command to get nats server is needed here at all?

Signed-off-by: Alex Bozhenko <alex@synadia.com>


wallyqs (approved):
LGTM

piotrpio (approved):
LGTM!

PR #1746, opened 2024-12-04
Update jsm.go

SinTod:
fix arm32 device NumDeleted value overflow

### DIFF
--- a/jsm.go
+++ b/jsm.go
@@ -1073,7 +1073,7 @@ type StreamState struct {
 	LastTime    time.Time         `json:"last_ts"`
 	Consumers   int               `json:"consumer_count"`
 	Deleted     []uint64          `json:"deleted"`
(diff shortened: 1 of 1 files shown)


piotrpio:
Hello @SinTod, thank you for the contribution. You're correct about this being a bug, but I'm not sure we can merge this PR as this would be a breaking change. Also, this is the same struct as in [nats-server](https://github.com/nats-io/nats-server/blob/main/server/store.go#L149), so if we were to introduce this breaking change at some point, we would need to do this in the server as well (we cannot have `int` on the server and `uint` in the client, what if the server starts sending e.g. `-1` at some point - for whatever reason).

SinTod:
hi @piotrpio  If i change NumDeleted to int64？

SinTod:
@piotrpio btw, can I  erase this value from nats  server ?

PR #1762, opened 2025-01-07
Correct Error Returned from DeleteObjectStore

samuelattwood:
Currently `DeleteObjectStore` incorrectly returns `ErrStreamNotFound` if the Object Store Bucket is not found


piotrpio (approved):
LGTM!

PR #1763, opened 2025-01-07
[CHANGED] Remove toolchain version from go.mod

piotrpio:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,8 +2,6 @@ module github.com/nats-io/nats.go
 
 go 1.20
 
-toolchain go1.23.2
-
 require (
 	github.com/klauspost/compress v1.17.9
 	github.com/nats-io/nkeys v0.4.9


Jarema (approved):
LGTM

go.mod
```diff
@@ -2,8 +2,6 @@ module github.com/nats-io/nats.go
 
 go 1.20
```
wallyqs, line 3: need to change this one to 1.20.0 then

piotrpio: Could you explain why is this the case?

ripienaar: The change to the first release of a go release branch being `1.x.0` and not `1.x` only happened in `1.21.0` so I think if we are compatible with `1.20` series then `1.20` is the correct version as `1.20.0` didnt exist?

wallyqs: good point... maybe for someone using Go 1.20 that is the case, but if it changes to `go 1.21` or later then it has to be `go 1.21.0`, @piotrpio full background is here: https://github.com/golang/go/issues/62278#issuecomment-1693538776
https://github.com/golang/go/issues/62278#issuecomment-1693538776

ripienaar: Yes aware for newer releases if needs the .0, but not for 1.20

piotrpio: @wallyqs thanks - so it looks good for now, after updating to newer version in `go.mod` we'll have to add patch version as well.

wallyqs: quite a bit of repos doing 1.20.0 so not 100% sure now, but we can wait and see: https://github.com/search?q=%22go+1.20.0%22+language%3A%22Go+Module%22&type=code&l=Go+Module

alexbozhenko: readme.md says:
> Additionally, this library always supports at least 2 latest minor Go versions. For example, if the latest Go version is 1.22, the library will support Go 1.21 and 1.22.

So should it be bumped to `1.22.0`?

piotrpio: Well, it says at least 2 minor versions, and we definitely could do that, we just did not need to. When go 1.24 comes out, we definitely want to bump the version in go.mod to 1.23 to stop hiding useful features (iterators) behind build tags.

That said, I think we could still bump to 1.22 for the next client release.

PR #1774, opened 2025-01-13
Switch to glangci lint

alexbozhenko:
Just checking to see if this will work...

The server does not install separate configs, but lets golangci manage all of it[1]

Since golangci config for jetstream package is a superset of linters that we individually run on the repo root, we can have two separate configs, one basic for the repo root, and one extended just for jetstream.

So it is only one tool to install and version, and all the linters and formatters can be configured in a single file...
https://golangci-lint.run/usage/linters/#govet
https://golangci-lint.run/usage/linters/#misspell
https://golangci-lint.run/usage/linters/#gofmt
https://golangci-lint.run/usage/linters/#staticcheck

[1]
https://github.com/nats-io/nats-server/blob/e485bfa816d7c62b687df23447ac28784f5b4f38/.golangci.yml#L17-L29
https://github.com/nats-io/nats-server/blob/e485bfa816d7c62b687df23447ac28784f5b4f38/.github/workflows/tests.yaml#L25-L31


.github/workflows/ci.yaml
```diff
@@ -30,11 +30,8 @@ jobs:
             - name: Run linters
               shell: bash --noprofile --norc -x -eo pipefail {0}
               run: |
-                $(exit $(go fmt -modfile=go_test.mod  ./... | wc -l))
-                go vet -modfile=go_test.mod ./... 
-                GOFLAGS="-mod=mod -modfile=go_test.mod" staticcheck ./...
-                find . -type f -name "*.go" | xargs misspell -error -locale US
-                golangci-lint run --timeout 5m0s ./jetstream/...
+                GOFLAGS="-mod=mod -modfile=go_test.mod" golangci-lint run -c .golangci.yaml --timeout 5m0s ./...
```
neilalexander, line 33: Probably better to use `golangci-lint-action` instead of shell actions, so that PRs are properly annotated with lint failures.

PR #1786, opened 2025-01-23
[FIXED] Invalid heartbeat timer for Consumer.Messages

piotrpio:
This fixes an issue where for `Consumer.Messages` the heartbeat error would be returned after 4*idle heartbeat (instead of 2*)


Jarema (approved):
LGTM

PR #1787, opened 2025-01-23
Remove obsolete build tags

alexbozhenko:
similar to https://github.com/nats-io/nats-server/pull/6402

Run:
`# go fix ./...`

It removed obsolete `// +build lines`: https://tip.golang.org/doc/go1.18#go-build-lines
Current build tag syntax: https://pkg.go.dev/cmd/go#hdr-Build_constraints

Since new fixes are rarely introduced, we shouldn't bother with running this automatically:
https://go.dev/blog/introducing-gofix
https://pkg.go.dev/cmd/fix@go1.23.5

```
go tool fix -h
Available rewrites are:

buildtag
        Remove +build comments from modules using Go 1.18 or later

cftype
        Fixes initializers and casts of C.*Ref and JNI types

context
        Change imports of golang.org/x/net/context to context

egl
        Fixes initializers of EGLDisplay

eglconf
        Fixes initializers of EGLConfig

gotypes
        Change imports of golang.org/x/tools/go/{exact,types} to go/{constant,types}

jni
        Fixes initializers of JNI's jobject and subtypes

netipv6zone
        Adapt element key to IPAddr, UDPAddr or TCPAddr composite literals.

        https://codereview.appspot.com/6849045/

printerconfig
        Add element keys to Config composite literals.
```

Signed-off-by: Alex Bozhenko <alex@synadia.com>


test/context_test.go
```diff
@@ -12,7 +12,6 @@
 // limitations under the License.
 
 //go:build go1.7
-// +build go1.7
```
piotrpio, line 15: We can remove this build tag as it's obsolete now.

util/tls.go
```diff
@@ -12,7 +12,6 @@
 // limitations under the License.
 
 //go:build go1.8
-// +build go1.8
```
piotrpio, line 15: This one can be removed too

util/tls_go17.go
```diff
@@ -12,7 +12,6 @@
 // limitations under the License.
 
 //go:build go1.7 && !go1.8
-// +build go1.7,!go1.8
```
piotrpio, line 15: And here I think we can safely remove the whole file, it was used only when building with go 1.7.

piotrpio (commented):
Looks good, just few possible improvements while we're on this topic.

util/tls_go17.go
alexbozhenko: Thanks, done!

piotrpio (approved):
LGTM!

PR #1788, opened 2025-01-23
Split dev dependencies into separate modules

piotrpio:


go.mod
```diff
@@ -3,10 +3,11 @@ module github.com/nats-io/nats.go
 go 1.22.0
 
 require (
-	github.com/klauspost/compress v1.17.9
+	github.com/klauspost/compress v1.17.11
 	github.com/nats-io/nkeys v0.4.9
 	github.com/nats-io/nuid v1.0.1
 	golang.org/x/text v0.21.0
+	google.golang.org/protobuf v1.36.2
```
wallyqs, line 10: would this one be needed if we put a `go.mod` under `encoders`?

piotrpio: Not really under encoders - `enc.go` imports `encoders/builtin`, so even if we put it in a separate module, it would have to be added to root go.mod (with protobuf as an indirect dependency), so does not make sense.

We could in theory make `encoders/protobuf` a separate module but then we'd have to release and version it separately, not sure it's a good solution.

wallyqs: ah I see, thanks for checking

eikemeier:
Why is this a good idea? Who else does it this way?

alexbozhenko:
I guess that's because we do not want circular dependency between server and nats.go.
And nats.go import server for tests
https://go.dev/wiki/Modules#why-does-go-mod-tidy-record-indirect-and-test-dependencies-in-my-gomod

Issue #1783, opened 2025-01-21
Parsing errors leading to disconnections when WS compression is enabled


rsafonseca:
### Observed behavior

After enabling the use of websocket connections with compression enabled, we have some client disconnections happening every few minutes. There's millions of messages of varying sizes and payloads passing by without issues, but once a client disconnects it causes a number of issues on our side, so this very problematic.
The issues do not occur when compression is disabled on ws connection

nats: Parse Error [26]: 'Regi'
nats: Parse Error [0]: '\"Frame'
nats: Parse Error [0]: ':{\"Ena'

### Expected behavior

No disconnections due to parsing errors

### Server and client version

server 2.10.24
client 1.38.0

### Host environment

Running on kubernetes, x86_64 arch nodes, using containerd runtime, using official nats helm chart

### Steps to reproduce

- Run a nats server with enabled websockets and compression
- Run a nats.go client which connects to server with websockets, using compression and a high ping rate for easier reproduction

For easy 100% reproducibility, I've used the nats-bench example, here's some simple steps to reproduce:
- Add the contents of this commit to existing nats repo https://github.com/rsafonseca/nats.go/commit/4994597f55fbc369aedb75bc1a021fbd8b24bd29
- cd to examples/nats-bench
- Run local nats container with the following command to mount the required websocket config:`docker run -it --mount type=bind,src=$(pwd)/nats.conf,dst=/etc/nats/nats-server.conf -p 8080:8080 nats -c /etc/nats/nats-server.conf `
- Compile the example nats-bench by running `go build`
- Run the following example command: `./nats-bench -s ws://localhost:8080 -n 1000000 -np 2 -ns 2 foo`
- lo and behold, a myriad of failures, parsing errors and disconnects

Disabling compression or increasing the ping interval makes for successful completion of the bench

wallyqs:
thanks for the report, do you modify the default `max_payload` setting by any chance?

rsafonseca:
nope, it's using the default value, the messages are varying sizes but not over 2Mb :)

wallyqs:
ok then the default would be using the 1Mb max payload

rsafonseca:
Yes, sorry 1Mb, idk why i had the 2Mb in my mind :)

rsafonseca:
I haven't dug too deep yet, but it may be possible due to some partial flush? https://github.com/klauspost/compress/discussions/993
support for this was added in the patch version after the one nats.go is using

[...]

PR #1790, opened 2025-01-29
fix: protocol parsing errors with ws compression and pongs

rsafonseca:
Fixes: https://github.com/nats-io/nats.go/issues/1783

This PR addresses a problem that happens when using WS compression and receiving PONGs from the server along with other compressed messages.

The byte slices added to websocketReader.pending are memory references to the original buffer, so when the buffer is re-written in websocketReader.drainPending([]byte) the value for the pending PONG message takes a value that is part of the re-written bytes in the buffer.

This only happens with compression, because the size of the messages in the buffer changes, and only happens when receiving PONG messages, because the value for the regular messages is a new []byte generated after decompression, and not a memory reference to a sub-slice of the original buffer

### DIFF
--- a/ws.go
+++ b/ws.go
@@ -76,14 +76,15 @@ var wsGUID = []byte("258EAFA5-E914-47DA-95CA-C5AB0DC85B11")
 var compressFinalBlock = []byte{0x00, 0x00, 0xff, 0xff, 0x01, 0x00, 0x00, 0xff, 0xff}
 
 type websocketReader struct {
-	r       io.Reader
-	pending [][]byte
-	ib      []byte
-	ff      bool
-	fc      bool
-	nl      bool
-	dc      *wsDecompressor
-	nc      *Conn
+	r        io.Reader
+	pending  [][]byte
+	compress bool
+	ib       []byte
+	ff       bool
+	fc       bool
+	nl       bool
+	dc       *wsDecompressor
+	nc       *Conn
 }
 
 type wsDecompressor struct {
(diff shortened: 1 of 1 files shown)


piotrpio:
Amazing find @rsafonseca, thank you. My only concern is the added allocations - maybe we could only copy the slice if it's necessary? What I mean is - if we already create a new byte slice when decompressing, this one does not have to be cloned again when appending to `r.pending`. As you said, we're only really concerned with frames coming from the server which are not compressed (when using compression). WDYT?

rsafonseca:
Yeah, i considered that, but when i ran the benchmark i didn't see any noticeable difference and just wanted to keep it simple for getting the issue fixed.
Right now there's no flag in the websocketReader to assert whether compression was negotiated with the server or not, only on the websocketWriter, the reader only checks this on a per-message basis, though that could be added for the purpose of this check, or we could find another way to implement it if you feel strongly about it

rsafonseca:
I've pushed a simple change for this, WDYT @piotrpio ?

piotrpio (approved):
Looks nice and simple enough, I see no reason not to avoid those allocations 🙂 

LGTM, but it would be good if @wallyqs could take a look as well.

Thanks for the contribution!

wallyqs (approved):
LGTM

rsafonseca:
qq, can we get this backported to 2.10.x and when do you expect a new tag version that includes it? :)

piotrpio:
@rsafonseca most probably second week of February

PR #1794, opened 2025-02-04
[IMPROVED] Use faster gha runners and fix leaking goroutines in Service API

piotrpio:
- Switched to using `ubuntu-latest-8-cores` in CI as it seems to improve on the `goleak` flakes previously happening only in CI
- Added missing goleak execution in Service API
- Fixed leaking goroutine issues in Service API


aricart (approved):
LGTM

PR #1797, opened 2025-02-05
[ADDED] Support for disabling queue groups at service, group, and endpoint levels

piotrpio:
Signed-off-by: Piotr Piotrowski piotr@synadia.com


Jarema (approved):
LGTM!

Issue #1802, opened 2025-02-14
kvs.ListKeysFiltered returns deleted keys


kruegernet:
### Observed behavior

I create a kv bucket with RF=3 and storage=file
I create a couple of entries, delete them, minutes pass.
Using ListKeysFiltered function on the bucket from application code returns keys of the long (in computer time) deleted keys. This is unexpected.

### Expected behavior

Users probably expect ListKeysFiltered will not return deleted keys. Alternatively ListKeysFiltered could take WatchOpt

### Server and client version

server 2.10.25, client 1.39.0

### Host environment

nats docker image on macos, on applie silicon

### Steps to reproduce

Create single node or 3-node cluster
Create a kv bucket with RF=1 or RF=3 and storage=file
Create a couple of entries, delete them, wait a reasonable amount of time (seconds)
Use ListKeysFiltered with empty filter list, or wildcard or globbing filter list.
Receive the keys of recently deleted values from KV

kruegernet:
PR
https://github.com/nats-io/nats.go/issues/1802

piotrpio:
Hello @kruegernet. Thank you for contribution, that was an oversight on our part! I reviewed the PR and have one additional ask.

PR #1803, opened 2025-02-14
Prevents KV ListKeysFiltered from returning deleted keys

kruegernet:
https://github.com/nats-io/nats.go/issues/1802

### DIFF
--- a/jetstream/kv.go
+++ b/jetstream/kv.go
@@ -1276,7 +1276,7 @@ func (kv *kvs) ListKeys(ctx context.Context, opts ...WatchOpt) (KeyLister, error
 
 // ListKeysWithFilters returns a channel of keys matching the provided filters using WatchFiltered.
 func (kv *kvs) ListKeysFiltered(ctx context.Context, filters ...string) (KeyLister, error) {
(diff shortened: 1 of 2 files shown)


jetstream/kv.go
```diff
@@ -1276,7 +1276,7 @@ func (kv *kvs) ListKeys(ctx context.Context, opts ...WatchOpt) (KeyLister, error
 
 // ListKeysWithFilters returns a channel of keys matching the provided filters using WatchFiltered.
 func (kv *kvs) ListKeysFiltered(ctx context.Context, filters ...string) (KeyLister, error) {
-	watcher, err := kv.WatchFiltered(ctx, filters)
+	watcher, err := kv.WatchFiltered(ctx, filters, IgnoreDeletes())
```
piotrpio, line 1279: While we're at it, could you please also add `MetaOnly()` option here? It'll add a `HeadersOnly` option which makes the operation more lightweight and is sufficient for listing keys. Thanks.

kruegernet: You bet, done.

piotrpio (approved):
LGTM!

PR #1804, opened 2025-02-15
[ADDED] ReconnectErrCB + handler function

sschleemilch:
Hi,

I am often using those connect options due to an offline first approach:

```go
opts := []nats.Option{
    nats.MaxReconnects(-1),
}
```
I noticed that **none** of the current handlers covers handling connection attempt errors when there never was an initial connection with those options.
This is probably a very narrow use case.

That is why I thought adding `ReconnectErrCB` and the corresponding `ReconnectErrHandler` makes sense.

Since this is my first contribution, please let me know whether I missed something or whether it should be solved
in a complete different manner. I am willing to put in more work if necessary of course.

Sorry for having a bit of additional diffs due to using `gopls` formatting. I did not find any formatting rules.
If there are any, I am happy to reformat accordingly.

Regards

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -274,7 +274,6 @@ type InProcessConnProvider interface {
 
 // Options can be used to create a customized connection.
 type Options struct {
-
 	// Url represents a single NATS server url to which the client
 	// will be connecting. If the Servers option is also set, it
 	// then becomes the first server in the Servers array.
@@ -421,6 +420,10 @@ type Options struct {
 
 	// AsyncErrorCB sets the async error handler (e.g. slow consumer errors)
 	AsyncErrorCB ErrHandler
+
+	// ReconnectErrCB sets the callback that is invoked whenever a
+	// reconnect attempt failed
+	ReconnectErrCB ConnErrHandler
 
 	// ReconnectBufSize is the size of the backing bufio during reconnect.
 	// Once this has been exhausted publish operations will return an error.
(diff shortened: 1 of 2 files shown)


sschleemilch:
Hi, do I need to do something to make that change happen? :)

piotrpio (commented):
Sorry it took so long to review. This looks great and I think adds a lot of value. The only thing I would add is one more test scenario when you're not using `RetryOnFailedConnect` but get disconnected and check the callbacks order:. I would imagine something like this:

```go

	t.Run("stop the server after initial connect", func(t *testing.T) {
		opts := test.DefaultTestOptions
		s := RunServerWithOptions(&opts)

		reconnectErr := make(chan bool)
		connected := make(chan bool)
		disconnected := make(chan bool, 1)

		nc, err := nats.Connect(nats.DefaultURL,
			nats.ReconnectErrHandler(handler(reconnectErr)),
			nats.MaxReconnects(-1),
			nats.ReconnectWait(100*time.Millisecond),
			nats.DisconnectErrHandler(func(_ *nats.Conn, _ error) {
				disconnected <- true
			}),
			nats.ReconnectHandler(func(_ *nats.Conn) {
				connected <- true
			}))
		if err != nil {
		}
		defer nc.Close()
		s.Shutdown()
		if err = Wait(disconnected); err != nil {
			t.Fatal("Timeout waiting for disconnect handler")
		}
		if err = Wait(reconnectErr); err != nil {
			t.Fatal("Timeout waiting for reconnect error handler")
		}
		if err = Wait(reconnectErr); err != nil {
			t.Fatal("Timeout waiting for reconnect error handler")
		}
		s = RunServerWithOptions(&opts)
		defer s.Shutdown()
		if err = Wait(connected); err != nil {
			t.Fatal("Timeout waiting for reconnect handler")
		}
	})
```

Let me know if you'll be able to add it, if not I'll merge your PR and will add the test separately.

piotrpio:
I'll merge the PR to include it in the next release and will add the test separately.

sschleemilch:
> I'll merge the PR to include it in the next release and will add the test separately.

Thanks a lot. I would have been able to write a test this weekend

PR #1811, opened 2025-02-28
fix : removed the unnecessary usage of  ErrInvalidMsg

Radovenchyk:


piotrpio (changes requested):
Hello @Radovenchyk, thanks for creating the PR. I'm not sure I understand this change, what's wrong with `ErrInvalidMsg` when the message is invalid (nil). Also, there is no such error as `ErrInvalid` so this does not compile.

Radovenchyk:
I jumped to conclusions, it really doesn't compile

PR #1812, opened 2025-03-03
[FIXED] Avoid ack id collision in PublishAsync

### LINKED ISSUE #1801: Deadlock on PublishMsgAsync
### Observed behavior

I have code like this:
```
   futures := make([]jetstream.PubAckFuture, len(b))

   for i, msg := range b {
		jsMsg := nats.NewMsg(subj)

		msgBytes, err := msg.AsBytes()
		if err != nil {
			return err
		}

		jsMsg.Data = msgBytes

		future, err := js.PublishMsgAsync(jsMsg)
		if err != nil {
			return err
		}

		futures[i] = future
	}

	timeOut := time.After(5 * time.Second)

	select {
	case <-js.PublishAsyncComplete():
		for _, future := range futures {
			select {
			case <-future.Ok():
			case err := <-future.Err():
				return err
			case <-timeOut: // This will deadlock without this case here
				return errors.New("async publish complete, but futures did not return")
			}
		}
	case <-ctx.Done():
		return ctx.Err()
	case <-timeOut:
		return errors.New("did not receive async publish complete signal")
	}
```
It dead locks when reading the result of the futures if you remove the time out case.

This happens because there are ID collisions in the map that tracks acks. I updated the code that updates the map with a panic
```
func (js *jetStream) registerPAF(id string, paf *pubAckFuture) (int, int) {
	js.publisher.Lock()
	if js.publisher.acks == nil {
		js.publisher.acks = make(map[string]*pubAckFuture)
	}
	if _, ok := js.publisher.acks[id]; ok {
		panic(fmt.Sprintf("collision occurred on id: %s", id))
	}
	js.publisher.acks[id] = paf
	np := len(js.publisher.acks)
	maxpa := js.publisher.asyncPublisherOpts.maxpa
	js.publisher.Unlock()
	return np, maxpa
}
```
and I hit the panic occasionally.

I'm able to consistently reproduce this, although it takes a long time (up to an hour sometimes). I am using a batch size of 10k messages.

### Expected behavior

No dead lock or some error condition to indicate a misconfiguration in the producer client.

### Server and client version

server `2.10.25` 
go sdk `v1.39.0`

### Host environment

_No response_

### Steps to reproduce

_No response_

piotrpio:
Fixes #1801 

### DIFF
--- a/js.go
+++ b/js.go
@@ -712,13 +712,19 @@ func (js *js) newAsyncReply() string {
 	}
 	var sb strings.Builder
 	sb.WriteString(js.rpre)
-	rn := js.rr.Int63()
-	var b [aReplyTokensize]byte
-	for i, l := 0, rn; i < len(b); i++ {
-		b[i] = rdigits[l%base]
-		l /= base
+	for {
+		rn := js.rr.Int63()
+		var b [aReplyTokensize]byte
+		for i, l := 0, rn; i < len(b); i++ {
+			b[i] = rdigits[l%base]
+			l /= base
+		}
+		if _, ok := js.pafs[string(b[:])]; ok {
+			continue
+		}
+		sb.Write(b[:])
+		break
 	}
-	sb.Write(b[:])
 	js.mu.Unlock()
 	return sb.String()
 }
--- a/jetstream/publish.go
+++ b/jetstream/publish.go
@@ -352,13 +352,20 @@ func (js *jetStream) newAsyncReply() (string, error) {
 	}
 	var sb strings.Builder
 	sb.WriteString(js.publisher.replyPrefix)
-	rn := js.publisher.rr.Int63()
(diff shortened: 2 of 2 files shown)


js.go
```diff
@@ -712,13 +712,20 @@ func (js *js) newAsyncReply() string {
...
-		l /= base
+	for {
+		rn := js.rr.Int63()
+		var b [aReplyTokensize]byte
+		for i, l := 0, rn; i < len(b); i++ {
+			b[i] = rdigits[l%base]
+			l /= base
+		}
+		replyID := string(b[:])
+		if _, ok := js.pafs[replyID]; ok {
```
wallyqs, line 723: I think it would be better to avoid keeping the string and let do Write use the bytes as it was before
```suggestion
		if _, ok := js.pafs[string(b[:])]; ok {
```

PR #1813, opened 2025-03-06
fix: kv regex key validation

leandrofars:
Allows ":", "$" characters to be used as key value store keys


ripienaar:
Unfortunately what makes up a valid key is a standard across all our clients and must be the same everywhere. 

It’s managed under a specification that we are unlikely to change. 

In a nearby future release you will be able to inject a codec to translate keys to the standard set.

piotrpio:
@leandrofars thanks for the PR. As mentioned by @ripienaar, that will be possible with the inclusion of codecs and cannot be changed in a single client, so I'll be closing this PR.

PR #1814, opened 2025-03-08
perf: use reflect to test equity of message headers

Aden-Q:
This patch uses reflection to test equality of message headers. This apporach is more efficient in the case when:
+ Both `m.Header` and `msg.Header` are non-nil
+ `m.Header` and `msg.Header` share the same underlying object, so the equality can be caught before testing the content of the map, compared to the original approach. For reference, [map reflect implementation](https://github.com/golang/go/blob/5817e650946aaa0ac28956de96b3f9aa1de4b299/src/reflect/deepequal.go#L142-L144) in go source code: 

```go
	case Map:
		if v1.IsNil() != v2.IsNil() {
			return false
		}
		if v1.Len() != v2.Len() {
			return false
		}
		if v1.UnsafePointer() == v2.UnsafePointer() {
			return true
		}
		iter := v1.MapRange()
		for iter.Next() {
			val1 := iter.Value()
			val2 := v2.MapIndex(iter.Key())
			if !val1.IsValid() || !val2.IsValid() || !deepValueEqual(val1, val2, visited) {
				return false
			}
		}
		return true
```

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -31,6 +31,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"reflect"
 	"regexp"
 	"runtime"
 	"strconv"
@@ -726,21 +727,7 @@ func (m *Msg) Equal(msg *Msg) bool {
 	if !bytes.Equal(m.Data, msg.Data) {
 		return false
 	}
-	if len(m.Header) != len(msg.Header) {
-		return false
-	}
-	for k, v := range m.Header {
-		val, ok := msg.Header[k]
-		if !ok || len(v) != len(val) {
-			return false
-		}
-		for i, hdr := range v {
-			if hdr != val[i] {
-				return false
-			}
-		}
-	}
-	return true
+	return reflect.DeepEqual(m.Header, msg.Header)
 }
 
 // Size returns a message size in bytes.


piotrpio:
Hello @Aden-Q - thanks for the contribution, but I am not sure about this - using reflection has its inherent cost, and while it's true that e.g. if we're comparing the same underlying object it may be faster, for more standard comparison (different header contents, it has worse performance.

Take this benchmark:
```go
func BenchmarkMsgEqual(b *testing.B) {
	b.StopTimer()
	s := RunDefaultServer()
	defer s.Shutdown()
	nc := NewDefaultConnection(b)
	defer nc.Close()
	m1 := &nats.Msg{
		Subject: "foo",
		Data:    []byte("Hello World"),
		Header: nats.Header{
			"foo": []string{"bar"},
		},
	}
	m2 := &nats.Msg{
		Subject: "foo",
		Data:    []byte("Hello World"),
		Header: nats.Header{
			"bar": []string{"baz", "quxx"},
		},
	}
	b.StartTimer()
	b.ReportAllocs()
	for i := 0; i < b.N; i++ {
		m1.Equal(m2)
	}
}
```

These are the results with the current implementation (no reflection):
```
goos: darwin
goarch: arm64
cpu: Apple M1 Pro
BenchmarkMsgEqual
BenchmarkMsgEqual-10    	31758997	        38.86 ns/op	       0 B/op	       0 allocs/op
```

And with your changes (using reflection):
```
goos: darwin
goarch: arm64
cpu: Apple M1 Pro
BenchmarkMsgEqual
BenchmarkMsgEqual-10    	 7698056	       156.5 ns/op	      40 B/op	       2 allocs/op
```
Also note that using reflection we also add allocations.

Aden-Q:
Hi @piotrpio, thanks for taking a look and benchmarking it. Agreed that reflection is a less efficient fallback for complex data structure due to reflection overhead. While the current implementation performs better. Closing this. Thanks for taking a look!

PR #1819, opened 2025-03-14
[ADDED] Timeout option for async publish

piotrpio:
This adds a timeout option (configurable on `JetStream`) for discarding stale acks to both legacy API and `jetstream` package.


jetstream/publish.go
```diff
@@ -302,9 +305,44 @@ func (js *jetStream) PublishMsgAsync(m *nats.Msg, opts ...PublishOpt) (PubAckFut
 				return nil, ErrTooManyStalledMsgs
 			}
 		}
+		if js.publisher.ackTimeout > 0 {
+			paf.timeout = time.AfterFunc(js.publisher.ackTimeout, func() {
```
wallyqs, line 309: This one does make a goroutine per publish which is kind of new territory for publishing in the client... but would be similar overhead as users passing context.WithTimeout per publish using the non async version.
I think there might be a more efficient way, we should revisit at some point.

wallyqs (approved):
LGTM, maybe try adding to natscli bench command as an option to compare perf, memory usage.

jetstream/publish.go
```diff
@@ -415,6 +453,10 @@ func (js *jetStream) handleAsyncReply(m *nats.Msg) {
 		}
 	}
 
+	if paf.timeout != nil {
+		paf.timeout.Stop()
```
wallyqs, line 457: from Go v1.23 onwards the timers are GC without explicit stop but do we need to add something extra to cleanup publishers to remove these too when called? https://github.com/nats-io/nats.go/blob/main/js.go#L766

piotrpio: I think what we have to do is make sure we don't try to resolve timeout after acquiring the lock if it's already been resolved elsewhere. I added that to `time.AfterFunc` now:
```go
paf.timeout = time.AfterFunc(js.publisher.ackTimeout, func() {
	js.publisher.Lock()
	defer js.publisher.Unlock()

	if _, ok := js.publisher.acks[id]; !ok {
		// paf has already been resolved
		// while waiting for the lock
		return
	}
```

mprimi:
Hot take: the default timeout should be a sensible value, as opposed to "never timeout".

I know "never timeout" is the current behavior and so you are rightfully trying to maintain it.

However my point is that the current behavior is bad. The reasonable expectation is that any remote requests will eventually timeout. Never timing out can deadlock client/user code.

Just voicing my opinion, not really trying to convince you to change it. That ship has sailed, but it's something to keep in mind going forward.

ripienaar:
> Hot take: the default timeout should be a sensible value, as opposed to "never timeout".
> 
> I know "never timeout" is the current behavior and so you are rightfully trying to maintain it.
> 
> However my point is that the current behavior is unexpected. The reasonable expectation is that any remote requests will eventually timeout. Never timing out can deadlock client/user code.
> 
> Just voicing my opinion, not really trying to convince you to change it. That ship has sailed, but it's something to keep in mind going forward.

Let’s say we had to pick a default. What would you pick and what would be your supporting motivations for the number that you suggest ?

mprimi:
It's hard to choose defaults that work for everyone, but we do just that in many places, and it's fine? 

[http.Get](https://pkg.go.dev/net/http#Get) sets a default timeout, and so does [approximately every] other client library in the universe.  People can tweak those to fit their situation.

Can you imagine if `Get` docs said: "this request may not complete, ever. It's up to you to handle timeout and cancellation in user code, unfortunately there's no way to cancel** and cleanup, so expect memory to grow for each request that does not get a server response".


(** maybe the new client API does have a way of cancelling via `ctx`, not sure. I'm still on the "classic" nats.go API which does not have a way of cancelling. i.e. each request that does not complete will never be released)

js.go
```diff
@@ -1050,11 +1066,52 @@ func (js *js) PublishMsgAsync(m *Msg, opts ...PubOpt) (PubAckFuture, error) {
...
+				defer js.mu.Unlock()
+
+				if _, ok := js.pafs[id]; !ok {
+					// paf has already been resolved
+					// while waiting for the lock
+					return
+				}
+
+				// ack timed out, remove from pending acks
+				delete(js.pafs, id)
```
wallyqs, line 1084: another thing that we could consider is adding `Cancel` to the PubAcks interface with a closure over `(js.pafs, id)` so that user can control releasing these resources when needed

wallyqs: this way the pub acks would behave more like futures, return the PubAck and in case the client no longer interested in the response they can then cancel to release the resources, similar to the context package

derekcollison:
I might be of the opposite opinion, at least for go this feels correct with the wait with optional timeout on the completion channel.

piotrpio:
@derekcollison the issue here was that even if the user was timing out on waiting for the ack, the ack was never removed from the map, which eventually exceeded max pending acks and resulted in each subsequent `PublishAsync` failing (it would only recover on reconnect).

piotrpio:
The client. We have `maxPending` async acks which we can be waiting for (default 4000), after passing which we return "stalled with too many outstanding async published messages". So if there were publishes for which the client never received an ack (because of e.g. network blips), those were never discarded internally and they could pile up.

js.go
wallyqs: Then the other part that could be missing in the future is being able to check whether the future is `IsDone()` or not plus its internal timestamp check if its stale, we could then be collecting all the futures inflight for example. I think surfacing that may give folks to manage this on their own in case they want to be more efficiently

Jarema (approved):
LGTM

PR #1822, opened 2025-03-15
fix: use resp.Error to show NATS error in deleteMsg

imariman:
The error message in deleteMsg was printing nil when resp.Success was false. Changed to use resp.Error, displaying the actual NATS error.


piotrpio (approved):

PR #1824, opened 2025-03-18
[FIXED] Deadlock when fetching keys from KV while messages are deleted/purged

piotrpio:
KV watcher will now send a `nil` marker if we've not reached 0 pending msgs and have not received any new key for more than JS timeout. This avoids a deadlock when messages are purged/deleted from the stream while e.g. listing keys.


Jarema (approved):
LGTM

PR #1828, opened 2025-03-19
[FIXED] Possible panic in Consumer.Fetch

### LINKED ISSUE #1823: Fetch tries to access metadata after parsing error
### Observed behavior

Looking at the code here: https://github.com/nats-io/nats.go/blob/0f6e9ef1c9c2d1ae2b107a39e1bbb72e23c693d4/jetstream/pull.go#L810-L814
you are setting res.err if there is an error parsing metadata, but then you go ahead and try and use meta.Sequence.Stream anyway. I don't know if this ever happens in practice as I didn't see it happen, but I was looking at the code and it seems to me this could cause a panic.

### Expected behavior

The code does not try to use the metadata after it fails to parse.

### Server and client version

nats server 2.10.26
nats client v0.1.6

### Host environment

_No response_

### Steps to reproduce

_No response_

piotrpio:
Resolves #1823

### DIFF
--- a/jetstream/pull.go
+++ b/jetstream/pull.go
@@ -810,6 +810,9 @@ func (p *pullConsumer) fetch(req *pullRequest) (MessageBatch, error) {
 				meta, err := msg.Metadata()
 				if err != nil {
 					res.err = fmt.Errorf("parsing message metadata: %s", err)
+					res.done = true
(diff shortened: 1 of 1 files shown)


wallyqs (approved):
LGTM

PR #1835, opened 2025-03-25
[FIXED] Set context from option when creating consumer in js.Subscribe

piotrpio:
Previously when using `nats.Context(ctx)` option in `js.Subscribe()`, `ctx` was not used when creating consumer.


derekcollison (approved):
LGTM

wallyqs (commented):
LGTM

PR #1837, opened 2025-03-29
cancel when connection status changes

alexbozhenko:


js.go
```diff
@@ -3129,6 +3130,15 @@ func (sub *Subscription) Fetch(batch int, opts ...PullOpt) ([]*Msg, error) {
 						hbLock.Unlock()
 						cancel()
 					})
+					go func() {
+						select {
+						case <-ctx.Done():
```
wallyqs, line 3135: I think also need to stop the timer?

js.go
```diff
@@ -3121,6 +3121,7 @@ func (sub *Subscription) Fetch(batch int, opts ...PullOpt) ([]*Msg, error) {
 			if err := nc.PublishRequest(nms, rply, req); err != nil {
 				return err
 			}
+			connStatusChanged := nc.StatusChanged()
 			if o.hb > 0 {
```
piotrpio, line 3125: I think this should be triggered regardless of whether `o.hb > 0`. We want to be able to bail early even without heartbeat set.

js.go
piotrpio, line 1: Same should be added for `FetchBatch()`

piotrpio (changes requested):

piotrpio:
@alexbozhenko closing in favor of #1840

PR #1839, opened 2025-03-31
[IMPROVED] Stop the watcher before performing the purge operations for PurgeDeletes.

derekcollison:
This is to avoid the server having to keep calculating numPending for the watcher itself. This can be especially penal if we purge more than one message and the stream / KV has a lot of keys. The server will also be improved in handling this situation, those changes are in flight.


piotrpio (approved):
LGTM!

PR #1840, opened 2025-04-01
[FIXED] Cancel Fetch and FetchBatch on reconnect 

piotrpio:
Implementation based on #1837 


Jarema (approved):
LGTM!

PR #1841, opened 2025-04-01
[FIXED] Data race when reading current status in sub.StatusChanged and nc.StatusChanged

### LINKED ISSUE #1815: Data race between subscription.StatusChanged and subscription.changeSubStatus
### Observed behavior

When calling subscription.Drain followed by subscription.StatusChanged, the current status is read from inside StatusChanged outside a mutex, but written inside changeSubStatus, inside the mutex, which causes a data race.

### Expected behavior

I expect the status read and write to be serialized.

### Server and client version

github.com/nats-io/nats.go v1.38.0

### Host environment

_No response_

### Steps to reproduce

_No response_

piotrpio:
Fixes #1815 


Jarema (approved):
LGTM!

PR #1843, opened 2025-04-02
[ADDED] DefaultTimeout option for JetStream API requests

piotrpio:
This adds a `WithDefaultTimeout` option when creating `JetStream`. This timeout is used when a context without deadline is passed to JetStream requests.


Jarema (approved):
LGTM

PR #1842, opened 2025-04-02
[FIXED] Reset channel after closing in ForceReconnect

piotrpio:
Set the `rqch` channel to `nil` after closing to avoid `close of closed channel` panic.
This channel is recreated in `doReconnect()` allowing to bypass wait time between reconnects.

Signed-off-by: Piotr Piotrowski <piotr@synadia.com>

PR #1846, opened 2025-04-02
[FIXED] Initialize rqch channel for immediate reconnection on ForceReconnect

piotrpio:
This is a follow-up to #1842 which introduced an issue where `nc.rqch`
could be cleared before reading it in `doReconnect`.
This happened e.g. when an auth error triggered reconnect close to
calling `ForceReconnect` (made apparent by `TestAuthExpiredForceReconnect`
starting to be flaky).


Jarema (approved):
LGTM

PR #1849, opened 2025-04-08
Remove `golang.org/x/text` dependency

neilalexander:
The only use of `x/text` is in one `String()` method. In the interests of not pulling more dependencies into the server than we need to, it'd be better if we didn't do that.


piotrpio (approved):
LGTM!

PR #1850, opened 2025-04-09
[FIXED] Use default timeout for ObjectStore.Get when no deadline is set on ctx

### LINKED ISSUE #1848: Goroutine and consumer leak on jetstream's ObjectStore.Get with timeoutless context when the server is unresponsive
### Observed behavior

The NATS server had a memory leak whose root cause is not yet known, which caused "context deadline exceeded" errors on the client that attempted to get an object from a Jetstream Object Store with [ObjectStore.Get](https://github.com/nats-io/nats.go/blob/v1.40.1/jetstream/object.go#L143). Go routines calling ObjectStore.Get passing a context that had no timeout were hanging and ephemeral consumers created for the operation were leaking. An upgrade to v1.41.0 did not fix the issue. A downgrade to v1.40.0 fixed the client's issue.

### Expected behavior

A default timeout should be used in cases that a context which has no timeout is passed and no resources should leak. The issue is not fixed by #1843 as in this case the subscription is done against the NATS connection directly as jetstream's ObjectStore.Get uses a legacy  nats.JetStreamContext.

### Server and client version

Server: 2.11.1-RC.2
Client: 1.41.0 (issue introduced in 1.40.1)

A downgrade to client 1.40.0 fixed the issue.

### Host environment

_No response_

### Steps to reproduce

Since the server issue resolved and I had no faulty server to test against I added the following test to nats.go/test/js_test.go and ran it:

```
func TestJetStreamSubscribeConsumerCreateNoTimeout(t *testing.T) {
	srv := RunBasicJetStreamServer()
	defer shutdownJSServerAndRemoveStorage(t, srv)
	nc, js := jsClient(t, srv)
	defer nc.Close()

	_, err := js.AddStream(&nats.StreamConfig{Name: "foo", Subjects: []string{"FOO.*"}})
	if err != nil {
	}
	_, err = js.SubscribeSync("", nats.BindStream("foo"), nats.Context(context.Background()))
	if err != nil {
		t.Fatalf("Expected error")
	}
}
```

The test runs successfully (in this case against e040f7139e0dc179f04fd61e1dd05c5c1fd07d60), but a goroutine leak is identified:
```
=== RUN   TestJetStreamSubscribeConsumerCreateNoTimeout
--- PASS: TestJetStreamSubscribeConsumerCreateNoTimeout (0.04s)
PASS
goleak: Errors on successful test run: found unexpected goroutines:
[Goroutine 19 in state chan receive, with github.com/nats-io/nats%2ego.(*js).subscribe.func3 on top of the stack:
github.com/nats-io/nats%2ego.(*js).subscribe.func3()
	..../nats.go/js.go:2056 +0x38
created by github.com/nats-io/nats%2ego.(*js).subscribe in goroutine 5
	..../nats.go/js.go:2055 +0x1308
]
```

The error points to nats.go/js.go:
```
	if ctx != nil {
		go func() {
			<-ctx.Done()
			sub.Unsubscribe()
		}()
	}
```

If the context has no timeout, the consumer is not deleted and the goroutine hangs until the given context is cancelled. There is no default timeout implemented.

piotrpio:
Fixes #1848

### DIFF
--- a/jetstream/object.go
+++ b/jetstream/object.go
@@ -488,7 +488,7 @@ const (
 )
 
 func (js *jetStream) CreateObjectStore(ctx context.Context, cfg ObjectStoreConfig) (ObjectStore, error) {
-	scfg, err := js.prepareObjectStoreConfig(ctx, cfg)
+	scfg, err := js.prepareObjectStoreConfig(cfg)
 	if err != nil {
 		return nil, err
 	}
@@ -511,7 +511,7 @@ func (js *jetStream) CreateObjectStore(ctx context.Context, cfg ObjectStoreConfi
 }
 
 func (js *jetStream) UpdateObjectStore(ctx context.Context, cfg ObjectStoreConfig) (ObjectStore, error) {
-	scfg, err := js.prepareObjectStoreConfig(ctx, cfg)
+	scfg, err := js.prepareObjectStoreConfig(cfg)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


Jarema (approved):
LGTM

PR #1856, opened 2025-04-16
[FIXED] Add RemoveStatusListener method and fixFetch memory leak

piotrpio:
This adds a new method, `Conn.RemoveStatusListener`, which allows to manually remove an unwanted status changed listener. Internally listeners were converted to map for more efficient lookup.

Fixes a memory leak in `Fetch` and `FetchBatch` which caused status listeners to never be cleaned up (they would only disappear after status is changed).


wallyqs (approved):
LGTM

PR #1863, opened 2025-04-22
Fix typo: TimeStampHeaer -> TimeStampHeader

atauov:
This pull request fixes a typo in the constant name:
TimeStampHeaer -> TimeStampHeader

Correcting this typo improves code readability and prevents potential issues when using this constant.


piotrpio:
I agree this should be fixed - I am afraid however that this, while minor, is a breaking change since the unfortunate `TimeStampHeaer` is exported.

PR #1866, opened 2025-04-25
Return deleted entry with ErrKeyDeleted

matthewdevenny:
If the latest KV entry is a Deleted entry return the deleted entry with an ErrKeyDeleted error rather than returning nil.


piotrpio (commented):
I do not think the deleted entries should be returned by `Get()`, for a few reasons:

1. While tombstones are stored on the stream and can be read using `Watch()`, I do not think it's a good idea to return deleted entries in `Get()` - this is very counter-intuitive to the expectation associated with deleting a key.

piotrpio:
I'll be closing this PR @matthewdevenny if you feel this warrants further discussion feel free to reopen.

PR #1877, opened 2025-05-27
Add nc.LocalAddr, similar to nc.ConnectedAddr

alexbozhenko:
Currently, the Go client library exposes the remote address of the underlying net.Conn.
It would be helpful if it also exposed the LocalAddr.
The usecase is to help determine which network was used to reach the remote NATS server.


wallyqs (approved):
LGTM

PR #1880, opened 2025-06-02
[FIXED] Remove Println from Service API Stop()

piotrpio:


alexbozhenko:
@derekcollison yes, we can enable `forbidigo` linter(like server does) once https://github.com/nats-io/nats.go/pull/1774 is merged

PR #1884, opened 2025-06-19
[FIXED] KeyValue Keys() and ListKeys() returning duplicates

piotrpio:
Use consumer pending count from creation time instead of first message
delta to fix race where concurrent updates could cause duplicate keys
during listing operations.

Fixed in both legacy and new JetStream APIs.


alexbozhenko:
@Jarema please review when you are back

Jarema (approved):
LGTM

PR #1886, opened 2025-06-22
[ADDED] JetStream: Expose ClientTrace in JetStreamOptions

piotrpio:
Make ClientTrace field publicly accessible through Options() method.
Returns a copy to maintain read-only contract.


Jarema (approved):
LGTM

PR #1892, opened 2025-06-30
[Added] option WithProtocolTrace to enable protocol tracing

alexbozhenko:
This was claude code generated/human verified,
based on https://github.com/nats-io/nats.go/pull/911
and https://github.com/nats-io/nats.go/pull/334#issuecomment-352613677
And similar to [debug option in nats.js](https://nats-io.github.io/nats.js/core/index.html#connection-options)

We can copy-paste the tracing snippet that @kozlovic shows in the second link(I found it extremely useful), but why don't we make it part of the client, so it is easy to do just:

```
nc, err := nats.Connect(nats.DefaultURL, nats.WithProtocolTrace(nats.NewStderrTrace()))
```
and easily see all the nitty-gritty details from the wire
image
in any program or test

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -24,6 +24,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"log"
 	"math/rand"
 	"net"
 	"net/http"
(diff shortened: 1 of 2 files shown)


nats.go
```diff
@@ -272,6 +272,48 @@ type InProcessConnProvider interface {
...
+type tracedConn struct {
+	net.Conn
+	trace *ConnTrace
+}
+
+// Read implements net.Conn.Read and traces data received.
+func (tc *tracedConn) Read(b []byte) (n int, err error) {
+	n, err = tc.Conn.Read(b)
+	if err == nil && tc.trace != nil && tc.trace.DataReceived != nil {
+		// Make a copy of the data to avoid race conditions
```
alexbozhenko, line 297: Not sure if this copying is actually needed

PR #1896, opened 2025-07-07
Fix timeoutWriter not recovering after first error

### LINKED ISSUE #1891: timeoutWriter stops writing after first error
### Observed behavior

After the first write error (e.g., timeout), the timeoutWriter permanently returns that error for all subsequent calls to Write(), even if the underlying connection is still valid and ready. This makes it impossible for the writer to recover from transient issues.

### Expected behavior

A timeout or transient error should only affect the specific write operation it occurred on. The writer should continue attempting to send data on subsequent calls, unless the connection has been closed or is otherwise invalid.

### Server and client version

https://github.com/nats-io/nats.go/tree/v1.43.0

### Host environment

_No response_

### Steps to reproduce

_No response_

wallyqs:
Fixes #1891

### DIFF
--- a/nats_test.go
+++ b/nats_test.go
@@ -1780,3 +1780,75 @@ func BenchmarkHeaderDecode(b *testing.B) {
 		})
 	}
 }
+
+// mockConn simulates a network connection that can fail and recover
+// after a number of attempts.
+type mockConn struct {
+	failures          int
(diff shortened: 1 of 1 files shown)


piotrpio (approved):
LGTM!

PR #1899, opened 2025-07-08
[ADDED] Service API: WithEndpointPendingLimits option

piotrpio:
Adds ability to configure pending message/byte limits on service endpoints to handle slow consumers gracefully.

* Add WithEndpointPendingLimits() option for endpoint configuration
* Support message count and/or byte limits (-1 for no limit)  
* Enhanced NATSError with error wrapping support for proper error chains

### DIFF
--- a/micro/test/service_test.go
+++ b/micro/test/service_test.go
@@ -1910,3 +1910,183 @@ func TestDisableQueueGroup(t *testing.T) {
 	wg.Wait()
 
 }
+
+func TestEndpointPendingLimits(t *testing.T) {
+	s := RunServerOnPort(-1)
+	defer s.Shutdown()
+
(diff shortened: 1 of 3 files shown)


micro/test/service_test.go
```diff
@@ -1910,3 +1910,183 @@ func TestDisableQueueGroup(t *testing.T) {
...
+			t.Fatalf("Error adding service: %v", err)
+		}
+		defer srv.Stop()
+
+		msgLimit := 10
+		bytesLimit := 1024
+
+		err = srv.AddEndpoint("test", handler,
+			micro.WithEndpointSubject("test.pending"),
+			micro.WithEndpointPendingLimits(msgLimit, bytesLimit))
```
aricart, line 1944: Awesome!!

PR #1903, opened 2025-07-22
Fix Subject prefix for the Create/Update operation in KV store.

SalvaChiLlo:
Create a valid Subject for mirrored streams by using the `kv.putPre` prefix which references the subject in the remote domain

Without the prefix it tries to push messages into a topic no one is interested in the domain making leaf nodes unusable for create and update operations over the KV store


piotrpio (approved):

PR #1909, opened 2025-07-24
[FIXED] Use timeout from JetStreamContext if no deadline is set on ctx

piotrpio:
Legacy JetStream API calls could block indefinitely if `nats.Context` was used without deadline. This PR changes this so that a timeout from JetStreamContext will be used instead.


wallyqs (approved):
LGTM

PR #1913, opened 2025-08-05
[FIXED] Track delivered count and auto-unsubscribe for channel subscriptions

### LINKED ISSUE #1890: AutoUnsubscribe leaves ChanSubscription active
### Observed behavior

I hope I'm not using the library wrong (but if I am I would appreciate it being better documented), but AFAICT when calling AutoUnsubscribe on a ChanSubscription, the subscription remains active even after the delivery count is reached.

Visual inspection of `nats.go` from the git repository revealed that `nc.removeSub` is called only in:
- `waitForMsgs`, which is started only for AsyncSubscriptions)
- `checkDrained`, which is not started by `AutoUnsubscribe` because it sets `drainMode` to `false`
- `unsubscribe`, when the unsubscription is immediate,
- `processNextMsgDelivered`, which handles SyncSubscriptions.

So as far as I can tell, only ChanSubscription are never removed automatically, which is consistent with `oldRequest` needing to defer `Unsubscribe` after calling `AutoUnsubscribe(1)`.

### Expected behavior

I expected AutoUnsubscribe to be equivalent to Unsubscribe after the specified number of messages, which would change the subscription status to closed. That's what happens to AsyncSubscription, I expected ChanSubscription to behave similarly (or to see the documentation of AutoUnsuscribe mention the difference of behavior).

### Server and client version

I witnessed it with nats.go v1.43.0, connecting to nats-server v2.10.23 (but I don't expect that to be relevant).

### Host environment

All this runs on FreeBSD hosts on amd64 processors.

### Steps to reproduce

Hopefully the explanations above are enough to consider the issue, but I can try writing a reproducer if it's really needed.

piotrpio:
Fixes #1890

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -3432,6 +3432,10 @@ func (nc *Conn) processMsg(data []byte) {
 		if sub.mch != nil {
 			select {
 			case sub.mch <- m:
+				// For ChanSubscribe, track delivered count here
+				if sub.typ == ChanSubscription {
+					sub.delivered++
+				}
 			default:
 				goto slowConsumer
 			}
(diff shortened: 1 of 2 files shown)


wallyqs (approved):
LGTM, nice find!

wallyqs:
btw this would have affected js.ChanSubscribe as well when used with Drain, right?

PR #1932, opened 2025-09-09
Counters

piotrpio:
- add `AllowMsgCounter` stream config field
- add `NoHeaders` option when fetching message from stream


Jarema (approved):
LGTM

PR #1941, opened 2025-09-19
[FIXED] Revert leftover no headers support changes

piotrpio:
This reverts leftover changes introduced when adding no headers support.


Jarema (approved):
LGTM

PR #1957, opened 2025-10-17
[FIXED] Fix function pointer check in js.apiRequestWithContext()

svenfoo:
Most likely a copy'n'paste bug, the code should check the function pointer that it is about to call.


piotrpio (approved):
LGTM! Thank you fro the contribution!

PR #1958, opened 2025-10-20
[ADDED] Reconnect to selected server

piotrpio:
Implements [ADR-40](https://github.com/nats-io/nats-architecture-and-design/pull/388) extensions for custom server pool and reconnect server selection.

- Add `ReconnectToServer()` callback option invoked on each reconnect attempt, allowing users to select a server from the pool and control the reconnect delay
- Add `SetServerPool()` to replace the server pool at runtime for reconnection attempts
- Add `ServerPool()` to retrieve a snapshot of the current server pool
- Export `Server` and `ServerInfo` types
- Enforce pool membership - returning a server not in the pool fires error handler with `ErrServerNotInPool` and falls back to default selection

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -6219,6 +6219,9 @@ func (nc *Conn) Barrier(f func()) error {
 }
 
 // ServerPool returns a copy of the current server pool for the connection.
+//
+// This function should not be called from within connection callbacks to avoid
+// potential deadlocks.
 func (nc *Conn) ServerPool() []Server {
 	nc.mu.RLock()
 	defer nc.mu.RUnlock()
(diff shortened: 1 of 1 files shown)


nats.go
```diff
@@ -6085,6 +6208,97 @@ func (nc *Conn) Barrier(f func()) error {
...
+	var firstWS *bool
+
+	for _, addr := range servers {
+		s, err := nc.parseServerURL(addr, false, false)
+		if err != nil {
+			return err
+		}
+		isWS := isWebsocketScheme(s.URL)
+
+		// Check websocket consistency
```
ripienaar: We're not checking if the current client mode is web socket and so we should deny non ws entries from being added, is that a concern?


Probably addURLToPool() guards against that but worth checking

nats.go
```diff
@@ -6085,6 +6208,97 @@ func (nc *Conn) Barrier(f func()) error {
...
+// This function does not trigger an immediate reconnect. The new server
+// pool will be used on the next reconnect attempt.
+// If you want to trigger an immediate reconnect to apply the new server pool,
+// you can call [Conn.ForceReconnect] after this function.
+//
+// Unless [IgnoreDiscoveredServers] is used true, the client will continue to
+// discover and add new servers to the pool as it receives INFO messages from
+// the server. If you want to prevent this behavior and only use the servers
+// provided in SetServerPool, use [IgnoreDiscoveredServers].
+func (nc *Conn) SetServerPool(servers []string) error {
```
ripienaar: if I am adding a entry that was already in the pool and it has >1 reconnect count, should adding it again reset that or preserve it?

ripienaar (approved):

PR #1960, opened 2025-10-23
[ADDED] Enable custom subject transforms on KV sourcing

piotrpio:
If user provides `SubjectTransformConfig` for a source in `KeyValueConfig`, use it instead of overwriting. 

This also enables using non-kv streams as KV sources.


Jarema (approved):
LGTM!

PR #1974, opened 2025-11-24
[IMPROVED] Add publish subject validation

piotrpio:
Only validate for subjects which break protocol.

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -3916,7 +3916,7 @@ func (nc *Conn) kickFlusher() {
 // argument is left untouched and needs to be correctly interpreted on
 // the receiver.
 func (nc *Conn) Publish(subj string, data []byte) error {
-	return nc.publish(subj, _EMPTY_, nil, data)
+	return nc.publish(subj, _EMPTY_, false, nil, data)
 }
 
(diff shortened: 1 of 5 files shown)


nats.go
```diff
@@ -4072,15 +4072,29 @@ func (nc *Conn) PublishRequest(subj, reply string, data []byte) error {
...
 const digits = "0123456789"
 
+// validateSubject checks if the subject contains characters that break the NATS protocol.
+func validateSubject(subj string) error {
+	if subj == "" {
+		return ErrBadSubject
+	}
+	for i := range len(subj) {
+		c := subj[i]
+		if c <= ' ' && (c == ' ' || c == '\t' || c == '\r' || c == '\n') {
```
wallyqs: Like the idea of skipping all control characters `c <= ' '`... wonder if just that check is not enough?

Jarema:
@piotrpio I wonder if we should validate reply subject sent by the client - I thionk that it only has to happen on custom inbox/prefix?

wallyqs (approved):
LGTM, impact from this is a few ns at most in the worst cases

PR #1990, opened 2026-01-02
[IMROVED] Bump dependencies

### LINKED ISSUE #1988: Crypto Lib CVEs
### Proposed change

Greetings,

I just wanted to report that the crypto lib being used by nats.go has several CVEs at the present version. It would be great if we could upgrade the crypto lib to >= v0.45.0 for the next release.

https://pkg.go.dev/golang.org/x/crypto?tab=versions

Many thanks!

### Use case

Cleaning up CVEs for the folks that run vulnerability scanners on their code.

### Contribution

_No response_

piotrpio:
Fixes #1988 


Jarema (approved):
Lgtm

PR #2019, opened 2026-02-03
fix bug in constant name TimeStampHeader

jjpinto:
Fixed misspelled exported constant TimeStampHeaer to TimeStampHeader, corrected a fe typos and improved doc comments.


jjpinto:
@piotrpio Understood! Should I close this PR and keep the doc comment fixes in a separate PR, or would you prefer to handle those separately as well?

PR #2065, opened 2026-04-29
New Nats Docs, better subjects single wildcard

scottf:
https://github.com/synadia-labs/new-nats.docs/issues/31


piotrpio (approved):
LGTM

piotrpio:
Linter is failing but that's because of an issue in `queue-groups-dynamic-scaling` example, will fix separately

PR #2070, opened 2026-05-06
[ADDED] StreamSource.Consumer config field and AckFlowControlPolicy

piotrpio:


piotrpio:
@claude re review considering the following:

I added AckFlowControlPOlicy as another missing feature from nats-server 2.14, as it was needed for an actual e2e test of the new config fields.


Answering your questions:

1. Added a basic test
2. These structs are 1-1 with nats-server
3. Fixed
4. I think this doc is clear in the context of push consumers
5. Legacy API was not updated on purpose

Jarema (approved):
LGTM

PR #2073, opened 2026-05-07
fix: add nil checks for empty JetStream API responses (#1642)

### LINKED ISSUE #1642: Add checks for empty response in JetStream API calls
### Proposed change

Add checks for `nil` in all JetStream API requests in `jetstream` package.

For example, in [upsertConsumer()](https://github.com/nats-io/nats.go/blob/main/jetstream/consumer.go#L229), a check could be added returning a new error kind:
```go
	if resp.ConsumerInfo == nil {
		return nil, ErrInvalidJetStreamResponse
	}
```

Package should be scanned for places where API requests are sent, e.g. CRUD operations on streams, consumers, `AccountInfo` etc.

### Use case

While unlikely, it may happen that the server responds to JetStream API requests with unexpected data, neither `apiResponse` (for errors) nor the actual expected type. 

While not common, this can happen e.g. due to server bug and in such cases the client should not panic but simply return an error indicating what happened.

https://github.com/nats-io/nats.go/issues/1641 was a motivation for this proposal.

### Contribution

_No response_

colecschmidt:
Fixes #1642.

Adds ErrInvalidJetStreamResponse checks across JetStream API calls in 
the jetstream package — CreateStream, UpdateStream, Stream, stream.Info, 
and stream.getMsg — following the pattern established in upsertConsumer().

Includes 5 tests using a mock subscriber on a custom API prefix to 
simulate empty server responses. All existing tests pass with -race.

I found this issue while building FlowQ, a distributed job queue built 
on JetStream: https://github.com/colecschmidt/flowq


jetstream/test/nil_response_test.go
```diff
@@ -0,0 +1,185 @@
+// Copyright 2025 The NATS Authors
```
piotrpio:
```suggestion
// Copyright 2026 The NATS Authors
```

jetstream/test/nil_response_test.go
```diff
@@ -0,0 +1,185 @@
...
+
+	ctx := context.Background()
+	s, err := js.CreateStream(ctx, jetstream.StreamConfig{Name: "test", Subjects: []string{"test"}})
+	if err != nil {
+		t.Fatalf("CreateStream: %v", err)
+	}
+
+	_, err = s.Info(ctx)
+	if !errors.Is(err, jetstream.ErrInvalidJetStreamResponse) {
+	js, cleanup := setupNilResponseMock(t, func(subj string) []byte {
```
piotrpio: I think the claude review suggestion resulted in broken test: https://github.com/nats-io/nats.go/pull/2073/commits/5db05135da82290f9c7a72935b120aa0c95c56a8

Claude reviewer is good at reviews, but quite bad at suggestions :)

piotrpio (changes requested):
The checks are good, but I am not sure the complexity of the fix justifies this kind of test setup - these guards are so straightforward I would honestly drop the tests entirely.

colecschmidt:
@piotrpio removed the test file entirely as suggested. All existing tests pass. Let me know if anything else is needed!

piotrpio (approved):

PR #2075, opened 2026-05-11
Add Client and Orbit section to README

Jarema:
Explains the split between the core nats.go client and the Orbit utility packages (https://github.com/synadia-io/orbit.go): what each is for, what goes where, and a small layering diagram. Mirrors the section added in nats.rs#1580.

Signed-off-by: Tomasz Pietrek <tomasz@synadia.com>


piotrpio (approved):
LGTM!

PR #2079, opened 2026-05-14
[ADDED] `micro.WithEndpointMetadataKey`

### LINKED ISSUE #2078: Allow a way to provide individual metadata keys for `micro.Endpoint`
### Proposed change

Two possible solutions:
* Make `endpointOpts` and its fields public
* Add a new `WithEndpointMetadataKey`

I will go with the second as that is a purely additive change to the `micro` API.

### Use case

I want to write custom `micro.EndpointOpt` functions that can be composed together. These functions will mainly set individual key-value pairs on the endpoint's `Metadata`.

### Contribution

Yes

joeriddles:
closes https://github.com/nats-io/nats.go/issues/2078

See also https://github.com/nats-io/nats.go/pull/2080 for an alternate solution to the issue.

### DIFF
--- a/micro/service.go
+++ b/micro/service.go
@@ -967,6 +967,9 @@ func WithEndpointMetadata(metadata map[string]string) EndpointOpt {
 
 func WithEndpointMetadataKey(key, value string) EndpointOpt {
 	return func(e *endpointOpts) error {
+		if e.metadata == nil {
+			e.metadata = map[string]string{}
+		}
 		e.metadata[key] = value
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


micro/service.go
```diff
@@ -965,6 +965,16 @@ func WithEndpointMetadata(metadata map[string]string) EndpointOpt {
 	}
 }
 
+func WithEndpointMetadataKey(key, value string) EndpointOpt {
```
piotrpio, line 971: Please add documentation for this option (also mention how it works together with `WithEndpointMetadata`). I noticed the 2 other options above also lack documentation, I can add that in a separate PR.

joeriddles: Done

piotrpio (approved):
LGTM!

PR #2080, opened 2026-05-14
[IMPROVED] Add `micro.AddEndpointWithOptions` and make `micro.EndpointOpts` public

### LINKED ISSUE #2078: Allow a way to provide individual metadata keys for `micro.Endpoint`

joeriddles:
closes https://github.com/nats-io/nats.go/issues/2078

See also https://github.com/nats-io/nats.go/pull/2079 for an alternate solution to the issue.

### DIFF
--- a/micro/service.go
+++ b/micro/service.go
@@ -66,16 +66,16 @@ type (
 		AddEndpoint(string, Handler, ...EndpointOpt) error
 	}
 
-	EndpointOpt func(*endpointOpts) error
+	EndpointOpt func(*EndpointOpts) error
 	GroupOpt    func(*groupOpts)
 
-	endpointOpts struct {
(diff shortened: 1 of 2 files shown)


piotrpio:
I feel like this solution is more complex than #2079 while not adding much value. Also, it does not work when adding endpoint to a group. Let's move on with #2079

PR #2082, opened 2026-05-23
[IMPROVED] Migrate tests to ntf

piotrpio:
Migrates the integration suites (`test/`, `jetstream/test/`, `micro/test/`) off the
embedded in-process `nats-server` and onto [ntf-server](https://hub.docker.com/r/synadia/ntf-server).

- Tests bring up servers through the `ntf-client` API (`withServer`, `withJSServer`,
  `withJSCluster`, …) instead of `RunDefaultServer` / `RunBasicJetStreamServer` /
  `setupJSClusterWithSize`.
- Inline `*.conf` string literals replaced with typed options (`WithAccounts`,
  `WithAuthorization`, `WithWebSocket`, `WithLeafNode`, …); TLS tests use per-instance
  generated certs.
- Lifecycle control (stop/start a node, restart, failover) goes through the tester's
  RPC rather than in-process server handles.
- `go_test.mod` drops `github.com/nats-io/nats-server/v2` and its transitive closure,
  and adds `github.com/synadia-io/orbit.go/ntf-client`.
- CI runs the tester as a service container on the job's docker network
- `make` targets for local runs (`tester-up-host`, `test`, `test-tester`, …) and a new
  `TESTING.md` covering both host-side and sibling-container modes.
- With `TESTER_NATS_URL` unset the integration suites skip, so a plain `go test ./...`
  on a fresh checkout still runs the unit tests.


piotrpio:
@claude review this PR

piotrpio:
@claude issues addressed, please re-review

Jarema (changes requested):
The approach is sound and a nice improvement we can adopt cross-clients.

There are some AI-review points

# PR #2082 — Deep Review: Migrate tests to ntf

**PR:** https://github.com/nats-io/nats.go/pull/2082 (`[IMPROVED] Migrate tests to ntf`, piotrpio)
**Scope:** 52 files, +31,847/−36,369. Migrates `test/`, `jetstream/test/`, `micro/test/` off the embedded in-process `nats-server` onto the dockerized `synadia/ntf-server` tester, driven via `github.com/synadia-io/orbit.go/ntf-client v0.0.3`.
**Review method:** 10-example sample review, then 4 parallel deep audits (semantic drift, lifecycle semantics, CI/tooling, timing/flakes), with live test runs against a real tester on macOS/Docker Desktop. Agent claims spot-verified against source.

---

## Verdict

Merge-worthy architecture. Approach sound, cross-language value real. But "faithful line-by-line port" is too generous: **one real readiness regression, ~6 genuine coverage losses in core-client tests (one with a factually wrong rationale comment), and a load-bearing undocumented dependency contract.** None block the architecture; all fixable in-place. Recommended pre-merge: fix items H1–H2, restore or correctly document H3 coverage.

## Empirically verified (live tester, macOS/Docker Desktop — worst-case topology)

- `go vet` clean across all packages with `internal_testing` tag; skip-path without `TESTER_NATS_URL` works as documented.
- All green: 4 core tests (`TestAuth`, `TestBasicReconnectFunctionality`, `TestBasicClusterReconnect`, `TestClientCertificate`), full micro suite, all KV tests, both cluster pull tests, 18-test reconnect family, `TestCustomReconnectDelay` ×5, `TestReconnectWaitJitter` ×5, ordered-consumer restart subtests ×5 (10/10).
- JetStream state **does** survive `StopServer`/`StartServer` today (observed, not contractual — see H2).
- Real CI logs compared: PR race job passes 1766 tests vs main's 1757, per-package coverage within 0.1pp of main. **CI currently runs the full integration suite — no silent skip.**
- `nc -z` health check verified inside the actual `synadia/ntf-server:v0.0.5-nats-2.14.3` image; alpine CGO/race setup (`gcc libc-dev`) verified by running it.

---

## High — act before merge

### H1. Cluster readiness regression
Main's `withJSClusterAndStream` (`jetstream/test/helper_test.go:249-286` on main) retried `AccountInfo` + `CreateStream` every 500ms for up to 10s. The new `withJSCluster` only calls `WaitForJetStream`, which does a raw `nc.Request("$JS.API.INFO")` and checks the **transport** error only:

- A leaderless cluster **replies** with `JSClusterNotAvail` in the response body → passes the check (server replies with the error payload once the meta group is >10s without quorum; within the helper's ~15s budget this window is reachable).
- Stream **placement** (Replicas:3) requires peers registered with the meta layer *after* election → `no suitable peers for placement` race on single-shot `CreateStream`.

Five migrated cluster tests now do single-shot `CreateStream`: `jetstream/test/pull_test.go:823, :859, :2694, :2734, :2798`.
**Fix:** poll `js.AccountInfo(ctx)` (which parses the response body and surfaces `JSClusterNotAvailErr`) in `withJSCluster`, restoring the retry.

### H2. Post-restart readiness undefined; storage guarantee assumed
Old `restartBasicJSServer` blocked until `ReadyForConnections`. `inst.StartServer` is an RPC that returns on ack — ntf-client v0.0.3 documents **no readiness semantics** and **no storage-preservation guarantee** for `StopServer`/`StartServer` (the only "keeps storage on disk" wording sits on `StopInstanceRequest`, an API type with no client method; `ntf_client.go:764` doc is one line).

- `jetstream/test/ordered_test.go:184, :632` publish immediately after restart with zero wait.
- `jetstream/test/pull_test.go:2837-2844` carries an explicit 10×500ms `js.Stream` poll before publishing — someone got bitten at exactly one site and fixed only that site.
- Comment at `test/js_test.go:3192` — "storage preserved by testservice across stop/start" — states as fact a guarantee the dependency never makes.

**Fix:** a shared post-restart readiness helper (the `pull_test.go:2837` poll, applied everywhere); soften the js_test comment to "observed behavior"; raise the doc gap upstream (mirror `StopInstanceRequest` wording onto `StopServer`/`StartServer`, document port stability).

### H3. Coverage losses in core-client tests (all verified in source)
1. **`nats.ErrMaxAccountConnectionsExceeded` sentinel mapping uncovered.** Exact-error assert downgraded to `strings.Contains(err.Error(), "maximum account active connections exceeded")` (`test/reconnect_test.go:1346, :1350`; main had `err != nats.ErrMaxAccountConnectionsExceeded` at `:1471, :1475`). No test anywhere now asserts the client maps the server error to the sentinel.
2. **JWT account-limit reconnect uncovered.** Same test's JWT phase (`s.LookupAccount` + `jwt.NewAccountClaims` + `s.UpdateAccountClaims`, clients nc4/nc5) deleted, replaced by a second config reload. `UpdateAccountClaims` appears in zero tests.
3. **AND→OR weakening**, same test: main `:1501` required *both* clients reconnected; PR `:1376` accepts *at least one*. Post-tighten re-verification (new connect must fail) also deleted.
4. **`TestRetryOnFailedConnect` Phase B dropped with a wrong rationale.** The deleted phase asserted: with `RetryOnFailedConnect(true)` + `MaxReconnects(-1)` against a server restarted WITH auth, `ClosedHandler` fires and the reconnect CB does **not** (auth error aborts despite *unlimited* reconnects). The in-code rationale (`test/conn_test.go:~2953`) claims `TestRetryOnFailedConnectWithAuthError` is equivalent — it is not: that test uses `MaxReconnects(2)`, so a close cannot distinguish auth-abort from retry exhaustion.
5. **DNS-randomization path orphaned.** `TestLookupHostResultIsRandomized`/`...NotRandomizedWithNoRandom` were rewritten from two servers sharing one port on 127.0.0.1/::1 (exercising the `net.LookupHost` result shuffle, `nats.go:2455-2476`) to two distinct URLs (exercising `shufflePool`, `nats.go:2128` — already covered by `TestServersRandomize`). Both tests are now redundant and the DNS path has zero coverage.
6. **Zero-coverage APIs:** `Conn.ConnectedClusterName()` (both asserts dropped from `TestConnectedServer`, with rationale) and the in-process-conn happy path (`TestInProcessConn` replaced by root-package tests that only verify option storage and stub-error propagation — no test reaches CONNECTED over an in-process conn and answers `RTT()`).

---

## Medium

- **`TestGetClientID` vacuous pass:** discovery-wait became `if err := Wait(discoveredCh); err == nil { …check… }` (`test/conn_test.go:2785`); main hard-failed on timeout. Slow gossip under docker → assertion silently skipped.
- **Widened tolerance bands:** `TestAuthErrorOnReconnect` exact `!= 2` → range 2..4 while its own comment says the new value is deterministically 3; `TestProperFalloutAfterMaxAttemptsWithAuthMismatch` upper bound doubled (`MaxReconnect` → `len(servers)*MaxReconnect`).
- **CI structural hole:** only `scripts/cov.sh:8-11` hard-fails on unset `TESTER_NATS_URL`; the raw `go test` steps (`ci.yaml:81, :89`) rely purely on the job-level `env:` block (`ci.yaml:65-67`). Delete/typo that block and CI greens on unit tests alone. Same gap in `latest-server.yaml:46, :51`. **Cheap fix:** helper does `t.Fatal` instead of `t.Skip` when `os.Getenv("CI") != ""`.
- **`latest-server.yaml` semantic regression:** was `go get nats-server/v2@main` built at run time; now consumes `synadia/ntf-server:nightly-latest-nats-main`. If upstream image publishing stalls, nightly silently tests a stale server and stays green (version-log step is `|| true`'d, asserts nothing).
- **Dependency split:** `go_test.mod` bumps `klauspost/compress` 1.18.5→1.19.1 and `nkeys` 0.4.15→0.4.16; `go.mod` stays older. `compress` is live production code (websocket compression) — CI no longer exercises the versions released users get. Likely MVS-driven by ntf-client; align or note.
- **`dialInstance` blanket `nats.MaxReconnects(-1)`** (default was 60): genuine reconnect failure now hangs to test timeout instead of surfacing `ErrMaxReconnects`. Overridable (later options win) and overridden where it matters (`pull_test.go:3341, :3370`), but it's a silent behavior flip across three packages.
- **New unguarded `time.Sleep(500ms)`** in `TestNewServers` ("let the two nodes settle") with nothing asserting the settled state — the only sleep-count increase in the PR.
- **`checkout` before `apk add git`** (`ci.yaml:70` vs `:74`): `golang:*-alpine` has no git, so checkout falls back to the REST API — no `.git` in workspace. Works today (Coveralls uses env vars); breaks anything needing git history. Move apk ahead of checkout.

## Flake outlook

Better than feared. **No timing bound was numerically tightened anywhere** (full diff of upper-bound assertions vs main: identical). CI runs on the docker network (no docker-proxy). Worst-case local runs passed everything, including the 100ms `TestBasicClusterReconnect` bound at 0.01s actual. Residual risk concentrates in:

1. Post-restart publishes with no readiness wait (H2 sites) — the mechanism `pull_test.go:2837` already guards against.
2. `TestReconnectWaitJitter` (`test/nats_test.go:843-849`): remote server spawn sits *inside* a randomized 800ms window; failure is probabilistic by construction, so 5 green runs is weak evidence.
3. The 10ms post-restart sleeps (`pull_test.go:1494, :2330`, `push_test.go:237`) calibrated against a restart that used to be complete on return.
4. `TestCustomReconnectDelay`: ~100ms slack across 4 failed connects (passed ×5 locally at 0.43-0.47s).
5. `TestTimeoutOnNoServers`: 1.5s ceiling with `opts.Timeout` left at the 2s default — one hanging connect blows it alone.

Benchmarks assert no throughput numbers and CI never runs them — zero CI risk; historical loopback comparisons are dead, and `BenchmarkPublishWithWriteBufferSize` now spawns 21 server processes (slow by hand).

## Clean bill

- **Scary line counts are innocent.** pull_test −800 / ordered_test −480 / js_test −1100 reconcile **exactly** to deleted `nats.Connect`/`jetstream.New` boilerplate and already-skipped dead bodies. Zero real assertions lost in any jetstream/micro file. All 4 "inherited" skips verified present in main. (Caveat: the PR deleted the bodies below 4 pre-existing skips, so re-enabling those means rewriting from scratch.)
- **Function/subtest preservation near-total:** 3 function removals, all with rationale comments (one rationale wrong — H3.4); 3 subtests dropped, all inside an already-skipped test; every other subtest list byte-identical.
- **Cleanup ordering correct.** LIFO gives `nc.Close` → `inst.Destroy` → `client.Close` — required since `Destroy` sends a request over the client's connection. The `t.Fatalf`-inside-`t.Cleanup` edge is safe (Go's `runCleanup` re-enters to run remaining cleanups). Create/Destroy pairing balances in every file.
- **Port stability across restart is real** (proven by upstream ntf-client tests: same URL reconnects after stop/start) though undocumented, and `StartServerResponse` carries no address — so it's load-bearing and invisible in the type system.
- **Two removed version-guard skips are strengthening**, not drift (`serverVersionAtLeast` helpers died with the linked-in server; image pins nats 2.14.3).

## Nits (batch commit)

- `CLAUDE.md:47` false: `make test-tester` *does* need `make tester-up` first (`TESTING.md` has it right). `CLAUDE.md:138` false: `configs/` is not mounted into the tester. `CLAUDE.md:32` example dropped `-tags=internal_testing` → doc-following `-run` of internal tests silently runs nothing.
- `dependencies.md` stale: still lists `nats-server/v2`, omits `ntf-client` (license attribution file).
- `scripts/cov.sh:24, :27` runs `./jetstream/...` and `./micro/...` without `-p=1`, contradicting the documented invariant; benign today, flaky the day a tester-using test lands outside `*/test`.
- `Makefile:24` sed breaks (two-word image) if a second `synadia/ntf-server:` line ever appears in ci.yaml; `Makefile:30` `GO_IMAGE` hardcoded while `TESTER_IMAGE` is parsed — will drift; `tester-up*` targets have no readiness wait (first `make test` right after can hard-fail the probe); `tester-up-host` publishes 1001 ports.
- 11 of 14 files in `test/configs/` now orphaned (the `.conf` files are unloadable nats-server configs; only `certs/{ca,key,server}.pem` still read by `test/conn_test.go`).
- `TestAccountInfo` (`test/js_test.go`): dead `skip string` table field — declared, checked, never set.
- `WithConnectOptions` is a documented silent no-op on the exact path all three helpers use (`CreateServer` + self-dial) — live trap; note it in the helper doc comments.
- Helper triplication (`newTester`/`testerProbe`/`dialInstance` ×3 packages) — forced by Go test-package rules, but the copies must stay in sync; `jetstream`/`micro` variants take `*testing.T` while `test/` takes `testing.TB`.
- `WaitForJetStream` doc says "fails after 5 seconds"; actual budget ~15s.

---

*Review conducted 2026-08-07 against `pr-2082-review` (fetch of `pull/2082/head`, head `6ea8b10`), baseline `main` @ `9d92e85`.*

piotrpio:
Addressed the comment:
- H1 - `withJSCluster` now polls `AccountInfo` after `WaitForJetStream`
- H2 - added a shared `waitForStream` helper
- H3.1 - restored the `ErrMaxAccountConnectionsExceeded` assertion, it was asserted nowhere on the branch.
- H3.4 - the comment was wrong - corrected. `TestRetryOnFailedConnectWithAuthError`
  caps `MaxReconnects` at 2, so it can't distinguish auth-abort from max retries. Now stated as a gap rather than equivalent coverage.

Deliberately not changed: H3.2 (JWT `UpdateAccountClaims` —
needs a tester API that doesn't and probably should not exist), H3.3, H3.5 (DNS-shuffle path), H3.6
(`ConnectedClusterName`, in-process conn).

Jarema (approved):
LGTM!

PR #2084, opened 2026-05-25
docs: document HTTP header name requirement for Msg.Header

### LINKED ISSUE #1276: Client encoding headers using http.Header.Write discards valid headers under ADR-4
## Defect

- nats.go -- 1.26.0
- nats-server -- 2.9.17 (should be reproducible under other versions since this is a client-side bug)

I've run this under both macOS and Linux, but the environment should be N/A.

1. Create a message.
2. Set a header on the message with a field-name that would be valid under ADR-4 but not a valid HTTP header. For example, `thing/with/slash` would comply with ADR-4 and RFC-822 but [is invalid for HTTP](https://go.dev/play/p/OcPNN2U2O9q).
3. Publish the message using the NATS Go client
    a. This shouldn't affect other clients like the Rust one where header validation / filtering doesn't happen.
5. Receive the message and it should be missing the header from step 2.

This can also be observed with a small test program (requires a running NATS server to test with -- in this case I'm using 2.9.17 but should work with earlier): https://gist.github.com/nilium/816538609069c235da697988717f3b1f

Additional context in the thread with @wallyqs on the NATS Slack: https://natsio.slack.com/archives/C069GSYFP/p1685066458211899

#### Expected result:

The header that complies with ADR-4 but not HTTP should be sent via NATS.

#### Actual result:

Headers that comply with ADR-4 but are invalid under HTTP are filtered client-side and not sent to NATS. This is due to [the client using http.Header's Write method](https://github.com/nats-io/nats.go/blob/main/nats.go#LL711C20-L711C20) to encode headers for the message payload, which intentionally discards headers that aren't valid for HTTP.

---

Sorry for submitting this later than I originally intended. Meant to get this written on Friday, but things got busier than I expected. Thankfully, I don't think this is urgent either way, so I doubt anyone lost sleep over it.

From discussion on the NATS Slack and suggested by @wallyqs, this probably just needs a note in the godocs for now to mention this as a minor compatibility problem. Since nobody has run into it before, at least that I'm aware, it just doesn't strike me as an urgent problem.

wyf027:
## Summary

- Documents that `Header` encoding uses `http.Header.Write`, which only emits HTTP-valid field names
- Notes that ADR-4-valid names with non-HTTP characters (e.g. `/`) are stored locally but omitted on publish
- Adds `TestHeaderNonHTTPFieldNameOmitted` to lock in the current wire behavior

Fixes #1276

## Context

Per discussion on the issue and NATS Slack, the immediate fix is documentation rather than changing the wire encoding format. Callers who need headers to survive a round trip should use HTTP-compatible field names.

## Test plan

- [x] `go test -run 'TestHeaderNonHTTPFieldNameOmitted|TestHeaderMultiLine' -v`

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -4230,12 +4230,22 @@ func (nc *Conn) Publish(subj string, data []byte) error {
 	return nc.publish(subj, _EMPTY_, false, nil, data)
 }
 
-// Header represents the optional Header for a NATS message,
-// based on the implementation of http.Header.
+// Header represents the optional Header for a NATS message.
+//
+// Header is a map-based type modeled after net/http.Header. When messages are
+// encoded for the wire, headers are serialized via http.Header.Write, which only
+// emits field names valid for HTTP (RFC 7230 token syntax). Field names that are
+// valid under NATS ADR-4 but invalid for HTTP—such as those containing '/'—are
+// stored in the map but silently omitted from the encoded message. Use
+// HTTP-compatible header names (e.g. "Thing-With-Slash" instead of
+// "thing/with/slash") when publishing headers that must survive a round trip.
 type Header map[string][]string
 
 // Add adds the key, value pair to the header. It is case-sensitive
 // and appends to any existing values associated with key.
+//
+// Keys must be valid HTTP header field names to be included when the message
+// is published; see Header for details.
 func (h Header) Add(key, value string) {
 	h[key] = append(h[key], value)
 }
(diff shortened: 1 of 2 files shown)


piotrpio:
After merging #2083, this PR is no longer valid. #2083, in addition to performance enhancements, removed `http.Header.Write` and now invalid headers are not dropped, but rather return an error. So as it stands the comment is invalid and the added test fails.

Since #2083 also improved the documentation, I'll be closing this PR.

PR #2085, opened 2026-05-25
fix: wrap invalid server INFO parse errors with context

### LINKED ISSUE #1103: Failed initial INFO message from server is hard to understand
## Defect

When connecting to a public NATS server from behind an WAF  (web application firewall) and through TLS, the unencrypted INFO initial message can be caught by the WAF, or truncated. 

In this case the `(*Conn).readOp` in  `(*Conn).processExpectedInfo()`  may read an empty buffer, later on leading to json.Unmarshal in `(*Conn)processInfo(c.args)` to return and error that simply reads "EOF".

In these cases it is difficult to diagnose what happens, especially since the error is returned "naked" all the way through 

https://github.com/nats-io/nats.go/blob/543e628f4bafe989ce2ba548af09e2067c67ea69/nats.go#L3221

https://github.com/nats-io/nats.go/blob/543e628f4bafe989ce2ba548af09e2067c67ea69/nats.go#L2220

https://github.com/nats-io/nats.go/blob/543e628f4bafe989ce2ba548af09e2067c67ea69/nats.go#L2088

https://github.com/nats-io/nats.go/blob/543e628f4bafe989ce2ba548af09e2067c67ea69/nats.go#L2135

https://github.com/nats-io/nats.go/blob/543e628f4bafe989ce2ba548af09e2067c67ea69/nats.go#L2175

https://github.com/nats-io/nats.go/blob/543e628f4bafe989ce2ba548af09e2067c67ea69/nats.go#L1369

https://github.com/nats-io/nats.go/blob/543e628f4bafe989ce2ba548af09e2067c67ea69/nats.go#L784

 - [X] Included nats.go version
 - [ ] Included a [Minimal, Complete, and Verifiable example] (https://stackoverflow.com/help/mcve)

nats.go 1.17.0

I honestly don't know how to reproduce it without a WAF. 

#### Expected result:

A transmission error stating the received initial INFO message from the server is invalid

#### Actual result:

"nats: error: EOF"

unclear-error

#####Side note :
the WAF may have been triggered because the NATS system we are trying to reach is exposed on port 443, in a (not so smart) attempt to use a well-known port, so that firewall would likely let the connection pass though.

wyf027:
## Summary

Fixes #1103

When the initial INFO payload is truncated or malformed (for example behind a WAF on TLS port 443), `json.Unmarshal` returned a bare `EOF` or syntax error that was difficult to diagnose.

This change wraps INFO parse failures with `nats: invalid INFO message from server: ...` so callers can distinguish protocol/handshake issues from other connection errors.

## Test plan

- [x] `go test -run TestProcessInfoInvalidMessage -v`
- [ ] Added tests for truncated JSON and whitespace-only INFO payloads

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -4052,7 +4052,7 @@ func (nc *Conn) processInfo(info string) error {
 	}
 	var ncInfo ServerInfo
 	if err := json.Unmarshal([]byte(info), &ncInfo); err != nil {
-		return err
+		return fmt.Errorf("nats: invalid INFO message from server: %w", err)
 	}
 
 	// Copy content into connection's info structure.
--- a/nats_test.go
+++ b/nats_test.go
@@ -1244,6 +1244,30 @@ func TestNoEchoOldServer(t *testing.T) {
 	}
 }
 
+func TestProcessInfoInvalidMessage(t *testing.T) {
+	nc := &Conn{}
+
+	tests := []struct {
+		name string
+		info string
+	}{
+		{name: "empty json object fragment", info: "{"},
+		{name: "whitespace only", info: "   "},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			err := nc.processInfo(test.info)
+			if err == nil {
+				t.Fatal("expected error")
+			}
+			if !strings.Contains(err.Error(), "invalid INFO message from server") {
+				t.Fatalf("unexpected error: %v", err)
+			}
+		})
+	}
+}
+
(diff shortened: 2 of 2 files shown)


piotrpio:
I don't believe this actually fixes #1103 - The issue was about the client returning EOF when WAF chops the socket. EOF is returned from reader, not at the point of message deserialization.

Additionally, the client does not wrap unmarshalling errors elsewhere, so this seems to be too targeted (especially since it does not fix the linked issue).

PR #2089, opened 2026-05-25
fix: populate object watch ModTime when metadata timestamp is zero

### LINKED ISSUE #2053: Object store mtime is alsways '0001-01-01T00:00:00Z'
### Observed behavior

When doing a watch or requesting info about an object from the object store, the mtime is always set to '0001-01-01T00:00:00Z'.

Example (typescript):
```typescript
import { connect } from "@nats-io/transport-node";

import { jetstream } from "@nats-io/jetstream";
import { Objm } from "@nats-io/obj";

async function main() {
    try {
        const nc = await connect({ servers: "localhost" });
        const js = jetstream(nc, { timeout: 10000 });
        const objm = new Objm(js);
        const os = await objm.open("sample");

        const watcher = await os.watch({ includeHistory: true });

        for await (const entry of watcher) {
            console.log(entry);
        }

        await nc.close();
    } catch (err) {
        console.log(`Some error: ${err}`);
    }
}

await main();
```

The output is:
```
{
  name: 'something-one',
  options: { max_chunk_size: 131072 },
  bucket: 'sample',
  nuid: 'oWRfe1FNB5IykFUFvvpFTy',
  size: 542850,
  mtime: '0001-01-01T00:00:00Z',
  chunks: 5,
  digest: 'SHA-256=mT0zapKIRNt6BNC1i3KDVjzgvhm3clbgG-GcQbxyjUQ=',
  isUpdate: false
}
{
  name: 'something-twp',
  options: { max_chunk_size: 131072 },
  bucket: 'sample',
  nuid: '7OaaaiDsfg9sFjlOAobPWK',
  size: 5971143,
  mtime: '0001-01-01T00:00:00Z',
  chunks: 46,
  digest: 'SHA-256=GdBhynQjYs48FVOghB7BvpwfDYbWo5eZVpv_ONoXZL4=',
  isUpdate: true
}
```

### Expected behavior

I expect to get a proper modification time in the result.

### Server and client version

Server: 2.12.4
@nats-io/nats-core: 3.3.1
@nats-io/obj: 3.3.1
@nats-io/transport-node: 3.3.1

### Host environment

_No response_

### Steps to reproduce

_No response_

wyf027:
## Summary

Object store metadata is published with a zero `mtime` JSON field; callers are expected to derive modification time from the JetStream message timestamp.

`GetInfo` already uses the stream message time from a direct get, but `Watch` only copied `MsgMetadata.Timestamp`, which can be zero on some ordered-consumer deliveries. That surfaced as `0001-01-01T00:00:00Z` mtime values in watch results (#2053).

When metadata timestamp is zero, fall back to a stream message lookup by sequence (same source as `GetInfo`).

Fixes #2053

## Test plan

- [x] `go build ./jetstream/...`
- [x] Added `TestObjectWatch/watch modtime` integration test (requires nats-server in CI)

### DIFF
--- a/jetstream/object.go
+++ b/jetstream/object.go
@@ -997,6 +997,23 @@ func publishMeta(ctx context.Context, info *ObjectInfo, js *jetStream) error {
 	return nil
 }
 
+func objectModTimeFromMeta(ctx context.Context, stream Stream, meta *nats.MsgMetadata) time.Time {
+	if meta == nil {
+		return time.Time{}
+	}
+	if !meta.Timestamp.IsZero() {
+		return meta.Timestamp
+	}
+	if meta.Sequence.Stream == 0 {
+		return time.Time{}
+	}
+	rm, err := stream.GetMsg(ctx, meta.Sequence.Stream)
+	if err != nil || rm == nil || rm.Time.IsZero() {
+		return time.Time{}
+	}
+	return rm.Time
+}
+
 // AddLink will add a link to another object if it's not deleted and not another link
 // name is the name of this link object
 // obj is what is being linked too
@@ -1300,7 +1317,7 @@ func (obs *obs) Watch(ctx context.Context, opts ...WatchOpt) (ObjectWatcher, err
 		}
 
 		if !o.ignoreDeletes || !info.Deleted {
-			info.ModTime = meta.Timestamp
+			info.ModTime = objectModTimeFromMeta(ctx, obs.stream, meta)
 			w.updates <- &info
 		}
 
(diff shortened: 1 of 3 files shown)


piotrpio:
This PR is built on an incorrect premise, its test doesn't exercise the new code, and it doesn't touch the actual cause of #2053. 

When you're saying:
> GetInfo already uses the stream message time from a direct get, but Watch only copied MsgMetadata.Timestamp, which can be zero on some ordered-consumer deliveries.

`MsgMetadata.Timestamp` can never be zero (it's parsed from `$JS.ACK` subject and returns an error if parsing is not successful) so this is a totally made up.

PR #2092, opened 2026-05-27
fix: websocket connection with path

joeriddles:
Fix WebSocket connections when the URL has a path

> [!IMPORTANT]  
> I also opened a PR to fix this solely in `natscli`, https://github.com/nats-io/natscli/pull/1672. I think the pure-CLI change is the more correct fix.

```shell
# before
❯ nats -s 'ws://localhost:8080/api/cp-nats/' pub 'xyz' test
nats: error: invalid websocket connection

# after
❯ nats -s 'ws://localhost:8080/api/cp-nats/' pub 'xyz' test
10:22:37 Published 4 bytes to "xyz"
````

### DIFF
--- a/nats.go
+++ b/nats.go
@@ -1561,8 +1561,9 @@ func Compression(enabled bool) Option {
 // ProxyPath sets a path added to every WebSocket connection URL. This is
 // useful when NATS is behind a reverse proxy that routes by path. Unlike a
 // path in the connect URL, ProxyPath is also applied to server-discovered
-// URLs (which are bare host:port), so reconnects through the proxy still
-// work. When set, it overrides any path in the connection URL.
+// URLs (which are bare host:port). Use ProxyPath for clustered setups
+// behind a proxy, so the client can connect to discovered servers on
+// failover. When set, it overrides any path in the connection URL.
 func ProxyPath(path string) Option {
 	return func(o *Options) error {
 		o.ProxyPath = path
--- a/nats.go
+++ b/nats.go
@@ -1558,8 +1558,11 @@ func Compression(enabled bool) Option {
 	}
 }
 
-// ProxyPath is an option for websocket connections that adds a path to connections url.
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by piotrpio** [comment]
It would be nice to add a doc on `nats.ProxyPath()` mentioning that it overwrites user-provided path.

Also, I agree with the bot suggestion about `RawQuery`. Other than that looks good!

**Comment by joeriddles** [comment]
@piotrpio I addressed claude's suggestion about query params and updated the ProxyPath docstring to clarify its usage a bit.

**Comment by piotrpio** [comment]
@claude re-review this PR

File: nats.go [hunk 1]
```diff
@@ -1558,8 +1558,11 @@ func Compression(enabled bool) Option {
 	}
 }
 
-// ProxyPath is an option for websocket connections that adds a path to connections url.
-// This is useful when connecting to NATS behind a proxy.
+// ProxyPath sets a path added to every WebSocket connection URL. This is
+// useful when NATS is behind a reverse proxy that routes by path. Unlike a
+// path in the connect URL, ProxyPath is also applied to server-discovered
+// URLs (which are bare host:port), so reconnects through the proxy still
+// work. When set, it overrides any path in the connection URL.
```
piotrpio: Now it reads as if reconnects don't work in general when pathc is supplied in the connect URL, but it only doesn't to discovered servers. Maybe we could improve with something like this:

```suggestion
// ProxyPath sets a path added to every WebSocket connection URL. This is
// useful when NATS is behind a reverse proxy that routes by path. Unlike a
// path in the connect URL, ProxyPath is also applied to server-discovered
// URLs (which are bare host:port). Use ProxyPath for clustered setups
// behind a proxy, so the client can connect to discovered servers on
// failover. When set, it overrides any path in the connection URL.
```

piotrpio (approved):
LGTM!

PR #2093, opened 2026-05-27
iter: don't yield a phantom (nil, nil) after MsgsTimeout's ErrTimeout

c-tonneslan:
Spotted while reading nats_iter.go. After NextMsg returns ErrTimeout, MsgsTimeout yields (nil, err), honors the caller's keep-going signal, and then falls through to `if !yield(msg, nil)` with `msg` still nil. So a `for msg, err := range sub.MsgsTimeout(...)` loop that continues past a timeout (per the doc: "the iterator will return nats.ErrTimeout but it will not be closed") gets two yields per tick: one `(nil, ErrTimeout)` and one phantom `(nil, nil)`.

The existing TestSubscribeIterator/with_timeout test breaks on the first ErrTimeout so it never observed the phantom.

Fix is a one-line `continue` after the timeout-recovery branch. Added a regression test that pulls three iterations on a quiet subject and asserts each one was ErrTimeout, no phantoms; fails on master.

### DIFF
--- a/nats_iter.go
+++ b/nats_iter.go
@@ -63,6 +63,7 @@ func (sub *Subscription) MsgsTimeout(timeout time.Duration) iter.Seq2[*Msg, erro
 				if !errors.Is(err, ErrTimeout) {
 					return
 				}
+				continue
 			}
 			if !yield(msg, nil) {
 				return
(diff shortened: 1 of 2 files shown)


piotrpio (approved):

PR #2096, opened 2026-06-09
Fix ObjectInfo.ModTime parity across clients - store mtime in metadata JSON so that the time isn't zeroed out in the JSON

### LINKED ISSUE #1252: Parity issue with `ObjectInfo.ModTime`
`ObjectInfo.ModTime` isn't actually returned from info, which causes a bit of parity issue with at-least the deno (based on a quick chat with @aricart and grok of the sources) and rust clients.

Instead `ModTime` gets assigned to the raw message time here:

https://github.com/nats-io/nats.go/blob/2857164a10901305a3fc559d9539e849ad5f5526/object.go#L930

When we read back from something that has been `put` with the Go client (including nats-cli), we end up reading back a `ModTime` of "0001-01-01T00:00:00Z".

```json
{
  "name": "object",
  "options": {
    "max_chunk_size": 131072
  },
  "bucket": "store",
  "nuid": "2TZS1wGF9T8XKxBp16fpxY",
  "size": 4,
  "mtime": "0001-01-01T00:00:00Z",
  "chunks": 1,
  "digest": "SHA-256=Om6weQ85rIfJTzhWst0sXREOaBFgImGpqSPTuyOtyLc="
}
```

Example that populated this:

```go
import (
	"fmt"
	"github.com/nats-io/nats.go"
)

func main() {
	nc, _ := nats.Connect("localhost:4222")
	defer nc.Drain()

	js, _ := nc.JetStream()
	os, _ := js.ObjectStore("store")

	info, _ := os.PutString("object", "data")
	fmt.Printf("mtime is %s", info.ModTime)
}
```

Prashant-koi:
Fixes #1252

## Problem
The Go client was storing `"mtime":"0001-01-01T00:00:00Z"` (zero time) in the object metadata JSON because `ModTime` was set after `json.Marshal`. The Go client worked around this on reads by using the server-assigned `m.Time` from the JetStream message metadata, but other clients (Rust, Deno, etc.) read `mtime` directly from the JSON and always got zero time. I did go ahead and confirm that the Rust client was indeed directly reading from the JSON.

## Changes
- Set `ModTime = time.Now().UTC()` before `json.Marshal` in `Put` and `publishMeta` in both legacy (`object.go`) and new JetStream API (`jetstream/object.go`)
- This ensures `mtime` is stored with a real timestamp in the metadata JSON, fixing cross-client parity

Signed-off-by: Prasant Koirala <koiralaprashanta10@gmail.com>

### DIFF
--- a/jetstream/object.go
+++ b/jetstream/object.go
@@ -763,6 +763,9 @@ func (obs *obs) Put(ctx context.Context, meta ObjectMeta, r io.Reader) (*ObjectI
 		}
 	}
 
+	// Set ModTime before marshaling so it is stored in the metadata and readable by other clients.
+	info.ModTime = time.Now().UTC()
+
 	// Prepare the meta message
 	metaSubj := fmt.Sprintf(objMetaPreTmpl, obs.name, encodeName(meta.Name))
 	mm := nats.NewMsg(metaSubj)
@@ -796,8 +799,6 @@ func (obs *obs) Put(ctx context.Context, meta ObjectMeta, r io.Reader) (*ObjectI
 	case <-ctx.Done():
 		return nil, nats.ErrTimeout
 	}
-
-	info.ModTime = time.Now().UTC() // This time is not actually the correct time
 
 	// Delete any original chunks.
 	if einfo != nil && !einfo.Deleted {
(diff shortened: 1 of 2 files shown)


piotrpio:
Thanks for creating the PR @Prashant-koi. However, `ObjectInfo.ModTIme` returned to the user should not be populated form the json, but rather from msg timestamp metadata - see [ADR-20 - Object Info Strage](https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-20.md#objectinfo-storage).

Looking at that, the bug here in nats.go is that `mtime` is written at all - but if we were to remove it we have to first make sure it won't break other clients.

As it stands, the fix should be applied in rust client so that it properly populates ModTime with msg timestamp.

PR #2098, opened 2026-06-16
[FIXED] Recognize error code 10164 for replicated KV CAS conflicts (#2097)

### LINKED ISSUE #2097: KeyValue.Update/Create returns "unknown" 10164 error code in case of bucket replicas > 1
### Observed behavior

When a KV bucket has Replicas > 1, concurrent Create/Update operations against a single key may fail with error code 10164 (JSStreamWrongLastSequenceConstantErr from the server) instead of 10071.
Because the Go client only recognizes code 10071, these errors cannot be handled by errors.Is(err, jetstream.ErrKeyExists) or by known error code validation (which only supports the constant jetstream.JSErrCodeStreamWrongLastSequence / 10071).

A similar issue was reported in the Python client: [nats.py#782](https://github.com/nats-io/nats.py/issues/782). It would be good to properly support error code 10164 in the Go client as well.
This may be related to [nats-server#7281](https://github.com/nats-io/nats-server/issues/7281), but I'm not certain, so I opened this issue in the client repository.

### Expected behavior

CAS operation related error codes (10164, 10071) — indicate wrong last sequence on the server side. The client should be able to handle 10164 alongside 10071 and can retry the CAS operation gracefully

### Server and client version

Server: nats:2.14.2-alpine
Client: github.com/nats-io/nats.go v1.52.0

### Host environment

MacOS
Docker

### Steps to reproduce

Use [docker-compose.yml](https://github.com/user-attachments/files/28848068/docker-compose.yml) file to spin up nats cluster and run following script:
```
package main

import (
	"context"
	"crypto/rand"
	"encoding/hex"
	"errors"
	"fmt"
	"log/slog"
	"os/signal"
	"sync"
	"syscall"
	"time"

	"github.com/nats-io/nats.go"
)

const (
	KVReplicas = 2 // Set >1 to reproduce unexpected errors with concurrent operations
	LoopCount  = 10
)

var (
	urls = []string{
		"nats://localhost:4222",
		"nats://localhost:4223",
		"nats://localhost:4224",
	}
	kvCfg = jetstream.KeyValueConfig{
		Bucket:         "test-nats-kv-bucket",
		MaxBytes:       10000000, // 10MB
		LimitMarkerTTL: time.Minute,
		Replicas:       KVReplicas,
		Description:    "KeyValue for test",
	}
)

func main() {
	ctx, cancel := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
	defer cancel()

	jss := must(createJSs(urls))

	kvs := make([]jetstream.KeyValue, len(urls))
	for i, js := range jss {
		kvs[i] = must(js.CreateOrUpdateKeyValue(ctx, kvCfg))
	}

	wg := runCreate(ctx, kvs, LoopCount)
	wg.Wait()

	wg = runUpdate(ctx, kvs, LoopCount)
	wg.Wait()
}

func runCreate(ctx context.Context, kvs []jetstream.KeyValue, loopCount int) *sync.WaitGroup {
	type create struct {
		key   string
		value []byte
	}

	createCh := make(chan create, len(kvs))
	defer close(createCh)

	createFn := func(ctx context.Context, kv jetstream.KeyValue) {
		for {
			select {
			case <-ctx.Done():
				return
			case c, ok := <-createCh:
				if !ok {
					return
				}

				if _, err := kv.Create(ctx, c.key, c.value); err != nil {
					if errors.Is(err, jetstream.ErrKeyExists) {
						slog.Info("create status code", slog.Any("error", err))
					} else {
						slog.Error("create status code", slog.Any("error", err))
					}
				}
			}
		}
	}

	wg := &sync.WaitGroup{}

	for _, kv := range kvs {
		wg.Go(func() {
			createFn(ctx, kv)
		})
	}

	kvsCount := len(kvs)

	for range loopCount {
		select {
		case <-ctx.Done():
			slog.Info("create: context done")
			return wg
		case <-time.After(250 * time.Millisecond):
		}

		newValue := create{
			key:   randomString(6),
			value: mustRandomBytes(6),
		}

		for range kvsCount {
			createCh <- newValue
		}
	}

	return wg
}

func runUpdate(ctx context.Context, kvs []jetstream.KeyValue, loopCount int) *sync.WaitGroup {
	expectedKey := randomString(6)

	must(kvs[0].Create(ctx, expectedKey, []byte("start value")))

	type update struct {
		key   string
		value []byte
	}

	updateCh := make(chan update, len(kvs))
	defer close(updateCh)

	updateFn := func(ctx context.Context, kv jetstream.KeyValue) {
		for {
			select {
			case <-ctx.Done():
				return
			case u, ok := <-updateCh:
				if !ok {
					return
				}

				valueEntry := must(kv.Get(ctx, u.key))

				if _, err := kv.Update(ctx, u.key, u.value, valueEntry.Revision()); err != nil {
					apiError, casted := errors.AsType[*jetstream.APIError](err)
					if !casted {
						slog.Error("unexpected error", slog.Any("error", err))
						continue
					}

					errorCode := apiError.ErrorCode

					if errorCode == jetstream.JSErrCodeStreamWrongLastSequence {
						slog.Info("update status code", slog.Int("code", int(errorCode)))
					} else {
						slog.Error("update status code", slog.Int("code", int(errorCode)))
					}
				}
			}
		}
	}

	wg := &sync.WaitGroup{}

	for _, kv := range kvs {
		wg.Go(func() {
			updateFn(ctx, kv)
		})
	}

	kvsCount := len(kvs)

	for range loopCount {
		select {
		case <-ctx.Done():
			slog.Info("update: context done")
			return wg
		case <-time.After(250 * time.Millisecond):
		}

		newValue := update{
			key:   expectedKey,
			value: mustRandomBytes(6),
		}

		for range kvsCount {
			updateCh <- newValue
		}
	}

	return wg
}

func createJSs(urls []string) ([]jetstream.JetStream, error) {
	jss := make([]jetstream.JetStream, len(urls))
	for i, url := range urls {
		conn, err := newNatsConnect(url)
		if err != nil {
			return nil, fmt.Errorf("new nats connection for %q: %w", url, err)
		}

		stream, err := jetstream.New(conn)
		if err != nil {
			return nil, fmt.Errorf("new jetstream for %q: %w", url, err)
		}

		jss[i] = stream
	}

	return jss, nil
}

func newNatsConnect(url string) (*nats.Conn, error) {
	opts := []nats.Option{
		nats.MaxReconnects(5),
		nats.ReconnectWait(3 * time.Second),
		nats.Timeout(5 * time.Second),
	}

	return nats.Connect(url, opts...)
}

func randomString(n int) string {
	return hex.EncodeToString(mustRandomBytes(n))
}

func mustRandomBytes(n int) []byte {
	b := make([]byte, n)

	must(rand.Read(b))

	return b
}

func must[T any](val T, err any) T {
	if err != nil {
		panic(err)
	}

	return val
}
```

piotrpio:
Fixes #2097

### DIFF
--- a/jetstream/kv.go
+++ b/jetstream/kv.go
@@ -1075,14 +1075,29 @@ func (kv *kvs) Create(ctx context.Context, key string, value []byte, opts ...KVC
 		return kv.updateRevision(ctx, key, value, e.Revision(), o.ttl)
 	}
 
-	// Check if the expected last subject sequence is not zero which implies
-	// the key already exists.
-	if errors.Is(err, ErrKeyExists) {
-		jserr := ErrKeyExists.(*jsError)
-		return 0, fmt.Errorf("%w: %s", err, jserr.message)
+	// A wrong-last-sequence response means the key already exists.
+	if isWrongLastSeqErr(err) {
+		// 10071 matches ErrKeyExists via its code and keeps its original
+		// message; 10164 (replicated streams) must wrap ErrKeyExists explicitly.
+		if errors.Is(err, ErrKeyExists) {
+			jserr := ErrKeyExists.(*jsError)
+			return 0, fmt.Errorf("%w: %s", err, jserr.message)
+		}
+		return 0, fmt.Errorf("%w: %w", err, ErrKeyExists)
 	}
 
 	return 0, err
+}
+
+// isWrongLastSeqErr reports whether err is a "wrong last sequence" API error.
+// Replicated (R>1) streams report CAS conflicts as 10164 instead of 10071.
+func isWrongLastSeqErr(err error) bool {
+	var apiErr *APIError
+	if !errors.As(err, &apiErr) {
+		return false
+	}
+	return apiErr.ErrorCode == JSErrCodeStreamWrongLastSequence ||
+		apiErr.ErrorCode == JSErrCodeStreamWrongLastSequenceConstant
 }
 
 // Update will update the value if the latest revision matches.
--- a/jetstream/kv.go
+++ b/jetstream/kv.go
@@ -134,7 +134,8 @@ type (
 		// will not remove any previous revisions from the underlying stream.
 		//
 		// [LastRevision] option can be specified to only perform delete if the
-		// latest revision the provided one.
+		// latest revision matches the provided one; if it does not,
+		// ErrKeyRevisionMismatch is returned.
 		Delete(ctx context.Context, key string, opts ...KVDeleteOpt) error
 
 		// Purge will place a delete marker and remove all previous revisions.
@@ -143,7 +144,8 @@ type (
 		// previous revisions from the underlying streams.
 		//
 		// [LastRevision] option can be specified to only perform purge if the
-		// latest revision the provided one.
+		// latest revision matches the provided one; if it does not,
+		// ErrKeyRevisionMismatch is returned.
 		Purge(ctx context.Context, key string, opts ...KVDeleteOpt) error
 
 		// Watch for any updates to keys that match the keys argument which
@@ -1195,7 +1197,7 @@ func (kv *kvs) Delete(ctx context.Context, key string, opts ...KVDeleteOpt) erro
 	}
 
 	_, err := kv.js.PublishMsg(ctx, m, pubOpts...)
-	return err
+	return mapRevisionMismatch(err)
 }
 
 // Purge will place a delete marker and remove all previous revisions.
--- a/jetstream/kv.go
+++ b/jetstream/kv.go
@@ -123,7 +123,8 @@ type (
 		Create(ctx context.Context, key string, value []byte, opts ...KVCreateOpt) (uint64, error)
 
 		// Update will update the value if the latest revision matches.
-		// If the provided revision is not the latest, Update will return an error.
+		// If the provided revision does not match the key's current revision,
+		// ErrKeyRevisionMismatch is returned.
 		// Update also resets the TTL associated with the key (if any).
 		Update(ctx context.Context, key string, value []byte, revision uint64) (uint64, error)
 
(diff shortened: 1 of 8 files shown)


piotrpio:
@claude the tests are now passing, please re-review

jetstream/kv.go
```diff
@@ -1075,16 +1075,31 @@ func (kv *kvs) Create(ctx context.Context, key string, value []byte, opts ...KVC
...
+		}
+		return 0, fmt.Errorf("%w: %w", err, ErrKeyExists)
 	}
 
 	return 0, err
 }
 
+// isWrongLastSeqErr reports whether err is a "wrong last sequence" API error.
+// Replicated (R>1) streams report CAS conflicts as 10164 instead of 10071.
+func isWrongLastSeqErr(err error) bool {
```
vdorokhin, line 1097: I think it might make sense to consider making this method public (or exposing a similar API). Right now, all client-side kv update code checks for specific error codes directly, and with the introduction of 10164 there will be yet another code to handle in case of WrongLastSequence type of errors. Like:
```
before:
	if err != nil {
		if natsErr, ok := errors.AsType[*jetstream.APIError](err); ok &&
			natsErr.ErrorCode == jetstream.JSErrCodeStreamWrongLastSequence {
			return ErrOutdatedRevision
		}

		return fmt.Errorf("kv update: %w", err)
	}
---
after:
	if err != nil {
		if natsErr, ok := errors.AsType[*jetstream.APIError](err); ok &&
			(natsErr.ErrorCode == jetstream.JSErrCodeStreamWrongLastSequence ||
				natsErr.ErrorCode == jetstream.JSErrCodeStreamWrongLastSequenceConstant) {
			return ErrOutdatedRevision
		}

		return fmt.Errorf("kv update: %w", err)
	}
```
It seems reasonable to provide a dedicated API that would allow determining whether an error is related to a WrongLastSequence errors

piotrpio: Thanks, good call on the growing code list. I went with a sentinel instead of exposing the predicate.

`KV.Update` now wraps both `10071` and `10164` in a new `ErrKeyRevisionMismatch` (`jetstream` and legacy), so you can drop the code checks:

```go
if errors.Is(err, jetstream.ErrKeyRevisionMismatch) {
    return ErrOutdatedRevision
}
```

The underlying `APIError` is still there, so it won't break anything. I opted for a sentinel rather than exported `isWrongLastSeqErr` since the package only uses sentinels plus `errors.Is`, and a public predicate would open the door to a whole family of `Is*` helpers.

vdorokhin: I actually like the idea of introducing a dedicated sentinel error even more. I agree that its the cleaner and more appropriate approach. Thanks!

Jarema (approved):
LGTM

PR #2099, opened 2026-06-24
[FIXED] MsgsTimeout iterator yields spurious (nil, nil) after a timeout

sueun-dev:
`MsgsTimeout` returns a range-over-func iterator. When `NextMsg` times out it
yields `(nil, ErrTimeout)`. `ErrTimeout` is not fatal, so a consumer is allowed
to keep iterating (e.g. to wait for more messages or do periodic work between
batches). In that case the loop falls through to `yield(msg, nil)` with
`msg == nil` and `err == nil`, so the consumer gets a spurious `(nil, nil)` pair
after every timeout. Code that treats `err == nil` as "got a message" then
operates on a nil `*Msg`.

The `Msgs()` iterator returns after any error and doesn't have this. The fix is
a `continue` after the non-fatal timeout branch so it loops back to `NextMsg`
instead of falling through.

The existing `with timeout` test didn't catch it because it breaks on the first
timeout, so the iterator returns before reaching the second yield. The added
test keeps iterating across timeouts:

```
# before the fix
--- FAIL: TestSubscribeIterator/timeout_keeps_yielding_only_ErrTimeout
    nats_iter_test.go:105: iteration 1: got msg=<nil> with nil error, expected ErrTimeout
# after
ok  github.com/nats-io/nats.go/test
```

`TestSubscribeIterator` and `TestQueueSubscribeIterator` pass with `-race`.

AI assistance disclosure: this fix and test were prepared with the help of an AI
coding assistant; I reviewed and verified the change and tests myself.

Signed-off-by: Sueun Cho <sueun.dev@gmail.com>

### DIFF
--- a/nats_iter.go
+++ b/nats_iter.go
@@ -63,6 +63,7 @@ func (sub *Subscription) MsgsTimeout(timeout time.Duration) iter.Seq2[*Msg, erro
 				if !errors.Is(err, ErrTimeout) {
 					return
 				}
+				continue
 			}
 			if !yield(msg, nil) {
 				return
--- a/test/nats_iter_test.go
+++ b/test/nats_iter_test.go
@@ -79,6 +79,41 @@ func TestSubscribeIterator(t *testing.T) {
 		}
 	})
 
+	t.Run("timeout keeps yielding only ErrTimeout", func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


piotrpio (approved):

PR #2105, opened 2026-07-13
[FIXED] micro: endpoint subject prefix over-match

vsaraikin:
## Bug

`matchEndpointSubject` (in `micro/service.go`, used by `matchSubscriptionSubject` to attribute an incoming subject to a registered endpoint) treats an endpoint subject that is a proper **prefix** of the literal subject as a match:

```go
matchEndpointSubject("foo", "foo.bar")          // true  (want false)
matchEndpointSubject("foo.bar", "foo.bar.baz")  // true  (want false)
matchEndpointSubject("foo.*", "foo.bar.baz")    // true  (want false)
```

This violates NATS subject-matching semantics: without a trailing `>`, `foo` matches only `foo`, and `foo.*` matches exactly one token after `foo`. The function only rejects endpoints that are *longer* than the subject (`len(endpointTokens) > len(subjectTokens)`); when the endpoint is a shorter prefix, the loop consumes all endpoint tokens and falls through to `return true`.

### Impact

A service with two endpoints that share a prefix — e.g. a wildcard endpoint `foo.>` and a literal endpoint `foo` — misattributes activity. When an async error arrives on `foo.bar.baz` (delivered via the `foo.>` subscription), `matchSubscriptionSubject` can match it against endpoint `foo` instead of `foo.>`, so the error count / `LastError` lands on the wrong endpoint's stats.

## Fix

After the token loop, require that the endpoint consumed every subject token (unless it ended with `>`, which returns early):

```go
return len(endpointTokens) == len(subjectTokens)
```

Exact matches, `*` single-token wildcards, and `>` full wildcards are unchanged; only the prefix-over-match cases now correctly return `false`.

## Testing

Added `TestMatchEndpointSubject` (white-box, `package micro`) covering exact, single/full wildcard, mismatch, endpoint-longer, and the prefix-over-match cases. The four over-match cases fail on the current code and pass with the fix. The full `./micro/...` suite (unit + integration, `-race`) passes; `gofmt`/`go vet` clean.

### DIFF
--- a/micro/service.go
+++ b/micro/service.go
@@ -646,7 +646,10 @@ func matchEndpointSubject(endpointSubject, literalSubject string) bool {
 			return false
 		}
 	}
-	return true
+	// Without a trailing ">", every subject token must be consumed; otherwise a
+	// shorter endpoint would over-match a longer subject (e.g. "foo" vs
+	// "foo.bar").
+	return len(endpointTokens) == len(subjectTokens)
 }
 
 // addVerbHandlers generates control handlers for a specific verb.
--- a/micro/service_internal_test.go
+++ b/micro/service_internal_test.go
@@ -1,0 +1,45 @@
+// Copyright 2026 The NATS Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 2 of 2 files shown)


piotrpio:
@claude review this PR

piotrpio (approved):

PR #2106, opened 2026-07-21
Add docs.nats.io examples to main

Jarema:
Moves the documentation examples from the `doc-examples`, `core-docs` and `jetstream-docs` branches into `examples/docs/` on main (70 example programs, additive union — no other changes).

Why: the new docs.nats.io build fetches these snippets at build time. On side branches they get zero CI (no workflow triggers on them) and can be broken by branch pruning. On main, `go vet ./...` / `go test ./...` compile them automatically — verified locally, gofmt-clean.

Note: please keep the three docs branches until the docs repo's fetch config is switched to `main` (follow-up PR there).

### DIFF
--- a/examples/docs/basics-publish/main.go
+++ b/examples/docs/basics-publish/main.go
@@ -1,0 +1,16 @@
+package main
+
+import "github.com/nats-io/nats.go"
+
+func main() {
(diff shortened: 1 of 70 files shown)


piotrpio (approved):
LGTM!

PR #2109, opened 2026-07-24
[ADDED] `WithPublishAsyncAckHandler` option for JetStream async publish

### LINKED ISSUE #2108: `WithPublishAsyncSuccessHandler` for JetStream Async Publish
### Proposed change

New type definitions would look like below in `publish.go`
```go
++	type MsgSuccessHandler func(js JetStream, mgs *nats.Msg, ack *PubAck)

type asyncPublisherOpts struct {
++	ascb MsgSuccessHandler
aecb MsgErrHandler
maxpa int
ackTimeout time.Duration
}
```

and `in jetstream_options.go`

```go
++ func WithPublishAsyncSuccessHandler(cb MsgSuccessHandler) JetStreamOpt {
++  return func(opts *JetStreamOptions) error {
++	opts.publisherOpts.ascb = cb
++	return nil
++  }
++ }
```

Implementation change in publish.go @ source (lines 575-580)
```go
// So here we have received a proper puback.
paf.ack = pa.PubAck
if paf.doneCh != nil {
  paf.doneCh <- paf.ack
}
++ cb := js.publisher.asyncPublisherOpts.ascb
js.publisher.Unlock()
++ if cb != nil {
++  cb(js, paf.msg, paf.ack)
++ }
```

This would eliminate extra work of tracking PubAckFuture's in user's buffered channels and makes it straight forward to instrument success scenarios for a very little minimal change. (less than 50+ diff)

### Use case

`WithPublishAsyncErrorHandler` is the only callback handler for `PublishAsync` and `PublishMsgAsync`, it fires exclusively on failure only. There is no equivalent of success, the only way to observe success is to check successful ack via that specific message's `PubAckFuture` 's OK case. Requiring caller to track and drain a channel per in-flight publish just to get success-side visibility (e.g. for latency metrics instrumentation from NATS client side)

I could still launch a listener goroutine and collect in-flight publishes, but this puts a pressure on listener goroutine's channel and latency gets time-drifts due to buffered channel to be passed around for a listener goroutine, so I believe easiest way to instrument async publishes would be through a success callback handler in the NATS client API. 

So I propose to add `WithPublishAsyncSuccessHandler` to NATS Go Client to be able to make instrumentation easier for success scenarios.

If more detail is needed or example is needed, please feel free to tag me! 😸 

### Contribution

yes, of course, very happy to do so

occamist:
resolves https://github.com/nats-io/nats.go/issues/2108

### DIFF
--- a/jetstream/jetstream.go
+++ b/jetstream/jetstream.go
@@ -465,6 +465,7 @@ var subjectRegexp = regexp.MustCompile(`^[^ >]*[>]?$`)
 // Available options:
 //   - [WithClientTrace] - enables request/response tracing.
 //   - [WithPublishAsyncErrHandler] - sets error handler for async message publish.
+//   - [WithPublishAsyncAckHandler] - sets ack handler for async message publish.
 //   - [WithPublishAsyncMaxPending] - sets the maximum outstanding async publishes
 //     that can be inflight at one time.
 func New(nc *nats.Conn, opts ...JetStreamOpt) (JetStream, error) {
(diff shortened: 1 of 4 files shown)


piotrpio:
@claude review this PR

occamist:
@piotrpio I think claude reviewer borked, can you re-run maybe?

piotrpio:
@occamist yeah, the workflow seems to be failing on PRs from forks, I'll fix it separately,

piotrpio (approved):

PR #2110, opened 2026-07-26
feat(jetstream): map JetStream error code 10054 (stream message exceeds maximum)

matehsan:
Add client-side mapping for JetStream API error code 10054, returned by
nats-server when a published message exceeds the stream's configured
MaxMsgSize.

- Add JSErrCodeStreamMessageExceedsMaximum (10054) to the ErrorCode enum
  in jetstream/errors.go
- Add ErrStreamMessageExceedsMaximum sentinel error wrapping the
  corresponding APIError
- Add test coverage in jetstream/test/errors_test.go verifying the
  error is returned end-to-end when publishing a message that exceeds
  a stream's MaxMsgSize, and that it matches via errors.Is/errors.As
  and direct APIError inspection

Ref: https://github.com/nats-io/nats-server/blob/main/server/jetstream_errors_generated.go#L564

### DIFF
--- a/jetstream/publish.go
+++ b/jetstream/publish.go
@@ -25,8 +25,9 @@ import (
 	"sync"
 	"time"
 
-	"github.com/nats-io/nats.go"
 	"github.com/nats-io/nuid"
+
+	"github.com/nats-io/nats.go"
 )
 
 type (
@@ -253,6 +254,9 @@ func (js *jetStream) PublishMsg(ctx context.Context, m *nats.Msg, opts ...Publis
 		return nil, ErrInvalidJSAck
 	}
 	if ackResp.Error != nil {
+		if ackResp.Error.ErrorCode == JSErrCodeStreamMessageExceedsMaximum {
+			return nil, ErrStreamMessageExceedsMaximum
+		}
 		return nil, fmt.Errorf("nats: %w", ackResp.Error)
 	}
 	if ackResp.PubAck == nil || ackResp.PubAck.Stream == "" {
--- a/jetstream/test/errors_test.go
+++ b/jetstream/test/errors_test.go
@@ -208,4 +208,101 @@ func TestJetStreamErrors(t *testing.T) {
 		}
 	})
 
+	t.Run("stream message exceeds maximum", func(t *testing.T) {
+		s := RunBasicJetStreamServer()
+		defer shutdownJSServerAndRemoveStorage(t, s)
+
+		nc, err := nats.Connect(s.ClientURL())
+		if err != nil {
+			t.Fatalf("Unexpected error: %v", err)
+		}
+
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+		js, err := jetstream.New(nc)
+		if err != nil {
+			t.Fatalf("Unexpected error: %v", err)
+		}
+		defer nc.Close()
+
+		_, err = js.CreateStream(ctx, jetstream.StreamConfig{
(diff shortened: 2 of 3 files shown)


piotrpio:
@claude review this PR

jetstream/publish.go
```diff
@@ -253,6 +254,9 @@ func (js *jetStream) PublishMsg(ctx context.Context, m *nats.Msg, opts ...Publis
 		return nil, ErrInvalidJSAck
 	}
 	if ackResp.Error != nil {
+		if ackResp.Error.ErrorCode == JSErrCodeStreamMessageExceedsMaximum {
+			return nil, ErrStreamMessageExceedsMaximum
+		}
```
piotrpio, line 259: Returning `ErrStreamMessageExceedsMaximum` explicitly here is not needed - as long as the error sentinel is added in `errors.go` as `APIError` with matching error code, it's enough for `errors.Is()` to match (which is how API errors should be matched, not by direct comparison), `errors.As()` will work as well. A doc in `errors.go` (and possibly on `Publish()`) stating that this error can be returned should suffice instead.

jetstream/test/errors_test.go
```diff
@@ -208,4 +208,101 @@ func TestJetStreamErrors(t *testing.T) {
...
+			t.Fatalf("Unexpected error: %v", err)
+		}
+
+		_, err = js.Publish(ctx, "maxsize.a", []byte("this payload is longer than 10 bytes"))
+		if err == nil {
+			t.Fatalf("Expected error, got nil")
+		}
+
+		// check directly to var (backwards compatible)
+		if err != jetstream.ErrStreamMessageExceedsMaximum {
```
piotrpio, line 243: The test does not have to pass direct comparison - the convention is to use `errors.Is()` across the client. Without this check you don't need to explicitly return the new error sentinel from `Publish()`.

jetstream/test/errors_test.go
```diff
@@ -208,4 +208,101 @@ func TestJetStreamErrors(t *testing.T) {
...
+		// matching via errors.Is
+		if ok := errors.Is(err, jetstream.ErrStreamMessageExceedsMaximum); !ok {
+			t.Fatal("Expected jetstream.ErrStreamMessageExceedsMaximum")
+		}
+
+		// matching wrapped via error.Is
+		err2 := fmt.Errorf("custom error: %w", jetstream.ErrStreamMessageExceedsMaximum)
+		if ok := errors.Is(err2, jetstream.ErrStreamMessageExceedsMaximum); !ok {
+			t.Fatal("Expected wrapped ErrStreamMessageExceedsMaximum")
+		}
```
piotrpio, line 256: I think this check is not necessary, it does not really check any client functionality, just go sdk behavior.

jetstream/test/errors_test.go
```diff
@@ -208,4 +208,101 @@ func TestJetStreamErrors(t *testing.T) {
...
+		if !ok {
+			t.Fatal("Expected a jetstream.JetStreamError")
+		}
+		expected := jetstream.JSErrCodeStreamMessageExceedsMaximum
+		if jserr.APIError().ErrorCode != expected {
+			t.Fatalf("Expected: %v, got: %v", expected, jserr.APIError().ErrorCode)
+		}
+		if jserr.APIError() == nil {
+			t.Fatal("Expected APIError")
+		}
```
piotrpio, line 269: Errors should be typecast via `errors.As()`, so the check below should suffice

piotrpio (changes requested):

PR #2111, opened 2026-07-28
[FIXED] Data race in resetOrderedConsumer when resets overlap

piotrpio:
`resetOrderedConsumer` mutated the `ConsumerConfig` through the shared pointer in `jsi.ccreq`. Mutate a value copy instead.


Jarema (approved):
LGTM

PR #2114, opened 2026-07-31
[ADDED] Per-publish callback option for async publishes

infastin:
Add `WithAsyncHandler` option allowing to set a callback for an individual `PublishAsync`/`PublishMsgAsync` call. The callback is invoked when the message is acknowledged by the server or when the publish fails, receiving the original message, the `PubAck` (nil on error) and the error (nil on success).

### Use case

I want to asynchronously publish a lot of messages, wait for ack (or error) and do some postprocessing for each message depending on whether it was successfully published or not.

Currently, you can call `PublishAsync` and either create a separate goroutine for each future or call `PublishAsyncComplete`. Both of those have some downsides.

### DIFF
--- a/jetstream/jetstream.go
+++ b/jetstream/jetstream.go
@@ -87,6 +87,9 @@ type (
 		// received by the server. It only guarantees that the message has been
 		// sent to the server and thus messages can be stored in the stream
 		// out of order in case of retries.
+		//
+		// A per-publish callback can be set with [WithAsyncHandler] to be
+		// notified when the publish is acknowledged or fails.
 		PublishAsync(subject string, payload []byte, opts ...PublishOpt) (PubAckFuture, error)
 
 		// PublishMsgAsync performs a publish to a stream and returns
@@ -98,6 +101,9 @@ type (
 		// received by the server. It only guarantees that the message has been
 		// sent to the server and thus messages can be stored in the stream
 		// out of order in case of retries.
+		//
+		// A per-publish callback can be set with [WithAsyncHandler] to be
+		// notified when the publish is acknowledged or fails.
 		PublishMsgAsync(msg *nats.Msg, opts ...PublishOpt) (PubAckFuture, error)
 
 		// PublishAsyncPending returns the number of async publishes outstanding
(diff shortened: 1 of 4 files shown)


piotrpio:
Thank you for creating the PR @infastin - I believe the feature you're proposing was merged into main a few days ago (#2109), so I'll be closing this one as a duplicate. New version of the client which will include `WithPublishAsyncAckHandler` option will be released soon.

infastin:
Thanks for looking into my PR @piotrpio. I've seen https://github.com/nats-io/nats.go/pull/2109, but this one is different. `WithAsyncHandler` attaches a callback to an individual publish, while `WithPublishAsyncAckHandler` attaches a single callback to a JetStream publisher (to all publishes). Does a user really need to create a separate JetStream publisher if all they want is to publish a single message with a callback?

piotrpio:
Thanks for clarifying — you're right, this is different from #2109, and I was wrong to close it as a duplicate. Sorry about that.

Having looked properly, I think the gap is narrower than it first appears. `WithPublishAsyncAckHandler` (from #2109) and `WithPublishAsyncErrHandler` already give you per-message notification: they fire once per publish, with no goroutine per message and in the order in which acks arrive (not in the order of publishes). So your original use case is already covered:
>I want to asynchronously publish a lot of messages, wait for ack (or error) and do some postprocessing for each message depending on whether it was successfully published or not.

As for your question:
>Does a user really need to create a separate JetStream publisher if all they want is to publish a single message with a callback?

When dealing with a single message (or just a handful of messages) - draining `PubAckFuture` would be the reasonable approach. Your original use case mentioned publishing a lot of messages  - which usually warrants a context-wide approach IMO.

The value added by your PR would be twofold:
1. Per-publish closure without the need to correlate a publish with an ack
2. A middle ground where you want to have an individual callback for many, but not all publishes.

I'm still not convinced it's worth a third way to learn a single publish outcome, though. We'd have the future, the context-wide handlers and a per-publish handler all firing for the same event.

infastin:
> When dealing with a single message (or just a handful of messages) - draining `PubAckFuture` would be the reasonable approach.

I agree. This is a bad case for the proposed feature. I meant that a user might want to publish a few messages without having to create a new publisher: the "middle ground" that you've mentioned.

> I'm still not convinced it's worth a third way to learn a single publish outcome, though. We'd have the future, the context-wide handlers and a per-publish handler all firing for the same event.

That's fine. I'll wait for the next release and use context-wide handlers for my case.

PR #2120, opened 2026-08-13
[FIXED] Typo in Publisher interface methods comment

sreeram77:
Fixed comment on method signature for PublishMsg and PublishMsgAsync which incorrectly mentions subject name as a param.


piotrpio (approved):

Issue #2068, opened 2026-05-05
Reduce overhead associated when publishing core NATS messages with headers


jonchammer:
### Proposed change

# Summary
`nats.go` demonstrates suboptimal performance when publishing Core Nats messages containing headers. In the worst case, observed throughput drops by about 85% compared to a similar situation where headers aren't in use. Even if accounting for the larger total message payload that headers necessarily require, the difference is still around 75%. See benchmarks in the appendices below.

While some amount of overhead is obviously going to be required when using headers, it looks like there might be a few potential optimization opportunities that could help to close that gap. Our application leverages headers extensively, so any optimizations that can be made to the Nats client in this scenario will have powerful downstream effects in our use case.

We've identified 3 potential improvements that can reduce the amount of overhead associated with publishing messages that use headers:
1. The current implementation stores the serialized representation of the headers in a `bytes.Buffer`. That buffer is not initialized with any fixed size, so it may allocate/reallocate multiple times with a single `Publish` call, depending on how many headers are present and how large they are. If the buffer size can be calculated ahead of time, we can reduce the number of required memory allocations per `Publish`.
2. The current implementation relies on `http.Header.Write`, which is potentially problematic for a couple of reasons. Firstly, it actually differs from [ADR-4](https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-4.md) in that the allowed character set (for both keys and values) differs from what's given in the spec[^1]. Secondly, it adds additional constraints to the implementation that are either unnecessary in this context (e.g. always writing keys in alphabetic order) or potentially confusing (silently dropping keys that contain invalid characters, rather than returning an error). Replacing `http.Header.Write` with our own implementation allows us much tighter control over both the behavior of how headers are written and the associated performance characteristics.
3. As mentioned in the first bullet point, headers are written to a `bytes.Buffer` that is allocated from scratch once per publish. Even if the buffer is sized properly and the serialization process is well-optimized, the code will still require a memory allocation for each message that is published. To work around that, we could adopt a pooling mechanism of some sort (e.g. leveraging a `sync.Pool`) to ensure that those buffer objects are only created when necessary and that they can be reused across multiple `Publish` calls.

[...]

PR #2125, opened 2026-08-29
Refactor headerBytes + headerSize with exact size calculation.

0xAozora:
`Size()` of `nats.Msg` calls `headerBytes()` which serializes the full header and then discards the bytes.
Since only the length is necessary, I extracted the validation and length count into a new function `headerSize()`.
While I don't think the naming is fitting, we can change it to something else if you have any suggestions.
I also moved the trimming of the trailing/leading whitespaces into `headerSize()` to get the exact size and avoid serializing the buffer.

It might also be a good first step towards removing the allocating a buffer for the headers mentioned in #2068 

Behavioral Change: Header Values are trimmed and replaced.


0xAozora:
I think I accidentally pushed an incomplete follow up improvement reserved for later here.

nats.go
```diff
@@ -855,37 +855,45 @@ func (m *Msg) Size() int {
...
 	for k, vs := range m.Header {
 		if !isHeaderKeyValid(k) {
-			return nil, ErrBadHeaderMsg
+			return 0, ErrBadHeaderMsg
 		}
-		for _, v := range vs {
-			hdrSize += len(k) + len(v) + 4 // 4 is for the colon, space, CR, and LF
+		for i, v := range vs {
+			vs[i] = textproto.TrimString(v)
+			hdrSize += len(k) + len(vs[i]) + 4 // 4 is for the colon, space, CR, and LF
```
piotrpio, line 883: This mutates user-supplied headers on `m`. Instead, you could just calculate the size with trimmed value - the actual published headers will be correctly normalized in `writeHeaderValue`.

```suggestion
			hdrSize += len(k) + len(textproto.TrimString(v)) + 4 // 4 is for the colon, space, CR, and LF
```

piotrpio (changes requested):

PR #2126, opened 2026-08-30
[FIXED] JetStream async publish can exceed WithPublishAsyncMaxPending

### LINKED ISSUE #1612: JetStream: the value for max pending async messages can be exceeded
### Observed behavior

When using the simplified JetStream client in Go, it is possible to exceed the limit of max pending async messages that is specified by the `WithPublishAsyncMaxPending` option.

### Expected behavior

The limit specified by `WithPublishAsyncMaxPending` should not be exceeded in practice.

### Server and client version

NATS server: 2.10.12
NATS Go client: 1.34.1

### Host environment

_No response_

### Steps to reproduce

The following code
```go
package main

import (
	"context"
	"fmt"
	"time"

	"github.com/nats-io/nats.go"
)

const (
	Subject = "foo.bar.baz"
	Stream  = "FOO"
)

func main() {
	ctx := context.Background()
	nc, err := nats.Connect(nats.DefaultURL)
	if err != nil {
		panic(err)
	}

	js, err := jetstream.New(
		nc,
		jetstream.WithPublishAsyncMaxPending(1),
	)
	if err != nil {
		panic(err)
	}

	_, _ = js.CreateStream(ctx, jetstream.StreamConfig{
		Name:     Stream,
		Subjects: []string{Subject},
	})
	defer js.DeleteStream(ctx, Stream)

	go func() {
		for {
			fmt.Printf("currently pending: %d\n", js.PublishAsyncPending())
		}
	}()

	for i := 0; i < 5; i++ {
		go func(i int) {
			_, err := js.PublishAsync(Subject, []byte{})
			if err != nil {
				fmt.Printf("%d  %v\n", i, err)
			}
		}(i)
	}

	time.Sleep(200 * time.Millisecond)
}
```
generates the following output for me:
```
currently pending: 0
currently pending: 4
currently pending: 4
currently pending: 4
2  nats: stalled with too many outstanding async published messages
3  nats: stalled with too many outstanding async published messages
0  nats: stalled with too many outstanding async published messages
1  nats: stalled with too many outstanding async published messages
currently pending: 0
```

ShravanthReddy:
## Bug

`registerPAF` inserted the pending ack, unlocked, then stalled if `len(acks)` was over `WithPublishAsyncMaxPending`. Concurrent `PublishAsync` callers could all insert before any of them waited, so `PublishAsyncPending()` could exceed the configured cap (issue #1612).

## Fix

Check the cap and insert under the same lock. If already at `maxPending`, wait on the stall channel (or `stallWait`) *before* inserting. Pending cannot exceed the cap.

## Testing

White-box `TestPublishAsyncMaxPendingNotExceeded` in `./jetstream` (no nats-server/v2, no Docker). 32 goroutines hit `registerPAF` with `maxPending=1` and assert pending never goes above 1.

```
go test ./jetstream -count=1 -timeout 60s -run TestPublishAsyncMaxPendingNotExceeded
```

`go.mod` is unchanged (no `nats-server/v2`). Nearby `./jetstream` unit tests pass without Docker.

Resolves #1612

Signed-off-by: Shravanth Reddy <shrvnthrdy@gmail.com>

### DIFF
--- a/jetstream/publish.go
+++ b/jetstream/publish.go
@@ -1,4 +1,4 @@
-// Copyright 2022-2025 The NATS Authors
+// Copyright 2022-2026 The NATS Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
@@ -360,15 +360,8 @@ func (js *jetStream) PublishMsgAsync(m *nats.Msg, opts ...PublishOpt) (PubAckFut
 		}
 		id = reply[js.opts.replyPrefixLen:]
 		paf = &pubAckFuture{msg: m, jsClient: js.publisher, maxRetries: o.retryAttempts, retryWait: o.retryWait, reply: reply}
-		numPending, maxPending := js.registerPAF(id, paf)
-
-		if maxPending > 0 && numPending > maxPending {
-			select {
-			case <-js.asyncStall():
-			case <-time.After(stallWait):
-				js.clearPAF(id)
-				return nil, ErrTooManyStalledMsgs
-			}
+		if err := js.registerPAF(id, paf, stallWait); err != nil {
+			return nil, err
 		}
 		if js.publisher.ackTimeout > 0 {
 			paf.timeout = time.AfterFunc(js.publisher.ackTimeout, func() {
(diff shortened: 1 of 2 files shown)


jetstream/publish.go
```diff
@@ -625,17 +618,43 @@ func (js *jetStream) resetPendingAcksOnReconnect() {
 	}
 }
```
piotrpio, line 594: This wipes acks but never closes stallCh, so a caller stalled in registerPAF isn't woken and sits out the full stallWait before returning ErrTooManyStalledMsgs with nothing pending. Same thing in clearPAF below. I'd add a small lock-held helper (close stallCh if non-nil and len(acks) < maxpa, set it to nil) and call it next to every delete from acks. handleAsyncReply and the ack timeout callback already do this inline so it dedupes those too.

jetstream/publish.go
```diff
@@ -653,16 +672,6 @@ func (js *jetStream) clearPAF(id string) {
 	js.publisher.Unlock()
```
piotrpio, line 675: Same here, a failed conn.PublishMsg frees the slot but nobody is signalled.

jetstream/publish.go
```diff
@@ -625,17 +618,43 @@ func (js *jetStream) resetPendingAcksOnReconnect() {
...
+				timer = time.NewTimer(stallWait)
+			}
+			select {
+			case <-stc:
+				continue
+			case <-timer.C:
+				return ErrTooManyStalledMsgs
+			}
+		}
+		js.publisher.acks[id] = paf
```
piotrpio, line 647: This re-enables as id-collision error we've fixed some time ago - `newAsyncReply` checks `acks[id]` before the wait, but the insert happens up to `stallWait` later, so a stalled caller can end up sharing an id with a later one. Draw the token under the same lock hold that inserts it (move the loop from `newAsyncReply` into the admit branch here and return the reply from `registerPAF`).

PR #2127, opened 2026-08-31
Prevent doErr closure escaping to heap

0xAozora:
In `handleAsyncReply` there is a closure `doErr` which escapes to the heap. I have introduced a `handleAsyncErrAndUnlock` function and call it instead of the closure.


piotrpio (approved):

0xAozora:
@piotrpio Do you know why the test failed?

piotrpio:
Yes, there was a bug in the test helper where we did not wait for stream leader. I merged it 2 days ago so it was not on your branch yet: https://github.com/nats-io/nats.go/pull/2141

PR #2129, opened 2026-09-02
Return ErrDisconnected when a request is interrupted by reconnect

### LINKED ISSUE #2128: Receiving `ErrConnectionClosed` from `RequestWithContext` during a reconnect
### Observed behavior

When a client reconnects to a NATS server and follows the CONNECTED -> RECONNECTING -> CONNECTED state transition, the call to `conn.RequestWithContext` - a function used during a JetStream publish - may return an `ErrConnectionClosed`, even if the connection is not actually closed.

### Expected behavior

If the connection is not closed, then `ErrConnectionClosed` should not be returned.

### Server and client version

server: `v2.14.6`
client: `v1.53.1`

### Host environment

Linux 7.1.9, x86_64, go 1.27.0

### Steps to reproduce

```go
package main

import (
	"context"
	"errors"
	natsserver "github.com/nats-io/nats-server/v2/server"
	nats "github.com/nats-io/nats.go"
	"log"
	"time"
)

func main() {
	ctx := context.Background()

	srv, err := natsserver.NewServer(&natsserver.Options{
		Host: "localhost",
		Port: 4222,
	})
	if err != nil {
		log.Fatal("cannot create embedded NATS server")
	}

	srv.Start()

	options := []nats.Option{
		nats.ConnectHandler(func(_ *nats.Conn) {
			log.Print("connected")
		}),
		nats.DisconnectHandler(func(_ *nats.Conn) {
			log.Print("disconnected")
		}),
		nats.ReconnectHandler(func(_ *nats.Conn) {
			log.Print("reconnected")
		}),
		nats.MaxReconnects(-1),
	}

	conn, err := nats.Connect(srv.ClientURL(), options...)
	if err != nil {
		log.Fatal("cannot create NATS connection")
	}

	go func() {
		for {
			time.Sleep(100 * time.Millisecond)
			log.Print("forcing a reconnect...")
			conn.ForceReconnect()
		}
	}()

	go func() {
		for {
			_, err := conn.RequestWithContext(ctx, "foo", nil)
			if errors.Is(err, nats.ErrConnectionClosed) {
				log.Fatal("received ErrConnectionClosed - this shouldn't happen")
			}
			time.Sleep(1 * time.Millisecond)
		}
	}()

	<-ctx.Done()
}
```

Sample output:
```
2026/09/01 10:42:49 connected
2026/09/01 10:42:49 forcing a reconnect...
2026/09/01 10:42:49 disconnected
2026/09/01 10:42:49 reconnected
2026/09/01 10:42:49 forcing a reconnect...
2026/09/01 10:42:49 disconnected
2026/09/01 10:42:49 reconnected
2026/09/01 10:42:49 forcing a reconnect...
2026/09/01 10:42:49 received ErrConnectionClosed - this shouldn't happen
exit status 1
```

AshSgDe29071999:
ForceReconnect / reconnect handling closes pending request channels. `RequestWithContext` (and `newRequest`) treated a closed channel as `ErrConnectionClosed` even when the client was only reconnecting.

That is what JetStream publish hits during CONNECTED → RECONNECTING → CONNECTED.

If the connection is not actually closed, return `ErrDisconnected` instead.

Fixes #2128

## Test
- Added `TestRequestWithContextReconnectNotClosed` in `test/context_test.go`.
- `go test` for the client package compiles; the new test needs a NATS server to run.

### DIFF
--- a/test/context_test.go
+++ b/test/context_test.go
@@ -15,6 +15,7 @@ package test
 
 import (
 	"context"
+	"errors"
 	"strings"
 	"sync"
 	"testing"
@@ -266,6 +267,27 @@ func testContextRequestWithCancel(t *testing.T, nc *nats.Conn) {
 	if !strings.Contains(err.Error(), expected) {
 		t.Errorf("Expected %q error, got: %q", expected, err.Error())
 	}
+}
+
+func TestRequestWithContextForceReconnectNotClosed(t *testing.T) {
+	withServer(t, func(t *testing.T, nc *nats.Conn) {
+		ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+		defer cancel()
+
+		errCh := make(chan error, 1)
+		go func() {
+			_, err := nc.RequestWithContext(ctx, "nobody-home", nil)
+			errCh <- err
+		}()
+		time.Sleep(20 * time.Millisecond)
+		if err := nc.ForceReconnect(); err != nil {
+			t.Fatal(err)
+		}
+		err := <-errCh
+		if errors.Is(err, nats.ErrConnectionClosed) {
+			t.Fatalf("RequestWithContext returned ErrConnectionClosed during reconnect: %v", err)
+		}
+	})
 }
 
 func TestContextOldRequestClosed(t *testing.T) {
--- a/context.go
+++ b/context.go
@@ -67,7 +67,10 @@ func (nc *Conn) requestWithContext(ctx context.Context, subj string, hdr, data [
 		select {
 		case m, ok = <-mch:
 			if !ok {
-				return nil, ErrConnectionClosed
+				if nc.IsClosed() {
+					return nil, ErrConnectionClosed
+				}
+				return nil, ErrDisconnected
 			}
 		case <-ctx.Done():
 			nc.mu.Lock()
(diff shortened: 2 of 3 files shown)


test/context_test.go
```diff
@@ -268,6 +269,27 @@ func testContextRequestWithCancel(t *testing.T, nc *nats.Conn) {
 	}
 }
 
+func TestRequestWithContextForceReconnectNotClosed(t *testing.T) {
```
piotrpio, line 272: This test does not exercise the fix - since Request is sent to "nobody-home" and there is no subscriber, server will respond immediately with 503 (no responders), way before the 20ms wait time.

Instead, you could change it to have a subscriber which does not respond and then trigger reconnect:
```go
sub, _ := nc.SubscribeSync("hang")
go func() { _, err := nc.RequestWithContext(ctx, "hang", nil); errCh <- err }()
if _, err := sub.NextMsg(time.Second); err != nil {
  t.Fatal(err) 
}
nc.ForceReconnect()
if err := <-errCh; !errors.Is(err, nats.ErrDisconnected) { 
  t.Fatalf(...) 
}
```

piotrpio (changes requested):

PR #2130, opened 2026-09-04
fix processHeaderMsgArgs allocations

0xAozora:
JetStream messages may contain 5 arguments, while `argsLenMax` is currently set to 4. Appending the fifth argument causes the slice to grow and allocate a new [][]byte with capacity for 8 elements.

Increased the size of the backing array to avoid the allocation.

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -430,6 +430,7 @@ func (nc *Conn) cloneMsgArg() {
 }
 
 const argsLenMax = 4
+const hdrArgsLenMax = 5
 
 func (nc *Conn) processMsgArgs(arg []byte) error {
 	// Use separate function for header based messages.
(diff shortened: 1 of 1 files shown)


parser.go
```diff
@@ -484,7 +484,7 @@ func (nc *Conn) processMsgArgs(arg []byte) error {
 // processHeaderMsgArgs is for a header based message.
 func (nc *Conn) processHeaderMsgArgs(arg []byte) error {
 	// Unroll splitArgs to avoid runtime/heap issues
-	a := [argsLenMax][]byte{}
+	a := [argsLenMax + 1][]byte{} // Account for Jetstream Reply Subject
```
piotrpio: At this point this has nothing to do with JetStream - HMSG is a core NATS concept. This needs larger size than MSG because it can also contain headers, in addition to optional reply subject. I would either rephrase or remove this comment.

Also, instead of `argsLenMax + 1` please extract a const for this, e.g.:
```go
const hdrArgsLenMax = 5
```

piotrpio (commented):
This is a very good catch - thanks for finding it! Just one suggestion to clean it up a little bit.

0xAozora:
Done.

piotrpio (approved):
LGTM!

PR #2132, opened 2026-09-05
Avoid extra allocations in DecodeHeaderMsg

0xAozora:
I found it unnecessary to wrap the whole data slice into a `bytes.Reader`, then a `bufio.Reader`, then a `textproto.Reader`:

```go
br := bufio.NewReaderSize(bytes.NewReader(data), 128)
tp := textproto.NewReader(br)
```

This causes unnecessary allocations. I understand the appeal of reusing something that already does the job, but it was simple enough to extract the parsing behavior into a small `nextHeaderLine` helper instead.

I also noticed that the whole message payload is already copied once in `processMsg`, so we can unsafely slice header keys/values directly out of that buffer instead of allocating a fresh string for each one. Since I have already seen this exact `unsafe` pattern used in the builtin encoders, I assumed it was acceptable to use here too.

Alternatively, if we'd rather avoid that aliasing, `processMsg` could instead copy only the payload portion, but then still incur multiple allocations for the header keys/values.


piotrpio (changes requested):
Getting rid of the three Readers is a clear win and I'd take it as-is. The unsafe aliasing I don't want on an exported function: the keys alias too, so a caller that reuses its buffer doesn't get stale values, it gets a map whose hash buckets no longer match their keys - `Get` returns "" for an entry that's there.

Your alternative (copy only the payload, `string()` each key/value) is worse than main on allocations - I measured 18 vs 15. But there's a middle option: copy the header block once, `s := string(data)` at the top, and parse `s` instead of `data`. Keys and values become substrings of a string nobody else can reach. That's what `textproto` was doing per line - doing it once per message beats both.

PR #2133, opened 2026-09-07
[FIXED] Ordered consumer reset resurrecting a removed subscription

Jarema:
Follow-up to the review discussion on #2122 (now merged; this is rebased on `main`).

## Problem

`applyNewSID` releases `sub.mu` before taking `subsMu`, and the protocol for the new sid is sent from a goroutine later still. An `Unsubscribe`/`Drain` (`removeSub`) or a connection close can land in either window.

**Window 1, inside `applyNewSID`.** The removed subscription was unconditionally re-registered under the new sid:

- a closed subscription stayed in `nc.subs` for the life of the connection,
- a `SUB` for it was sent to the server and a new ordered consumer was created for a subscription that will never deliver anything,
- when `close()` had already set `nc.subs` to `nil`, the write panicked with `assignment to entry in nil map` (reachable from the readLoop or the heartbeat `activityCheck` timer).

Re-checking `sub.closed` after `applyNewSID` returns is too late: the map write already happened. The guard has to sit inside `applyNewSID` under `subsMu`.

**Window 2, between `applyNewSID` and the reset goroutine taking `nc.mu`.** `Unsubscribe` removed the subscription under the new sid and sent an `UNSUB` for a sid the server had never seen, while the goroutine still sent `SUB` for the new sid and created a new consumer: interest for a dead subscription stayed on the server until the connection closed, and the consumer was left for the inactivity threshold.

**Window 3, a reset racing with `Drain`.** A draining subscription is a live entry in `nc.subs`: `unsubscribe` skips `removeSub` in drain mode, and `checkDrained` removes the entry later *without* sending an `UNSUB`. A registration check alone therefore passes for it, so a reset racing with `Drain` sent a `SUB` for the new sid *after* the drain's `UNSUB` had already gone out. That interest was never removed and stayed on the server for the life of the connection, and the server kept pushing messages into a subscription that was supposed to have stopped receiving them.

## Fix

- `applyNewSID` returns `ok == false` and leaves everything untouched when `nc.subs[osid] != sub`; `resetOrderedConsumer` bails out.
- The protocol part of the reset moves to `rewireOrderedSub`, which under `nc.mu` checks that the subscription is still registered under the new sid **and not draining**, and otherwise only sends the `UNSUB` for the old sid. `unsubscribe` runs under `nc.mu` too, so the check is exact.
- If the subscription got closed or drained while the consumer create request was in flight, the goroutine deletes the consumer it just created instead of leaving it for the inactivity threshold.
- `resetOrderedConsumer` also bails out at entry for a subscription that is already draining, so it is not re-keyed at all.

## Tests

- `TestApplyNewSIDUnregisteredSub` (white-box, no server): failed before the fix with the resurrected map entry and the nil-map panic.
- `TestRewireOrderedSub` (white-box, no server): asserts the exact protocol sent for a registered, a removed, a draining, and a closed-connection subscription.
- `TestJetStreamOrderedConsumerSIDRace` additionally asserts `nc.NumSubscriptions()` returns to its baseline after every round.
- Not covered by a test: the post-create `sub.closed || sub.draining` guard (needs a server-side race in the middle of a consumer create request).

https://claude.ai/code/session_01FZSznPKF6L5UrYgryNRFsd

### DIFF
--- a/js.go
+++ b/js.go
@@ -2220,7 +2220,7 @@ func (sub *Subscription) applyNewSID() (osid, nsid int64, ok bool) {
 // Lock should be held.
 func (sub *Subscription) resetOrderedConsumer(sseq uint64) {
 	nc := sub.conn
-	if sub.jsi == nil || nc == nil || sub.closed {
+	if sub.jsi == nil || nc == nil || sub.closed || sub.draining {
 		return
 	}
 
@@ -2337,9 +2337,9 @@ func (sub *Subscription) resetOrderedConsumer(sseq uint64) {
 		}
 
 		sub.mu.Lock()
-		if sub.closed {
-			// Unsubscribed while the consumer was being created. The
-			// consumer is not attached to anything anymore, so delete it
+		if sub.closed || sub.draining {
+			// Unsubscribed or drained while the consumer was being created.
+			// The consumer is not attached to anything anymore, so delete it
 			// rather than waiting for the inactivity threshold.
 			sub.mu.Unlock()
 			go js.DeleteConsumer(jsi.stream, cinfo.Name)
@@ -2350,22 +2350,24 @@ func (sub *Subscription) resetOrderedConsumer(sseq uint64) {
 	}()
 }
 
-// rewireOrderedSub sends the protocol that moves an ordered consumer
-// subscription from its old sid to its new sid and deliver subject: UNSUB for
-// the old sid, SUB for the new one and, if the subscription has a max set,
-// the UNSUB carrying the adjusted max. Returns whether the subscription is
-// still registered on the connection. If it is not (it was unsubscribed, or
-// the connection was closed, after applyNewSID re-keyed it), only the UNSUB
-// for the old sid is sent and false is returned, so that no interest is
-// created on the server for a subscription that no longer exists. Since
-// unsubscribe also runs under nc.mu, the check here is exact.
+// rewireOrderedSub moves the subscription from osid to nsid on the server.
+// The old sid is always unsubscribed; the new one is only subscribed if the
+// subscription is still registered and not draining, so that no interest is
+// created for a subscription that will not accept messages anymore. Returns
+// whether that happened. unsubscribe runs under nc.mu too, so the check is
+// exact.
 func (nc *Conn) rewireOrderedSub(sub *Subscription, osid, nsid int64, deliver, maxStr string) bool {
 	nc.mu.Lock()
 	defer nc.mu.Unlock()
 
 	nc.bw.appendString(fmt.Sprintf(unsubProto, osid, _EMPTY_))
 	nc.subsMu.RLock()
-	registered := nc.subs[nsid] == sub
+	sub.mu.Lock()
+	// A draining subscription is only removed from nc.subs once the drain
+	// completes, and that removal sends no UNSUB, so subscribing the new sid
+	// here would leave interest on the server for the life of the connection.
+	registered := nc.subs[nsid] == sub && !sub.draining
+	sub.mu.Unlock()
 	nc.subsMu.RUnlock()
 	if registered {
 		nc.bw.appendString(fmt.Sprintf(subProto, deliver, _EMPTY_, nsid))
--- a/js.go
+++ b/js.go
@@ -2267,14 +2267,11 @@ func (sub *Subscription) resetOrderedConsumer(sseq uint64) {
 		// Unsubscribe and subscribe with new inbox and sid.
 		// Remap a new low level sub into this sub since its client accessible.
 		// This is done here in this go routine to prevent lock inversion.
-		nc.mu.Lock()
-		nc.bw.appendString(fmt.Sprintf(unsubProto, osid, _EMPTY_))
-		nc.bw.appendString(fmt.Sprintf(subProto, newDeliver, _EMPTY_, nsid))
-		if maxStr != _EMPTY_ {
-			nc.bw.appendString(fmt.Sprintf(unsubProto, nsid, maxStr))
+		if !nc.rewireOrderedSub(sub, osid, nsid, newDeliver, maxStr) {
+			// Unsubscribed (or connection closed) since applyNewSID:
+			// nothing to recreate.
+			return
 		}
-		nc.kickFlusher()
-		nc.mu.Unlock()
 
 		pushErr := func(err error) {
 			nc.handleConsumerSequenceMismatch(sub, fmt.Errorf("%w: recreating ordered consumer", err))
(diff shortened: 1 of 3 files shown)


js.go
```diff
@@ -2324,11 +2337,46 @@ func (sub *Subscription) resetOrderedConsumer(sseq uint64) {
...
 
+// rewireOrderedSub sends the protocol that moves an ordered consumer
+// subscription from its old sid to its new sid and deliver subject: UNSUB for
+// the old sid, SUB for the new one and, if the subscription has a max set,
+// the UNSUB carrying the adjusted max. Returns whether the subscription is
+// still registered on the connection. If it is not (it was unsubscribed, or
+// the connection was closed, after applyNewSID re-keyed it), only the UNSUB
+// for the old sid is sent and false is returned, so that no interest is
+// created on the server for a subscription that no longer exists. Since
+// unsubscribe also runs under nc.mu, the check here is exact.
```
piotrpio: Most of this comment is just narrating code, trim it to keep the important part, e.g.:
```
// rewireOrderedSub moves the subscription from osid to nsid on the server.
// The old sid is always unsubscribed; the new one is only subscribed if the
// subscription is still registered, so that no interest is created for a
// subscription that no longer exists. Returns whether that happened.
```

js.go
```diff
@@ -2239,7 +2250,12 @@ func (sub *Subscription) resetOrderedConsumer(sseq uint64) {
 	}
 
 	// Quick unsubscribe. Since we know this is a simple push subscriber we do in place.
-	osid, nsid := sub.applyNewSID()
+	osid, nsid, ok := sub.applyNewSID()
+	if !ok {
+		// The subscription was unsubscribed or the connection was closed
+		// while sub.mu was released: there is nothing left to reset.
```
piotrpio: Similar comment repeated multiple times, I think it's clear enough from `applyNewSID` doc and can be dropped here.

js.go
```diff
@@ -2251,14 +2267,11 @@ func (sub *Subscription) resetOrderedConsumer(sseq uint64) {
...
 		// Remap a new low level sub into this sub since its client accessible.
 		// This is done here in this go routine to prevent lock inversion.
-		nc.mu.Lock()
-		nc.bw.appendString(fmt.Sprintf(unsubProto, osid, _EMPTY_))
-		nc.bw.appendString(fmt.Sprintf(subProto, newDeliver, _EMPTY_, nsid))
-		if maxStr != _EMPTY_ {
-			nc.bw.appendString(fmt.Sprintf(unsubProto, nsid, maxStr))
+		if !nc.rewireOrderedSub(sub, osid, nsid, newDeliver, maxStr) {
+			// Unsubscribed (or connection closed) since applyNewSID:
+			// nothing to recreate.
```
piotrpio: Similar comment repeated multiple times, I think it's clear enough from `applyNewSID` doc and can be dropped here.

js.go
```diff
@@ -2184,14 +2184,25 @@ func (sub *Subscription) checkOrderedMsgs(m *Msg) bool {
...
+func (sub *Subscription) applyNewSID() (osid, nsid int64, ok bool) {
 	nc := sub.conn
 	sub.mu.Unlock()
 
 	nc.subsMu.Lock()
 	osid = sub.sid
+	// While sub.mu was released, the subscription may have been removed by
+	// removeSub (Unsubscribe/Drain) or the connection may have been closed
+	// (nc.subs is nil then). Registering it under a new sid would resurrect
+	// a closed subscription (or panic on the nil map), so refuse the swap.
```
piotrpio: Trim to e.g
```
// removeSub or close() may have run while sub.mu was released;
// re-registering would resurrect the sub or write to a nil nc.subs.
```

Same for the two lines added to the function doc - "Returns ok == false, leaving everything untouched, if the subscription is no longer registered on the connection" covers it in one.

js.go
```diff
@@ -2324,11 +2337,46 @@ func (sub *Subscription) resetOrderedConsumer(sseq uint64) {
...
+// created on the server for a subscription that no longer exists. Since
+// unsubscribe also runs under nc.mu, the check here is exact.
+func (nc *Conn) rewireOrderedSub(sub *Subscription, osid, nsid int64, deliver, maxStr string) bool {
+	nc.mu.Lock()
+	defer nc.mu.Unlock()
+
+	nc.bw.appendString(fmt.Sprintf(unsubProto, osid, _EMPTY_))
+	nc.subsMu.RLock()
+	registered := nc.subs[nsid] == sub
+	nc.subsMu.RUnlock()
```
piotrpio, line 2365: This will also return true if a subscription is draining (sub is not removed from `nc.subs` until draining completes) so we should include draining check:
```
nc.subsMu.RLock()
sub.mu.Lock()
registered := nc.subs[nsid] == sub && !sub.draining
sub.mu.Unlock()
nc.subsMu.RUnlock()
```

js.go
```diff
@@ -2324,11 +2337,46 @@ func (sub *Subscription) resetOrderedConsumer(sseq uint64) {
 		}
 
 		sub.mu.Lock()
+		if sub.closed {
```
piotrpio:
```suggestion
		if sub.closed || sub.draining {
```

piotrpio (changes requested):

js.go
Jarema: Trimmed to your wording, and the two extra doc lines collapsed into "Returns ok == false, leaving everything untouched, if the subscription is no longer registered on the connection."

js.go
Jarema: Done — used your wording, with the draining condition folded in and one line kept for the non-obvious part ("unsubscribe runs under nc.mu too, so the check is exact").

js.go
Jarema: Dropped.

js.go
Jarema: Dropped.

js.go
Jarema: Good catch, applied as written.

Confirming the mechanism, because it is worse than an extra `SUB`: `unsubscribe` skips `removeSub` in drain mode, and `checkDrained` later removes the entry from `nc.subs` **without** sending an `UNSUB`. So the `SUB` for the new sid went out *after* the drain's own `UNSUB`, and nothing ever took that interest down again — it stayed on the server for the life of the connection, with the server pushing into a subscription that was supposed to have stopped receiving.

Also added an entry guard in `resetOrderedConsumer` (`sub.closed || sub.draining`, `sub.mu` is held there) so a subscription that is already draining is not re-keyed at all. That does not replace this check — `draining` can be set after that point — so both are in.

Covered by a new `TestRewireOrderedSub` subtest ("draining sub only gets the old sid unsubscribed"); it fails without the `!sub.draining` term.

js.go
Jarema: Applied. Worth spelling out why it matters on the drain path: `checkDrained` calls `sub.deleteConsumer()`, which deletes whatever `jsi.consumer` holds — so without this guard the goroutine would overwrite it with the freshly created name, the drain would delete the new consumer, and the pre-reset one would be left for its inactivity threshold instead.

js_test.go
```diff
@@ -131,3 +131,61 @@ func TestJetStreamConvertDirectMsgResponseToMsg(t *testing.T) {
...
+		}
+		if osid != 1 || nsid != 2 {
+			t.Fatalf("expected sids 1 -> 2, got %d -> %d", osid, nsid)
+		}
+		if sub.sid != nsid || nc.subs[nsid] != sub || len(nc.subs) != 1 {
+			t.Fatalf("sub not registered under the new sid only: sid=%d subs=%v", sub.sid, nc.subs)
+		}
+	})
+
+	t.Run("removed sub is not re-registered", func(t *testing.T) {
```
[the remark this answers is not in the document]
Jarema: Correct — `removeSub` does set `s.closed = true` (under `s.mu`, held for the whole function). Leaving the test as is: the `nc.subs[osid] != sub` guard fires first, so setting `closed` would not change what is being asserted.

piotrpio (approved):
LGTM!

piotrpio (approved):
LGTM!

PR #2134, opened 2026-09-07
[FIXED] Migrate golangci-lint config to v2

piotrpio:
The v1 binary cannot decode export data from current Go toolchains, so typechecking collapses and the lint job fails on every branch.

Config migrated with `golangci-lint migrate`. The merged staticcheck linter enables check families this repo never ran, so `checks` is pinned to the previous set — no new findings, no code changes.


Jarema (approved):
LGTM

PR #2135, opened 2026-09-07
[FIXED] Messages(): handle a reconnect that completes between Next calls

1995parham:
## Problem

A durable pull consumer driven by `Messages()` goes silent after a NATS server restart when `Next()` is called with a short `NextMaxWait`. The connection reconnects and stays healthy, `Next()` keeps returning `nats.ErrTimeout`, and the server reports `num_waiting: 0` with `num_pending` growing. Nothing is logged, no error is returned, and only recreating the iterator recovers it. We hit this in production on a 3-node cluster rolling restart with nats.go v1.53.1 and `Next(NextMaxWait(5 * time.Second))`.

## Cause

Reconnect handling for `Messages()` was scoped to a single `Next()` call. The status goroutine started by `Messages()` forwards `RECONNECTING` and `CONNECTED` into `sub.errs`, but `isConnected` was a local of `Next`, so only a `Next` that had observed the `RECONNECTING` event reset the pending counters when the matching `CONNECTED` arrived. If the disconnect was observed by call N and the reconnect landed after call N returned, call N+1 started with `isConnected = true` and discarded `errConnected`. `pending.msgCount` stayed at `MaxMessages`, `checkPending` never sent a new `CONSUMER.MSG.NEXT`, and the iterator waited forever on a request the server no longer held.

`Consume()` is not affected: its status goroutine owns the connection state for the lifetime of the subscription.

## Fix

Track `isConnected` in the status goroutine that lives as long as the iterator and only forward a `CONNECTED` that follows a `RECONNECTING`. `Next` now treats every `errConnected` as a real reconnect and resets the pending counters (or returns it, for `notifyOnReconnect` used by the ordered consumer, whose behaviour is unchanged). The forwards also select on `done`, so the goroutine can no longer block forever on a full `errs` channel after `Stop`.

## Test

`TestPullConsumerMessages/with_server_restart_and_short_Next_timeout` restarts the server while `Next` is polled with a 100ms `NextMaxWait`. On `main` it fails with `Timeout waiting for messages after restart; got 5 of 10`; with this change it passes. `TestPullConsumerMessages`, `TestPullConsumerConsume`, `TestPullConsumerNext` and `TestOrderedConsumer*` pass with `-race` against the tester.

Note: the missed-heartbeat monitor is still armed per `Next` call, so a `NextMaxWait` shorter than `2 * Heartbeat` never lets it fire. That is a separate, smaller gap; happy to follow up on it if you want a subscription-level monitor for `Messages()` too.

### DIFF
--- a/jetstream/pull.go
+++ b/jetstream/pull.go
@@ -560,17 +560,37 @@ func (p *pullConsumer) Messages(opts ...PullMessagesOpt) (MessagesContext, error
 	go sub.pullMessages(subject)
 
 	go func() {
+		// Connection state is tracked here, for the lifetime of the
+		// iterator, rather than inside Next: a reconnect that completes
+		// between two Next calls (for example when Next is driven with a
+		// short NextMaxWait) must still reset the pull request accounting,
+		// otherwise no new pull request is ever sent and the iterator goes
+		// silent on a healthy connection.
+		isConnected := true
 		for {
 			select {
 			case status, ok := <-sub.connStatusChanged:
 				if !ok {
 					return
 				}
-				if status == nats.CONNECTED {
-					sub.errs <- errConnected
-				}
-				if status == nats.RECONNECTING {
-					sub.errs <- errDisconnected
+				switch status {
+				case nats.RECONNECTING:
+					isConnected = false
+					select {
+					case sub.errs <- errDisconnected:
+					case <-sub.done:
+						return
+					}
+				case nats.CONNECTED:
+					if isConnected {
+						continue
+					}
+					isConnected = true
+					select {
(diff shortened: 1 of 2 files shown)


piotrpio (approved):
LGTM! Thank you for finding this and thanks a lot for the contribution.

As for the subscription-level heartbeat monitor for `Messages()` - we could give it a go, although this requires a bit more thought than e.g. with Consume(), as with `Messages()` it's possible that there is no active pull request as these are user-driven (so server will not be sending idle heartbeats) - so we don't want to always have the timer running. That's why I designed it to run in `Next()` in the first place, although your point about the `NextMaxWait` is valid.

PR #2137, opened 2026-09-09
[FIXED] Ordered consumer silently losing messages on slow consumer

### LINKED ISSUE #2107: Ordered consumer silently loses messages on a client-side slow-consumer episode (no gap detection, no reset)
## Observed behavior

A push-based ordered consumer (`Subscribe(..., nats.OrderedConsumer())`, which KV and Object Store watchers use internally) can silently drop messages during a client-side slow-consumer episode. The dropped messages are never refetched, and no gap is detected, so the ordered consumer's reset does not fire. The only signal is one generic `ErrSlowConsumer` callback per episode.

Root cause: `checkOrderedMsgs` runs before the slow-consumer guard and advances the sequence tracker for a message that is then dropped.

- `nats.go:3766`: `msgHandler` calls `checkOrderedMsgs(m)` before the pending-limit check.
- `js.go:2172`: for an in-order message, `checkOrderedMsgs` sets `jsi.dseq, jsi.sseq = dseq+1, sseq` and returns `false` (no gap).
- `nats.go:3788-3791` and `:3807`: the pending-limit and async-full checks `goto slowConsumer`.
- `nats.go:3880`: `slowConsumer` drops the message, fires `ErrSlowConsumer` only on episode transition (`sc := !sub.sc`), and does not roll back the tracker or reset.

The tracker advanced before the drop, so the next message matches the advanced `dseq`, no gap is detected, and recovery never fires.

Scope: the legacy push path (`nats.OrderedConsumer()`), including KV and Object Store watchers. The `jetstream` package ordered consumer is not affected: it has no slow-consumer drop path, pull flow control bounds in-flight messages below the pending limit, and gap detection is on the consumer sequence checked before the cursor advances.

## Expected behavior

Either the dropped messages are redelivered (the ordered consumer resets and refetches from the last delivered sequence, as it does on a detected gap), or the loss is surfaced to the application.

## Server and client version

- `nats.go` v1.51.0
- `nats-server` v2.14.3 (embedded via the `server` package)

## Host environment

Go 1.26, macOS arm64, embedded `nats-server`. Not environment-specific; the mechanism is in the client delivery path.

## Steps to reproduce

Start an embedded JetStream server, add a stream, publish 100 messages. Run an ordered consumer whose first callback stalls (so the async queue backs up) and whose pending-msg limit is lowered so the slow-consumer guard trips:

```go
gate := make(chan struct{})
var once sync.Once
var mu sync.Mutex
var got []uint64

cb := func(m *nats.Msg) {
    meta, _ := m.Metadata()
    mu.Lock(); got = append(got, meta.Sequence.Stream); mu.Unlock()
    once.Do(func() { <-gate }) // stall the first callback so the queue backs up
}

sub, _ := js.Subscribe("foo.>", cb, nats.OrderedConsumer())
sub.SetPendingLimits(20, 8*1024*1024) // low msg limit -> slow-consumer drops

// wait until sub.Dropped() > 0, then:
close(gate)
// let it drain, then publish one more in-order message (stream seq 101)

// Result: got == {1..20, 101}; stream seq 21..100 were dropped and are never
// redelivered; sub.Dropped() == 80; no reset fired. A recovered consumer would
// have refetched 21..100 (from sseq+1) before delivering 101.
```

Deterministic across runs (80 of 100 lost each time). A runnable version, including a check that the `jetstream` package is not affected, is available on request.

## Possible fix directions

1. On a slow-consumer drop for an ordered subscription, roll back `jsi.dseq/jsi.sseq` so the next message trips gap detection and the existing reset path recovers.
2. Move `checkOrderedMsgs` after the pending-limit check so a dropped message reads as a gap.
3. Treat a slow-consumer drop on an ordered consumer as an explicit reset trigger.

piotrpio:
Fixes #2107

`processMsg` validated ordering before deciding whether the message could be delivered. `checkOrderedMsgs` advanced jsi.dseq/jsi.sseq, and the pending-limit guard could then drop the message, leaving the tracker past a message that was never delivered, so the next one matched and no gap was ever detected.

This fix decouples checking the sequences and gap detection from actually moving the trackers forward.

To keep that from becoming a reset storm:

- While the subscription is full, neither a gap nor a heartbeat resets the consumer, since a new one could only refetch into the same wall.
- An unbuffered channel has no capacity signal, so while nothing has been delivered since the last reset it waits for the heartbeat instead.

### DIFF
--- a/js.go
+++ b/js.go
@@ -2230,7 +2230,10 @@ func (sub *Subscription) checkOrderedDelivery(m *Msg) (jsMsgAction, orderedSeqs)
 	jsi := sub.jsi
 	// An unbuffered channel has no capacity signal. With nothing delivered
 	// since the last reset, resetting again would just refetch into the same
-	// wall, so leave it to the heartbeat.
+	// wall, so leave it to the heartbeat. Once something has been delivered
+	// the reader is there and a drop means it was merely busy, so reset at
+	// once; deferring to the heartbeat here would drain a backlog one message
+	// per interval.
 	if sub.mch != nil && cap(sub.mch) == 0 && jsi.dseq == 1 {
 		return jsMsgDrop, orderedSeqs{}
 	}
--- a/js.go
+++ b/js.go
@@ -1437,6 +1437,8 @@ type jsSub struct {
 	dseq    uint64
 	sseq    uint64
 	ccreq   *createConsumerRequest
+	// A reset waits for the buffer to drain, see tryResetOrderedConsumer.
+	resetPending bool
 
 	// Heartbeats and Flow Control handling from push consumers.
 	hbc    *time.Timer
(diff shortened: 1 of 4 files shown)


piotrpio:
@claude re-review this PR, there were a few new commits since your review

piotrpio:
@claude re-review looking at the recent changes. The last commit changed the trigger on which consumer is reset when buffer frees

wallyqs (approved):
LGTM

PR #2138, opened 2026-09-09
[IMPROVED] Fix list consumers test to work agains latest server

piotrpio:
Server now has a default max consumers limit per stream. This adjusts tests which create more than 1000 consumers to work against these changes.


wallyqs (approved):
LGTM

PR #2145, opened 2026-09-15
Add `nats.MultipathTCP()` for enabling or disabling MPTCP

neilalexander:
The server already has support for MPTCP connections when running on a system that has it supported and enabled, but Go does not default to using it for outbound dials. I think we should maintain that default but allow it to be enabled when a user requests it.


neilalexander:
@claude review

piotrpio (approved):
LGTM!

PR #2150, opened 2026-09-20
fix(jetstream): report pending publish errors during reconnect

H-XX-D:
## Problem

Both JetStream async publishers register error callbacks with `defer` inside the persistent reconnect-status loop. On reconnect, pending futures receive `ErrDisconnected` and are removed, but the callbacks wait until the worker exits, delaying application error handling and retaining message arguments until shutdown or publisher cleanup. Completion must also be signaled after the callbacks return, so callers waiting on `PublishAsyncComplete` can observe the callbacks' completed error handling.

## Solution

Collect affected message pointers and capture and clear the completion channel while holding the publisher lock. After unlocking, invoke the configured error callbacks, close the captured completion channel, and then wait for the next connection-status event. Apply the same change to the legacy and current JetStream implementations, matching the callback-before-completion ordering in `handleAsyncReply`.

## Verification

Two native regressions use `testing/synctest` to verify callback delivery before shutdown, callback reentry through `PublishAsyncPending`, and completion-channel ordering. The new ordering assertions fail against the preceding PR head (`9f1fe7a`) in both packages; with this update, both regressions pass 50 repetitions each with the race detector.

Passed with Go 1.26.0 on macOS/arm64:

```sh
go test -modfile=go_test.mod -race -run '^TestJetStreamReconnectErrorCallback$' -count=50 -timeout=60s . ./jetstream
go test -v -modfile=go_test.mod -tags=internal_testing -race -p=1 -count=1 -timeout=10m ./...
go test -v -modfile=go_test.mod -p=1 -count=1 -run TestNoRace -timeout=5m ./...
go build -modfile=go_test.mod ./...
go vet -modfile=go_test.mod ./...
git diff --cached --check
```

The full integration suite uses the repository's documented in-process tester. No module files or dependencies were changed.

AI disclosure: Codex assisted with orchestrating a proprietary static analysis tool and preparing structured regression tests.

Signed-off-by: Todd Hendricks <tjhendrx@icloud.com>


jetstream/publish.go
```diff
@@ -630,6 +631,9 @@ func (js *jetStream) resetPendingAcksOnReconnect() {
 			js.publisher.doneCh = nil
 		}
 		js.publisher.Unlock()
+		for _, msg := range msgs {
+			errCb(js, msg, nats.ErrDisconnected)
+		}
```
piotrpio, line 634: I think callbacks should be invoked before closing `doneCh` (used for `PublishAsyncComplete`), so that `<-PublishAsyncComplete` only notified after all callbacks are invoked - this is the order also `handleAsyncReply` uses. Same applies to the fix in legacy API.

piotrpio (changes requested):
Thank you for catching that and creating a PR - just one nit before merging.

jetstream/publish.go
H-XX-D: Thanks for pointing this out. Updated both APIs to invoke the error callbacks before closing the captured completion channel, while keeping callbacks outside the publisher lock.

The regressions now check that completion stays open during the callback and closes afterward. Both fail before this update and pass 50 race-enabled repetitions with it. The full race-enabled suite, separate NoRace tests, build, and vet also pass.

PR #2152, opened 2026-09-20
[FIXED] Protect jetstream cached Info with atomic.Pointer

### LINKED ISSUE #2151: jetstream: data race on cached info between concurrent Info calls (pullConsumer, pushConsumer, stream)
## Observed behavior

`Info` stores the fetched info on the consumer, and `CachedInfo` reads it, with no synchronization. Two goroutines calling `Info` on one consumer therefore race on that field. The same shape is in `stream`.

```
jetstream/consumer.go:209   p.info = resp.ConsumerInfo   // pullConsumer.Info
jetstream/consumer.go:217   return p.info                // pullConsumer.CachedInfo
jetstream/consumer.go:242   p.info = resp.ConsumerInfo   // pushConsumer.Info
jetstream/consumer.go:250   return p.info                // pushConsumer.CachedInfo
jetstream/stream.go:498     s.info = &cached             // stream.Info
jetstream/stream.go:511     return s.info                // stream.CachedInfo
```

We hit it where one goroutine exports a durable's pending count on a ticker while another reads the same consumer, which is an ordinary shape for a service that both consumes and reports.

## Expected behavior

`Consumer` and `Stream` handles are safe for concurrent use, or the docs say they are not.

## Server and client version

nats.go v1.54.0, nats-server v2.15.0, go1.27.1 darwin/arm64.

## Steps to reproduce

`go test -race`, module requiring only those two:

```go
func TestConsumerInfoRace(t *testing.T) {
	srv, err := server.NewServer(&server.Options{
		Port: -1, JetStream: true, StoreDir: t.TempDir(), NoLog: true, NoSigs: true,
	})
	if err != nil {
		t.Fatal(err)
	}
	go srv.Start()
	defer srv.Shutdown()
	if !srv.ReadyForConnections(10 * time.Second) {
		t.Fatal("server did not start")
	}

	nc, err := nats.Connect(srv.ClientURL())
	if err != nil {
		t.Fatal(err)
	}
	defer nc.Close()
	js, err := jetstream.New(nc)
	if err != nil {
		t.Fatal(err)
	}
	ctx := context.Background()
	if _, err := js.CreateStream(ctx, jetstream.StreamConfig{Name: "T", Subjects: []string{"t"}}); err != nil {
		t.Fatal(err)
	}
	c, err := js.CreateOrUpdateConsumer(ctx, "T", jetstream.ConsumerConfig{
		Durable: "d", AckPolicy: jetstream.AckExplicitPolicy,
	})
	if err != nil {
		t.Fatal(err)
	}

	var wg sync.WaitGroup
	for range 2 {
		wg.Go(func() {
			for range 20 {
				if _, err := c.Info(ctx); err != nil {
					t.Error(err)
					return
				}
			}
		})
	}
	wg.Wait()
}
```

```
WARNING: DATA RACE
  github.com/nats-io/nats.go/jetstream.(*pullConsumer).Info()
      jetstream/consumer.go:209
  github.com/nats-io/nats.go/jetstream.(*pullConsumer).Info()
      jetstream/consumer.go:209
--- FAIL: TestConsumerInfoRace (0.08s)
```

A mutex around the store and the `CachedInfo` read would cover it. Callers can serialize instead, but only where they own every caller, which a library-internal reader rules out.

vulragrag-star:
Concurrent `Info` and `CachedInfo` on the same pull/push consumer or stream raced on the cached info pointer (writes in `Info`, reads in `CachedInfo` with no synchronization).

This change stores Stream / Consumer infos in `atomic.Pointer` so cache updates and `CachedInfo` reads are safe without taking the consumer/stream mutex. That avoids nesting with `orderedConsumer.reset()`, which already holds the pull-consumer lock while reading the cached name (the previous shared-mutex approach deadlocked `TestOrderedConsumerConsume`).

```bash
go test -modfile=go_test.mod -race -count=1 ./jetstream/ -run 'TestCachedInfo'
```

Fixes #2151

Signed-off-by: Jason Wang <vulragrag@gmail.com>


piotrpio (changes requested):
The fix makes `TestOrderedConsumerConsume/base_usage,_delete_consumer` hang indefinitely - you're takes the same mutex as `orderedConsumer.reset()` does.

While I agree it would be good to make this operation safe to use concurrently, I would not re-use the same lock which is used across network operations. Instead, what we could do is wrap Stream / Consumer infos in `atomic.Pointer`.

vulragrag-star:
Switched Stream/Consumer cached infos to `atomic.Pointer` so `CachedInfo` no longer nests with `orderedConsumer.reset()`'s consumer lock. Tip: `f8cfda77e9d754253ea3b1c634c114e9cac86bd0`.

PR #2153, opened 2026-09-21
Fix JetStream async publish pending recovery under concurrency

### LINKED ISSUE #1612: JetStream: the value for max pending async messages can be exceeded

tsukiyoz:
## Summary

Fixes #1612.

Concurrent PublishMsgAsync calls currently register their futures before checking WithPublishAsyncMaxPending. Unsent, stalled callers therefore occupy pending slots. Even after all published messages receive ACKs, these unsent entries can keep the count above the limit and prevent stalled callers from resuming until stall timeouts expire.

## Changes

- Wait for capacity before registering a future.
- Check capacity and subscribe to the shared wakeup channel under the same lock, then recheck capacity after each wakeup.
- Preserve the WithStallWait timeout budget across wakeups.
- Wake stalled callers when pending entries are cleared by synchronous send failures, reconnects, connection close, or publisher cleanup.
- Add regression tests for concurrent recovery, stall/ACK timeouts, cleanup and reuse, connection close, and synchronous send failure.

No new dependencies or public API changes. This change targets the jetstream package; the legacy root-package JetStream implementation is not modified. PublishAsyncPending counts registered publishes, not callers still waiting to acquire capacity.

## Reproduction

With ACKs deliberately held and 16 additional concurrent publishers:
- max pending 1: the original code reports 17 pending.
- max pending 5: the original code reports 21 pending.
- Releasing the ACKs does not restore progress before stall timeouts on the original code.

The regression test passes with this change and verifies that pending remains within the configured limit.

## Validation

Local in-process tester: nats-server 2.15.0.

Passed:
- New concurrent recovery and wait-exit tests: 20 repetitions with -race.
- Publish/reconnect-related tests with -race, including synchronous send failure.
- Full jetstream/test race suite, root package, micro packages, and internal packages via make test.
- make test-norace.

Full-suite caveat:
- make test stopped in the Core NATS test package at TestForceReconnectSubsequentCalls (unexpected extra reconnect status).
- A complete rerun of ./test/... without failfast instead failed TestContextRequestWithDeadline (context deadline exceeded).
- Both failures also occurred intermittently in repeated targeted runs. These tests do not directly exercise the modified JetStream publisher, but they have not been checked against an unmodified baseline, so this PR does not claim a fully green suite or proven pre-existing failures.

Signed-off-by: tsukiyo <tsukiyo6@163.com>


tsukiyoz:
I reproduced the same failure on the unmodified base commit 7a8404a, using the coverage flags from CI. The test exercises the legacy nc.JetStream() ordered-consumer path, whereas this PR only changes the newer jetstream package. This appears to be a pre-existing failure independent of this change.

